PR #7, opened 2014-11-06
FIX: Spotlight: searching in user homes, bug #543

hat001:
Spotlight cannot search in user homes.
When netatalk(8) starts, it sets up paths for Spotlight.
Since user has not logged in yet, nobody sets up home's path.


slowfranklin:
Looks like this never materialized. Closing now.

PR #13, opened 2015-10-11
remove #541 workaround patch. There was this problem with only early …

hat001:
…Fedora 20.


slowfranklin:
Merged as a41fbc87fd9b0926f123c009109001cdb39873b3.

PR #14, opened 2015-10-26
fix manpage of afp.conf

hat001:


slowfranklin:
Merged as e4a4fdb18870da7914acde437ebf7dfc890b13d1.

PR #15, opened 2015-10-30
Spotlight fixes

dbuckley:
I'm not sure if it is helpful but this fork contains a couple of Spotlight fixes I attempted to implement.  

The first commit is a simple fix for character escaping which resolves problems searching in sub-directories which contain spaces.

The second commit is an attempt to rework the Spotlight file-type search logic to drop unknown type restrictions rather than silently failing.  In at least Yosemite and El Capitan it appears many in-application dialogs silently restrict search to specific file types supported by the application.  With the previous behavior Netatalk would silently fail and no results would be returned, now (hopefully) unknown types will be ignored and results returned as if that file type was not specified.

I'm most definitely not a competent C programmer and only briefly tested these changes, but the limited testing I was able to do seemed successful.  Unfortunately they don't seem to help the larger issue we're facing with Netatalk (high server load when Finder windows with many files are open), so I'm not able to justify spending any more time on this.  

-David

### DIFF
--- a/etc/afpd/spotlight.c
+++ b/etc/afpd/spotlight.c
@@ -855,6 +855,9 @@ static int sl_rpc_openQuery(AFPObj *obj,
     sl_query = dalloc_value_for_key(query, "DALLOC_CTX", 0,
                                     "DALLOC_CTX", 1,
                                     "kMDQueryString");
+
+    LOG(log_debug, logtype_sl, "Spotlight query pre-conversion: \"%s\"", sl_query);
+
     if (sl_query == NULL) {
         EC_FAIL;
     }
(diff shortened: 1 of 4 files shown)


slowfranklin:
Thanks for contributing! There are a few issues with the patches (patch1 leaking memory, patch2 can probably we rewritten and simplified) I'd like to address before merging. Stay tuned... :)

andy-js:
Here is a fixed-up version of the first change:

<pre>
diff --git a/etc/afpd/spotlight.c b/etc/afpd/spotlight.c
index 4aabeca..fdc0d2d 100644
--- a/etc/afpd/spotlight.c
+++ b/etc/afpd/spotlight.c
@@ -831,6 +831,7 @@ static int sl_rpc_openQuery(AFPObj *obj,
     GError *error = NULL;
     bool ok;
     sl_array_t *scope_array;
+    gchar *scope;
 
     array = talloc_zero(reply, sl_array_t);
 
@@ -891,10 +892,22 @@ static int sl_rpc_openQuery(AFPObj *obj,
     scope_array = dalloc_value_for_key(query, "DALLOC_CTX", 0, "DALLOC_CTX", 1,
                                        "kMDScopeArray");
     if (scope_array == NULL) {
-        slq->slq_scope = talloc_strdup(slq, v->v_path);
+        scope = g_uri_escape_string(v->v_path,
+                                    G_URI_RESERVED_CHARS_ALLOWED_IN_PATH,
+                                    NULL);
     } else {
-        slq->slq_scope = talloc_strdup(slq, scope_array->dd_talloc_array[0]);
+        scope = g_uri_escape_string(scope_array->dd_talloc_array[0],
+                                    G_URI_RESERVED_CHARS_ALLOWED_IN_PATH,
+                                    NULL);
     }
+    if (scope == NULL) {
+        LOG(log_error, logtype_sl, "failed to escape search scope");
+        EC_FAIL;
+    }
+
+    slq->slq_scope = talloc_strdup(slq, scope);
+    g_free(scope);
+
     if (slq->slq_scope == NULL) {
         LOG(log_error, logtype_sl, "failed to setup search scope");
         EC_FAIL;
</pre>

slowfranklin:
I had this commit in branched and pushed it now. At the time I alos looked into integrating the other one, but that wasn't quite as simple.

slowfranklin:
This landed as 90aa43d1eb75d62452dafa0c55ab8455c548c379.

PR #29, opened 2016-01-17
configure option "--with-init-style=" for Gentoo.

hat001:
"gentoo" is renamed to "gentoo-openrc".
"gentoo-openrc" is same as "openrc".
"gentoo-systemd" is same as "systemd".


slowfranklin:
Merged, thanks!

PR #39, opened 2016-02-19
Add missing license based on BSD origin.

hat001:
Since it is derived from the BSD sources written by Elz, it is under the
CA regents copyright and usual license. But we need the more formal lanuage
included for it to be valid.

This addresses one of the remaining license FIXME issues:
https://github.com/adiknoth/netatalk-debian/blob/master/debian/copyright#L59
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=685878#56

### DIFF
--- a/etc/afpd/nfsquota.c
+++ b/etc/afpd/nfsquota.c
@@ -1,13 +1,33 @@
 /*
- *
- * parts of this are lifted from the bsd quota program and are
- * therefore under the following copyright:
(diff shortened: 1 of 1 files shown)


slowfranklin:
Thanks!

Will try to merge all the open pull requests on Monday.

slowfranklin:
Merged. Thanks!

PR #50, opened 2016-04-18
Use start-stop-daemon for all functions. Add 'status' function.

papamoose:
Rely completely on 'start-stop-daemon' since it is already being
used to stop the daemon.

The intention here was to make the init script work reliably with
automation and ensure idempotency.


slowfranklin:
Merged, thanks for contributing.

PR #55, opened 2016-04-27
Tracker fix2

hat001:


andy-js:
First changeset:

You need to check in "$ac_cv_tracker_install_prefix"/bin for tracker/tracker-contol.

Second changeset

I think it might make more sense to change to simply change the description of the --with-tracker-prefix option "Tracker prefix (default: none)"

hat001:
Go to
https://github.com/Netatalk/Netatalk/pull/58

Issue #56, opened 2016-04-29
[PATCH] Spotlight searches can cause afpd to segfault


andy-js:
Seems like a NULL pointer is being passed to strcmp.

<pre>
diff --git a/etc/spotlight/sparql_parser.y b/etc/spotlight/sparql_parser.y
index 9d60997..7128a01 100644
--- a/etc/spotlight/sparql_parser.y
+++ b/etc/spotlight/sparql_parser.y
@@ -119,6 +119,8 @@ BOOL {
         yyerror("Spotlight queries with logic expressions are disabled");
         YYABORT;
     }
+    if ($1 == NULL || $3 == NULL)
+        YYABORT;
     if (strcmp($1, $3) != 0)
         $$ = talloc_asprintf(ssp_slq, "{ %s } UNION { %s }", $1, $3);
     else
</pre>

rdmark:
@andy-js Is there a reason this patch wasn't submitted as a PR? Still a work in progress?

rdmark:
Merged to main

PR #57, opened 2016-04-29
Spotlight searches can cause afpd to segfault #56

andy-js:


etc/spotlight/sparql_parser.y
```diff
@@ -119,6 +119,8 @@ BOOL {
         yyerror("Spotlight queries with logic expressions are disabled");
         YYABORT;
     }
+    if ($1 == NULL || $3 == NULL)
```
slowfranklin: Can we add a yyerror() here?

etc/spotlight/sparql_parser.y
```diff
@@ -119,6 +119,8 @@ BOOL {
...
+        YYABORT;
```
slowfranklin: Looks like we should employ the same check above for the expr AND expr case?

andy-js: I suppose so, though I didn't run into problems there.

etc/spotlight/sparql_parser.y
andy-js: Maybe.  I just copied what the code about did.

slowfranklin:
This got lost somehow. Do you still have the patches somewhere and can create a new merge request?

jonassmedegaard:
The patch is [here](https://salsa.debian.org/netatalk-team/netatalk/-/blob/debian/latest/debian/patches/100~20160429~b30a2ba.patch), in case that's of some use.

rdmark:
Merged to main in https://github.com/Netatalk/netatalk/pull/243

Issue #59, opened 2016-04-30
Configure fails if tracker isn't installed


andy-js:
Title says it all.

PR #60, opened 2016-04-30
Configure fails if tracker isn't installed #59

andy-js:


andy-js:
How about pulling this one in?

PR #77, opened 2016-12-23
Switch openrc init script to openrc-run

jlec:
Openrc has been switching from /sbin/runscript to /sbin/openrc-run
a long time ago. Currently there is a big push to curate all init scripts.

Signed-off-by: Justin Lecher <jlec@gentoo.org>


hat001:
Thanks.

PR #85, opened 2017-02-19
Fix manual include paths for Berkeley DB

manover:
configure parameter `--with-bdb=<manual path>` does not work like
it's supposed to, i.e. it is impossible to point configure to
the exact directory you want. It would always append the trailing
`/include` component even if it already exists in your path, so
for example, debian triplex multi-arch includes like
`/usr/include/x86_64-linux-gnu` are out of the question, the script
would turn it into `/usr/include/x86_64-linux-gnu/include`...
This pull request fixes this problem without breaking compatibility
with other use cases.

### DIFF
--- a/macros/db3-check.m4
+++ b/macros/db3-check.m4
@@ -86,7 +86,7 @@ AC_DEFUN([AC_NETATALK_PATH_BDB],[
 if test "x$bdb_required" = "xyes"; then
     trybdbdir=""
     dobdbsearch=yes
-    bdb_search_dirs="/usr/local /usr"
+    bdb_search_dirs="/usr/local/include /usr/include"
     search_subdirs="/ /db6.1 /db6 /db5 /db5.3 /db5.2 /db5.1 /db51 /db5.0 /db50 /db4.8 /db48 /db4.7 /db47 /db4.6 /db46 /db4"
 
     bdbfound=no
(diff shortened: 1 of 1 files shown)


slowfranklin:
Breaks the use case where BerkeleyDB was configured with a private --prefix, eg --prefix=/usr/local/db62.

[slow@kazak scratch]$ ls /usr/local/db62/
bin  docs  include  lib
[slow@kazak scratch]$ cat ../configure-netatalk
#!/bin/sh
CFLAGS='-O0 -g' \
./configure \
    --prefix=/opt/netatalk \
    --with-pkgconfdir=/opt/netatalk/etc \
    --with-uams-path=/opt/netatalk/lib \
    --with-pam-confdir=/etc/pam.d \
    --enable-maintainer-mode \
    --with-docbook=/home/slow/git/netatalk/docbook-xsl-1.78.0 \
    --with-tracker-pkgconfig-version=1.0 \
    --with-bdb=/usr/local/db62 \
    --enable-developer $@
[slow@kazak scratch]$ ../configure-netatalk
checking build system type... x86_64-unknown-linux-gnu
checking host system type... x86_64-unknown-linux-gnu
checking target system type... x86_64-unknown-linux-gnu
...
checking for Berkeley DB headers in /usr/local/db62/... no
checking for Berkeley DB headers in /usr/local/db62/db6.1... no
checking for Berkeley DB headers in /usr/local/db62/db6... no
checking for Berkeley DB headers in /usr/local/db62/db5... no
checking for Berkeley DB headers in /usr/local/db62/db5.3... no
checking for Berkeley DB headers in /usr/local/db62/db5.2... no
checking for Berkeley DB headers in /usr/local/db62/db5.1... no
checking for Berkeley DB headers in /usr/local/db62/db51... no
checking for Berkeley DB headers in /usr/local/db62/db5.0... no
checking for Berkeley DB headers in /usr/local/db62/db50... no
checking for Berkeley DB headers in /usr/local/db62/db4.8... no
checking for Berkeley DB headers in /usr/local/db62/db48... no
checking for Berkeley DB headers in /usr/local/db62/db4.7... no
checking for Berkeley DB headers in /usr/local/db62/db47... no
checking for Berkeley DB headers in /usr/local/db62/db4.6... no
checking for Berkeley DB headers in /usr/local/db62/db46... no
checking for Berkeley DB headers in /usr/local/db62/db4... no
configure: error: Berkeley DB library required but not found!

andy-js:
Most --with-whatever options I've come across work the way --with-bdb does.  I don't think changing it is an option.  If you need more granularity we should add some new options, such as --with-bdb-headers and --with-bdb-libs.

PR #88, opened 2017-03-29
Remove unused function 'evutil_secure_rng_add_bytes', fixing build on SmartOS.

mattconnolly:
This function was not used anywhere in netatalk. Additionally, it
was calling arc4random_addrandom() which is apparently notoriously
difficult to use correctly and has since been removed from OpenBSD
and SmartOS.

See: https://www.illumos.org/issues/5949


hat001:
Let's argue at the following place.
https://github.com/Netatalk/Netatalk/issues/89

slowfranklin:
Closing this as we're soon going to remove the bundled libevent.

PR #94, opened 2017-07-14
Fix nfs detection, allow to use on remote fuse systems

sapfeer0k:
Hi! I would like to use Netatalk with mounted fuse filesystem. By default, afpd detects it as a NFS share , if mount point has a colon in the name. But my partition is not an NFS share. So, I think it must be corrected.


slowfranklin:
Lgtm. Can I get a second team-member review?

PR #99, opened 2018-03-19
Branch netatalk 3 1 fork2

seiman:
The Analysis of the `fce_api.c` routines has show that during the creation and deletion of many small files a huge amount of udp events get lost during the communication with the udp server.
Although the udp server is written as performant as possible we still get a udp buffer overflow resulting in package loss.
Increasing the buffer size helps to avoid the problem but we still added a waiting parameter between sending events to reduce the risk of event loss on low memory server.

### DIFF
--- a/etc/afpd/fce_api.c
+++ b/etc/afpd/fce_api.c
@@ -36,6 +36,7 @@
 #include <arpa/inet.h>
 #include <netdb.h>
 #include <stdbool.h>
+#include <unistd.h>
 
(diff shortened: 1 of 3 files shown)


slowfranklin:
Patch doesn't apply on current master. Please resubmit in a new MR. Sorry for leaving the lingering for so long.

rdmark:
This changeset has been rebased and filed with https://github.com/Netatalk/netatalk/pull/849

Issue #67, opened 2016-08-09
Bootstrap has warnings


KalleDK:
Seems like the bootstrap script could need an update

```
macros/netatalk.m4:4: warning: underquoted definition of AX_CHECK_DOCBOOK
```

```
test/afpd/Makefile.am:12: warning: source file '$(top_srcdir)/etc/afpd/afp_config.c' is in a subdirectory,
test/afpd/Makefile.am:12: but option 'subdir-objects' is disabled
```

```
libatalk/dsi/Makefile.am:3: warning: 'INCLUDES' is the old name for 'AM_CPPFLAGS' (or '*_CPPFLAGS')
```

```
automake: warning: autoconf input should be named 'configure.ac', not 'configure.in'
configure.in:14: warning: AM_INIT_AUTOMAKE: two- and three-arguments forms are deprecated.  
```

Full run

```
root@633f46b695fc:/usr/local/src/netatalk-code# ./bootstrap 
+ rm -rf autom4te*.cache
+ LIBTOOLIZE=libtoolize
+ uname
+ SYSNAME=Linux
+ [ xLinux = xDarwin ]
+ aclocal -I macros
macros/netatalk.m4:4: warning: underquoted definition of AX_CHECK_DOCBOOK
macros/netatalk.m4:4:   run info Automake 'Extending aclocal'
macros/netatalk.m4:4:   or see http://www.gnu.org/software/automake/manual/automake.html#Extending-aclocal
+ autoheader
+ libtoolize --force --copy
libtoolize: putting auxiliary files in `.'.
libtoolize: copying file `./ltmain.sh'
libtoolize: putting macros in `macros'.
libtoolize: copying file `macros/libtool.m4'
libtoolize: copying file `macros/ltoptions.m4'
libtoolize: copying file `macros/ltsugar.m4'
libtoolize: copying file `macros/ltversion.m4'
libtoolize: copying file `macros/lt~obsolete.m4'
libtoolize: Consider adding `AC_CONFIG_MACRO_DIR([m4])' to configure.ac and
libtoolize: rerunning libtoolize, to keep the correct libtool macros in-tree.
+ automake --include-deps --add-missing --foreign --copy
configure.ac:9: warning: AM_INIT_AUTOMAKE: two- and three-arguments forms are deprecated.  For more info, see:
configure.ac:9: http://www.gnu.org/software/automake/manual/automake.html#Modernize-AM_005fINIT_005fAUTOMAKE-invocation
[...]

PR #101, opened 2018-04-26
Autotools cleanup

rubenk:
Cleanup some warnings while running autoreconf and remove / replace some stuff which has long been deprecated.


andy-js:
Looks pretty good to me.

rubenk:
This fixes most of the warnings from https://github.com/Netatalk/Netatalk/issues/67.
I haven't looked at the subdir-objects warning yet, but I plan to.

slowfranklin:
Pushed to master after dropping the libevent changes. libevent is going to get dropped soon see MR #132.

Thanks!

PR #106, opened 2018-05-13
docs: manual: Remove wrong TCP-over-TCP info; minor copy editing

mfwitten:
This commit is intended to apply cleanly as a **fast&#x2011;forward&nbsp;merge**. The reason for this is that a future pull request will merge this commit's changes into `branch-netatalk-3-1`, so as to record the nature of the development and to avoid the duplication of cherry-picking.

Of course, if this is not possible, that future pull request can be re-configured.

---------------

The main reason for this commit is to remove this text:

> Tunneling TCP over TCP (as SSH does) is not the best idea. There exist better solutions like VPNs based on the IP layer.

It's true that TCP-over-TCP is a bad idea, because it can lead to TCP "meltdown"; however, it is *not* true that tunneling via SSH implies TCP&#x2011;over&#x2011;TCP. Indeed, because a VPN is a much more general solution to tunneling, it is likely to perform worse than SSH for the simple case of connecting to a single, specific AFP&nbsp;Server.

Consider the following port forwarding (using OpenSSH):

&nbsp;&nbsp;`ssh -N -L localhost:1234:localhost:548 example.com`

When an AFP Client communicates with an AFP&nbsp;Server through this tunnel, the data they exchange is encapsulated in TCP and decapsulated from TCP such that TCP is never encapsulated by TCP; that is, TCP&#x2011;over&#x2011;TCP *never* occurs (view the following diagram with a monospace font and with at least 93 coloumns of text):

                       localhost:1234           example.com:22           localhost:548
                             |                        |                        |
                             |                        |                        |
                             V   encrypt(AFP)--+      V                        V
                                               |
    +--------------+         +--------------+  V      +--------------+         +--------------+
    |          AFP | AFP/TCP |     AFP      | ???/TCP |     AFP      | AFP/TCP | AFP          |
    |   data <=====o=========o==============o=========o==============o=========o=====> data   |
    |              |         |              |         |              |         |              |
    |  AFP Client  |         |  SSH Client  |         |  SSH Server  |         |  AFP Server  |
    +--------------+         +--------------+         +--------------+         +--------------+

Anyway, while I was in the source rummaging around, I took the opportunity to perform some copy editing, including a few mild additions and clarifications.

### DIFF
--- a/doc/manual/configuration.xml
+++ b/doc/manual/configuration.xml
@@ -1091,9 +1091,9 @@ uams_dhx2.so -&gt; uams_dhx2_pam.so</programlisting> then you're using PAM,
           forwarded to the remote server's AFP port (we used the default 548)
           over SSH.</para>
 
-          <para>These sorts of tunnels are an ideal solution if you've to
-          access an AFP server providing weak authentications mechanisms
-          through the Internet without having the ability to use a "real" VPN.
+          <para>This sort of tunnel is an ideal solution if you must access
+          an AFP server through the Internet without having the ability
+          or desire to use a "real" VPN.
           Note that you can let <command>ssh</command> compress the data by
           using its "-C" switch and that the tunnel endpoints can be different
           from both AFP client and server (compare with the SSH documentation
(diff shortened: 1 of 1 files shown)


slowfranklin:
Pushed to master. Thanks!

PR #113, opened 2018-12-12
make afpstats python 3 compatible

knight-of-ni:
Fedora is deprecating Python 2, requiring us to update any packages containing Python 2 scripts to Python 3. This simple PR appears all that is necessary to make the afpstats script compatible with Python 3.

It fixes this error:
```
$afpstats
  File "/usr/bin/afpstats", line 26
    print name
             ^
SyntaxError: Missing parentheses in call to 'print'. Did you mean print(name)?
```


slowfranklin:
Pushed to master. Thanks!

PR #116, opened 2019-01-17
afpd: Fix bogon client name.

michael-skynorth:
Not sure why I get bogon as hostname, replace it with ip address


slowfranklin:
Hm, this looks more like a local DNS resolving issue?

PR #120, opened 2019-02-27
@seiman seiman NETATALK: Added Ignore Directories Feature

seiman:
We currently have the problem that we receive multiple events of a hidden directory, which we do not want to receive.
Since the filename contains an Uuid like pattern, the current ignore files feature does not suffice.
To guarantee that the runtime does not increase too much (like using regular expression on the full filename) we implemented the specification of comma separated directory names, separated by ",".

Could you please check and accept our pull request!


slowfranklin:
Generally looks good, but the manpage changes belong to `doc/manpages/man5/afp.conf.5.xml`. Can you update the patchset accordingly?

rdmark:
@slowfranklin The current state of the patchset includes the updated ```doc/manpages/man5/afp.conf.5.xml``` as per your request. This should be good to be merged?

slowfranklin:
Pushed, thanks!

PR #127, opened 2020-03-30
Fix garbage read in bsd_attr_list

Jose123456:
Calling memmove(3) for one byte more than what was retrieved by
extattr_list_link(2) causes clang on FreeBSD to complain about a heap
buffer overflow when we try to read that byte at line 362.

Found with Clang's address sanitizer. Report attached:
[sanitize-report2.txt](https://github.com/Netatalk/Netatalk/files/4406008/sanitize-report2.txt)


slowfranklin:
Pushed to master. Thanks for contributing!

PR #128, opened 2020-03-31
Fix use after free in get_tm_used()

Jose123456:
If get_tm_used encounters a directory with a name ending in
"sparsebunlde", and the logged-in user does not have execute permission
on that directory, we destroy the infoplist bstring we created, and
move on to the next entry. Unfortunately, we do not set infoplist to
NULL, and trying to bdestroy infoplist at cleanup time causes an
attempted read of bstring->slen in a region that was freed.

Found with Clang's address sanitizer.
[sanitize-report.txt](https://github.com/Netatalk/Netatalk/files/4406013/sanitize-report.txt)


slowfranklin:
Pushed to master. Thanks for contributing!

PR #131, opened 2021-10-11
Bug and compile time fixes for openssl 1.1

rdmark:
- Fix for atalkd bug "setifaddr: eth0 (1-65534): Invalid argument. try specifying a smaller net range. "
- Fix for compile time errors when using libssl 1.1

Credits goes to the authors of the patches referenced in the commit messages

### DIFF
--- a/etc/atalkd/main.c
+++ b/etc/atalkd/main.c
@@ -296,10 +296,10 @@ static void as_timer(int sig _U_)
 		    LOG(log_info, logtype_atalkd, "config for no router" );
 		      
 		    if ( iface->i_flags & IFACE_PHASE2 ) {
-			iface->i_rt->rt_firstnet = htons(1);
+			iface->i_rt->rt_firstnet = 0;
 			iface->i_rt->rt_lastnet = htons( STARTUP_LASTNET );
 			setaddr( iface, IFACE_PHASE2, iface->i_addr.sat_addr.s_net, iface->i_addr.sat_addr.s_node,
-				htons(1), htons( STARTUP_LASTNET ));
+				0, htons( STARTUP_LASTNET ));
 		    }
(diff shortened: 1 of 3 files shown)


slowfranklin:
Thanks for contributing! It seems the patch is for Netatalk 2.2 which is EOL. Eg atalkd has been removed in newer releases.  The second patch is already part of master in a similar form. Not sure about the third patch, but it doesn't apply.

rdmark:
@slowfranklin Thanks for following up on this PR. Yes, you are correct that the patches are for Netatalk 2.2 in particular. Is there any chance you could revisit the decision to EOL 2.2? There is an ever increasing userbase of classic Macs out there now, with the boom in retro computing and all, and the deprecation of AppleTalk/DDP in 3.x you effectively cut off support for anything prior to Mac OS 8 (although 7.5 and 7.6 have upgrade paths by installing OpenTransport).

For context, I'm a contributor to the [RaSCSI project](https://github.com/akuker/RASCSI), a device that emulates various SCSI devices for vintages Macs and other computers that support SCSI. We've introduced close integration with Netatalk in the latest stable release, for running an AppleShare server in parallel with emulated SCSI network adapters and hard drives. We've been relying on local patches and unofficial forks of Netatalk to make this work, and I'd love to be able to pivot to official Netatalk releases instead. However, Netatalk 2.2 just doesn't compile anymore with modern Linux toolchains.

slowfranklin:
People find interesting ways to torture themselves. :) We can of course integrate patches if that helps.

slowfranklin:
Pushed to 2.2 branch. Thanks for contributing!

rdmark:
@slowfranklin Thanks for taking a second look at this! I appreciate it.

PR #136, opened 2021-11-23
[2.2] Fix 'multiple definition' compile time errors

rdmark:
gcc 10 on Linux throws "multiple definition" errors; declares two variables as external to avoid this.


slowfranklin:
Lgtm, thanks!

PR #137, opened 2021-11-23
Resolve gcc 10 compile time warnings on Linux

rdmark:


etc/afpd/file.c
```diff
@@ -2243,7 +2243,7 @@ int afp_exchangefiles(AFPObj *obj, char *ibuf, size_t ibuflen _U_, char *rbuf _U
      * NOTE: the temp file will be in the dest file's directory. it
      * will also be inaccessible from AFP. */
     memcpy(temp, APPLETEMP, sizeof(APPLETEMP));
-    if (!mktemp(temp)) {
+    if ( (mkstemp(temp)) == -1) {
         err = AFPERR_MISC;
```
slowfranklin, line 2247: This leaks a file descriptor.

etc/afpd/volume.c
```diff
@@ -2226,7 +2226,7 @@ static void check_ea_sys_support(struct vol *vol)
             return;
         }
 
-        mktemp(eaname);
+        mkstemp(eaname);
```
slowfranklin, line 2230: Same here.

slowfranklin (changes requested):

etc/afpd/file.c
rdmark: Good catch! See if this change resolves the issue properly (I basically looked at the 3.1 branch and copied what was done there.)

etc/afpd/volume.c
rdmark: Code updated.

rdmark:
@slowfranklin What's your guidance with regards to a PR like this where incremental commits were made following code review? Do you want a fresh new PR that meets the "chain of pearls" standard?

slowfranklin:
Each MR should only contain the relevant patches on-top of the target branch that are a genuine part of the MR, but no patches from other MRs -- if that's what you're asking,

rdmark:
@slowfranklin No, I meant rather this concept:

> The end result of a MR is like [a string of pearls](https://upload.wikimedia.org/wikipedia/commons/thumb/c/cd/Spine_Of_The_Centipede_using_Swarvoski_Pearls.jpg/698px-Spine_Of_The_Centipede_using_Swarvoski_Pearls.jpg). Each pearl (commit) individually polished, then assembled as a string. The quality of the end result is almost entirely due to the quality of each element.

So, earlier commits contained issues and are no longer individually polished per se.

rdmark:
Oh, I meant, say, [3ec6798](https://github.com/Netatalk/Netatalk/pull/137/commits/3ec67982d0b351954acf26a71fa5259752102563) which had a file descriptor leak bug that you pointed out. As such, that commit arguably isn't a "pearl" suited to be kept in the patchset, is it?

slowfranklin:
Please squash 7f5711aebba1c91ad10706266b3ebc3bd381d003 into a5a5da946e01054cedbde084e34ca369c8cf401b.

rdmark:
@slowfranklin Created a clean PR in https://github.com/Netatalk/Netatalk/pull/165

PR #138, opened 2021-12-08
Handle special FIRSTNET behavior in NetBSD

rdmark:
With PR https://github.com/Netatalk/Netatalk/pull/131 merged, compatibility with NetBSD that was originally fixed in https://github.com/Netatalk/Netatalk/commit/bbc0d89c056be3a2b5efa6bf5428f7ba8f07a1d0 is again broken.

This captures this special case inside an ifndef macro.


rdmark:
Deprecated by cleaner PR https://github.com/Netatalk/Netatalk/pull/154

PR #140, opened 2022-01-02
[2.2] systemd unit configurations for atalkd et.al.

rdmark:
This enables you to run atalkd and the other legacy daemons as systemd services.

It leverages the pre-existing "--enable-systemd" configuration option. It will generate systemd unit configurations for each of the 6 daemons, but only enable the service for those that were compiled.


rdmark:
@slowfranklin I've made the assumption here that papd, timelord, and a2boot are inherently forking daemons. Did not see any sign of the opposite.

rdmark:
Clean PR in https://github.com/Netatalk/Netatalk/pull/157

PR #141, opened 2022-01-02
[2.2] Improve compatibility with GS/OS and LocalTalk bridge hardware

rdmark:
```
        # Patch the source so file dates are preserved during a GS/OS folder copy,
        #   and the AsanteTalk bridge consistently starts up in AppleTalk Phase 2
        #   and the Dayna bridge doesn't crash GS/OS
        # props to Steven Hirsch for these
        sed -i ':a;N;$!ba;s/case FILPBIT_ATTR :\n *change_mdate = 1;\n/case FILPBIT_ATTR :\n/g' etc/afpd/file.c
        sed -i 's/rtmp->rt_iface == iface/rtmp->rt_iface != iface/g' etc/atalkd/main.c
```

Source: http://appleii.ivanx.com/a2server/scripts/a2server-3-sharing.txt

### DIFF
--- a/etc/afpd/file.c
+++ b/etc/afpd/file.c
@@ -895,7 +895,6 @@ int setfilparams(struct vol *vol,
         }
         switch(  bit ) {
         case FILPBIT_ATTR :
-            change_mdate = 1;
             memcpy(&ashort, buf, sizeof( ashort ));
             buf += sizeof( ashort );
             break;
--- a/etc/afpd/file.c
+++ b/etc/afpd/file.c
@@ -895,6 +895,9 @@ int setfilparams(struct vol *vol,
         }
         switch(  bit ) {
         case FILPBIT_ATTR :
+            /* Preserve file dates after a GS/OS folder copy
+             *
+             * change_mdate = 1; */
             memcpy(&ashort, buf, sizeof( ashort ));
             buf += sizeof( ashort );
             break;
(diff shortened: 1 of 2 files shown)


etc/afpd/file.c
```diff
@@ -895,7 +895,6 @@ int setfilparams(struct vol *vol,
         }
         switch(  bit ) {
         case FILPBIT_ATTR :
-            change_mdate = 1;
             memcpy(&ashort, buf, sizeof( ashort ));
```
slowfranklin, line 901: Are you sure this is correct? IIrc this breaks AFP 3 conformance.

etc/afpd/file.c
```diff
@@ -895,6 +895,9 @@ int setfilparams(struct vol *vol,
         }
         switch(  bit ) {
         case FILPBIT_ATTR :
+            /* Preserve file dates after a GS/OS folder copy
+             *
+             * change_mdate = 1; */
             memcpy(&ashort, buf, sizeof( ashort ));
```
slowfranklin, line 901: Both commits should be squashed.

etc/afpd/file.c
rdmark: I see, that's good insight. So there's a conflict between AFP3 specs and GS/OS expectations. I think it makes the most sense to discard this patch since the issue it's trying to fix for GS/OS is trivial.

etc/afpd/file.c
rdmark: Alright, let me put together a new branch that only takes in the AsanteTalk fix.

rdmark:
@slowfranklin New PR in https://github.com/Netatalk/Netatalk/pull/166

PR #146, opened 2022-01-09
General stability patches from netatalk-classic etc.

rdmark:
Patches from the 20200806 release tag: https://github.com/christopherkobayashi/netatalk-classic/releases/tag/20200806

### DIFF
--- a/etc/afpd/status.c
+++ b/etc/afpd/status.c
@@ -671,7 +671,11 @@ server_signature_auto:
                     goto server_signature_random;
                 } else {
                     fseek(fp, -1L, SEEK_END);
-                    if(fgetc(fp) != '\n') fputc('\n', fp); /* last char is \n? */
+                    if(fgetc(fp) != '\n')
+                    {
+                        fflush(fp);
+                        fputc('\n', fp); /* last char is \n? */
+                    }
                     goto server_signature_random;
                 }                    
             } else {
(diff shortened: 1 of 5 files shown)


etc/afpd/status.c
```diff
@@ -671,7 +671,11 @@ void set_signature(struct afp_options *options) {
                     goto server_signature_random;
                 } else {
                     fseek(fp, -1L, SEEK_END);
-                    if(fgetc(fp) != '\n') fputc('\n', fp); /* last char is \n? */
+                    if(fgetc(fp) != '\n')
+                    {
+                        fflush(fp);
+                        fputc('\n', fp); /* last char is \n? */
```
slowfranklin, line 677: Why? This doesn't seem to make sense.

rdmark: As I understand it, this is in order to flush the bidirectional i/o stream when switching direction.

rdmark: I had a chat with Chris. We would suggest using fseek here instead for C11 conformance. Please me know if this looks good to you now. I'll make a clean PR.

slowfranklin: Ah, that one, now I remember. So either please squash the two commits or drop the last one.

rdmark: New PR in https://github.com/Netatalk/Netatalk/pull/168

PR #147, opened 2022-01-09
[2.2] Make timelord work on non-big-endian systems, and sync localtime

rdmark:
- Patch for non-big-endian systems by cheesestraws: https://68kmla.org/bb/index.php?threads/doctorin-the-tardis-and-timelord-the-revived-appletalk-time-server.38553/
- Patch for localtime also by cheesestraws
- Introduces the '-l' command line option to sync with the server's localtime. Default behavior is still GMT.
- Adds a proper man page template, while documenting the new option


cheesestraws (approved):
My only thought is that logging the time zone per request even though that configuration is done once for the whole process seems like it'd result in wasteful log noise.  This probably doesn't matter but it feels kind of wasteful of the log-reader's attention to me?  A tiny thing, and not worth arguing about, at any rate: if you like it like this leave it like this.

cheesestraws:
(Hang on, how can I approve this for merging into _someone else's_ repository?)  That seems a bit odd.

rdmark:
@cheesestraws Thanks for reviewing! After sleeping on it, I agree with your commend on logging. I pushed a small update that logs time zone info only once.

cheesestraws:
Glad I could help! :-)

slowfranklin:
I haven't started formal review yet, just wanted to point out that in case you have patches in a patchset that modify changes done by earlier ones, then please squash and consolidate patches into smaller units. The Samba wiki has some nice documentation on the subject: https://wiki.samba.org/index.php/Creating_a_Samba_patch_series

slowfranklin:
This is more about patches *within* one patchset of an MR!

rdmark:
@slowfranklin Ah, I understand your point now. There are definitely patchsets here that were created incrementally, as feedback and testing revealed additional improvement areas over the last 3 months.

Just to understand the end goal here: What you want to see is a clean and appropriately verbose commit log in a PR, that can be merged as-is without further squashing?

slowfranklin:
Yup, that would be the goal. :)

rdmark:
@slowfranklin For starters, I redid this timelord PR, distilling the patchset down to four logical increments that each stand on their own. Can you please have a look at this and let me know if I'm on the right track? https://github.com/Netatalk/Netatalk/pull/151

PR #149, opened 2022-01-09
[2.2] make papd work with Apple LaserWriter drivers on Mac and GS/OS

rdmark:
Downstream patches from the A2SERVER project: https://github.com/NJRoadfan/a2server/blob/ivanx-1.5.x/files/netatalkpatches.tar.gz


rdmark:
@NJRoadfan Thanks for the breakdown of the fixes!
Does this also cover what the patch for print_cups.c does?

PR #159, opened 2022-01-18
[2.2] Enable all Netatalk daemons to be run as systemd services

rdmark:
Enable all Netatalk daemons to be run as systemd services using the existing --enable-systemd config flag

- Converted the experimental unit configurations for afpd and cnid into templates
- Renamed netatalk.service to afpd.service to follow the daemon to service naming convention
- Made the systemd templates sensitive to SBINDIR and NETATALK_VERSION autoconf variables
- Created new systemd templates for atalkd, papd, timelord, and a2boot
- Introduced a USE_CUPS autoconf variable to use in the Makefile
- Install and enable systemd services only for daemons that are compiled
- Updated READMEs for systemd and timelord
- Removed deprecated initscript templates


rdmark:
Closing for now to clean up and improve the patchset.

PR #169, opened 2022-03-01
NEWS changelog for 2.2.6 and 2.2.7

rdmark:


rdmark:
@slowfranklin I took the liberty of drafting a changelog, with the assumption that the 2.2.x versioning scheme will continue.
Also, I couldn't find any record online of a full changelog for 2.2.6, so I put together a highly summarized one based on the commit log leading up to that release.

slowfranklin:
Pushed, thanks!

PR #170, opened 2022-03-02
[2.2] Preserve file dates during copying in GS/OS, in an AFP3 compliant way

rdmark:
AFP2 and earlier did not expect file date stamps to get updated when copied. This patch makes this line conditional of AFP3.

See discussion in https://68kmla.org/bb/index.php?threads/yet-another-netatalk-2-2-fork.39889/post-435799


slowfranklin:
Pushed, thanks!

PR #171, opened 2022-03-23
[2.2] .gitignore cleanup

rdmark:
- Remove .gitignore from all .gitignore files. They were added as a temporary measure during the cvs to git migration. Right now, they're causing issues with working with .gitignore files in git.
- Add global patterns for common temporary files to the root .gitignore.


slowfranklin:
Pushed, thanks!

PR #172, opened 2022-03-27
Fix setting of LD_LIBRARY_FLAGS ($shlibpath_var).

ffontaine:
`LD_LIBRARY_PATH` should not be set when cross-compiling, because it adds the cross-libraries to the build's LD-path.

Also the restoring of `LD_LIBRARY_PATH` was done incorrectly: it would set `LD_LIBRARY_PATH=LD_LIBRARY_PATH`.

Signed-off-by: Arnout Vandecappelle (Essensium/Mind) <arnout@mind.be>
[Retrieved from:
https://git.buildroot.net/buildroot/tree/package/netatalk/0001-Fix-setting-of-LD_LIBRARY_FLAGS-shlibpath_var.patch]
Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>

### DIFF
--- a/macros/db3-check.m4
+++ b/macros/db3-check.m4
@@ -94,7 +94,7 @@ if test "x$bdb_required" = "xyes"; then
     savedldflags="$LDFLAGS"
     savedcppflags="$CPPFLAGS"
     savedlibs="$LIBS"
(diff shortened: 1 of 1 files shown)


slowfranklin:
Pushed, thanks!

PR #176, opened 2022-04-14
Fix some compile errors

JensKSP:
This fixes some compile issues that happen if some obsolete features are being turned on

### DIFF
--- a/contrib/a2boot/a2boot.c
+++ b/contrib/a2boot/a2boot.c
@@ -27,6 +27,7 @@
 #include <sys/types.h>
 #include <sys/file.h>
 #include <sys/uio.h>
+#include <sys/ioctl.h>
 
 #include <stdlib.h>
 #include <unistd.h>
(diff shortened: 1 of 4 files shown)


rdmark:
@JensKSP May I ask which compiler and architecture you're using to compiler? I've been working in the Netatalk 2.2 codebase for some time and not run into these particular errors, so was just curious what the difference with your environment might be.

JensKSP:
I guess you found the Dockerfile already?
It's Alpine Linux on x86_64.
I am running Netatalk on my Synology NAS to share data with the classic Macs of my collection.

rdmark:
@JensKSP What version of gcc does Alpine Linux come with? ```gcc --version```

JensKSP:
It uses gcc (Alpine 10.3.1_git20211027) 10.3.1 20211027 but more importantly musl libc.

slowfranklin (approved):

Issue #114, opened 2018-12-15
Unicode 11.0.0


hat001:
libatalk shoud support Unicode 11.0.0 .

hat001:
patch is here.
[0001-libatalk-Update-Unicode-support-to-version-11.0.0.patch.txt](https://github.com/Netatalk/Netatalk/files/2682084/0001-libatalk-Update-Unicode-support-to-version-11.0.0.patch.txt)

rdmark:
@hat001 It seems Unicode 14.0 is available at this point. Would it make sense to revise this for the latest standard?

rdmark:
I put up a PR for Unicode 14 in master: https://github.com/Netatalk/Netatalk/pull/177

rdmark:
This ticket can be closed now. Unicode 14 updates have been merged to master, 3.1 and 2.2 branches.

PR #177, opened 2022-04-18
Unicode 14 update

rdmark:
Updated the generated code based on the latest stable data from Unicode:

https://www.unicode.org/Public/UNIDATA/UnicodeData.txt

Issue ticket: https://github.com/Netatalk/Netatalk/issues/114


slowfranklin:
Pushed, thanks!

PR #174, opened 2022-04-11
add handling for cases where ad_entry() returns NULL

anodos325:
With recent CVE fixes, ad_enty() may now return NULL. This
commit adds basic error handling for these cases and asserting
where such a return is totally unexpected. In case of
ad_getid() and ad_forcegetid(), return CNID_INVALID rather
than 0 to clarify for future people investigating this that
a 0 here is an indication of error.

In case of new_ad_header(), the valid_data_len of the
adouble data may still be zero. This causes subsequent
ad_entry() calls within new_ad_header() to fail. As such,
overwrite valid_data-Len with AD_DATASZ2 or AD_DATASZ_EA
depending on how adouble data is stored on disk.

Another side-effect of the fix is that ad_entry() for
ADEID_DID on ea-backed adouble data will return NULL. In
this case, add explicit check for the backend before
attempting to get the DID.

Signed-off-by: Andrew Walker <awalker@ixsystems.com>

PR #178, opened 2022-06-08
Fix CVE-2022-23121, CVE-2022-23123 regression

Synology-andychen:
- Added guard check before access ad_entry()
- Allow zero length entry, for AppleDouble specification

### DIFF
--- a/etc/afpd/desktop.c
+++ b/etc/afpd/desktop.c
@@ -859,7 +859,7 @@ static int ad_addcomment(const AFPObj *obj, struct vol *vol, struct path *path, 
         return( AFPERR_ACCESS );
     }
 
-    if (ad_getentryoff(adp, ADEID_COMMENT)) {
+    if (ad_getentryoff(adp, ADEID_COMMENT) && ad_entry(adp, ADEID_COMMENT)) {
         if ( (ad_get_MD_flags( adp ) & O_CREAT) ) {
             if ( *path->m_name == '\0' ) {
                 name = (char *)curdir->d_m_name->data;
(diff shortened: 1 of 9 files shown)


rdmark:
@Synology-andychen Have you seen this PR: https://github.com/Netatalk/Netatalk/pull/174 -- it has code for fixing the same regressions in a somewhat different way. You may want to evaluate that code and see if there's overlap with yours or not.

Synology-andychen:
- Fix merge conflicts, some `AFP_ASSERT()` removed

slowfranklin:
Thanks for contributing. Question: is this still needed?

rdmark:
Discarding this in favor of the patch for the same regression that was already merged.

@Synology-andychen if you disagree, please reopen and let us know what outstanding issues this PR addresses.

rdmark:
It turns out that this PR addresses at least two issues that users are facing right now: https://github.com/Netatalk/netatalk/issues/357 and https://github.com/Netatalk/netatalk/issues/368

I will go ahead and reopen & merge it.

rdmark:
Validated that an entry length of 0 is allowed as per the AppleDouble spec: https://nulib.com/library/AppleSingle_AppleDouble.pdf

rdmark:
Interestingly, the one case of a zero length entry in the wild that has been empirically proven, are for the four private netatalk entries that are technically not covered by the AD spec.

southbay82:
What is the frequency new builds are created that will have this fix?

rdmark:
@southbay82 This fix alone is enough to justify a new netatalk release, so I've started planning it now. The netatalk release will be a source code release, mind you. The binary build updates will come later when the package managers for each OS catch up.

What OS or Linux distribution are you running netatalk on?

PR #184, opened 2023-01-13
[2.2] Use daemon icon as the default for mounted volumes

rdmark:
This aligns 2.2 with the default icon in 3.x


rdmark:
netatalk-icons

Just for fun, these are the 7 icons encoded in `icons.h` -- all 32x32 1 bit with a 1 bit mask.

slowfranklin:
Thanks!

PR #187, opened 2023-01-27
fixes for misc compiler issues

nekopsykose:
correct main definition is required as otherwise the test fails with implicit-int

include is required for:

```
netatalk_conf.c:84:32: error: 'LOCK_EX' undeclared (first use in this function)
   84 |     result = flock(fileno(fp), LOCK_EX);
```

which comes from sys/file.h: https://linux.die.net/man/2/flock

### DIFF
--- a/libatalk/util/netatalk_conf.c
+++ b/libatalk/util/netatalk_conf.c
@@ -24,6 +24,7 @@
 #include <utime.h>
 #include <errno.h>
 #include <string.h>
+#include <sys/file.h>
 #include <sys/param.h>
 #include <sys/socket.h>
 #include <netinet/in.h>
--- a/macros/largefile-check.m4
+++ b/macros/largefile-check.m4
@@ -60,7 +60,7 @@ if test "$enable_largefile" != no; then
     AC_TRY_RUN([#include <stdio.h>
 #include <stdlib.h>
 #include <sys/stat.h>
-main() { exit((sizeof(off_t) == 8) ? 0 : 1); }],
+int main() { exit((sizeof(off_t) == 8) ? 0 : 1); }],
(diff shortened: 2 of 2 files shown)


rdmark:
@nekopsykose Which compiler are these fixes for?

nekopsykose:
clang 16 and gcc 14 do not permit implicit function declarations and so e.g. the configure test will always unconditionally fail

nekopsykose:
er, apologies, that one is implicit-int, not implicit-function-decl

rdmark:
Aha, I'm using the ancient gcc 10 which comes standard with Raspbian Bullseye, which I why I didn't encounter the issues.

nekopsykose:
to be fair, neither of those are even released yet. closest would be clang16 (-rc1 currently), it was a reverted change in 15 (see https://discourse.llvm.org/t/configure-script-breakage-with-the-new-werror-implicit-function-declaration/65213)

this is just an ahead of time fix i found via some `-Werror` flags to help forward-port things a bit. it shouldn't break anything existing, just keep things building in the future.

nekopsykose:
the main motivation for it (being invalid) is mainly that c23 forbids the k&r function decl (amongst other things), and so these are technically not legal declarations (even since c89 iirc!). but i'm a bit rusty, the above endless thread has plenty more info, i'm just a janitor here :-)

rdmark:
Right, I've observed in the past the ebb and flow of compiler strictness leading up to new gcc/clang releases. I agree that it makes sense to do a modicum of forward porting when the rcs start hitting.

If you don't mind, I'll steal your patch for my fork (https://github.com/rdmark/Netatalk-2.x) -- although only the large file macro fix is relevant for the 2.2 branch.

nekopsykose:
go for it :)

slowfranklin:
Thanks for contributing!

PR #190, opened 2023-02-05
[2.2] Provide MNTTYPE_NFS on OmniOS

rdmark:
This enables compilation on OmniOSce r151044k. Patch by Hauke Fath.


slowfranklin:
Thanks!

PR #192, opened 2023-02-06
[2.2] fix largefile-check macro for largefile with clang16

rdmark:
main() without a return type fails with Werror=implicit-int, which makes it fail to detect largefile support with clang 16

Backport of https://github.com/Netatalk/Netatalk/commit/a08e6137afaf8f6ec2139c250482c8515ceb3e2f


rdmark:
Merged this without code review since it's a trivial change, as well as a backport of a reviewed PR.

Issue #194, opened 2023-02-06
[2.2.7] timelord: tm_gmtoff not compatible with OmniOS


rdmark:
OmniOS, and probably all OpenSolaris descendants, do not support tm_gmtoff in time.h. Another more portable way to determine time zone adjusted time should be researched.

As per https://www.gnu.org/software/libc/manual/html_node/Broken_002ddown-Time.html it is a BSD extension and not available in strict ISO C.

rdmark:
I tested that the fix 1) leads to a compiled binary in OmniOS that executes and outputs debug info, and 2) that the newly introduced code path leads to a correct time passed to a tardis client.

The full e2e flow on Solarish has not been tested.

PR #195, opened 2023-02-07
[2.2] Fall back to timezone when tm_gmtoff is unavailable

rdmark:
On Solaris/OmniOS `tm_gmtoff` is not available in `struct tm` in time.h, unlike *BSD/Linux.
https://www.gnu.org/software/libc/manual/html_node/Broken_002ddown-Time.html

As a fallback, use `timezone` (also from time.h). The main drawback is that timezone does not take daylight savings into consideration.
https://www.gnu.org/software/libc/manual/html_node/Time-Zone-Functions.html

### DIFF
--- a/contrib/timelord/timelord.c
+++ b/contrib/timelord/timelord.c
@@ -243,7 +243,11 @@ int main( int ac, char **av )
             exit( 1 );
             }
 
-            mtime = tv.tv_sec + EPOCH + tm->tm_gmtoff;
+#if defined (HAVE_STRUCT_TM_GMTOFF)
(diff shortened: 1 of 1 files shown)


rdmark:
Addresses https://github.com/Netatalk/Netatalk/issues/194

Keeping in draft state since I haven't tested it properly yet. I haven't gotten the DDP kernel module to work with amd64 OmniOS yet, since the Makefile is hard coded for SPARC.

The code does compile again on OmniOS now so it's an improvement at least. :)

slowfranklin (approved):

PR #197, opened 2023-02-09
[2.2] Update contents and fix syntax errors in man pages

rdmark:
- Align syntax and format of timelord man page with the Netatalk standard
- Fix several syntax errors that caused `man -l` to throw warnings
- Update the description of `netatalk.conf` to reflect the changes made when introducing systemd unit configurations
- Correct the Netatalk version number
- Bump dates for updated man pages
- Delete two deprecated man pages


slowfranklin:
Hm.... shouldn't this be done via updating https://github.com/Netatalk/netatalk-doc and then importing the generated files?

rdmark:
Aha, that's how it works. I'll look into it.

rdmark:
I'm going to attempt to backport https://github.com/Netatalk/Netatalk/commit/530376ceef7d0d5e4187dbf7246bd217782d9e5a into 2.2 and then generate the manpages from there.

rdmark:
Discarding this in favor of https://github.com/Netatalk/Netatalk/pull/210

PR #200, opened 2023-02-11
[2.2] Disable ACL and quota by default; remove a few antiquated checks

rdmark:


slowfranklin:
Why disable ACL support by default?

rdmark:
I thought it is an adjacent feature with little practical use today (especially for 2.2 users which are entirely hobby related). It's not causing build failures though so I'm not in a hurry to make the change. I can update the PR and remove the ACL changes.

slowfranklin:
> I thought it is an adjacent feature with little practical use today (especially for 2.2 users which are entirely hobby related). It's not causing build failures though so I'm not in a hurry to make the change. I can update the PR and remove the ACL changes.

 I'm ok with disabling it by default in 2.2. Whatever you prefer. :)

rdmark:
For the purpose of keeping the scope narrow and solving the actual problem, let's focus on quota in this PR. Updated!

slowfranklin (approved):

PR #201, opened 2023-02-11
[2.2] Add missing license based on BSD origin.

rdmark:
Since it is derived from the BSD sources written by Elz, it is under the CA regents copyright and usual license. But we need the more formal lanuage included for it to be valid.

This addresses one of the remaining license FIXME issues: https://github.com/adiknoth/netatalk-debian/blob/master/debian/copyright#L59 https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=685878#56

Reviewed-by: Ralph Boehme <slow@samba.org>


rdmark:
Merged without code review, since this is a backport of a trivial changeset.

PR #205, opened 2023-02-13
Hotfix: Don't try to bootstrap doc/Makefile

rdmark:
Hotfix for breakage introduced with https://github.com/Netatalk/Netatalk/pull/203
`doc/Makefile` should not be bootstrapped in configure.ac anymore, since we have a static Makefile imported from netatalk-docs


rdmark:
Merged PR without code review, since it's a trivial change for an issue that was blocking bootstrapping altogether

PR #207, opened 2023-02-13
Fix a typo in debug message

m-emelchenkov:
While debugging one of Netatalk installs I scrolled through logs and noticed unknown term 'came', dug into source code and found out it was just a typo, so I fixed it.


rdmark:
Thanks for contributing the PR!

PR #211, opened 2023-02-15
Fix more typos

m-emelchenkov:
Four more debug string fixes of typos in "cname". Today I looked at code more deeper than 2 days ago trying to fix (understand) one thing so found that these typos were not only in 1 string, but in more places. Now I searched through all sources to make sure no more left.

PS: Sorry, I don't know for sure why all three commits are submitted in this PR, I can't find how to submit only 
a955dc7 commit. But, anyway, changes in this PR are only from this commit.

### DIFF
--- a/etc/afpd/directory.c
+++ b/etc/afpd/directory.c
@@ -1216,14 +1216,14 @@ struct path *cname(struct vol *vol, struct dir *dir, char **cpath)
                  * this will terminate clean in while (1) because len == 0,
                  * probably afp_createfile|dir
                  */
-                LOG(log_maxdebug, logtype_afpd, "came('%s'): {leave-cnode ENOENT (possile create request): '%s'}",
+                LOG(log_maxdebug, logtype_afpd, "cname('%s'): {leave-cnode ENOENT (possile create request): '%s'}",
                     cfrombstr(dir->d_fullpath), ret.u_name);
                 continue; /* 10 */
             }
 
             switch (ret.st.st_mode & S_IFMT) {
             case S_IFREG: /* 11 */
-                LOG(log_debug, logtype_afpd, "came('%s'): {file: '%s'}",
+                LOG(log_debug, logtype_afpd, "cname('%s'): {file: '%s'}",
                     cfrombstr(dir->d_fullpath), ret.u_name);
                 if (len > 0) {
                     /* it wasn't the last part, so we have a bogus path request */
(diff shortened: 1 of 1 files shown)


rdmark:
Yes it would be cleaner if you can squash these changes into a single commit. My strategy is to do a soft reset back to the commit *before* your first change, add all changed files again and make one new commit for the whole changeset. Rebase against the target branch if needed. Then force push the branch to this PR. Viola, clean commit log!

rdmark:
See also: https://github.com/Netatalk/Netatalk/wiki/Developer-Notes

m-emelchenkov:
@rdmark Thank you for your instructions. Could you please check it, now better?

etc/afpd/directory.c
```diff
@@ -1216,14 +1216,14 @@ struct path *cname(struct vol *vol, struct dir *dir, char **cpath)
                  * this will terminate clean in while (1) because len == 0,
                  * probably afp_createfile|dir
                  */
-                LOG(log_maxdebug, logtype_afpd, "came('%s'): {leave-cnode ENOENT (possile create request): '%s'}",
+                LOG(log_maxdebug, logtype_afpd, "cname('%s'): {leave-cnode ENOENT (possile create request): '%s'}",
```
rdmark, line 1219: For the record: There's actually one more typo on this line "possile create request" which I assume is supposed to say "possible".... which is still awkward English. Maybe the author meant to say "possible to create request"?

rdmark:
@m-emelchenkov It looks much cleaner now, thanks!

One more piece of advice for the development process: I suggest you create feature branches where you do your code changes. If you keep committing directly to the master branch in your fork, you will eventually run into messy merge conflicts that will make it hard for you to submit upstream patches. There's a section on this in the Developer FAQ that I linked above.

etc/afpd/directory.c
m-emelchenkov: Yes, you are right, there are a lot of typos in the code, not only in this place. I fixed only the one that seemed important to me for understanding debug output. Honestly, I even did not notice that the word "possible" was with a typo too :).

m-emelchenkov:
Thanks for the advice, @rdmark.

etc/afpd/directory.c
m-emelchenkov: I even found a Debian-specific patch in their distro, exclusively related to typos: https://sources.debian.org/patches/netatalk/3.1.14~ds-1/103_fix_typos.patch/
I was surprised that they started patching the source inside the distro instead of just committing it to upstream.

rdmark: Wow, good job tracking down that patch! I made a PR for upstream (while making a few more typo fixes that I have noticed) -- https://github.com/Netatalk/Netatalk/pull/218

PR #227, opened 2023-02-26
Remove release notes code since it's no longer used

rdmark:
- Removes the doc/www/ dir, which contains the scripts and templates that generates the traditional release notes
- Removes the two release notes actions from the parent Makefile
- Removes the css file from doc/gfx_and_css, and renames the dir to doc/gfx (the logo images are intact)


slowfranklin:
I wonder whether we may want to keep the logo in the tree? Don't remember if we have it in another secure and accessible place.

rdmark:
@slowfranklin I agree that we should keep the logo files in the tree (this PR doesn't remove them, but simply renames the dir they're stored in.) When I did some digging a few months ago, the only other public place where the logos are stored is the `netatalk-doc` repo.

Well, actually I converted the pdf version to svg and posted it to [Wikimedia Commons](https://commons.wikimedia.org/wiki/File:Netatalk_logo.svg) last month. So that's one place where it's backed up as well.

Issue #232, opened 2023-02-28
atalkd fails to start during boot on Raspberry Pi Zero W


rmcardle:
atalkd fails to start during boot on Raspberry Pi Zero W with the error, `atalkd: zero interfaces, exiting.` It retries 5 times but still fails.

It seems that it can't start until Wi-Fi has finished connecting. Changing `After=syslog.target network.target` to `After=syslog.target network-online.target` in `/usr/lib/systemd/system/atalkd.service` fixes the problem.

Running Raspberry Pi OS Lite released on February 21st 2023 based on Debian 11.6 Bullseye. Linux raspberrypi 5.15.84+ #1613 Thu Jan 5 11:58:09 GMT 2023 armv6l GNU/Linux

rmcardle:
I'm currently running branch-netatalk-2-2 at commit c1aac12cd51ad2f43d9b401c43ba99258680b5ff.

https://github.com/Netatalk/netatalk/commit/c1aac12cd51ad2f43d9b401c43ba99258680b5ff

rdmark:
@rmcardle Thanks for reporting, and finding the solution! I have observed the same on my Pi Zero Ws. PR here https://github.com/Netatalk/netatalk/pull/233

rdmark:
Merged to branch-netatalk-2-2

PR #233, opened 2023-03-01
[2.2] Improve systemd services dependencies

rdmark:
atalkd will fail to start unless a network connection is established. All other DDP services depend on atalkd to have started first, so make them do the same.

This is critical for when netatalk is running on a device that only has a wlan interface, such as an RPi Zero W.

Addresses https://github.com/Netatalk/netatalk/issues/232

Also pulls in the improvement from downstream https://sources.debian.org/src/netatalk/3.1.14~ds-1/debian/patches/104_modernize_systemd.patch/

### DIFF
--- a/distrib/initscripts/a2boot.service.tmpl
+++ b/distrib/initscripts/a2boot.service.tmpl
@@ -1,10 +1,8 @@
-# This is experimental service file.
-# See distrib/systemd/README
 # This file is part of Netatalk :NETATALK_VERSION:.
 
 [Unit]
 Description=Apple II boot daemon
-After=syslog.target network.target atalkd.service
+After=network-online.target atalkd.service
 Requires=atalkd.service
 
 [Service]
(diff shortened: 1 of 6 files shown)


rdmark:
> where can i find distrib/initscripts/afpd.service.tmpl and edit it ?

@marcuskuenzel1971 I suggest you edit the *.service files that are installed on your system already. If you run the command `systemctl status atalkd` you can find where atalkd.service is located, for instance. In my case, they're under /lib/systemd/system/

```
$ systemctl status atalkd
● atalkd.service - Netatalk AppleTalk daemon
     Loaded: loaded (/lib/systemd/system/atalkd.service; enabled; vendor preset: enabled)
     Active: active (running) since Sun 2023-02-26 19:26:42 PST; 2 days ago
    Process: 5620 ExecStart=/usr/local/sbin/atalkd (code=exited, status=0/SUCCESS)
      Tasks: 1 (limit: 1596)
        CPU: 2.868s
     CGroup: /system.slice/atalkd.service
             └─5621 /usr/local/sbin/atalkd
```

Edit the file (requires sudo) then save and run the command `sudo systemctl daemon-reload`

PR #247, opened 2023-03-27
Tweak Sonar scanning config

rdmark:
Scan only PRs coming from the present repo. Github does not pass tokens to forks, so Sonar jobs will fail for PRs originating from forks. See https://docs.github.com/en/actions/security-guides/encrypted-secrets#using-encrypted-secrets-in-a-workflow

Scan the two active stable branches in addition to main.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -3,6 +3,8 @@ on:
   push:
     branches:
       - main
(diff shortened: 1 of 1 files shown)


rdmark:
Self-merging, since this is a trivial configuration change, and it is causing false negative static analysis scans.

PR #252, opened 2023-03-27
Sonar Scan: Use a different logic for determining the originating repo

rdmark:


rdmark:
Self-merging, since this is a trivial configuration change, and it is causing false negative static analysis scans.

PR #273, opened 2023-04-05
Update pthread macro

dgsga:


rdmark:
@dgsga Just for future reference, I assume this is simply the latest version of this macro from https://www.gnu.org/software/autoconf-archive/ax_pthread.html as-is?

dgsga:
Absolutely...

PR #275, opened 2023-04-05
Run scans only when C code is changed

rdmark:


rdmark:
Merging this trivial configuration change to try to restore the main branch scanning

PR #279, opened 2023-04-05
Drop log level to warning for non-lethal errors

rdmark:
Drops the log level to warning for three very chatty non-fatal errors.


rdmark:
Working on improving the underlying issue

rdmark:
Fix for the immediate issue in https://github.com/Netatalk/netatalk/pull/292

However dropping the log level to warning still makes sense I think. It's a fairly common occurrence.

PR #284, opened 2023-04-07
github workflow: Use branch and path-ignore together

rdmark:


rdmark:
Self-merging a trivial workflow config change. Hopefully this will restore branch scanning.

PR #285, opened 2023-04-07
github workflow: Remove repo conditional

rdmark:


rdmark:
Self-merging a trivial config change after updating the Org level settings on secret sharing.

PR #290, opened 2023-04-07
[2.2] git workflow: bring push and fork rules up to date

rdmark:


rdmark:
Self merging github workflow config

PR #291, opened 2023-04-08
Restore the importing of compat.h for Linux compatibility

rdmark:


rdmark:
Self-merging trivial hotfix to restore functionality on Linux

PR #294, opened 2023-04-09
C99 compatibility fixes

fweimer-rh:
Avoid build failures with future compilers.

Related to:

* https://fedoraproject.org/wiki/Changes/PortingToModernC
* https://fedoraproject.org/wiki/Toolchain/PortingToModernC


rdmark:
Thanks for contributing the PR! Verified that this resolves compiler warnings that I've observed on Debian /gcc10 as well.

PR #295, opened 2023-04-09
Resolve gcc10 compiler warnings

rdmark:
- Use strlcpy() in spotlight code 
https://linux.die.net/man/3/strlcpy
- glib init calls have been deprecated and the system starts up automatically
https://www.manpagez.com/html/gobject/gobject-2.56.0/gobject-Type-Information.php#g-type-init
- glib thread creation has changed
https://docs.gtk.org/glib/ctor.Thread.new.html
- declare as const to math the return data type of the later assignment


rdmark:
This should resolve all remaining compiler warnings with gcc10 on Debian

PR #301, opened 2023-04-11
libatalk: Remove dequoting of double quotes in the iniparser, again

rdmark:
Partially reverts https://github.com/Netatalk/netatalk/commit/b568955f3f970cffe70289bcaa7e7cf9ddaf0687
Restores the change originally made in https://github.com/Netatalk/netatalk/commit/3343c2

### DIFF
--- a/libatalk/iniparser/iniparser.c
+++ b/libatalk/iniparser/iniparser.c
@@ -587,7 +587,7 @@ static line_status atalk_iniparser_line(
         sscanf(line, "[%[^]]", section);
         strstrip(section);
         sta = LINE_SECTION;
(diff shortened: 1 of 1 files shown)


dgsga:
Does this PR mean you have found an error in the coding of ndevilla's upstream ini parser repo? The two sscanf calls are present there, do they not both serve a function?

rdmark:
@dgsga It's not an upstream error per se, but a modification that we did locally in this project: Not allowing quotation marks around ini options.

@slowfranklin May have more context since he's the one who made the change in https://github.com/Netatalk/netatalk/commit/3343c2

slowfranklin:
Sorry, but I'm having a hard time remembering things from 24 hours ago and this is 11 years.... :)))

PR #303, opened 2023-04-11
libatalk: remove const keyword for variable that changes

rdmark:
gcc10 complains about discarded qualifiers, so removing the const keyword from a variable that has multiple assignments throughout the function

This was a minor regression introduced by https://github.com/Netatalk/netatalk/pull/299


rdmark:
Self merged a trivial code change.

PR #304, opened 2023-04-11
manpages: Clarify that guest users must be an actual user

rdmark:


rdmark:
Self merged a trivial code change.

PR #308, opened 2023-04-14
Add support for macOS hosts

dgsga:
This initial changeset adds support for macOS hosts which do not have a built-in afp server (macOS Big Sur onwards) or which do not support AFP 2.2, the version required to connect to vintage Macs running classic Mac OS. In summary:
- is compatible with modern Intel and Apple Silicon based Macs
- fixes compilation errors when using the macOS native compiler. Apple's version of clang is picky about header orders, it likes the system headers to be iterated before the atalk and same-dir headers.
- fixes macro to enable Zeroconf support on macOS
- fixes EA macro to ensure sys EA support
- fixes the lock file location on macOS
- fixes db3 and ssl macros to ensure OpenSSL and Berkeley DB dependencies installed via Homebrew are found
- fixes PAM macro to enable authentication on macOS
- disables ACL and dtrace support as Apple implementations are currently not supported
- installs a macOS LaunchDaemon to start netatalk on boot and restart from the command line when required
- requires dependencies to be installed via Homebrew:
`brew install berkeley-db libevent libgcrypt mysql libressl openssl@1.1 pkg-config`

What is currently unsupported in this changeset (but possible, proof of concept please see my fork https://github.com/dgsga/netatalk):

- building documentation using docbook (WIP)
- DBUS support / afpstats (WIP)

I have tested these code changes extensively on my Macs over a two year period. I have also tested the PR branch on Debian 11 to check for inadvertent regressions, none found yet. Please test if you have Mac hardware available

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,3 +29,5 @@ test-driver
 *.diff.gz
 *~
 *.o
+**/.DS_Store
+build-aux
--- a/bin/ad/ad_cp.c
+++ b/bin/ad/ad_cp.c
@@ -821,7 +821,7 @@ static int setfile(const struct stat *fs, int fd)
     islink = !fdval && S_ISLNK(fs->st_mode);
     mode = fs->st_mode & (S_ISUID | S_ISGID | S_ISVTX | S_IRWXU | S_IRWXG | S_IRWXO);
 
-#if defined(__FreeBSD__)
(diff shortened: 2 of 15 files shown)


rdmark:
This is a surprisingly light-weight changeset! I expected something like 10k+ changed lines ;)
Let me run some tests in various environments...

PR #309, opened 2023-04-15
configure.ac: Remove extraneous forward slash in UAMS path

dgsga:
This is evident when you type afpd -V


rdmark:
I did notice this too!

Weirdly enough, the trailing slashes were deliberately added in https://github.com/Netatalk/netatalk/commit/ed9a339b481cbe160d08aa2cd0a6d76307d29dcf

I wonder if it's a cross platform thing? Seems unlikely though.

PR #310, opened 2023-04-17
Add support for macOS hosts

dgsga:
This initial changeset adds support for macOS hosts which do not have a built-in afp server (macOS Big Sur onwards) or which do not support AFP 2.2, the version required to connect to vintage Macs running classic Mac OS. In summary:

is compatible with modern Intel and Apple Silicon based Macs
fixes compilation errors when using the macOS native compiler. Apple's version of clang is picky about header orders, it likes the system headers to be iterated before the atalk and same-dir headers.
fixes macro to enable Zeroconf support on macOS
fixes EA macro to ensure sys EA support
fixes the lock file location on macOS
fixes db3 and ssl macros to ensure OpenSSL and Berkeley DB dependencies installed via Homebrew are found
fixes PAM macro to enable authentication on macOS
disables ACL and dtrace support as Apple implementations are currently not supported
installs a macOS LaunchDaemon to start netatalk on boot and restart from the command line when required
requires dependencies to be installed via Homebrew:
brew install berkeley-db libevent libgcrypt mysql libressl openssl@1.1 pkg-config
What is currently unsupported in this changeset (but possible, proof of concept please see my fork https://github.com/dgsga/netatalk):

building documentation using docbook (WIP)
DBUS support / afpstats (WIP)
I have tested these code changes extensively on my Macs over a two year period. I have also tested the PR branch on Debian 11 to check for inadvertent regressions, none found yet. Please test if you have Mac hardware available

### DIFF
--- a/macros/netatalk.m4
+++ b/macros/netatalk.m4
@@ -1045,6 +1045,19 @@ case "$this_os" in
   *freebsd4* | *dragonfly* )
     AC_DEFINE(BROKEN_EXTATTR, 1, [Does extattr API work])
   ;;
+  
+  *macosx*)
+	AC_SEARCH_LIBS(getxattr, [attr])
+    if test "x$neta_cv_eas_sys_found" != "xyes" ; then
+       AC_CHECK_FUNCS([getxattr fgetxattr listxattr],
+                      [neta_cv_eas_sys_found=yes],
+                      [neta_cv_eas_sys_not_found=yes])
+	   AC_CHECK_FUNCS([flistxattr removexattr fremovexattr],,
+                      [neta_cv_eas_sys_not_found=yes])
+	   AC_CHECK_FUNCS([setxattr fsetxattr],,
+                      [neta_cv_eas_sys_not_found=yes])
+    fi
+  ;;
 
   *)
 	AC_SEARCH_LIBS(getxattr, [attr])
--- a/macros/netatalk.m4
+++ b/macros/netatalk.m4
@@ -486,10 +486,17 @@ AC_ARG_ENABLE(shell-check,
 dnl Check for optional initscript install
 AC_DEFUN([AC_NETATALK_INIT_STYLE], [
     AC_ARG_WITH(init-style,
-                [  --with-init-style       use OS specific init config [[redhat-sysv|redhat-systemd|suse-sysv|suse-systemd|gentoo-openrc|gentoo-systemd|netbsd|debian-sysv|debian-systemd|solaris|openrc|systemd]]],
+                [  --with-init-style       use OS specific init config [[redhat-sysv|redhat-systemd|suse-sysv|suse-systemd|gentoo-openrc|gentoo-systemd|netbsd|debian-sysv|debian-systemd|solaris|openrc|systemd|macos-launchd]]],
                 init_style="$withval", init_style=none
     )
-    case "$init_style" in 
+    case "$host_os" in
+    *darwin*)
+    if test x"$init_style" = x"none" ; then
+    init_style=macos-launchd
+    fi
+    ;;
+    esac
+    case "$init_style" in
     "redhat")
 	    AC_MSG_ERROR([--with-init-style=redhat is obsoleted. Use redhat-sysv or redhat-systemd.])
         ;;
@@ -550,6 +557,10 @@ AC_DEFUN([AC_NETATALK_INIT_STYLE], [
 	    AC_MSG_RESULT([enabling general systemd support])
 	    ac_cv_init_dir="/usr/lib/systemd/system"
         ;;
+    "macos-launchd")
+    	AC_MSG_RESULT([enabling macOS-style launchd initscript support])
+    	ac_cv_init_dir="/Library/LaunchDaemons"
+        ;;    
     "none")
 	    AC_MSG_RESULT([disabling init-style support])
 	    ac_cv_init_dir="none"
@@ -566,6 +577,7 @@ AC_DEFUN([AC_NETATALK_INIT_STYLE], [
     AM_CONDITIONAL(USE_DEBIAN_SYSV, test x$init_style = xdebian-sysv)
     AM_CONDITIONAL(USE_SYSTEMD, test x$init_style = xsystemd || test x$init_style = xredhat-systemd || test x$init_style = xsuse-systemd || test x$init_style = xgentoo-systemd)
     AM_CONDITIONAL(USE_DEBIAN_SYSTEMD, test x$init_style = xdebian-systemd)
+    AM_CONDITIONAL(USE_MACOS_LAUNCHD, test x$init_style = xmacos-launchd)
     AM_CONDITIONAL(USE_UNDEF, test x$init_style = xnone)
 
     AC_ARG_WITH(init-dir,
--- a/macros/netatalk.m4
+++ b/macros/netatalk.m4
@@ -52,6 +52,13 @@ AC_DEFUN([AC_NETATALK_DTRACE], [
     [WDTRACE=$withval],
     [WDTRACE=auto]
   )
+  case "$host_os" in
+  *darwin*)
+  if test x"$WDTRACE" = x"auto" ; then
+  WDTRACE=no
+  fi
+  ;;
+  esac
   if test "x$WDTRACE" = "xyes" -o "x$WDTRACE" = "xauto" ; then
     AC_CHECK_PROG([atalk_cv_have_dtrace], [dtrace], [yes], [no])
     if test "x$atalk_cv_have_dtrace" = "xno" ; then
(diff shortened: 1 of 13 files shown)


dgsga:
Re-submitted after source repo cleanup.

macros/db3-check.m4
```diff
@@ -86,7 +86,7 @@ AC_DEFUN([AC_NETATALK_PATH_BDB],[
 if test "x$bdb_required" = "xyes"; then
     trybdbdir=""
     dobdbsearch=yes
-    bdb_search_dirs="/usr/local /usr"
+    bdb_search_dirs="/usr/local /usr /usr/local/opt/berkeley-db /opt/homebrew/opt/berkeley-db"
```
rdmark, line 89: Rather than hard coding these obscure paths, would it be possible to ask macOS users to pass the `--with-bdb=/path/to/bdb/installation/` configure parameter instead?

macros/ssl-check.m4
```diff
@@ -45,7 +45,7 @@ dnl	compile_ssl=no
 
 	if test "$tryssl" = "yes"; then
 		AC_MSG_CHECKING([for SSL])
-		for ssldir in "" $tryssldir /usr /usr/local/openssl /usr/$atalk_libname/openssl /usr/local/ssl /usr/$atalk_libname/ssl /usr/local /usr/pkg /opt /opt/openssl /usr/local/ssl ; do
+		for ssldir in "" $tryssldir /usr /usr/local/openssl /usr/$atalk_libname/openssl /usr/local/ssl /usr/$atalk_libname/ssl /usr/local /usr/pkg /opt /opt/openssl /usr/local/ssl /usr/local/opt/openssl@1.1 /opt/homebrew/opt/openssl@1.1; do
```
rdmark, line 48: Similarly to above, there is the `--with-ssl-dir=PATH` configure option that should work in lieu of hard coding here. Or is there a reason this won't work?

libatalk/bstring/bstrlib.c
```diff
@@ -2737,7 +2737,7 @@ struct genBstrList g;
 #define START_VSNBUFF (256)
 #else
 
-#ifdef __GNUC__
+#if defined(__GNUC__) && !defined(__clang__)
```
rdmark, line 2740: I think this inline prototype is bitrotted. Just now I put up a PR to remove it. https://github.com/Netatalk/netatalk/pull/316

macros/netatalk.m4
```diff
@@ -486,10 +493,17 @@ AC_ARG_ENABLE(shell-check,
 dnl Check for optional initscript install
 AC_DEFUN([AC_NETATALK_INIT_STYLE], [
     AC_ARG_WITH(init-style,
-                [  --with-init-style       use OS specific init config [[redhat-sysv|redhat-systemd|suse-sysv|suse-systemd|gentoo-openrc|gentoo-systemd|netbsd|debian-sysv|debian-systemd|solaris|openrc|systemd]]],
+                [  --with-init-style       use OS specific init config [[redhat-sysv|redhat-systemd|suse-sysv|suse-systemd|gentoo-openrc|gentoo-systemd|netbsd|debian-sysv|debian-systemd|solaris|openrc|systemd|macos-launchd]]],
```
rdmark, line 496: The html documentation should to be updated with this new init-style option: https://github.com/Netatalk/netatalk/blob/main/doc/manual/install.xml

macros/netatalk.m4
```diff
@@ -52,6 +52,13 @@ AC_DEFUN([AC_NETATALK_DTRACE], [
     [WDTRACE=$withval],
     [WDTRACE=auto]
   )
+  case "$host_os" in
+  *darwin*)
+  if test x"$WDTRACE" = x"auto" ; then
+  WDTRACE=no
+  fi
+  ;;
+  esac
```
rdmark, line 61: Might be worth leaving a comment here to explain why dtrace is being disabled.

macros/db3-check.m4
dgsga: It certainly would be possible but these paths are not obscure if you use a Mac as your primary machine for compilation of open source software.

macros/ssl-check.m4
dgsga: As above...

libatalk/bstring/bstrlib.c
dgsga: Great, thanks for this. This particular patch has been discarded from my PR branch...

macros/netatalk.m4
dgsga: Will add and upload to my PR branch.

macros/netatalk.m4
dgsga: As above!

macros/netatalk.m4
dgsga: **Update:** done

macros/netatalk.m4
dgsga: **Update:** done

macros/db3-check.m4
rdmark: The location is obscure in the sense that it's whatever the homebrew team decided to use at one time, and might be subject to change in the future. Another package manager might become popular and take over, like what happened when macports lost popularity.

I think the --with-bdb override exists just for this kind of usecase -- an unofficial package manager and non-standard library locations.

Of course I do see the convenience factor for macos users, but I wanted to discuss the sustainable way to manage this macro. It's a bit of a slippery slope to try to support every non standard location...

macros/ssl-check.m4
rdmark: Would it at least be possible to not lock openssl to v1.1 in the macro? The world is rapidly moving towards openssl 3...

macros/db3-check.m4
rdmark: Case in point: OmniOS (an OpenSolaris descendent) installs the Berkeley libs in /opt/ooce so configuring netatalk with `--with-bdb=/opt/ooce` becomes mandatory.

dgsga: Fair do's, I have removed the SSL and db3 macro mods from the changes. By the way, OpenSSL3 breaks DHX authentication when logging in from classic Mac OS clients because it enforces a minimum DH modulus size of 512 bits. DHX (otherwise known as DHCAST128) uses a modulus size of 128 bits so you have to compile against OpenSSL 1.1 or libreSSL to login successfully.

macros/ssl-check.m4
dgsga: Please see above...

macros/db3-check.m4
dgsga: I have also updated install.xml to prompt users to specify a non standard SSL install location in the same way as the manual already does for BDB

rdmark: Looking good, thanks for making those changes.

Well noted on the breaking change in openssl3; TIL.

FWIW I have the habit of disabling DHX and using clrtxt for auth with classic Mac OS clients, DHX2 with macOS clients, since DHX was buggy on ARM for the longest time (works now I think.)

AFAICT the upcoming Debian Bookworm has dropped openssl1.1 altogether now so it's a bridge that we have to cross sooner or later...

rdmark:
A few thoughts for after this is merged:
- I see you have a github workflow set up in your fork to build on macOS. I would like to add this to the CI checks, in parallel with the Ubuntu job.
- A wiki page that covers the required homebrew packages, and other quirks of macOS would be helpful.

dgsga:
> A few thoughts for after this is merged:
> 
> * I see you have a github workflow set up in your fork to build on macOS. I would like to add this to the CI checks, in parallel with the Ubuntu job.

Fell free to copy and paste.

dgsga:
> A few thoughts for after this is merged:
> 
> * A wiki page that covers the required homebrew packages, and other quirks of macOS would be helpful.

Done!

macros/db3-check.m4
dgsga: LibreSSL is a good alternative and is likely to maintain support for older protocols for longer.

rdmark: Good point... netatalk3 already supports libressl I think.

rdmark:
> > A few thoughts for after this is merged:
> > 
> > * A wiki page that covers the required homebrew packages, and other quirks of macOS would be helpful.
> 
> Done, will be updated as required!

Looks good! Just a few thoughts...
- Link to homebrew's installation page, in case the installation procedure changes in the future
- Don't you need also xcode and/or Apple's development packages installed first?
- How do you launch netatalk after installing?

rdmark:
Got netatalk running on my MBP M1 Max! That was very seamless. Great job smoothing out all of the bumps.

rdmark:
As a side note, I keep running into error -50 on Mac clients when copying files with EA onto shared volumes. This is not specific to macOS hosts, but thought I'd mention it in case you've seen this too and know how to avoid it. I bet it's some kind of permissions / configuration issue.

rdmark:
@slowfranklin This changeset has turned out less complex / intrusive than I anticipated, so I'm leaning towards merging it sooner rather than later and include it in the next patch release. Thoughts?

rdmark:
I enabled and did a squashed merge for this changeset, for a cleaner commit log.

@dgsga Thanks for your hard work making this happen!

PR #311, opened 2023-04-17
configure.ac: Remove extraneous forward slash in UAMS path

dgsga:
[This is evident when you type afpd -V


dgsga:
Re-submitted after source repo cleanup

PR #312, opened 2023-04-18
Introducing a github workflow job to run integration tests

rdmark:
This introduces a separate job to the github workflow that builds the code and runs the integration test suite in test/


rdmark:
Self merging a trivial workflow config change.

PR #314, opened 2023-04-18
[2.2] Introduce a job to run tests in the github workflow

rdmark:
This introduces a separate job to the github workflow that builds the code and runs the integration test suite in test/


rdmark:
Self merging a trivial github workflow configuration change

PR #317, opened 2023-04-19
Add build status and license badges to README; improve verbiage

rdmark:


rdmark:
Self merge documentation update.

PR #320, opened 2023-04-23
[2.2] Fix incorrect type in printer status check

smagoun:
The `printer-is-accepting-jobs` attribute is a boolean, not an integer, so check for Boolean. 

Before this fix the client's driver would detect the printer but would fail to print, complaining that the printer was rejecting jobs. There would be a corresponding message in the `papd` log like `appletalk papd: CUPS_PAP_OPEN: Canon_MF632CDW is not accepting jobs`.

This fix causes `cups_get_printer_status` to return a non-zero `status` when the printer is ready, allowing printing to work. Tested on Mac OS 8.6 and 9.2.2 and 10.2.8

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,6 +1,7 @@
 Changes in 2.2.9
 ================
 
+* FIX: Fix printing from old Mac OS LaserWriter drivers
 * FIX: CVE-2022-45188
 * FIX: Improve systemd service dependencies GH#232
 * FIX: macusers: Fix output for long usernames
--- a/etc/papd/print_cups.c
+++ b/etc/papd/print_cups.c
@@ -307,7 +307,7 @@ cups_get_printer_status (struct printer *pr)
 
 	if ((attr = ippFindAttribute(response, "printer-is-accepting-jobs", IPP_TAG_BOOLEAN)) != NULL)
 	{
-		if ( ippGetInteger(attr, 0) == 0 )
+		if ( ippGetBoolean(attr, 0) == 0 )
 			status = 0;
 	}
 		


rdmark:
Thanks for contributing a PR!

Were you able to have a look at surrounding code to make sure this was the only type mismatch?

For the record these checks were introduced last year with https://github.com/Netatalk/netatalk/commit/b88b006f28b7d8829d5dff3ae4dbd544506d4cf4 as part of the CUPS API update. Printing to my own HP printer over CUPS has been working since then, but perhaps other printers are stricter with their queue validation?

Also, can you please update NEWS with a line for this fix as part of your changeset!

smagoun:
Thanks for the review! I've updated the NEWS file.

I did a cursory check for other type mismatches and didn't see any. While looking into that, I've also come across a [different patch](https://github.com/rdmark/netatalk-2.x/commit/6c35bb780e83e7a83fad19a2cd14fe2b29c3f54a) for the issue on a different fork of netatalk 2.2. That patch includes the change from integer-->boolean, but also has some other changes in it related to requests + handling status codes. I'm not sure what the material differences are between that fork and this one.

rdmark:
Oh yeah you're right, I completely forgot that I didn't get the papd patch from netatalk 2.x (originally from the a2server project) ported upstream. We should definitely get those changes over here too, since they fix compatibility issues with very old systems and laserwriter drivers. Do you want to take a stab at porting this?

NJRoadfan:
I don't know how this snuck into the main 2.2 tree. The original patches I made against 2.2.6 used ippGetBoolean.

rdmark:
> I don't know how this snuck into the main 2.2 tree. The original patches I made against 2.2.6 used ippGetBoolean.

That was me porting a downstream patch from NetBSD early on. https://github.com/Netatalk/netatalk/pull/152

I'm open to (partially) reverting that changeset.

NJRoadfan:
No need to revert the change set. This PR corrects the one difference between my patch and the NetBSD one. The NetBSD patch was more through then my work anyway as it handled the deprecated httpConnect().

rdmark:
@NJRoadfan Thanks for looking into that. I'll merge this then and leave the rest as is.

@smagoun FYI for future reference: The convention in NEWS is to prefix each change with the component that it's touching. F.e. in this case `papd: ...`

smagoun:
> @smagoun FYI for future reference: The convention in NEWS is to prefix each change with the component that it's touching. F.e. in this case `papd: ...`

Noted - thanks for the tip!

PR #322, opened 2023-04-24
Create GitHub workflow for macOS runner

dgsga:


dgsga:
Hi @rdmark
As requested :)
@dgsga

dgsga:
All of the macOS specific build warnings are easily fixed but I'm not sure how this would affect building on other platforms...

.github/workflows/macos.yml
```diff
@@ -0,0 +1,21 @@
+name: macos
+
+on:
+  push:
+    branches: [ "main" ]
```
rdmark, line 5: I think we want to add "branch-*" to this so that the job will run when merged to branch-netatalk-3-1 for instance.

.github/workflows/macos.yml
```diff
@@ -0,0 +1,21 @@
...
+  pull_request:
+    branches: [ "main" ]
+
+jobs:
+  build-macos:
+    runs-on: macos-latest
```
rdmark, line 11: Can an argument be made to lock this to `macos-13` for now, and then deliberately bump it when a new version is released? I'm always a bit wary of `*-latest` runners since a future rollout of later OS runners may lead to unexpected failures.

.github/workflows/macos.yml
```diff
@@ -0,0 +1,21 @@
+name: macos
+
+on:
+  push:
+    branches: [ "main" ]
+  pull_request:
```
rdmark, line 6: FWIW I'm also doing a `paths-ignore` in the Linux jobs but I'm not sure it's working 100%. I'm trying to avoid having the build run for documentation changes, or non-C code changes...

rdmark: Although I'm not confident that the current paths-ignore configuration actually has the intended effect. I'm still seeing jobs kicked off when NEWS and README.me are changed.

rdmark:
> All of the macOS specific build warnings are easily fixed but I'm not sure how this would affect building on other platforms...

We can certainly test it out (in another PR). There are a multitude of build warnings on the latest ubuntu runner too, mostly to do with stricter standards compliance in later gcc, and major toolchain updates.

I have VMs set up with Fedora, CentOS, openSUSE, OmniOS, FreeBSD, Debian Bookworm.

rdmark:
@dgsga Is there a particular reason to keep this yml file separate from build.yml? I was thinking that defining the macos job as one of the jobs in build.yml, and rename "Integration Tests" to "Ubuntu".

dgsga:
That was quick! I was going to suggest a change, that the build step be broken down into bootstrap, configure, make and test as it makes it easier to see where any issue lies. What do you think?

rdmark:
Sure! It’s worth a try. As long as it doesn’t slow down the job
significantly…

PR #327, opened 2023-04-27
configure: Strip newline from VERSION if needed

rdmark:
vi/vim or even the github web code editor silently adds newlines to any file that they edit, breaking the bootstrapping process when a newline enters VERSION. This makes sure the newline gets stripped when bootstrapping the configure script.


rdmark:
I updated the developer notes wiki page with a note to not let a newline get into VERSION

PR #329, opened 2023-04-27
[2.2.] bootstrap: Avoid using -echo n for better portability

rdmark:
`echo -n` isn't supported by all shells, notably macOS `sh`


rdmark:
Covered by https://github.com/Netatalk/netatalk/pull/340

Issue #313, opened 2023-04-18
Integration tests linking fails when dtrace is enabled


rdmark:
The integration tests in test/ fail when dtrace is enabled.

Run them with `make check`. On Debian the `systemtap-sdt-dev` package is required for dtrace support.

```
Making check in include
make[1]: Entering directory '/home/runner/work/netatalk/netatalk/include'
Making check in atalk
make[2]: Entering directory '/home/runner/work/netatalk/netatalk/include/atalk'
make  check-am
make[3]: Entering directory '/home/runner/work/netatalk/netatalk/include/atalk'
make[3]: Nothing to be done for 'check-am'.
make[3]: Leaving directory '/home/runner/work/netatalk/netatalk/include/atalk'
make[2]: Leaving directory '/home/runner/work/netatalk/netatalk/include/atalk'
make[2]: Entering directory '/home/runner/work/netatalk/netatalk/include'
make[2]: Nothing to be done for 'check-am'.
make[2]: Leaving directory '/home/runner/work/netatalk/netatalk/include'
make[1]: Leaving directory '/home/runner/work/netatalk/netatalk/include'
Making check in libatalk
make[1]: Entering directory '/home/runner/work/netatalk/netatalk/libatalk'
Making check in acl
make[2]: Entering directory '/home/runner/work/netatalk/netatalk/libatalk/acl'
make[2]: Nothing to be done for 'check'.
make[2]: Leaving directory '/home/runner/work/netatalk/netatalk/libatalk/acl'
Making check in adouble
make[2]: Entering directory '/home/runner/work/netatalk/netatalk/libatalk/adouble'
make[2]: Nothing to be done for 'check'.
make[2]: Leaving directory '/home/runner/work/netatalk/netatalk/libatalk/adouble'
Making check in bstring
make[2]: Entering directory '/home/runner/work/netatalk/netatalk/libatalk/bstring'
make[2]: Nothing to be done for 'check'.
make[2]: Leaving directory '/home/runner/work/netatalk/netatalk/libatalk/bstring'
Making check in compat
make[2]: Entering directory '/home/runner/work/netatalk/netatalk/libatalk/compat'
make[2]: Nothing to be done for 'check'.
make[2]: Leaving directory '/home/runner/work/netatalk/netatalk/libatalk/compat'
Making check in cnid
make[2]: Entering directory '/home/runner/work/netatalk/netatalk/libatalk/cnid'
Making check in last
[...]

PR #330, opened 2023-04-28
configure: Revert the subdir-objects option

rdmark:
Temporarily reverting the subdir-objects option for bootstrapping. A more comprehensive solution will be required to make this work with the test/afpd/

https://github.com/Netatalk/netatalk/issues/313


rdmark:
Merging in order to proceed with 3.1 release

PR #332, opened 2023-04-29
[2.2] bootstrap: test objects not generated with subdir-objects, #313

rdmark:


rdmark:
A better fix is in the works

PR #339, opened 2023-05-10
[2.2] Fix byte order detection on OpenIndiana

rdmark:
Backport of 3.x patch: https://github.com/Netatalk/netatalk/commit/a0b26922623b9de65f227a3ee3a47819d8530613

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,3 +1,9 @@
+Changes in 2.2.10
+================
+
+* FIX: afpd: Byte order detection. Fixes an error where Netatalk on
+       OpenIndiana returned wrong volume size information.
+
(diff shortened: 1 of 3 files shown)


rdmark:
Found out that the issue didn't affect 2.2.x in the first place: https://sourceforge.net/p/netatalk/mailman/netatalk-devel/thread/CAB-D_%3Db11XmeB74UPQeUA0YGpcyjpZ11xzRKoyjbAei83WSXxw%40mail.gmail.com/#msg29156103

Also, I tested dynamically on OpenIndiana 2023.04 and the volume size calculation seems roughly correct.

Issue #220, opened 2023-02-22
Reintroduce AppleTalk / DDP support


rdmark:
To investigate the feasibility and break down the requirements for reintroducing the AppleTalk into the main Netatalk development branch.

# Background
AppleTalk / DDP and associated daemons and tooling were deprecated with Netatalk 3.0, in order to streamline the codebase and making it more maintainable / extendable. Fast forward a few years, and AFP over TCP has become a legacy protocol with Apple switching to SMB as the file sharing protocol in macOS since Maverick. This changes the equation for Netatalk in two notable ways: 1, no need to continuously iterate on the capabilities as Apple adds new functionality, and 2, the primary value of Netatalk becomes to support vintage Mac networks. In this context, reinstating AppleTalk makes sense.

# Solutions
Two approaches can be considered

1. Add `atalkd` code back to Netatalk 3.1
2. Backport all Netatalk 3.1 improvements to Netatalk 2.2

Pros with option 1
- Retains all improvements and features in 3.1 by default (AFP 3.4, OSX style EA, Spotlight, ini style config files etc.)

Cons with option 1
- Unknown how compatible atalkd 2.2 is with afpd 3.1
- Parts of the codebase changed significantly since 2.2 so not all original deprecation changes can be simply reverted

Pros with option 2
- AFP over TCP and AppleTalk already works in harmony in 2.2

Cons with option 2
- A decade worth of code changes to backport

# Requirements
- Restore module code
  - atalkd
  - papd
  - timelord
  - a2boot
  - AT protocols in libatalk
  - AT tools
- Reimplement branches for AT and older AFP/AD versions in afpd and other modules
- Update autotool hooks
- Update Makefiles
- Update init scripts
- Update documentation

rdmark:
In the case of option 1, I went through the 2.2 code tree and picked out modules that I think should be brought over, and a few that I think are safe to discard. Just to get an idea of the scope of effort.

Save
- bin/aecho -- useful AT tool
- bin/getzones -- useful AT tool
- bin/nbp -- useful AT tool
- bin/pap -- useful AT tool
- config/atalkd.conf
- config/papd.conf
- contrib/a2boot
- contrib/timelord
- etc/atalkd
- etc/papd
- libatalk/asp
- libatalk/atp
- libatalk/nbp
- libatalk/netddp
- macros/cups.m4
- sys/solaris -- has the Solaris AT kernel module

Discard
[...]

PR #341, opened 2023-05-15
Cleanup filetree and fix syntax error in netatalk executable makefile

dgsga:
This commit removes two redundant files in etc/afpd and fixes a syntax error in the Netatalk makefile. Both libs and cflags for libevent now appear in the configuration summary.


dgsga:
> @dgsga Thanks for the PR! I did notice that the libevent check behaved weirdly when bootstrapping on FreeBSD and OpenIndiana in particular, so good job cleaning up that part of the makefile.
> 
> Would you mind creating an issue ticket for this so that we can add it to the next 3.1 release candidate? 
Issue ticket created #342
>The makefile fix should definitely make it into the next patch release, IMO. Last time it got a bit messy when preparing the 3.1.15 release candidate so I'd like to do a bit more structured approach moving forward:
> 
> 1. Cut ticket and assign target-netatalk-3-1 label
> 2. Describe problem and solution in ticket
> 3. Raise PR
> 4. Review & merge
> 5. At the time of release tagging, pull the list of release candidate PRs using the label, and then cherry-pick into branch-netatalk-3-1
> 
> What do you think?
I know very little about GitHub release protocol but very happy to go with a more structured approach. While we are on our way to 3.1.16 shouldn't the development version be 3.1.16dev?

rdmark:
Thanks for the ticket!

And sure, let's bump the 3.1 branch version to 3.1.16dev. I think keeping the main branch dev version to a blanket 3.1.99dev still makes sense though. But I'm open to counter arguments!

dgsga:
TBH I don't 'get' having main and 3.1 branches, it seems like unnecessary duplication to me. Could we not create releases from main as it *is* the 3.1 branch. I may be missing something obvious though :)

rdmark:
It's a good question! Are you subscribed to the mailing lists by any chance? There was some discussion on this topic a few months back. The one legitimate reason to keep branch-netatalk-3-1 right now, I think, is because of https://github.com/Netatalk/netatalk/issues/220 . If major breaking / intrusive changes start happening on `main`, in preparation for a new feature release (3.2? 4.0?), it makes sense to have a stable branch for patch releases. That said, no coding has happened yet for bringing appletalk back in `main` so it's kind of a moot point.... (my fault; life has gotten in the way of that project)

Issue #345, opened 2023-05-19
[2.2] Multiple warnings persist on running bootstrap


dgsga:
These warnings are due to obsolete macros remaining in configure.ac and the m4 macros

rdmark:
Thanks for getting the 2.2 bootstrap cleaned up as well!

PR #346, opened 2023-05-19
[2.2] Fix all warnings on running  bootstrap

dgsga:
These commit fixes all the obsolete macros and syntax in bootstrap, configure.ac and the m4 macros #345

### DIFF
--- a/bootstrap
+++ b/bootstrap
@@ -5,21 +5,7 @@ set -x
 rm -rf autom4te*.cache
 
 # build it all.
-aclocal -I macros $ACLOCAL_FLAGS || exit 1
-autoheader || exit 1
-libtoolize --force --copy
-automake --include-deps --add-missing --foreign --copy || exit 1
-autoconf || exit 1
-
-# Original configure call was:
-#./configure --enable-maintainer-mode "$@"
-# However, according to:
-#  http://sources.redhat.com/autobook/autobook/autobook_43.html#SEC43
-# we need to have AM_MAINTAINER_MODE in configure.in in order to use
-# this feature.  It's not there at the moment, so I'm changing the
-# configure call.
-# And sometimes I don't want this to run.
-# [ -z "$NOEXECCONFIGURE" ] && ./configure "$@"
+autoreconf -fi || exit 1
 
 # Let's not fall off the end...
 exit 0
--- a/configure.ac
+++ b/configure.ac
@@ -1,25 +1,23 @@
 dnl configure.in for netatalk
 
-AC_INIT([netatalk], [m4_esyscmd([echo -n `cat VERSION`])])
+AC_INIT([netatalk],m4_normalize(m4_include([VERSION])))
 
 NETATALK_VERSION=`cat $srcdir/VERSION`
 AC_SUBST(NETATALK_VERSION)
 
-AC_CANONICAL_SYSTEM
+AC_CANONICAL_HOST
 AM_INIT_AUTOMAKE([subdir-objects])
-AM_CONFIG_HEADER(config.h)
+AC_CONFIG_HEADERS(config.h)
 AC_CONFIG_MACRO_DIRS([macros])
 AM_MAINTAINER_MODE([enable])
 
 dnl Checks for programs.
 AC_PROG_AWK
 AC_PROG_CC
-AC_PROG_CC_C99
 AC_PROG_INSTALL
 AC_PROG_LN_S
 AC_PROG_MAKE_SET
-AC_LIBTOOL_DLOPEN
-AC_PROG_LIBTOOL
+LT_INIT([dlopen])
 AC_PROG_PERL
 AC_PROG_GREP
 AC_PROG_PS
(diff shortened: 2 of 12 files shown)


configure.ac
```diff
@@ -1,25 +1,23 @@
 dnl configure.in for netatalk
 
-AC_INIT([netatalk], [m4_esyscmd([echo -n `cat VERSION`])])
+AC_INIT([netatalk],[m4_esyscmd([echo -n `cat VERSION`])])
```
rdmark, line 3: I think we can simplify this to `AC_INIT([netatalk],m4_normalize(m4_include([VERSION])))`

For the record: `echo -n` isn't portable

distrib/initscripts/Makefile.am
```diff
@@ -105,7 +105,7 @@ install-data-hook:
 
 uninstall-startup:
 	-systemctl disable $(service_DATA)
-	rm -f $(addprefix $(DESTDIR)$(servicedir)/, $(service_DATA))
+	rm -f $(DESTDIR)$(servicedir)/$(service_DATA))
```
rdmark, line 108: Did you test that this syntax works for uninstalling all of the service files? Unlike netatalk3, $(service_DATA) is a list of space separated files.

dgsga: Well spotted, will revert the makefile change until I've found a fix...

rdmark: When I tinkered with this I went with the brute force approach:
```
	rm -f $(DESTDIR)$(servicedir)/a2boot.service \
	      $(DESTDIR)$(servicedir)/afpd.service \
	      $(DESTDIR)$(servicedir)/atalkd.service \
	      $(DESTDIR)$(servicedir)/cnid.service \
	      $(DESTDIR)$(servicedir)/papd.service \
	      $(DESTDIR)$(servicedir)/timelord.service
```

The repetition of code is not very elegant, but OTOH I don't think this list of services will change any time soon.

dgsga: Fixed in a (slightly) more elegant way...

rdmark: This works!

I'll run some tests on other platforms now and then merge if nothing broke.

rdmark:
@dgsga The largefile check seems to be failing, at least on Raspbian Bullseye and OpenIndiana:

```
configure:15222: checking for _FILE_OFFSET_BITS value needed for large files
configure:15240: gcc -c -g -O2  conftest.c >&5
conftest.c: In function 'main':
conftest.c:130:1: error: 'WX_SYS_LARGEFILE_TEST' undeclared (first use in this function)
  130 | WX_SYS_LARGEFILE_TEST
      | ^~~~~~~~~~~~~~~~~~~~~
conftest.c:130:1: note: each undeclared identifier is reported only once for each function it appears in
```

rdmark:
Functionality of the largefile check is restored when bringing over the syntax from the main branch:
```
    AC_CACHE_CHECK([for $1 value needed for large files], [$3],
        [
          AC_COMPILE_IFELSE([AC_LANG_SOURCE([[#define $1 $2
                          #include <sys/types.h>],
                         WX_SYS_LARGEFILE_TEST])],
                         [$3=$2],
                         [$3=no])
        ]
```

It uses AC_LANG_SOURCE instead of AC_LANG_PROGRAM, plus slight variations to nesting.

rdmark:
The next problem is that something broke with the ACL lib check. Configuration claims to succeed but linking with libacl fails:
```
a  cache.lo unix.lo uuid.lo ldap.lo ldap_config.lo -lacl -lattr -lpthread  -lcrack -L../../libatalk
libtool: link: rm -fr  .libs/libacl.a .libs/libacl.la
libtool: link: (cd .libs/libacl.lax/libacl.a && ar x "/home/dmark/dev/netatalk2/libatalk/acl/.libs/libacl.a")
ar: /home/dmark/dev/netatalk2/libatalk/acl/.libs/libacl.a: No such file or directory
```

Raspbian Bullseye

macros/largefile-check.m4
```diff
@@ -20,11 +20,8 @@ define(WX_SYS_LARGEFILE_MACRO_VALUE,
 [
     AC_CACHE_CHECK([for $1 value needed for large files], [$3],
         [
-          AC_TRY_COMPILE([#define $1 $2
-                          #include <sys/types.h>],
-                         WX_SYS_LARGEFILE_TEST,
-                         [$3=$2],
-                         [$3=no])
+          AC_COMPILE_IFELSE([AC_LANG_PROGRAM([[#define $1 $2
+                          #include <sys/types.h>]], [[WX_SYS_LARGEFILE_TEST]])],[$3=$2],[$3=no])
```
rdmark, line 24: This makes the largefile check functional:

```
          AC_COMPILE_IFELSE([AC_LANG_PROGRAM([[#define $1 $2
                          #include <sys/types.h>], WX_SYS_LARGEFILE_TEST])],[$3=$2],[$3=no])
```

bootstrap
```diff
@@ -5,11 +5,7 @@ set -x
...
 
 # build it all.
-aclocal -I macros $ACLOCAL_FLAGS || exit 1
-autoheader || exit 1
-libtoolize --force --copy
-automake --include-deps --add-missing --foreign --copy || exit 1
-autoconf || exit 1
+autoreconf -fi || exit 1
 
 # Original configure call was:
```
rdmark, line 10: Side note: I think we can remove this very legacy comment

macros/largefile-check.m4
dgsga: Fix applied!

bootstrap
dgsga: Removed.

Issue #347, opened 2023-05-20
Extra token warning in largefile-check.m4 macro


rdmark:
The largefile check contains this block:
```
    AC_CACHE_CHECK([for $1 value needed for large files], [$3],
        [
          AC_COMPILE_IFELSE([AC_LANG_SOURCE([[#define $1 $2
                          #include <sys/types.h>],
                         WX_SYS_LARGEFILE_TEST])],
                         [$3=$2],
                         [$3=no])
        ]
```

It uses AC_LANG_SOURCE incorrectly, since according to the [documentation](https://www.gnu.org/software/autoconf/manual/autoconf-2.63/html_node/Generating-Sources.html), the macro takes only one argument, while we pass it two. In fact, we see this warning in config.log:

```
configure:15222: checking for _FILE_OFFSET_BITS value needed for large files
configure:15237: gcc -c -g -O2  conftest.c >&5
conftest.c:126:49: warning: extra tokens at end of #include directive
  126 |                           #include <sys/types.h>,
      |    
```

Therefore, the correct solution seems to be using the AC_LANG_PROGRAM macro.

rdmark:
PR in https://github.com/Netatalk/netatalk/pull/348

rdmark:
The proposed solution doesn't actually resolve the warning. I may have misread the bracket nesting.

rdmark:
For the record: This needs to be properly fixed in the 2.2 branch as well. The incorrect fix was merged there as part of a another PR.

rdmark:
I think we can live with the warning. Closing.

PR #348, opened 2023-05-20
macros: Correct macro for largefile check, #347

rdmark:


rdmark:
Reverting to draft. This doesn't actually seem to fix the issue.

rdmark:
@dgsga FYI I think we need to actually revert back to AC_LANG_SOURCE in the 2.2 branch

PR #349, opened 2023-05-21
[2.2] Enable compilation on macOS hosts v2

dgsga:
This commit enables the 2.2 branch to be compiled on macOS hosts. The DDP, a2boot and timelord configure options can be enabled in preparation for a possible development of a userland AppleTalk stack for modern macOS iterations. Note that compiling for macOS is _not_ the same as compiling for Mac OS X Server, the kernel has long since discarded all AppleTalk/DDP related code (xnu/netat) and use of the /netat header files is not indicated.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -687,7 +687,7 @@ if test x"$this_os" = "xmacosx"; then
 	dnl AC_DEFINE(NO_DLFCN_H)
 	dnl AC_DEFINE(NO_DDP, 1, [Define if DDP should be disabled])
 	AC_DEFINE(NO_QUOTA_SUPPORT, 1, [Define if Quota support should be disabled])
-	AC_DEFINE(MACOSX_SERVER, 1, [Define if compiling for MacOS X Server])
+	dnl AC_DEFINE(MACOSX_SERVER, 1, [Define if compiling for MacOS X Server])
 fi
 
 dnl ----- NetBSD specific -----
--- a/configure.ac
+++ b/configure.ac
@@ -682,7 +682,7 @@ fi
 dnl ----- Mac OSX specific -----
 if test x"$this_os" = "xmacosx"; then 
 	AC_MSG_RESULT([ * Mac OSX specific configuration])
-	AC_DEFINE(BSD4_4, 1, [BSD compatiblity macro])
+	dnl AC_DEFINE(BSD4_4, 1, [BSD compatiblity macro])
 	AC_DEFINE(HAVE_2ARG_DBTOB, 1, [Define if dbtob takes two arguments])
 	dnl AC_DEFINE(NO_DLFCN_H)
 	dnl AC_DEFINE(NO_DDP, 1, [Define if DDP should be disabled])
--- a/configure.ac
+++ b/configure.ac
@@ -682,12 +682,7 @@ fi
 dnl ----- Mac OSX specific -----
 if test x"$this_os" = "xmacosx"; then 
 	AC_MSG_RESULT([ * Mac OSX specific configuration])
-	dnl AC_DEFINE(BSD4_4, 1, [BSD compatiblity macro])
-	AC_DEFINE(HAVE_2ARG_DBTOB, 1, [Define if dbtob takes two arguments])
-	dnl AC_DEFINE(NO_DLFCN_H)
-	dnl AC_DEFINE(NO_DDP, 1, [Define if DDP should be disabled])
 	AC_DEFINE(NO_QUOTA_SUPPORT, 1, [Define if Quota support should be disabled])
-	dnl AC_DEFINE(MACOSX_SERVER, 1, [Define if compiling for MacOS X Server])
 fi
 
 dnl ----- NetBSD specific -----
--- a/configure.ac
+++ b/configure.ac
@@ -1158,6 +1158,19 @@ case "$this_os" in
   *freebsd4* | *dragonfly* )
     AC_DEFINE(BROKEN_EXTATTR, 1, [Does extattr API work])
   ;;
+	
+	*macosx*)
+	AC_SEARCH_LIBS(getxattr, [attr])
+    if test "x$neta_cv_eas_sys_found" != "xyes" ; then
+       AC_CHECK_FUNCS([getxattr fgetxattr listxattr],
+                      [neta_cv_eas_sys_found=yes],
+                      [neta_cv_eas_sys_not_found=yes])
+	   AC_CHECK_FUNCS([flistxattr removexattr fremovexattr],,
+                      [neta_cv_eas_sys_not_found=yes])
+	   AC_CHECK_FUNCS([setxattr fsetxattr],,
+                      [neta_cv_eas_sys_not_found=yes])
+    fi
+  ;;
 
   *)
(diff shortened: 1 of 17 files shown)


dgsga:
@rdmark I have two more things to do on this PR. I need to remove two empty commits and also enable configure.ac to distinguish between macOS and Mac OS X Server as we don't want to lose the latter as a possible host.

dgsga:
Another question, do you think the MACOSX_SERVER conditional relates to Mac OS X Server 1.0-1.2v3 or to post 10.0 OSes?

rdmark:
> @rdmark I have two more things to do on this PR. I need to remove two empty commits and also enable configure.ac to distinguish between macOS and Mac OS X Server as we don't want to lose the latter as a possible host.

I would personally not mind if we deprecated OSX Server support altogether, and repurposed whatever code is still relevant for macOS. If someone wants to run Netatalk on that OS, I would suggest they run a contemporaneous Netatalk release version, similarly to how we don't actively support legacy UNIX distros like Ultrix etc. anymore. What do you think?

rdmark:
> Another question, do you think the MACOSX_SERVER conditional relates to Mac OS X Server 1.0-1.2v3 or to post 10.0 OSes?

I only have a surface level understanding of autoconf, but doesn't this suggest that any version of OSX will pass the MACOSX_SERVER check?

```
dnl ----- Mac OSX specific -----
if test x"$this_os" = "xmacosx"; then 
	AC_MSG_RESULT([ * Mac OSX specific configuration])
	AC_DEFINE(BSD4_4, 1, [BSD compatiblity macro])
	AC_DEFINE(HAVE_2ARG_DBTOB, 1, [Define if dbtob takes two arguments])
	dnl AC_DEFINE(NO_DLFCN_H)
	AC_DEFINE(NO_DDP, 1, [Define if DDP should be disabled])
	AC_DEFINE(NO_QUOTA_SUPPORT, 1, [Define if Quota support should be disabled])
	AC_DEFINE(MACOSX_SERVER, 1, [Define if compiling for MacOS X Server])
fi
```

rdmark:
Never mind, I see that you're commenting out that line in your changeset, so this is a moot point.

rdmark:
BTW one interesting thing that I spotted is that the lockfile path selection logic in 2.2 uses the MACOSX_SERVER conditional. It might be interesting to test which branch macOS falls under here.

```
/* lock file path. this should be re-organized a bit. */
#if ! defined (_PATH_LOCKDIR)
#  if defined (FHS_COMPATIBILITY) || defined (__NetBSD__) || defined (__OpenBSD__)
#    define _PATH_LOCKDIR	"/var/run/"
#  elif defined (BSD4_4)
#    ifdef MACOSX_SERVER
#      define _PATH_LOCKDIR	"/var/run/"
#    else
#      define _PATH_LOCKDIR	"/var/spool/lock/"
#    endif
#  elif defined (linux)
#    define _PATH_LOCKDIR	"/var/lock/"
#  else
#    define _PATH_LOCKDIR	"/var/spool/locks/"
#  endif
#endif
```

include/atalk/paths.h

rdmark:
BTW, did you deliberately leave out the Zeroconf mDNS stuff for macOS? Would be handy to have that part of the changeset unless there's a good reason!

dgsga:
> > @rdmark I have two more things to do on this PR. I need to remove two empty commits and also enable configure.ac to distinguish between macOS and Mac OS X Server as we don't want to lose the latter as a possible host.
> 
> I would personally not mind if we deprecated OSX Server support altogether, and repurposed whatever code is still relevant for macOS. If someone wants to run Netatalk on that OS, I would suggest they run a contemporaneous Netatalk release version, similarly to how we don't actively support legacy UNIX distros like Ultrix etc. anymore. What do you think?

I agree, I think it's extremely unlikely that legacy OSes would be used as netatalk hosts, it kind of defeats the object of the project (file sharing between modern OSes and legacy Macs)

dgsga:
> BTW, did you deliberately leave out the Zeroconf mDNS stuff for macOS? Would be handy to have that part of the changeset unless there's a good reason!

Done!

dgsga:
PR now complete AFAICT

configure.ac
```diff
@@ -682,12 +682,12 @@ fi
...
 	AC_MSG_RESULT([ * Mac OSX specific configuration])
-	AC_DEFINE(BSD4_4, 1, [BSD compatiblity macro])
+	dnl AC_DEFINE(BSD4_4, 1, [BSD compatiblity macro])
 	AC_DEFINE(HAVE_2ARG_DBTOB, 1, [Define if dbtob takes two arguments])
 	dnl AC_DEFINE(NO_DLFCN_H)
-	AC_DEFINE(NO_DDP, 1, [Define if DDP should be disabled])
+	dnl AC_DEFINE(NO_DDP, 1, [Define if DDP should be disabled])
 	AC_DEFINE(NO_QUOTA_SUPPORT, 1, [Define if Quota support should be disabled])
-	AC_DEFINE(MACOSX_SERVER, 1, [Define if compiling for MacOS X Server])
+	dnl AC_DEFINE(MACOSX_SERVER, 1, [Define if compiling for MacOS X Server])
```
rdmark, line 690: Nitpick: What's your reason for commenting out these lines over removing them altogether? It might be worth leaving a comment why they're commented out at least. :)

rdmark (approved):

rdmark:
> PR now complete AFAICT. Let me know if you would like all the commits squashed (apart from the workflow change?)

A little bit of squash action would be great, to get a cleaner commit log!

configure.ac
dgsga: Shall I remove all MACOSX_SERVER conditionals from the codebase, it makes more sense if we're removing OS X Server host support?

dgsga: I thought it might be worth leaving them in in case they are required by a future userland AppleTalk network stack??

dgsga:
> > Another question, do you think the MACOSX_SERVER conditional relates to Mac OS X Server 1.0-1.2v3 or to post 10.0 OSes?
> 
> I only have a surface level understanding of autoconf, but doesn't this suggest that any version of OSX will pass the MACOSX_SERVER check?
> 
> ```
> dnl ----- Mac OSX specific -----
> if test x"$this_os" = "xmacosx"; then 
> 	AC_MSG_RESULT([ * Mac OSX specific configuration])
> 	AC_DEFINE(BSD4_4, 1, [BSD compatiblity macro])
> 	AC_DEFINE(HAVE_2ARG_DBTOB, 1, [Define if dbtob takes two arguments])
> 	dnl AC_DEFINE(NO_DLFCN_H)
> 	AC_DEFINE(NO_DDP, 1, [Define if DDP should be disabled])
> 	AC_DEFINE(NO_QUOTA_SUPPORT, 1, [Define if Quota support should be disabled])
> 	AC_DEFINE(MACOSX_SERVER, 1, [Define if compiling for MacOS X Server])
> fi
> ```

It does. This is true only for versions of macOS up to and including Mac OS X 10.5 (Leopard). All netat code was removed from the 10.6 (Snow Leopard) kernel (sadly).

rdmark:
Merged; thanks for giving 2.2 some love. :)

PR #353, opened 2023-06-06
Revert "Temporary workaround for -lzlib not found issue in upstream mysql version 8.0.33"

dgsga:
The bad linker flags in mysql_config are now fixed the upstream Homebrew formula for mysql.


rdmark:
@dgsga I'm still seeing failures. Perhaps github's runners aren't updated yet?

PR #354, opened 2023-06-06
workflow: Split build steps

rdmark:


rdmark:
Self merging trivial configuration change

Issue #358, opened 2023-06-17
DHX Incompatible with OpenSSL 3.0


Windoze345:
I've built Netatalk 3 on Debian 12 Bookworm. Debian currently uses OpenSSL 3.0.9.
Building and running works without problem but when i try to connect with a Mac OS 9.2.2 client I get the following error in the log:
uams_dhx_pam.c :PAM: Err Generating Key (OpenSSL error code: 41943166, error:0280007E:Diffie-Hellman routines::modulus too small)
The login fails after that.

As I understand it, this limit is hardcoded in OpenSSL >= 3.0 See this quote from the OpenSSL 3 migration guide:
> Enforce a minimum DH modulus size of 512 bits
> Smaller sizes now result in an error.

Any ideas or workarounds?

rdmark:
The workaround is to use LibreSSL rather than OpenSSL. DHX needs a 128 bit modulus size.

rdmark:
Or, to disable DHX and use the clrtxt UAM for authentication with classic Mac OS. :)

Windoze345:
Disabling DHX and enabling Clrtxt works for me.
Maybe it would be nice to detect OpenSSl >= 3.0 at build time and don't build DHX then.

rdmark:
That's an interesting option. I agree that it'd be good to have a default configuration that works out of the box. OTOH it wouldn't be very transparent to the user the criteria for when DHX is available and not.

Another, more brute-force, option that I could think of is to make the default afpd UAM configuration clrtxt+DHX2, rather than DHX+DHX2. It would be less secure, of course. But as an additional upside, this would make netatalk work out of the box with Mac OS 8 and earlier clients.

Yet another option is to make LibreSSL a dependency and use it by default over OpenSSL.

Thoughts?

NJRoadfan:
LibreSSL isn't available as a package for most Linux distributions, particularly Debian and Ubuntu. There is also the possibility that 128bit key sizes will be dropped in the future with LibreSSL as well. On a related note, 2-way randnum UAMs still work fine on classic MacOS clients, despite being ancient and weak. The downside is that you need to maintain a afppasswd file.

rdmark:
What I propose is that we make DHX2 the only UAM enabled by default on afpd runtime. Neither clrtxt nor randnum is secure at all so it seems prudent to keep the default configuration as secure as possible. https://github.com/Netatalk/netatalk/pull/359

What do you think?

pbobbenb:
[...]

PR #359, opened 2023-06-25
libatalk: Use only DHX2 UAM by default. Issue #358

rdmark:
As discussed in https://github.com/Netatalk/netatalk/issues/358 -- the DHX UAM will no longer work properly when using the latest OpenSSL 3.0, which is the default with the newly released Debian Bookworm. You can still build and execute the DHX code, but it leads to cryptic (no pun intended) errors when attempting to authenticate with the UAM with a client such as Mac OS 9.

There are workarounds such as building with LibreSSL or WolfSSL, but DHX is inherently insecure in this day and age, so I think it is advisable to not have it enabled by default.

### DIFF
--- a/doc/manpages/man5/afp.conf.5.xml
+++ b/doc/manpages/man5/afp.conf.5.xml
@@ -5,7 +5,7 @@
 
     <manvolnum>5</manvolnum>
 
-    <refmiscinfo class="date">11 Apr 2023</refmiscinfo>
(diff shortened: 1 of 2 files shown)


rdmark:
@slowfranklin What do you think about this change? Do you have a better idea how to deal with the original issue?

rdmark:
Discarding this in favor of a more constructive solution.
I encourage downstream distros to patch locally if OpenSSL 3.0 is causing trouble, which I've done for Debian now for instance.

PR #362, opened 2023-07-02
Use non-interactive PAM session, fixes #361 

### LINKED ISSUE #361: afpd: systemd-logind ReleaseSession rejected by dbus-daemon
As reported at [https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1018106](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1018106)

I am using netatalk for time machine backups. After every session I see this line in /var/log/auth.log
```
2023-07-01T22:31:47.223949+02:00 my-server dbus-daemon[1538889]: [system] Rejected send message, 2 matched rules; type="method_call", sender=":1.153636" (uid=145 pid=2690475 comm="/usr/sbin/afpd -d -F /etc/netatalk/afp.conf") interface="org.freedesktop.login1.Manager" member="ReleaseSession" error name="(unset)" requested_reply="0" destination="org.freedesktop.login1" (uid=0 pid=1538900 comm="/lib/systemd/systemd-logind")
```
Today I finally found the solution at [https://bugs.launchpad.net/ubuntu/+source/netatalk/+bug/1538004](https://bugs.launchpad.net/ubuntu/+source/netatalk/+bug/1538004)

It is simple really, in `/etc/pam.d/netatalk` replace this line:
```
@include common-session
```
with this line:
```
@include common-session-noninteractive
```
This fixed the issue for me.

RichieB2B:
`pam_systemd.so` should not be loaded, as it causes dbus errors. See #361 and [here](https://bugs.launchpad.net/ubuntu/+source/netatalk/+bug/1538004) for details.
```
/etc/pam.d$ diff -u common-session common-session-noninteractive
--- common-session      2023-06-16 14:50:01.333615224 +0200
+++ common-session-noninteractive       2023-06-16 14:50:01.337615060 +0200
@@ -1,9 +1,10 @@
 #
-# /etc/pam.d/common-session - session-related modules common to all services
+# /etc/pam.d/common-session-noninteractive - session-related modules
+# common to all non-interactive services
 #
 # This file is included from other service-specific PAM config files,
 # and should contain a list of modules that define tasks to be performed
-# at the start and end of interactive sessions.
+# at the start and end of all non-interactive sessions.
 #
 # As of pam 1.0.1-6, this file is managed by pam-auth-update by default.
 # To take advantage of this, it is recommended that you configure any
@@ -21,5 +22,4 @@
 session        required                        pam_permit.so
 # and here are more per-package modules (the "Additional" block)
 session        required        pam_unix.so 
-session        optional        pam_systemd.so 
 # end of pam-auth-update config
```

### DIFF
--- a/macros/pam-check.m4
+++ b/macros/pam-check.m4
@@ -82,7 +82,13 @@ AC_DEFUN([AC_NETATALK_PATH_PAM], [
            PAM_AUTH=common-auth
            PAM_ACCOUNT=common-account
            PAM_PASSWORD=common-password
-           PAM_SESSION=common-session
+           if test -f "$pampath/common-session-noninteractive" ; then
+               PAM_SESSION=common-session-noninteractive
+           elif test -f "$pampath/common-session-nonlogin" ; then
+               PAM_SESSION=common-session-nonlogin
+           else
+               PAM_SESSION=common-session
+           fi
         dnl RHEL/FC
         elif test -f "$pampath/system-auth" ; then
            PAM_DIRECTIVE=include


rdmark:
So interestingly, there is a difference between how openSUSE and Debian has implemented this. In openSUSE, the analog to *common-session-noninteractive* is named *common-session-nonlogin* as per the docs: https://en.opensuse.org/openSUSE:PAM_configuration

The two share the same goal though: same as *common-session* except they do not load `pam_systemd.so`

I tested this in my fairly fresh openSUSE installation and auth seems to work fine when using *common-session-nonlogin*.

How about we add another nested check, perhaps first for noninteractive, second for nonlogin, with a fallback to *common-session* just in case?

RichieB2B:
Something like this? I wasn't sure what indentation style to use, that file is not very consistent with tabs and spaces.

rdmark:
Thanks for this! It's exactly the logic that I had in mind, too. 

The indentation looks fine to me. This codebase doesn't have any style guide and the indentation can be all over the place.

BTW, I would like to backport this to the `branch-netatalk-2-2` branch as well. If you're up for it please create a PR for that branch, but if you can't be bothered I'll do it myself later. :)

rdmark:
Oh and one nitpick for future reference: The developer guidelines has rules around commit messages. For instance, they should be prefixed by the module that has been changed. In this case f.e.:

`macros: Use non-interactive PAM session when available`

https://github.com/Netatalk/netatalk/wiki/Developer-Notes#commit-messages

This helps with the creation and readability of the changelog.

PR #363, opened 2023-07-05
libatalk: Remove special case for COMMENT metadata validation

rdmark:
Remove special handling of COMMENT ad entries, since empty entries are now allowed as of the code changes in https://github.com/Netatalk/netatalk/pull/178

### DIFF
--- a/libatalk/adouble/ad_open.c
+++ b/libatalk/adouble/ad_open.c
@@ -804,9 +804,6 @@ static int ad_header_read_ea(const char *path, struct adouble *ad, const struct 
     }
 
     /*
-     * It is possible for AFP metadata to contain a zero-length
-     * comment. This will cause ad_entry(ad, ADEID_COMMENT) to return NULL
-     * but should not be treated as an error condition.
      * Since recent CVE fixes have introduced new behavior regarding
      * ad_entry() output. For now, we will AFP_ASSERT() in EC_CLEANUP to prevent
      * altering on-disk info. This does introduce an avenue to DOS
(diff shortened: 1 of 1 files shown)


southbay82:
Can't wait to see this in a release!

rdmark:
@southbay82 For the record, the more important fix which addresses the fatal internal errors is in https://github.com/Netatalk/netatalk/pull/178 which I merged to main yesterday.

The current PR has been repurposed for some minor cleanup of the logic to consolidate the two rounds of fixes for the same CVE regression.

rdmark:
@slowfranklin I'm proposing we restore the cleanup of xattr metadata that fail the validity check now. The safeguard in https://github.com/Netatalk/netatalk/commit/f90031aecde6372a26055541f63072ebbc821de6 has been live now since 3.1.13 in January, and we've gotten about half a dozen user reports. They all point towards empty entries on certain directories as the culprit. So with the fix in https://github.com/Netatalk/netatalk/pull/178 allowing empty entries again, I think it should be safe to go back to cleaning up anything that fails the check.

What do you think?

slowfranklin:
> What do you think?

To be honest, I don't know. I would have to spend much more time to drill into this than I can affort atm. Sorry!

rdmark:
@slowfranklin Fair enough. To hedge the risk, I propose that we only merge the change to COMMENT entry handling, and keep the invalid entry assertion intact for one more release cycle. This will help us catch any potential remaining corner cases.

I've repurposed this PR for the smaller change, adding a TODO, and raised https://github.com/Netatalk/netatalk/issues/400 to track the next step.

PR #373, opened 2023-08-15
[2.2] papd: Future-proof CUPS API usage

rdmark:


NJRoadfan:
Looks good.

The SonarCloud warning is due to using "strcpy()". I would have used the safer "strlcpy()", but that is not portable due to it being missing from glibc.

rdmark:
Thanks for the review! Do you think the Sonar static analysis reports are being helpful in general? I've kept them enabled because they do catch bugs in new code every now and then. But the noise level is high.

PR #396, opened 2023-08-20
Rename asip-status.pl to asip-status

rdmark:
Rename the script to asip-status and deemphasize the fact that's it's a Perl script under the hood


rdmark:
@dgsga Would you have any objection against making this change?

This was originally flagged by Debian's lintian, with the justification that the naming of an application shouldn't be specific to its current implementation under the hood.

dgsga:
Not at all! :)

rdmark:
Cheers! Off it goes.

PR #399, opened 2023-08-21
[2.2] Fix user visible typos and improve English grammar

rdmark:
- Removed two cases of explicit reference to the underlying implementation (Perl script)
- Improve English grammar for a few sentences
- Several typos reported by Debian's lintian
- Capitalization fixes
- Word choice improvements

### DIFF
--- a/doc/manual/man/man1/ad.1.xml
+++ b/doc/manual/man/man1/ad.1.xml
@@ -35,7 +35,7 @@
   <refsect1>
     <title>Description</title>
 
-    <para><command>ad</command> is a UNIX file utlity suite with Netatalk
+    <para><command>ad</command> is a UNIX file utility suite with Netatalk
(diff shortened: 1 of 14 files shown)


rdmark:
@NJRoadfan While correcting the papd man page it struck me that some of the info there may have to be updated following your changes to API, notable concerning PPD files. What do you think?

NJRoadfan:
Nothing needs to be changed. The CUPS interface always handled PPD files automatically, so you never needed the "pd" option unless you needed to override it for some reason. The whole reason why papd uses PPD files is that it was an easy source to provide information to driver queries.

With the revisions, IPP queries are used to pull that information from printers and stored in a generated PPD. This is to avoid extensively rewriting papd and to retain the ability to manually add printer queues the old fashioned way if someone isn't using CUPS.

PR #410, opened 2023-08-22
Create man page for cnid2_create tool

rdmark:
The cnid2_create tool is used to convert a Netatalk 1 CNID database to a Netatalk 2 compatible CNID database. I don't think it ever had a man page, so here's that's based on the Upgrade guide.


rdmark:
I'm on the fence whether I want to make this change in netatalk3 or not. A better solution yet might be to disable the tool's distribution like we did with `megatron`.

rdmark:
Doing the alternative solution in https://github.com/Netatalk/netatalk/pull/412

PR #411, opened 2023-08-22
[2.2] Create man page for cnid2_create tool

rdmark:
The cnid2_create tool is used to convert a Netatalk 1 CNID database to a Netatalk 2 compatible CNID database. I don't think it ever had a man page, so here's one that's based on the Upgrade guide.


rdmark:
The value of this tool is negligible in this day and age, but I'd like to add this man page for completion's sake. Before I started reading up on it, its usage wasn't clear to me.

PR #414, opened 2023-08-23
[2.2] Do not install at.h when glibc header is present

rdmark:
The at.h header conflicts with at.h distributed with glibc. Therefore, install the header only on systems where glibc isn't distributed by default. We still want to distribute it in the tarball, however.

As a side note, our at.h is probably relevant only for NetBSD presently, which is the one non-Linux OS where an appletalk kernel module is being kept alive.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -692,6 +692,11 @@ dnl	AC_COMPILE_IFELSE([
 		AC_MSG_RESULT([no])
 	])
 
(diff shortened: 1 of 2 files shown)


rdmark:
BTW, the current solution is a little awkward with the code duplication when assigning the same values twice to `pkginclude_HEADERS`. For some reason I couldn't make any changes to this variable after the first assignment and have it picked up by automake.

PR #417, opened 2023-08-24
[2.2] Configurable systemd prefix

rdmark:
This changeset makes it possible to choose a systemd dir prefix on configure time.

- Introduce `--with-systemd-prefix` option that takes PATH to prefix (typically /lib or /usr/lib)
- Make `--enable-system` option detect the most likely prefix (/lib or /usr/lib)
- Deprecate platform specific systemd options


rdmark:
TODO: Update the documentation

PR #421, opened 2023-08-26
[2.2] Correct install hook for static config files

rdmark:
The source path of static config files must be prefixed with `$(srcdir)` for out of tree builds to work.

It was already fixed for netatalk3 with this commit: https://github.com/Netatalk/netatalk/commit/a67d6457dcfeba36068952929d170056da95210e


VorpalBlade:
This seems to work. I also diffed two packages (one in tree the other out of tree) and there are no differing files.

There are however differences in the compiled binaries, which is strange. So, the builds are not fully reproducible, though that is a certainly a separate issue.

Issue #383, opened 2023-08-19
Variables not substituted in docbook man/html pages


rdmark:
The variables in documents is not replaced with the appropriate value. The example below is for the apf.conf man page, but it affects all man and html pages generated by docbook in netatalk3.

```
$ ./configure --with-docbook=/usr/share/sgml/docbook/xsl-stylesheets-1.78.1
$ make
$ make html

$ sudo make install

$ man afpd
...
FILES
@pkgconfdir@/afp.conf
configuration file used by afpd

   ${prefix}/var/netatalk/afp_signature.conf
       list of server signature

   ${prefix}/var/netatalk/afp_voluuid.conf
       list of UUID for Time Machine volume

   @pkgconfdir@/extmap.conf
       file name extension mapping

   @pkgconfdir@/msg/message.pid
       contains messages to be sent to users.
```

Originally reported in https://sourceforge.net/p/netatalk/bugs/626/

rdmark:
Being fixed in https://github.com/Netatalk/netatalk/pull/422

rdmark:
@dgsga There seems to be some residual problems remaining after your PR, when running `make html` to generate the docbook styled pages.

Do you want me to revert your change or do you want to roll forward with a fix?

```
Making html in manual
make[2]: Entering directory '/home/dmark/dev/netatalk/doc/manual'
Writing table-toc.html for book(netatalk-manual)
Writing example-toc.html for book(netatalk-manual)
Writing pr01.html for preface
Writing intro.html for chapter(intro)
Writing installation.html for chapter(installation)
Writing configuration.html for chapter(configuration)
Writing upgrade.html for chapter(upgrade)
Writing ad.1.html for refentry(ad.1)
Writing afp.conf.5.html for refentry(afp.conf.5)
Writing afp_signature.conf.5.html for refentry(afp_signature.conf.5)
Writing afp_voluuid.conf.5.html for refentry(afp_voluuid.conf.5)
Writing afpd.8.html for refentry(afpd.8)
Writing afpldaptest.1.html for refentry(afpldaptest.1)
Writing afppasswd.1.html for refentry(afppasswd.1)
Writing afpstats.1.html for refentry(afpstats.1)
Writing apple_dump.1.html for refentry(apple_dump.1)
Writing asip-status.1.html for refentry(asip-status.1)
Writing cnid_dbd.8.html for refentry(cnid_dbd.8)
Writing cnid_metad.8.html for refentry(cnid_metad.8)
[...]

PR #422, opened 2023-08-26
Fix variable substitution in manpages and docbook man/html pages #383

dgsga:
This commit fixes the failure to correctly substitute variables in both the manpages files and docbook man/html files

### DIFF
--- a/etc/afpd/Makefile.am
+++ b/etc/afpd/Makefile.am
@@ -50,7 +50,7 @@ afpd_LDADD =  \
 afpd_LDFLAGS = -export-dynamic
 
 afpd_CFLAGS = \
-	@GSSAPI_CFLAGS@ @KRB5_CFLAGS@ @PTHREAD_CFLAGS@\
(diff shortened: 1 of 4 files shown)


rdmark:
Thanks for filing the PR!

One thing that sprung to mind: wouldn't it be ideal to actually fix the autoconf substitutions of `pkgconfdir` and `localstatedir` (or similar)? Making assumptions about those paths in multiple locations might cause maintenance headaches in the future.

rdmark:
Interestingly, I found that uppercase `PKGCONFDIR` is being substituted in one of the macros, however in action the substitution leads to `${prefix}/etc/afp.conf` (in plain text) which isn't ideal. https://github.com/Netatalk/netatalk/blob/557d0bf2f6e9274c3a454a9f9ab1774e171910a6/macros/config-checks.m4#L30

dgsga:
I'm currently working on revising the PR as I agree with your earlier comment...

rdmark:
I had no idea it was this involved to do substitutions in man pages! Sorry for making it a bigger project for you than anticipated.

rdmark:
BTW I have merged all the other pending PRs now which introduced a mild merge conflict here. I suggest rebasing.

PR #424, opened 2023-08-28
[2.2] autoconf: Fold a2boot and timelord under the appletalk conditional

rdmark:
- Remove the distinct --enable-a2boot and --enable-timelord config options, and fold them under --enable-ddp.
- Simplify appletalk conditionals in man makefiles

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -505,26 +505,6 @@ AC_ARG_ENABLE(install-privileged,
 	]
 )
 
-dnl ----- timelord compilation (enabled by default)
-AC_MSG_CHECKING([whether timelord should be compiled])
-compile_timelord=yes
-AC_ARG_ENABLE(timelord,
-	[  --enable-timelord       enable compilation of timelord server],
-	[compile_timelord="$enableval"],
-	[compile_timelord="yes"]
-)
-AC_MSG_RESULT([$compile_timelord])
-
-dnl ----- a2boot compilation (enabled by default)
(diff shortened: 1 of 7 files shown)


dgsga (approved):
Looks good to me Daniel as a2boot and timelord have no relevance outside of the AppleTalk universe.

rdmark:
Thanks for reviewing.

The one drawback that I can think of with this change, is that whenever timelord or a2boot code won't compile on a niche platform you can't easily disable it at configure time anymore. But OTOH in the long run this is preferable since it incentivizes everyone to maintain the code.

dgsga:
> Thanks for reviewing.
> 
> The one drawback that I can think of with this change, is that whenever timelord or a2boot code won't compile on a niche platform you can't easily disable it at configure time anymore. But OTOH in the long run this is preferable since it incentivizes everyone to maintain the code.

And fix compile failures as and when required. It would be good to have an idea of which host OS'es netatalk users are currently using. Maybe this is a way forward? [https://github.blog/changelog/2022-04-12-discussions-polls/](url)

rdmark:
I like the idea of collecting user data. Ideally we should have (opt-in) user analytics that tells us what people are actually running it on rather what they say they're running it on. :)

One challenge is that the community is split between Github and the mailing lists, so I'd say engage with both.

rdmark:
Started a poll here https://github.com/Netatalk/netatalk/discussions/435

Please suggest any changes.

Issue #408, opened 2023-08-22
netatalk2: Out of tree builds broken


VorpalBlade:
If you try to build netatalk 2.x out of tree, some of the config files are missing.

E.g:

```bash
mkdir build && cd build
../netatalk-2.2.10/configure [...]
make
make install
```

Then all of these will be missing:
* /etc/netatalk/AppleVolumes.system
* /etc/netatalk/afp_ldap.conf
* /etc/netatalk/netatalk.conf
* /etc/netatalk/atalkd.conf
* /etc/netatalk/papd.conf

I'm not an expert at automake/autoconf (I have avoided them for over a decade, having moved over to cmake and other newer build systems for my own things), so I don't know exactly what is wrong here. Presumably something is set to relative the build dir instead of the source dir by mistake.

rdmark:
Good catch! Do they end up anywhere else on the file system, by any change? Any other files that don't get installed properly, or only these five?

Maybe something like $DESTDIR isn't expanded properly somewhere... Not obvious what's wrong from a read of the makefiles.

VorpalBlade:
> Good catch! Do they end up anywhere else on the file system, by any change?

No, I think they are just gone

> Any other files that don't get installed properly, or only these five?

Didn't really look into it. Would be quite easy though: make install with `DESTDIR` two different builds and diff the results.

rdmark:
I think the problem is that automake doesn't copy the config files into the build dir (unless I misunderstand the mechanism).

For the record, this affects netatalk3 as well.

rdmark:
Found the fix. You were right: a missing $(srcdir) prefix was the culprit. The bug was fixed in netatalk3 in March of 2012 but not ported to netatalk2, which was the norm back then.

PR in https://github.com/Netatalk/netatalk/pull/421 -- please give it a spin!

rdmark:
As a side note, I think this is a bug that we want `make distcheck` to catch in the future. The mechanism of the distcheck target is to do an out of tree build and installation into a temporary dir. I wonder if some flag need to be set for it to stop when it encounters a missing file.

The logs from a distcheck pre-patch does say there's an error but happy chugs along. This isn't right I think.
```
for f in AppleVolumes.system netatalk.conf afp_ldap.conf atalkd.conf papd.conf afpd.conf AppleVolumes.default; do \
[...]

PR #425, opened 2023-08-28
Fix variable substitution in manpages and docbook man/html pages #383

dgsga:
This commit fixes the failure to substitute installation path variables in manages and docbook man/html pages.

### DIFF
--- a/doc/manual/Makefile.am
+++ b/doc/manual/Makefile.am
@@ -49,6 +49,12 @@ HTML_PAGES = \
 	uniconv.1.html \
 	upgrade.html
 
+do_subst =  sed \
+  -i '' \
+  -e 's|@pkgconfdir[@]|$(sysconfdir)|g' \
+  -e 's|@localstatedir[@]|$(localstatedir)|g' \
+  -e 's|@prefix[@]|$(prefix)|g'
+
 DISTCLEANFILES = manual.xml
 
 if HAVE_XSLTPROC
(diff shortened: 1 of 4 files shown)


dgsga:
Hmm, failing on make distcheck. So still a WIP after all

rdmark:
You may have to prefix static source files with `$(srcdir)` for out-of-tree builds to work. I learned about this while dealing with https://github.com/Netatalk/netatalk/issues/408

rdmark:
Those are some handy sed substitutions you have there! Does it work as expected now?

Just for the sake of my understanding: Did you try using automake substitution macros, and couldn't get them to work for some reason? When working correctly, macros should use less code and be easier to maintain.

dgsga:
> Those are some handy sed substitutions you have there! Does it work as expected now?
> 
> Just for the sake of my understanding: Did you try using automake substitution macros, and couldn't get them to work for some reason? When working correctly, macros should use less code and be easier to maintain.

It does work as expected (finally...). For the sake of consistency I used the sed substitution style found elsewhere in the repo e.g initscripts. We could update the syntax for all the sed substitutions as per the example at [https://www.gnu.org/software/autoconf/manual/autoconf-2.67/html_node/Installation-Directory-Variables.html](url)

rdmark:
BTW, now when you're a Developer on the project, you have the privileges to push work branches to remote directly here on the mother project, no need to use your fork (unless you prefer to). 

There are two benefits of working in the mother project: It's easier for others to check out your work branches, and the Static Analysis CI job will run.

dgsga:
Great!

PR #439, opened 2023-08-30
Change sed substitution syntax so it works for both GNU and BSD versions

dgsga:
This commit fixes the GNU sed syntax error resulting from commit ffbe839

### DIFF
--- a/doc/manual/Makefile.am
+++ b/doc/manual/Makefile.am
@@ -50,7 +50,7 @@ HTML_PAGES = \
 	upgrade.html
 
 do_subst =  sed \
-  -i '' \
+  -i.'' \
   -e 's|@pkgconfdir[@]|$(sysconfdir)|g' \
   -e 's|@localstatedir[@]|$(localstatedir)|g' \
   -e 's|@prefix[@]|$(prefix)|g'


rdmark:
There's some funkiness going on with the configure summary (Debian) -- $sysconfdir not expanded to the actual path

```
Configure summary:
    INIT STYLE:
         none
    AFP:
         Extended Attributes: ad | sys
         ACL support: yes
         Spotlight: no
    CNID:
         backends:  dbd last tdb mysql
    UAMS:
         DHX     (PAM SHADOW)
         DHX2    (PAM SHADOW)
         RANDNUM (afppasswd)
         clrtxt  (PAM SHADOW)
         guest
    Options:
         Zeroconf support:        yes
         tcp wrapper support:     no
         quota support:           no
         valid shell check:       yes
         cracklib support:        no
         ACL support:             auto
         Kerberos support:        yes
         LDAP support:            yes
         AFP stats via dbus:      yes
         dtrace probes:           yes
    Paths:
         Netatalk lockfile:       /var/lock/netatalk
         dbus system directory:   ${sysconfdir}/dbus-1/system.d
         pam config directory:    ${sysconfdir}/pam.d
    Documentation:
         Docbook:                 yes
```

Maybe the summary macro needs updating?

rdmark:
The error is gone now so that part is fixed.
But, dunno if I'm doing something wrong but the substitution doesn't actually seem to be working with this spot check of afpd.8:

```
FILES
       @pkgconfdir@/afp.conf

       ${prefix}/var/netatalk/afp_signature.conf
           list of server signature

       ${prefix}/var/netatalk/afp_voluuid.conf
           list of UUID for Time Machine volume

       @pkgconfdir@/extmap.conf
           file name extension mapping

       @pkgconfdir@/msg/message.pid
           contains messages to be sent to users.
```

rdmark:
Anyways, let me merge this so that the error gets resolved at least.

rdmark (approved):

PR #442, opened 2023-08-31
Fix remaining issues with variable substitution in manpage and docbook man/html files

dgsga:


dgsga:
This was a PITA to fix because GNU and BSD sed use different syntax for in-place substitution. Finally found a solution that works across both platforms, you have to create a backup with the -i command then delete it immediately after variable substitution has taken place. BSD sed -i allows for a zero length extension `-i'',` GNU sed does not. The GNU equivalent is just `-i`, which BSD sed doesn't tolerate!!! So it's a Catch22. The issue with the manpage file generation in the man directory was fixed by changing the sed syntax to bring it into line with that used by the initscript generator.

Have tested on Debian and macOS...

rdmark:
Wohoo, it finally works for the man pages! :)

```
FILES
       /usr/local/etc/afp.conf

       /usr/local/var/netatalk/afp_signature.conf
           list of server signature

       /usr/local/var/netatalk/afp_voluuid.conf
           list of UUID for Time Machine volume

       /usr/local/etc/extmap.conf
           file name extension mapping

       /usr/local/etc/msg/message.pid
           contains messages to be sent to users.
```

rdmark (approved):
I wonder why they changed from colon notation to at-mark notation for substitution in 3.x?

PR #444, opened 2023-08-31
Update URL for css stylesheet in manual pages

rdmark:
The css stylesheet for html manual pages is currently hard coded to an old sourceforge domain that isn't used anymore. As a result, the stylesheet isn't found and the manual pages stay unstyled. I think this is because the user agent refuses to load resources over insecure http in a https context.

### DIFF
--- a/doc/html.xsl.in
+++ b/doc/html.xsl.in
@@ -4,7 +4,7 @@
 	<xsl:param name="use.id.as.filename" select="'1'"/>
 	<xsl:param name="chunk.section.depth" select="0"></xsl:param>
 	<xsl:param name="chunk.separate.lots" select="1"></xsl:param>
-	<xsl:param name="html.stylesheet" select="'http://netatalk.sourceforge.net/css/netatalk.css'"/>
+	<xsl:param name="html.stylesheet" select="'https://netatalk.sourceforge.io/css/netatalk.css'"/>
 
 	<xsl:template name="user.header.navigation">
 		<xsl:variable name="codefile" select="document('netatalk.html',/)"/>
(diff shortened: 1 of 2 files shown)


dgsga (commented):
I might be missing something obvious here but the doc/css/netatalk.css file does not appear to exist in the repo??

rdmark:
That's a good question! The static web sources live in a different repo: https://github.com/Netatalk/netatalk-web/blob/main/css/netatalk.css

rdmark:
But yes you're right, netatalk.css used to live in the main repo, under https://github.com/Netatalk/netatalk/tree/branch-netatalk-3-0/doc/gfx_and_css

I had it removed when I set up the netatalk-web repo and salvaged all the PHP code that had been living dangerously on the SF servers without revision control.

dgsga:
> But yes you're right, netatalk.css used to live in the main repo, under https://github.com/Netatalk/netatalk/tree/branch-netatalk-3-0/doc/gfx_and_css
> 
> I had it removed when I set up the netatalk-web repo and salvaged all the PHP code that had been living dangerously on the SF servers without revision control.

Strange, I find on Debian 12 and macOS that the pages are already styled without this PR. Styling fails when this PR is applied. Hmmm...

rdmark:
Right, this fix makes it so that the styles gets applied only when manual html pages are loaded from the remote server. e.g. https://netatalk.sourceforge.io/3.1/htmldocs/installation.html

The alternative solution is to just update the protocol and domain in the URL and keep it loading from remote... With the drawback that it might break in the future if sourceforge does another big change to their domain scheme.

rdmark:
Force pushed a change to keep the absolute URL.

I think the reason it's not working remotely right now is that the user agent disallows loading resources over http for a site that's otherwise https

rdmark:
FYI I went in a cleaned up all other http URLs in the website now. Also fixed a few errors to make it XHTML 1.0 Strict compliant

dgsga (approved):

PR #448, opened 2023-09-01
More elegant solution to variable substitution in manpages

dgsga:
This is a much cleaner approach to substitution in the manpages, it uses the AC_SUBST macro in configure.ac. Have reverted the colon notation to the original at-mark notation. Have tested in Debian 12 and macOS


rdmark (approved):
Happy to see a solution with less code in Makefiles!

Issue #441, opened 2023-08-31
Persistent errors in variable substitution in manpages and docbook man/html pages


dgsga:

rdmark:
For the record I was testing the latest code on FreeBSD and was getting this in generated man pages...

```
NETATALK(8)                   :NETATALK_VERSION:                   NETATALK(8)



NAME
       netatalk - Netatalk AFP server service controller daemon
```

rdmark:
With the latest fixes from today, NETATALK_VERSION works again.

There's still one kink left on FreeBSD... seemingly falling back to `NONE` instead of nothing when no substitution happens?

```
FILES
       NONE/etc/afp.conf
           configuration file used by afpd

       NONE/var/netatalk/afp_signature.conf
           list of server signature

       NONE/var/netatalk/afp_voluuid.conf
           list of UUID for Time Machine volume

       NONE/etc/extmap.conf
           file name extension mapping

       NONE/etc/msg/message.pid
           contains messages to be sent to users.
```

rdmark:
This is with `--sysconfdir=/usr/local` on FreeBSD:

```
FILES
       /usr/local/afp.conf
           configuration file used by afpd

       NONE/var/netatalk/afp_signature.conf
           list of server signature

       NONE/var/netatalk/afp_voluuid.conf
           list of UUID for Time Machine volume

       /usr/local/extmap.conf
           file name extension mapping

       /usr/local/msg/message.pid
           contains messages to be sent to users.
```

rdmark:
Hmm I'm seeing the same `NONE` problem in Debian Bookworm too when not explicitly defining any prefixes.

rdmark:
You know, this situation is kind of like a showcase why you want to move to the Meson build system. ;)

I butted my head against this for a few hours today. Cannot understand where the "NONE" comes from.

Reading up on stackoverflow this seems to be a long-standing limitation with autotools: it does not like to expand variables within variables in macros, hence that `sed` nonsense we see in Makefiles all over the place.

I'm starting to think that you very first proposal is the way to go... namely avoid double variable expansion with substitution. Just use the first layer of ${prefix} and hard-code the rest in the manual pages...

rdmark:
This one is what I was thinking of: https://github.com/Netatalk/netatalk/commit/1d00aacfaeb04ef1db945fa8d2484fdd954c322c

dgsga:
> With the latest fixes from today, NETATALK_VERSION works again.
> 
[...]

PR #451, opened 2023-09-03
Fix failures with double variable expansion when running configure #441

dgsga:
Please wake me up from this autotools nightmare...

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -88,10 +88,19 @@ AC_SEARCH_LIBS(connect, socket)
 AC_CHECK_FUNCS(getifaddrs) dnl comes after gethostbyname and connect so it picks up the libs
 
 dnl We substitute these in the manpage templates.
+
+if test x$prefix = xNONE ; then
+prefix=/usr/local
+sysconfdir=/usr/local/etc
+localstatedir=/usr/local/var
+fi
+
 localstatedir=`eval echo $localstatedir`
 AC_SUBST(localstatedir)
 pkgconfdir=`eval echo $sysconfdir`
 AC_SUBST(pkgconfdir)
+sysconfdir=`eval echo $sysconfdir`
+AC_SUBST(sysconfdir)
 
 AX_PTHREAD(, [AC_MSG_ERROR([missing pthread_sigmask])])
 
--- a/macros/netatalk.m4
+++ b/macros/netatalk.m4
@@ -113,12 +113,6 @@ AC_DEFUN([AC_NETATALK_DBUS_GLIB], [
     AC_MSG_ERROR([afpstats requested but dbus-glib not found])
   fi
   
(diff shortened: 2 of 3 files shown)


rdmark:
So the NONE problem is definitely gone!

However I think the default fallback substitution to `/usr/local` is malfunctioning when using configure overrides now.

For instance, this is a snippet from what the Debian package script does:

```
		--localstatedir=/var/lib \
		--sysconfdir=/etc \
		--with-pkgconfdir=/etc/netatalk \
```

However the generated paths in ./Makefile are all wrong:

```
localstatedir = /usr/local/var
sysconfdir = /usr/local/etc
pkgconfdir = /usr/local/etc
```

Similarly, the `--enable-fhs` overrides aren't taking effect either. As an experiment, I put the three substitutions *inside* the AC_ARG_ENABLE macro for --enable-fhs, and *then* the substitution did happen as expected. So the substitutions mustn't be happening at the right time in the macro right now...

dgsga:
> So the NONE problem is definitely gone!
> 
> However I think the default fallback substitution to `/usr/local` is malfunctioning when using configure overrides now.
> 
> For instance, this is a snippet from what the Debian package script does:
> 
> ```
> 		--localstatedir=/var/lib \
> 		--sysconfdir=/etc \
> 		--with-pkgconfdir=/etc/netatalk \
> ```
> 
> However the generated paths in ./Makefile are all wrong:
> 
> ```
> localstatedir = /usr/local/var
> sysconfdir = /usr/local/etc
> pkgconfdir = /usr/local/etc
> ```
> 
> Similarly, the `--enable-fhs` overrides aren't taking effect either. As an experiment, I put the three substitutions _inside_ the AC_ARG_ENABLE macro for --enable-fhs, and _then_ the substitution did happen as expected. So the substitutions mustn't be happening at the right time in the macro right now...

I think this is getting far too complex. I'm going to try your suggestion throughout the codebase, namely:
>I'm starting to think that your very first proposal is the way to go... namely avoid substitution with double variable >expansion. Just use the first layer of ${prefix} substitution and hard-code the rest in the manual pages...

rdmark:
Way too complex with potential for layers of cross-platform headaches...

dgsga:
It's looking good so far. Which version of Debian and FreeBSD are you using so we have consistent test environments?

rdmark:
Debian 12.0 Bookworm and FreeBSD 13.1

dgsga:
Please can you pass on the ./configure args you are using for FreeBSD 13.1, make is failing on 13.2 with a dtrace error :(

rdmark:
Weeeird. It compiles cleanly for me with zero args to ./configure

Maybe a missing dependency? I've left a few cursory notes at the wiki https://github.com/Netatalk/netatalk/wiki/Installation-Notes#freebsd-131

PR #461, opened 2023-09-04
Use single substitution for all variables in macros, manpages and docbook man/html files #441

dgsga:
Tested on FreeBSD, Debian and macOS:
Config summary works
Substitution in man/man1, man/man5 and man/man8 works
Substitution in manual html files works
make distcheck works in Debian and FreeBSD
No NONE's in sight
Never again...

### DIFF
--- a/macros/netatalk.m4
+++ b/macros/netatalk.m4
@@ -116,14 +116,12 @@ AC_DEFUN([AC_NETATALK_DBUS_GLIB], [
   if test x$prefix = xNONE ; then
     prefix=/usr/local
   fi
-  sysconfdir=`eval echo $sysconfdir`
-  AC_SUBST(sysconfdir)
-
+  
   AC_ARG_WITH(
       dbus-sysconf-dir,
-      [AS_HELP_STRING([--with-dbus-sysconf-dir=PATH],[Path to dbus system bus security configuration directory (default: ${sysconfdir}/dbus-1/system.d/)])],
+      [AS_HELP_STRING([--with-dbus-sysconf-dir=PATH],[Path to dbus system bus security configuration directory (default: ${sprefix}/etc/dbus-1/system.d/)])],
       ac_cv_dbus_sysdir=$withval,
-      ac_cv_dbus_sysdir=${sysconfdir}/dbus-1/system.d
+      ac_cv_dbus_sysdir=${prefix}/etc/dbus-1/system.d
   )
   DBUS_SYS_DIR=""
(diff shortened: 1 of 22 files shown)


macros/netatalk.m4
```diff
@@ -116,14 +116,12 @@ AC_DEFUN([AC_NETATALK_DBUS_GLIB], [
...
-  AC_SUBST(sysconfdir)
-
+  
   AC_ARG_WITH(
       dbus-sysconf-dir,
-      [AS_HELP_STRING([--with-dbus-sysconf-dir=PATH],[Path to dbus system bus security configuration directory (default: ${sysconfdir}/dbus-1/system.d/)])],
+      [AS_HELP_STRING([--with-dbus-sysconf-dir=PATH],[Path to dbus system bus security configuration directory (default: ${sprefix}/etc/dbus-1/system.d/)])],
       ac_cv_dbus_sysdir=$withval,
-      ac_cv_dbus_sysdir=${sysconfdir}/dbus-1/system.d
+      ac_cv_dbus_sysdir=${prefix}/etc/dbus-1/system.d
```
rdmark, line 124: Is is necessary to change these? I think they're internal to autoconf and not part of substitutions. I could be wrong though.

macros/pam-check.m4
```diff
@@ -154,14 +154,12 @@ AC_DEFUN([AC_NETATALK_PATH_PAM], [
...
-    
+
     AC_ARG_WITH(
         pam-confdir,
-        [AS_HELP_STRING([--with-pam-confdir=PATH],[Path to PAM config dir (default: ${sysconfdir}/pam.d)])],
+        [AS_HELP_STRING([--with-pam-confdir=PATH],[Path to PAM config dir (default: ${prefix}/etc/pam.d)])],
         ac_cv_pamdir=$withval,
-        ac_cv_pamdir=${sysconfdir}/pam.d
+        ac_cv_pamdir=${prefix}/etc/pam.d
     )
```
rdmark, line 163: Same as above.

rdmark (approved):

macros/pam-check.m4
dgsga: Ditto. If you try to add fixes for {sysconfdir} then you're down the rabbit hole again with cross-platform madness. You're welcome to try but TBH I think this issue has taken up too much time already...

rdmark: Ah, those were the very last vestiges of the sysconfdir variable. Let's kill them with fire.

PR #474, opened 2023-09-11
3.1.16 release

rdmark:


macros/netatalk.m4
```diff
@@ -113,11 +113,15 @@ AC_DEFUN([AC_NETATALK_DBUS_GLIB], [
     AC_MSG_ERROR([afpstats requested but dbus-glib not found])
   fi
 
+  if test x$prefix = xNONE ; then
+    prefix=/usr/local
+  fi
```
rdmark, line 118: Are there known states when $prefix resolves to NONE still that we need to hedge against?

macros/pam-check.m4
```diff
@@ -151,11 +151,15 @@ AC_DEFUN([AC_NETATALK_PATH_PAM], [
 	    AC_DEFINE(USE_PAM, 1, [Define to enable PAM support])
 	fi
 
+  if test x$prefix = xNONE ; then
+    prefix=/usr/local
+  fi
```
rdmark, line 156: Same as above

rdmark:
@dgsga Please do a final sanity check if you have a moment. In particular the changelog. I'll merge this first thing tomorrow and then tag the release!

rdmark:
BTW, one functional difference with the big removal of nested variable substitution, is that we now have to explicitly pass `--prefix=` (with an empty argument) to get "FHS compatible" dirs in the generated manual pages, i.e. /etc/afp.conf etc. Otherwise it defaults to /usr/local/etc/afp.conf in the manual page substitution, even if you use `--with-pkgconfdir` to have netatalk actually use a different dir. So we have the unfortunate situation of a potential discrepancy between the manual and actual paths.

rdmark:
Additionally, I just realized that we can never have absolutely correct manual pages on f.e. Debian now, since they use `/etc/netatalk` as pkgconfig prefix but `/etc` as sysconfdir prefix. Therefore the afpd manual will incorrectly say
```
FILES
       /etc/afp.conf
```

rdmark:
I wonder if the best solution is to stop pretending that the manual pages are aware of the exact location of config files?

Taking Samba as an example again, this is what the `samba` man page says about config files:

```
FILES
       /etc/rc
           or whatever initialization script your system uses.

           If running the server as a daemon at startup, this file will need to contain an appropriate startup sequence for the server.

       /etc/services
           If running the server via the meta-daemon inetd, this file must contain a mapping of service name (e.g., netbios-ssn) to service port
           (e.g., 139) and protocol type (e.g., tcp).

       /usr/local/samba/lib/smb.conf
           This is the default location of the smb.conf(5) server configuration file. Other common places that systems install this file are
           /usr/samba/lib/smb.conf and /etc/samba/smb.conf.

           This file describes all the services the server is to make available to clients. See smb.conf(5) for more information.
```

*The* authoritative source of truth for config file locations is `netatalk -V`

rdmark:
For now I revised the changeset and removed the man page substitution fix (sorry!)
My reasoning is that the old behavior is obfuscating but accurate (you have to know what the unresolved variable means), while the new behavior is potentially inaccurate while introducing an unexpected change in behavior for downstream packagers.

What do you think?

For the record I saved the commit hash for the change after squishing all of those commits: https://github.com/Netatalk/netatalk/pull/474/commits/a1dfe2b046819a11e919a6faf931809e368fa572

dgsga (approved):

dgsga:
> For now I revised the changeset and removed the man page substitution fix (sorry!) My reasoning is that the old behavior is obfuscating but accurate (you have to know what the unresolved variable means), while the new behavior is potentially inaccurate while introducing an unexpected change in behavior for downstream packagers.
> 
> What do you think?
> 
> For the record I saved the commit hash for the change after squishing all of those commits: [a1dfe2b](https://github.com/Netatalk/netatalk/commit/a1dfe2b046819a11e919a6faf931809e368fa572)

I'm fine with this. Look how Samba deals with the issue:

		<varlistentry>
		<term><filename>/usr/local/samba/lib/smb.conf</filename></term>
		<listitem><para>This is the default location of the <citerefentry><refentrytitle>smb.conf</refentrytitle>
		<manvolnum>5</manvolnum></citerefentry> server configuration file. Other common places that systems 
		install this file are <filename>/usr/samba/lib/smb.conf</filename> 
		and <filename>/etc/samba/smb.conf</filename>.</para>

So no variable substitution at all.

rdmark:
Raised https://github.com/Netatalk/netatalk/issues/475 to follow up...

PR #482, opened 2023-09-13
Default to xz compression for tarballs

dgsga:


rdmark:
That was easy! The gzip target can still be explicitly chosen if someone would want to:

```
-rw-r--r--  1 dmark dmark 1.3M Sep 13 13:54 netatalk-3.1.99dev.tar.bz2
-rw-r--r--  1 dmark dmark 1.7M Sep 13 13:54 netatalk-3.1.99dev.tar.gz
-rw-r--r--  1 dmark dmark 957K Sep 13 13:53 netatalk-3.1.99dev.tar.xz
```

rdmark (approved):

PR #483, opened 2023-09-13
[2.2] Default to xz compression for tarballs

rdmark:


rdmark:
FWIW the Ubuntu job failure is a false positive. This is caused by libcups packages that are missing from http://azure.archive.ubuntu.com/ubuntu ... I've seen this exact failure intermittently in the past. Seems to be a flakiness with Githubs cloud runners.

dgsga (approved):

PR #490, opened 2023-09-17
Remove unused pap directory in libatalk

dgsga:
Speaks for itself...


rdmark (approved):
FWIW if/when we forward-port all the appletalk functionality, the pap dir will make a comeback :)

PR #492, opened 2023-09-18
No quota support on macOS

dgsga:
This commit fixes a new compilation error on macOS 14 (Sonoma)

### DIFF
--- a/macros/netatalk.m4
+++ b/macros/netatalk.m4
@@ -701,6 +701,12 @@ if test x"$this_os" = "xsolaris"; then
 	init_style=solaris
 fi
 
+dnl ----- macOS specific -----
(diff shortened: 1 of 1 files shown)


rdmark (approved):
Shall we remove the Quota feature altogether? :)

dgsga:
> Shall we remove the Quota feature altogether? :)

I think so as storage capacity is not an issue with current SSD's and HDD's. If you're in agreement I'll strip it out (I can port a commit from my fork)

rdmark:
> > Shall we remove the Quota feature altogether? :)
> 
> I think so as storage capacity is not an issue with current SSD's and HDD's. If you're in agreement I'll strip it out (I can port a commit from my fork)

Yes I think so. Please create a ticket first.

PR #496, opened 2023-09-19
Remove quota support from netatalk 3

dgsga:


rdmark:
I gave it to the community until 9/25 to respond and object if they have actual use for Quota.

rdmark (approved):

PR #499, opened 2023-09-19
Dtrace support in netatalk fails to compile on aarch64 FreeBSD

dgsga:
Make distcheck now passes on aarch64 FreeBSD

### DIFF
--- a/macros/netatalk.m4
+++ b/macros/netatalk.m4
@@ -60,6 +60,18 @@ AC_DEFUN([AC_NETATALK_DTRACE], [
   fi
   ;;
   esac
+  dnl the version of dtrace on aarch64 FreeBSD is currently unsupported
+  case "$host_cpu" in
+  aarch64)
+  case "$host_os" in
+  *freebsd*)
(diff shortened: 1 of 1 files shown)


rdmark:
Do you know why it's unsupported? Any plans for it to be supported in the future?

FWIW without dtrace I think most the integration tests are moot.

rdmark:
Right, I get what you're saying. We also have the option to let the error be. There's a bit of a risk in hiding these things, I think. Harder for future contributors to know what's going on, and may smooth over other potential issues...

Do you think it's important that we can run the distcheck target on FreeBSD? I can be convinced either way really.

rdmark:
@dgsga https://wiki.freebsd.org/arm64#Known_Limitations <-- This states that "Userland SDT DTrace probes (usdt) do not work" which seems related. I only have surface level understanding though...

rdmark:
Approved. The only thing I'd like to suggest is to leave enough context in the commit message for future contributors to easily understand and trace the underlying limitation in FreeBSD to make it easy to revert if dtrace gets better support in the future.

PR #502, opened 2023-09-20
Replace gfx files corrupted by commit 06e427e

dgsga:
Original files restored


rdmark:
May I suggest simply reverting the change that broke the png, as well as the changes to the pdf file which are unwanted I think?

rdmark:
Alternative solution: remove the png from revision control. It's arguably a generated file.

dgsga:
> Alternative solution: remove the png from revision control. It's arguably a generated file.

I have replaced all three gfx files with the originals...

rdmark (approved):

PR #509, opened 2023-10-05
CVE-2022-22995: Harden create_appledesktop_folder()

rdmark:
* Check that source path is a dir before moving .AppleDesktop
* Use C APIs for renaming files rather than a system call to mv
* Call become_root() before the first lstat() call
Signed-off-by: Daniel Markstedt <daniel@mindani.net>

### DIFF
--- a/etc/afpd/desktop.c
+++ b/etc/afpd/desktop.c
@@ -12,8 +12,10 @@
 #endif /* HAVE_CONFIG_H */
 
 #include <stdio.h>
+#include <stdlib.h>
 #include <string.h>
 #include <ctype.h>
+#include <sys/stat.h>
 
 #include <errno.h>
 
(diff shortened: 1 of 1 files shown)


dgsga:
> * Check that source path is a dir before moving .AppleDesktop
> * Use C APIs for renaming files rather than a system call to mv
> * Call become_root() before the first lstat() call
>   Signed-off-by: Daniel Markstedt [daniel@mindani.net](mailto:daniel@mindani.net)

@rdmark Thanks for dealing with this one, TBH I knew how to start the fix but not complete it.

dgsga (approved):

rdmark:
@dgsga Happy to! Preparing the 3.1.18 release now but every step is slow because I'm running Linux under qemu on my MBP.

dgsga:
> @dgsga Happy to! Preparing the 3.1.18 release now but every step is slow because I'm running Linux under qemu on my MBP.

FYI Asahi Linux now runs with GPU acceleration on M2 Macs :) Also, check this out:
https://asahilinux.org/2023/08/fedora-asahi-remix/

rdmark:
This looks awesome. I will try to install this weekend.

PR #514, opened 2023-10-06
Remove substitutions for configurable file paths in man pages

rdmark:
Rewrite paragraphs that referred to specific configurable file system paths.


dgsga:
> Rewrite paragraphs that referred to specific configurable file system paths.

My only request would be that we revisit this design decision if we do eventually change to the meson build system. I do like the idea of creating  manpage and manual html pages that are system specific...

rdmark:
@dgsga Happy to revisit it at any time. :)

I agree that it's handy to have the man pages remind me where particular config files are on the file system. The one challenge to solve would be how to make cross-platform html manual pages. Perhaps a macro that hides particular elements when using the html style sheet (if that is possible)?

PR #522, opened 2023-10-10
[2.2] Rearrange the dummy file system paths for the tests

rdmark:
The new default location for autoconf (libdir) isn't valid on macOS.

### DIFF
--- a/test/afpd/.gitignore
+++ b/test/afpd/.gitignore
@@ -5,3 +5,7 @@ Makefile.in
 test
 test.conf
 test.default
+test.log
+test.trs
+test.sh.*
+test-suite.log
--- a/test/afpd/Makefile.am
+++ b/test/afpd/Makefile.am
@@ -1,6 +1,6 @@
 # Makefile.am for test/afpd/
 
-pkgconfdir = @PKGCONFDIR@
+tmpconfdir = /tmp/netatalk
 
 TESTS = test.sh test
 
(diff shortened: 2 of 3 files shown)


VorpalBlade:
Will this fix still place the UAMS in a sensible location on Linux? Or will they start going into /etc again?

dgsga:
> The new default location for autoconf (libdir) isn't valid on macOS.

Strange, make check on the current main branch runs fine on my M2 Mac...

rdmark:
> Will this fix still place the UAMS in a sensible location on Linux? Or will they start going into /etc again?
> 
> If so: What flags do I need to append to configure to make it sensible again?

This is only for the optional integration tests: `make check`

rdmark:
> > The new default location for autoconf (libdir) isn't valid on macOS.
> 
> Strange, make check on the 2.2 branch on my M2 Mac works fine. The duplicate library linker warning (one of many) is due to the new ld64 in Xcode 15`

That's bizarre. I wonder what's different on my M1 Mac. I'm on Ventura 13.6. Let me install the full Xcode suite and see if that makes a difference.

rdmark:
Installing Xcode 15.0 proper did not make a difference.

The error happens as the test script is generated, it seems:

```
# Test script is run here. We create the file first, then append to it,
# to ameliorate tests themselves also writing to the log file. Our tests
# don't, but others can (automake bug#35762).
: >"$log_file"
"$@" >>"$log_file" 2>&1
estatus=$?
```

rdmark:
Hang on, I didn't look at the log file. (Getting tunnel vision when tinkering late at night.) The error is actually happening when it tries to create afp_signature.conf. Back to the drawing board!

rdmark:
Even when addressing the permissions issue the test fails and the logs aren't helpful. 

```
==================================================
   netatalk 2.2.11dev: test/afpd/test-suite.log
==================================================

# TOTAL: 2
# PASS:  1
# SKIP:  0
# XFAIL: 0
# FAIL:  1
# XPASS: 0
# ERROR: 0

.. contents:: :depth: 2

FAIL: test
==========

Oct 11 21:35:38.029201 [5370] {afp_config.c:364} (N:AFPDaemon): AFP/TCP started, advertising 172.16.201.29:10548 (2.2.11dev)
FAIL test (exit status: 139)
```

rdmark:
Reading the code in test/afpd, I can only conclude that these file paths are dummy values that are just needed to compile afpd, and not relevant for actually running the tests. Therefore I went a slightly different direction and moved the paths to /tmp in this PR.

dgsga (approved):

PR #524, opened 2023-10-13
add return 0 into become_user_permanently()

### LINKED ISSUE #523: become_user_permanently(uid_t uid, gid_t gid) return value
... should be zero at the very end of the function. Right?

pgajdos:
Closes #523.


rdmark (approved):

rdmark:
Thanks for contributing!

PR #525, opened 2023-10-13
[2.2] add return 0 into become_user_permanently()

rdmark:


rdmark:
Not correct according to Sonar:

> void function 'become_user_permanently' should not return a value

PR #528, opened 2023-10-14
Deprecate legacy cnid backends

dgsga:
This commit removes legacy CNID backends that complicate the user experience without providing significant functional gain


doc/manpages/man5/afp.conf.5.xml
```diff
@@ -2214,19 +2214,6 @@
...
-      <varlistentry>
-        <term>cdb</term>
-
-        <listitem>
-          <para>"Concurrent database", backend is based on Oracle Berkeley DB.
-          With this backend several <command>afpd</command> daemons access the
-          CNID database directly. Berkeley DB locking is used to synchronize
-          access, if more than one <command>afpd</command> process is active
-          for a volume. The drawback is, that the crash of a single
-          <command>afpd</command> process might corrupt the database.</para>
```
rdmark, line 2226: As a side note, when we remove this paragraph the following paragraph for dbd reads kind of awkwardly. We may want to borrow the first sentence?

rdmark (approved):

PR #530, opened 2023-10-15
Remove code for obsolete host operating systems

dgsga:


etc/afpd/auth.h
```diff
@@ -31,11 +31,7 @@ static const struct afp_versions  afp_versions[] = {
 #define USERIBIT_ALL   (USERIBIT_USER | USERIBIT_GROUP | USERIBIT_UUID)
 
 extern uid_t    uuid;
-#if defined( sun ) && !defined( __svr4__ ) || defined( ultrix )
```
rdmark, line 34: I think the `__svr4__` macro is used for Solaris, which we do still support.

include/atalk/globals.h
```diff
@@ -10,10 +10,7 @@
 #include <grp.h>
 #include <sys/types.h>
 #include <stdbool.h>
-
-#ifdef HAVE_NETDB_H
-#include <netdb.h>  /* this isn't header-protected under ultrix */
-#endif /* HAVE_NETDB_H */
```
rdmark, line 16: Last I checked there were tons more `HAVE_NETDB_H` checks elsewhere in the codebase. I wonder if all of them can be removed, or if they still serve a purpose for another OS?

etc/afpd/auth.h
rdmark: Oops, I didn't notice the negative check on that macro! Please by all means get rid of this.

include/atalk/globals.h
dgsga: netdb.h is a widely used header in posix oses, and contains definitions for network database operations. macOS and Ubuntu have it for example. So good to keep I think...

rdmark: Right, that's what I thought!

So what I wanted to suggest is that we keep the ifdef and just remove the code comment.
I suppose the macro is there for non-posix systems that may not have this header...

```
#ifdef HAVE_NETDB_H
...
#endif /* HAVE_NETDB_H */
```

dgsga: Done.

rdmark (approved):

PR #546, opened 2023-10-21
Validate size of ace_count in FPSetACL request

rdmark:
Avoiding a potential buffer overrun if the value is too large

Credits: Robert Morris


rdmark:
Keeping this as Draft until it can be properly tested

rdmark:
Did a smoketest on Ubuntu 22.04.3 -- the ext4 file system has ACLs enabled. The afp_setacl function is called in the `UAM_AFPSERVER_POSTAUTH` phase. Authentication and connection to the shared volume works without error.

Beyond that, more ACL scenarios should probably be tested, but I have no use cases/experiences with ACLs myself. Ideas welcome!

dgsga (approved):

rdmark:
Backported to 2.x https://github.com/Netatalk/netatalk/pull/551

Issue #364, opened 2023-07-05
an FPSetACL request can crash afpd if the ace_count is too big


rtmrtmrtmrtm:
afp_setacl() reads the client's ace_count without checking against
the message size:

        uint32_t ace_count;
        memcpy(&ace_count, ibuf, sizeof(uint32_t));
        ace_count = htonl(ace_count);
        ibuf += 8;      /* skip ACL flags (see acls.h) */

        ret = set_acl(vol,
                      s_path->u_name,
                      (bitmap & kFileSec_Inherit),
                      (darwin_ace_t *)ibuf,
                      ace_count);

set_acl() passes ibuf (as darwin_aces) and ace_count to
map_aces_darwin_to_posix(), which will run off the end of the message
buffer and perhaps crash if ace_count is too big:

    for ( ; ace_count != 0; ace_count--, darwin_aces++) {
        darwin_ace_flags = ntohl(darwin_aces->darwin_ace_flags);

Here's a backtrace from a core dump:

#0  0x0000000000253ccb in map_aces_darwin_to_posix
    (darwin_aces=0x8423e9ff6, def_aclp=0x820273930, acc_aclp=0x820273928, ace_count=268391768, default_acl_flags=0x820273914) at acls.c:823
#1  0x0000000000251f90 in set_acl
    (vol=0x839b3ee20, name=0x25dca0 <mtoupath[upath]> "1", inherit=16, daces=0x8422ea036, ace_count=268435456) at acls.c:1344
#2  0x0000000000251a7a in afp_setacl
    (obj=0x2641a0 <obj>, ibuf=0x8422ea036 "", ibuflen=112, rbuf=0x8411736f0 "\001", rbuflen=0x8411836f0) at acls.c:1746
#3  0x0000000000219e27 in afp_over_dsi (obj=0x2641a0 <obj>) at afp_dsi.c:627
#4  0x0000000000242dc0 in dsi_start
    (obj=0x2641a0 <obj>, dsi=0x841173000, server_children=0x839b178c0)
    at main.c:474
#5  0x00000000002423f6 in main (ac=2, av=0x820273c88) at main.c:417

rdmark:
Thanks for reporting the bug!

Do you have concrete reproduction steps?

And if you have a fix in mind please by all means file a PR. We very much welcome code contributions. :)

rtmrtmrtmrtm:
To reproduce, you could temporarily modify afp_setacl() to set
[...]

PR #551, opened 2023-10-21
[2.3] Validate size of ace_count in FPSetACL request

rdmark:
Avoid potential crash in FPSetACL, https://github.com/Netatalk/netatalk/issues/364


dgsga (approved):

rdmark:
FWIW I tried to run a negative test case to see if the check caught a mismatched buffer size. However on Ubuntu I couldn't see evidence that the setacl function is ever called. I referenced this article to check that Linux ACLs were activated on the file system, and then used setfacl to set and revoke ACLs on dirs that were shared by afpd.

Before release some additional cross platform testing of ACL functionality should be carried out. (I never use it personally)

PR #556, opened 2023-11-09
Fix build issue with system talloc

dgsga:


rdmark:
@dgsga Interesting, so the linker now expects system talloc even when spotlight is disabled, if you disable bundled talloc? Or maybe that was always the case (bundled talloc was always built even when spotlight was disabled)...?

PR #578, opened 2023-12-08
Compile spotlight.h only when Spotlight is enabled

rdmark:
The dalloc.h header depends on talloc so it shouldn't be imported without the Spotlight toolchain.

Also, the symbols in spotlight.h aren't needed without Spotlight to let's ifdef them away (unless there's a more graceful way to do this?)

### DIFF
--- a/etc/afpd/spotlight.c
+++ b/etc/afpd/spotlight.c
@@ -27,16 +27,12 @@
 #include <inttypes.h>
 #include <time.h>
 #include <utime.h>
+#include <talloc.h>
 
 #include <atalk/list.h>
 #include <atalk/errchk.h>
 #include <atalk/util.h>
 #include <atalk/logger.h>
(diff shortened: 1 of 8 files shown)


dgsga:
@rdmark, please can we use the cleaner fix in PR #579 as the code changes are more consistent with how other headers are treated e.g those associated with ACL support?

rdmark:
@dgsga Sure, we can use your fix.

In fact, why don't we do both? My approach is consistent with f.e. ldapconfig.h and the `HAVE_LDAP` macro: https://github.com/Netatalk/netatalk/blob/main/include/atalk/ldapconfig.h

There may be a good reason not to, but I thought it was worth a conversation. :)

dgsga (approved):

dgsga:
@rdmark, I have merged your PR and only the macOS workflow fix component of mine as this seems the best of both worlds.

PR #583, opened 2023-12-15
Introduce option to control the format of file log timestamp

rdmark:
Introduces a boolean afp.conf option "log microseconds", enabled by default, which turns on and off microsecond timestamps in file logs.

### DIFF
--- a/bin/ad/ad.c
+++ b/bin/ad/ad.c
@@ -25,6 +25,7 @@
 #include <signal.h>
 #include <string.h>
 #include <errno.h>
+#include <stdbool.h>
 
 #include <atalk/cnid.h>
 #include <atalk/logger.h>
@@ -56,7 +57,7 @@ int main(int argc, char **argv)
     if (afp_config_parse(&obj, "ad") != 0)
         return 1;
 
-    setuplog("default:note", "/dev/tty");
+    setuplog("default:note", "/dev/tty", true);
 
     if (load_volumes(&obj, LV_DEFAULT) != 0)
         return 1;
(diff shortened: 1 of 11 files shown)


rdmark:
The four code level bugs aren't newly introduced, I believe. Sonar reports them because the surrounding code has been changed.

bin/ad/ad.c
```diff
@@ -56,7 +56,7 @@ int main(int argc, char **argv)
     if (afp_config_parse(&obj, "ad") != 0)
         return 1;
 
-    setuplog("default:note", "/dev/tty");
+    setuplog("default:note", "/dev/tty", NULL);
```
slowfranklin, line 59: Isn't this a bool parameter and should be "false", not "NULL"?

slowfranklin: Other then that, lgtm.

rdmark: Thanks for reviewing. I pushed a fix that uses bool "false" for each call of setuplog. Needed to import stdbool.h for a few of the source files.

rdmark: I will squash the commits shortly.

rdmark:
@slowfranklin After sleeping on it, I propose that we flip the boolean: Name the option "log microseconds", make it "yes" by default. (Still maintain the same default behavior as before this change.)

It makes more logical sense that the option is additive: When enabled something gets appended. When disabled something gets removed.

It's a minor nuance, but I think it will make it easier for users to understand the option.

Thoughts?

slowfranklin:
On 12/16/23 11:16, Daniel Markstedt wrote:
> Thoughts?

fine with me. Thansk!

rdmark:
@slowfranklin Commits squished and ready approval!

slowfranklin (approved):

dgsga (approved):

PR #597, opened 2023-12-18
atalkd: Introduce 3rd party bridge quirks mode

rdmark:
- Introduce a `-q` parameter for atalkd which stops AppleTalk packet propagation from subnets under other routers. This improves compatibility with Asante and Dayna hardware AppleTalk bridges.
- Removed the unused quiet and chatty parameters
- A more helpful error message when using unknown parameters
- Fleshed out the atalkd man page

### DIFF
--- a/etc/atalkd/main.c
+++ b/etc/atalkd/main.c
@@ -82,7 +82,7 @@ static int		atservNATSERV = elements( atserv );
 
 struct interface	*interfaces = NULL, *ciface = NULL;
 
-static int		debug = 0, quiet = 0, chatty = 0;
+static int		debug = 0, quirks = 0;
 static char		*configfile = NULL;
 static int		ziptimeout = 0;
 static int		stable = 0, noparent = 0;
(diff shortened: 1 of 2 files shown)


IvanExpert:
I'll check this out this coming weekend with Asante and Dayna bridges.

dgsga:
@rdmark, I'm hesitant to approve this simply because I know next to nothing about AppleTalk and the DDP protocol. Maybe we could invite @NJRoadfan or @IvanExpert to review?

NJRoadfan:
This is causing more problems then the previous patch. Now I'm randomly losing the ZIP broadcasts on networks. It appears to be a logic error. The line `} else if (rtmp->rt_iface == iface) {` is likely still executing when "quirks mode" is enabled.

etc/atalkd/main.c
```diff
@@ -547,7 +547,14 @@ static void as_timer(int sig _U_)
...
 		    /* split horizon */
-		    if (rtmp->rt_iface == iface) {
+
+		    /* Makes the AsanteTalk bridge consistently start up in
+		     * AppleTalk Phase 2, and stop the Dayna bridge from
+		     * crashing GS/OS.
+		     */
+		    if (quirks && (rtmp->rt_iface != iface)) {
+		        continue;
+		    } else if (rtmp->rt_iface == iface) {
```
NJRoadfan, line 557: This line is likely still executing when "quirks mode" is enabled.

Alter line to `} else if (!quirks && rtmp->rt_iface == iface) {`

rdmark: Ah, of course. The dangers of coding late at night.

Fixed.

NJRoadfan (approved):
Revisions appear to be working in test environment. Running atalkd with and without the "quirks" mode argument tested and appear consistent with previous behavior of the daemon.

PR #600, opened 2023-12-18
Fix Spotlight support detection and compilation on Fedora

dgsga:
Tested on Fedora39 x86_64 and aarch64 with configuration as per the wiki. Also tested on FreeBSD14, Debian12

### DIFF
--- a/etc/afpd/afp_dsi.c
+++ b/etc/afpd/afp_dsi.c
@@ -42,7 +42,9 @@
 #include <atalk/fce_api.h>
 #include <atalk/globals.h>
 #include <atalk/netatalk_conf.h>
+#ifdef WITH_SPOTLIGHT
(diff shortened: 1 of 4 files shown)


rdmark (approved):

rdmark:
@dgsga Is there an argument to made to remove the HAVE_TRACKER macro in favor of WITH_SPOTLIGHT? There is no other use for Tracker beyond Spotlight functionality, is there?

dgsga:
> @dgsga Is there an argument to made to remove the HAVE_TRACKER macro in favor of WITH_SPOTLIGHT? There is no other use for Tracker beyond Spotlight functionality, is there?

There is, but I think we should wait until we decide how to manage earlier versions of tracker. This will have a big effect on the Spotlight macro.

PR #601, opened 2023-12-19
Improve logger_test and remove outdated comments

rdmark:
Improved logger_test:
- log type identifiers should be lower case
- Uncommended the syslog tests, and brought their logging API calls up to date
- Added a switch to the default log scope when changing from syslog to file log, otherwise the file logging didn’t take effect.
- Fleshed out and numbered the test case outputs

Additionally: removed inline documentation of the log API that was outdated. The code itself is readable enough to serve as documentation.

### DIFF
--- a/bin/misc/.gitignore
+++ b/bin/misc/.gitignore
@@ -7,5 +7,4 @@ logger_test
 .deps
 .libs
 *.o
-
-
+*.log
--- a/bin/misc/Makefile.am
+++ b/bin/misc/Makefile.am
@@ -11,7 +11,7 @@ netacnv_LDADD = $(top_builddir)/libatalk/libatalk.la @MYSQL_LIBS@
 logger_test_SOURCES = logger_test.c
 logger_test_LDADD = $(top_builddir)/libatalk/libatalk.la @MYSQL_LIBS@
 
-fce_SOOURCE = fce.c
+fce_SOURCES = fce.c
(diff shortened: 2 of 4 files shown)


dgsga:
@rdmark, I'm probably missing something obvious here but when I compile and run logger_test on macOS it gives no visible output. Hmm...

rdmark:
@dgsga the first four tests should output to syslog, and the rest of them should output to a file called ./test.log

It’s probably a good idea to have the app output something to this effect to stdout. 

Are you able to see logs in both places on macOS?

dgsga:
> @dgsga the first four tests should output to syslog, and the rest of them should output to a file called ./test.log
> 
> It’s probably a good idea to have the app output something to this effect to stdout.
> 
> Are you able to see logs in both places on macOS?

I get the test.log output but no changes in Console.app -> system.log

rdmark:
That’s curious. The syslogs show up on Debian. 

I assume that syslogs generated by afpd can be seen on macOS? I don’t think I ever checked.

dgsga:
> That’s curious. The syslogs show up on Debian.
> 
> I assume that syslogs generated by afpd can be seen on macOS? I don’t think I ever checked.

No afpd generated syslog viewable on macOS. I have approved the PR so you can push if you're OK with this...

dgsga (approved):

rdmark:
@dgsga I filed https://github.com/Netatalk/netatalk/issues/615 for logging on macOS.

PR #611, opened 2023-12-23
[2.3] Remove cdb and tdb CNID backends

rdmark:
- Removed cdb and tdb CNID backends
- Removed in-memory tdb backend fallback
- Updated documentation and code comments
- Removed dead code


dgsga (approved):

rdmark:
Unfortunately the error handling when the cnid_metad process is dead isn't good. You can connect to the shared volume, but the volume is shown as empty. Some more work is needed here...

Screenshot 2023-12-24 at 22 59 13

dgsga (approved):

dgsga (approved):

PR #628, opened 2023-12-26
Fixing bugs that can cause system errors

neko-hat:
It seems unlikely that a bug will occur, but it seems that a patch is needed just in case.

```
diff --git a/etc/afpd/auth.c b/etc/afpd/auth.c
index 15e16441..89a2a3b3 100644
--- a/etc/afpd/auth.c
+++ b/etc/afpd/auth.c
@@ -827,7 +827,7 @@ int afp_logout(AFPObj *obj, char *ibuf _U_, size_t ibuflen  _U_, char *rbuf  _U_
  */
 int afp_changepw(AFPObj *obj, char *ibuf, size_t ibuflen, char *rbuf, size_t *rbuflen)
 {
-    char username[MACFILELEN + 1], *start = ibuf;
+    char username[MAXUSERLEN], *start = ibuf;
     struct uam_obj *uam;
     struct passwd *pwd;
     size_t len;
```
It could be overflowed when uam_getname is called in afp_changepw. (that function is in etc/afpd/uam.c)
(name buf size is 32 but MAXUSERLEN is 255)

### DIFF
--- a/etc/afpd/auth.c
+++ b/etc/afpd/auth.c
@@ -827,7 +827,7 @@ int afp_logout(AFPObj *obj, char *ibuf _U_, size_t ibuflen  _U_, char *rbuf  _U_
  */
 int afp_changepw(AFPObj *obj, char *ibuf, size_t ibuflen, char *rbuf, size_t *rbuflen)
 {
-    char username[MACFILELEN + 1], *start = ibuf;
+    char username[MAXUSERLEN], *start = ibuf;
     struct uam_obj *uam;
     struct passwd *pwd;
     size_t len;


etc/afpd/auth.c
```diff
@@ -827,7 +827,7 @@ int afp_logout(AFPObj *obj, char *ibuf _U_, size_t ibuflen  _U_, char *rbuf  _U_
  */
 int afp_changepw(AFPObj *obj, char *ibuf, size_t ibuflen, char *rbuf, size_t *rbuflen)
 {
-    char username[MACFILELEN + 1], *start = ibuf;
+    char username[MAXUSERLEN + 1], *start = ibuf;
```
rdmark, line 830: I don’t think you need the + 1 here since MAXUSERLEN is defined as 256 already.

rdmark (changes requested):

etc/afpd/auth.c
neko-hat: Why did I see it as 255? I modified it

neko-hat: Why did I see it as 255? I modified it.
@rdmark

neko-hat:
@rdmark  I completed the revision by reflecting your feedback.
Please review it.

rdmark:
@neko-hat I think the code looks good now, thanks.

Now, I have one more ask for you: In this project we have policies for the changeset and commit messages. The goal is to have a descriptive and high quality historical commit log in the project. So, can you please squash this changeset into one commit, and write a descriptive commit message in the format described in the below wiki page?

https://github.com/Netatalk/netatalk/wiki/Developer-Notes

Basically, the commit message should say something like "afpd: Use correct username length in afp_changepw, Issue #XYZ" (with the actual issue number)

Are you familiar with how to reset the HEAD and force push in git?

neko-hat:
@rdmark  I'm sorry I'm not good at it. I changed it as you said, is it okay?

rdmark (approved):
LGTM, thanks!

PR #631, opened 2023-12-28
[2.3] Clean up residual svn commit headers

rdmark:
These subversion generated code comments were removed long ago in 3.x /head branches.


rdmark:
Self-merging trivial non-code change.

PR #634, opened 2023-12-30
Create OmniOS, FreeBSD OpenBSD and NetBSD workflows

dgsga:


dgsga:
Notes:
Jobs run in parallel by default :)
This is a WIP as have yet to add Solaris & DragonflyBSD. Solaris is proving troublesome...
make check fails for some OSes in the VMactions VM's but not in VM's run on a bare metal hypervisor. Not sure why yet.

rdmark (approved):
Approving in advance. Please go ahead and merge or continue to work on the additional targets. 

I assume you’re aiming for Solaris 11 in the workflow?

PR #635, opened 2023-12-31
Create DragonflyBSD and Solaris workflows

dgsga:


dgsga:
Notes:
The Solaris VM has out-of-date versions of autoconf and automake, hence need to update them. I'll post this as an issue with the maintainer.
Have to compile Netatalk --without-afpstats as otherwise the build fails. I think this is due to an outdated version of glib on the VM which would take too long to replace.

rdmark (approved):
You made it happen!

PR #637, opened 2024-01-01
636 test program should only be compiled and run on make check/distcheck

dgsga:


.github/workflows/build.yml
```diff
@@ -4,6 +4,7 @@ on:
     branches:
       - main
       - branch-*
+      - 636-*
```
rdmark, line 7: I assume this line is for testing and will be reverted before merging?

dgsga: Fixed

rdmark (approved):

PR #646, opened 2024-01-05
[2.3] afpd: Use explicity sized variable types

rdmark:


rdmark:
I *think* the newly introduced static analysis issues are bogus. The `u_int64_t` data type evaluates to `unsigned long long` on 64 bit systems, and not `unsigned long` as Sonar claims. Similar situation with the code smell issue in uam.c – `tv
.tv_sec` evaluates to long and not int as Sonar claims.

This will depend on the standard library implementation on each platform of course, but isn't it simply safer to use a larger data type in these cases?

dgsga (approved):

rdmark:
Discarding this PR. Solaris 11.4 doesn't like the explicit data types. Importing stdint.h didn't help.

PR #651, opened 2024-01-05
[2.3] papd: Update cups_print_job() to use CUPS destination API

NJRoadfan:
The function cupsPrintFile() is being removed in CUPS 3.0. Use the CUPS destination API to send print jobs from papd from now on.

### DIFF
--- a/etc/papd/print_cups.c
+++ b/etc/papd/print_cups.c
@@ -534,6 +534,9 @@ cups_get_printer_status (struct printer *pr)
 
 int cups_print_job ( char * name, char *filename, char *job, char *username, char * cupsoptions )
 {
+	http_t          *http;          /* HTTP connection to server */
+	cups_dest_t	*dest = NULL;	/* Destination */
+	cups_dinfo_t 	*info = NULL;
 	int 		jobid;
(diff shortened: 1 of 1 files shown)


NJRoadfan:
I have tested this and it appears to work. I would like others to test this and see if it is working ok with their setups, particularly if they have an AirPrint compatible printer using dynamic printer queues.

To fully test this, properly:

- Stop `cups-browsed` if it is installed and running with `sudo systemctl stop cups-browsed`
- Restart papd. Make sure all the printers on your network appear when running `nbplkup`
- Attempt to print from a Macintosh client.

rdmark:
I have an HP Tango X that I can test on tomorrow. Dunno if it has support for dynamic printer queues though.

rdmark (approved):

rdmark:
@NJRoadfan Seems to work in my environment when cups-browsed is shut down (and when it's running). Text and bitmap images. Laserwriter 7 on System 7.1.1 and Laserwriter 8 on MacOS 9.2.

rdmark:
The host is running Raspbian Bullseye and cups 2.3.3op2

PR #653, opened 2024-01-07
[2.3] workflow: Compile on Debian with cracklib, ldap, quota

rdmark:
Adding a bunch of deb packages to enable configuring and compiling cracklib and ldap/acl on Ubuntu/Debian.

The quota autoconf macro doesn't seem to work anymore in Debian bookworm (or Ubuntu 22.04) so using Debian bullseye

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -50,7 +50,7 @@ jobs:
     name: Debian
     runs-on: ubuntu-22.04
     container:
-      image: debian:oldstable
+      image: debian:latest
     steps:
       - uses: actions/checkout@v3
       - name: Install dependencies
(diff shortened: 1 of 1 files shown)


dgsga:
@rdmark, just added the missing --with-libtirpc flag and quota support compiles fine on Debian Bookworm :)

dgsga (approved):

rdmark:
@dgsga Right, thanks for the reminder. That option had temporary slipped my mind, despite it being me who added it a few months ago… in fact, —with-libtirpc got quota compiling on Ubuntu too!

Adding a Debian job to the v3 branch seems like a decent idea. I think it tends to be slightly more conservative with library version bumps compared to Ubuntu. Maybe even reconsidering using the oldstable image to capture an older tech stack that a lot of users are surely still using.

PR #655, opened 2024-01-07
[2.3] papd: Update cups_autoadd_printers() to use current API call.

NJRoadfan:
Replace deprecated cupsGetDests() with cupsGetDests2(). The original cupsGetDests() function is due to be removed in CUPS 3.0. Confusingly, cupsGetDest2() will be renamed to cupsGetDest() in libcups3, so we'll need a define macro for this in the future in order to maintain compatibility with both libcups2 and libcups3.

### DIFF
--- a/etc/papd/print_cups.c
+++ b/etc/papd/print_cups.c
@@ -641,7 +641,7 @@ cups_autoadd_printers ( struct printer	*defprinter, struct printer *printers)
 	char 	    	name[MAXCHOOSERLEN+1], *p;
 
         language  = cupsLangDefault();		/* needed for conversion */
-        num_dests = cupsGetDests(&dests);	/* get the available destination from CUPS */
+        num_dests = cupsGetDests2(CUPS_HTTP_DEFAULT, &dests);	/* get the available destination from CUPS */
 
         for  (i=0; i< num_dests; i++)
         {


rdmark (approved):
So if I understand it correctly, `cupsGetDests2()` exists in all versions of cups 2.x, but not in cups 1.x? 

I wonder if 3.x will have both variants but flag the one with the “2” postfix as to-be-deprecated?

rdmark:
Another remark: this is probably another nail in the coffin for cups 1.x backwards compatibility.

NJRoadfan:
`cupsGetDests2()` was added to CUPS 1.1.21. Here are the differences.

CUPS 1.x-2.x

`int cupsGetDests (
    [cups_dest_t](cups_dest_t) **dests
);`

`int cupsGetDests2 (
    http_t *http,
    [cups_dest_t](cups_dest_t) **dests
);`

All the `cupsGetDests()` function does is call `cupsGetDests2()` with `CUPS_HTTP_DEFAULT` as the http connection variable.

CUPS 3.x

`int cupsGetDests (
    http_t *http,
    [cups_dest_t](cups_dest_t) **dests
);`

All code in papd's `print_cups.c` should be fully compatible with CUPS 1.6 to 2.5. A short #define macro will smooth out the differences between 3.x.

PR #658, opened 2024-01-08
Add Debian VM and improve readability of GitHub workflow

dgsga:


dgsga:
Needed some surprising additions to both configure flags and apt-get install for compilation to succeed on the oldstable container...

rdmark:
@dgsga Just a thought: how about “hard coding” the list of packages to install with apt for oldstable, and keep the global list of packages as-is? The former list is kind of a temporary workaround for the current Debian release cycle. I think it will look cleaner and be easier to maintain. What do you think?

dgsga:
@rdmark, old stable packages are now hardcoded. I have also beautified the entire yaml document so it is now easy to see which packages are installed and which configure flags are used on each operating system.

rdmark (approved):
Gorgeous. :)

PR #661, opened 2024-01-08
Remove unneeded HAVE_TRACKER conditional and warn if Spotlight components are missing at configure time

dgsga:
The HAVE_TRACKER3 conditional remains to distinguish whether tracker v3 is present vs v1 / v2


dgsga:
PR closed as need to deal with an issue

PR #664, opened 2024-01-09
Remove unneeded HAVE_TRACKER conditional

dgsga:
Refactor Spotlight macro

Warn if any Spotlight support components are missing at configure time

### DIFF
--- a/etc/netatalk/netatalk.c
+++ b/etc/netatalk/netatalk.c
@@ -369,18 +369,18 @@ static void show_netatalk_version( void )
 
 static void show_netatalk_paths( void )
 {
-	printf( "              afp.conf:\t%s\n", _PATH_CONFDIR "afp.conf");
-	printf( "                  afpd:\t%s\n", _PATH_AFPD);
-	printf( "            cnid_metad:\t%s\n", _PATH_CNID_METAD);
+  printf( "              afp.conf:\t%s\n", _PATH_CONFDIR "afp.conf");
+  printf( "                  afpd:\t%s\n", _PATH_AFPD);
+  printf( "            cnid_metad:\t%s\n", _PATH_CNID_METAD);
+  printf( "           dbus-daemon:\t%s\n", DBUS_DAEMON_PATH);
   
 #ifdef WITH_SPOTLIGHT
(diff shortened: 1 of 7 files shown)


dgsga:
I have refactored the Spotlight macro so it meets all the following conditions:

1. If only tracker installed: Spotlight support = no (configure warns talloc and tracker SPARQL are not found)
2. If only tracker SPARQL installed: Spotlight support = no (configure warns talloc and tracker are not found)
3. If only talloc installed: Spotlight support = no (configure warns tracker and tracker SPARQL are not found)
4. If tracker & tracker SPARQL installed: Spotlight support = no (configure warns talloc is not found)
5. If only tracker & talloc installed: Spotlight support = no (configure warns tracker SPARQL is not found)
6. If only talloc & tracker SPARQL installed: Spotlight support = no (configure warns tracker not found)
7. If all three are present: Spotlight support=yes (no warnings at configure time

etc/netatalk/netatalk.c
```diff
@@ -369,18 +369,14 @@ static void show_netatalk_version( void )
...
 static void show_netatalk_paths( void )
 {
+	printf( "              afp.conf:\t%s\n", _PATH_CONFDIR "afp.conf");
 	printf( "                  afpd:\t%s\n", _PATH_AFPD);
 	printf( "            cnid_metad:\t%s\n", _PATH_CNID_METAD);
-
-#ifdef HAVE_TRACKER
-	printf( "       tracker manager:\t%s\n", TRACKER_PREFIX "/bin/" TRACKER_MANAGING_COMMAND);
+  
+#ifdef WITH_SPOTLIGHT
```
rdmark, line 376: Isn’t dbus strictly speaking used for other functionality too, even if you’re not building spotlight? E.g. afpstats

rdmark (approved):
I like that we’re being more verbose now when other spotlight dependencies are missing!

etc/netatalk/netatalk.c
dgsga: Well spotted. I'll look into it as just moving bus-daemon path outside of the Spotlight conditional causes compilation failure.
dbus-session.conf is Spotlight only.

PR #673, opened 2024-01-13
[2.3] afpd: Remove support for AppleDouble osx and sfm

rdmark:
The osx and sfm AppleDouble variants are long-obsoleted. The former was a late-era AD format that OSX 10.3 used, but isn't backwards or forwards compatible. The latter was an early hack for Samba interoperability, but relied on a hacked version of Samba.

Both were removed in v3.0 of netatalk


rdmark:
I tested that we can still read and write with AppleDouble v1 and v2

dgsga (approved):

Issue #679, opened 2024-01-18
[2.3] atalkd: The atalkd service is not sending RTMP broadcasts if "-router" switch is used.


NJRoadfan:
If the machine is configured for one network interface and the `-router` switch is used to seed that network, atalkd is not sending out RTMP broadcasts as it should. The "quirks" mode switch however fixes this.

This really should be fixed as it is a bug in atalkd. If one is using multiple interfaces and seeding networks with atalkd with `-seed`, those networks are getting RTMP broadcasts as they should. This would also likely explain why various applications and hardware are crashing with atalkd seeding a network without quirks mode.

This can easily be reproduced with a Macintosh computer as well. Approximately 80 seconds after connecting to a network managed by atalkd with the '-router' switch, the computer will display a message that your connection to the network has been interrupted. At this point the machine may act erratic, lose track of AppleTalk zones and devices, and sometimes crash.

NJRoadfan:
Look back at the commit history, the "split horizon" code wasn't introduced until 2009. This is well after AppleTalk was in production use in most networks, so this bug likely went unnoticed.

PR #680, opened 2024-01-18
[2.3] atalkd: Fix for RTMP broadcast bug.

NJRoadfan:
If the (only) interface was configured with the `-router` switch, automatically enable quirks mode. Otherwise don't. This is a hack as the split horizon logic is clearly broken here. If a single interface is configured as a router, it should ALWAYS broadcast RTMP packets no matter what.

Note: This patch doesn't remove the "quirks mode" command line switch or manpage documentation.

### DIFF
--- a/doc/manual/man/man8/atalkd.8.xml
+++ b/doc/manual/man/man8/atalkd.8.xml
@@ -5,7 +5,7 @@
 
     <manvolnum>8</manvolnum>
 
-    <refmiscinfo class="date">19 Dec 2023</refmiscinfo>
+    <refmiscinfo class="date">18 Jan 2024</refmiscinfo>
 
     <refmiscinfo class="source">:NETATALK_VERSION:</refmiscinfo>
   </refmeta>
@@ -51,8 +51,6 @@
       <arg>-2</arg>
 
       <arg>-d</arg>
-
-      <arg>-q</arg>
 
       <arg>-t</arg>
 
@@ -154,24 +152,6 @@
         <listitem>
           <para>Specifies the file in which <command>atalkd</command> stores its
           process id.</para>
-        </listitem>
-      </varlistentry>
-
-      <varlistentry>
-        <term>-q</term>
-
-        <listitem>
-          <para>Stops <command>atalkd</command> from broadcasting RTMP routing tuples
-	      to other AppleTalk routers on the network. This improves compatibility with
-	      AsantéTalk and Dayna EtherPrint bridge devices. However,
-	      clients behind other routers on the network can't see or communicate
-	      with network segments that <command>atalkd</command> is routing.</para>
-
-	  <note>
-	    <para>Do NOT use this option unless you absolutely need it.
-	        It breaks the AppleTalk specification.
-	        Ref. page 5-11 in "Inside AppleTalk".</para>
-	  </note>
         </listitem>
       </varlistentry>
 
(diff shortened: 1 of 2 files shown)


NJRoadfan:
Fixes #679. I have tested this patch in various configurations in a multi-router environment and it appears to work ok.

rdmark:
@NJRoadfan Do I understand it correctly that the issue that Steven attempted to resolve with his original split horizon hack — Asante and Dayna bridges causing trouble in GS/OS and not starting up in the right AppleTalk phase mode — is happening only with the `-router` switch enabled in the first place?

Meaning that the quirks mode doesn’t have be manually controlled ever again after this provisional fix?

NJRoadfan:
That would be correct. I discovered the problem when running in `-router` mode because my ethernet connected Mac would just throw up the "You've lost your connection to your AppleTalk network" dialog. So I traced network traffic, and noticed that RTMP packets weren't going out and the Mac would always display that message about 80 seconds after bringing up the network. Clearly it was timing out.

As for the DaynaPrint and AsantéTalk bridges, I have a hunch on each.

First off, the Dayna issue was likely a GS/OS problem. The OS likely got confused that it got zone data, but no RTMP information. Heck, it causes weird behavior on MacOS 8 and 9, like locking up the AppleTalk control panel and sending out corrupt NBP lookups!

As for the AsantéTalk, it is known to look for network traffic on boot to determine whether to run in Phase 1 or Phase 2 mode. @IvanExpert said the patch wasn't required, but made the device consistently work. Having regular Phase 2 RTMP updates every 10 seconds on the network certainly fits that theory.

I also have a theory why the maintainers never hit this bug. If you run in `-router` mode and have another router on the network, you would never notice a problem as that router would be broadcasting RTMP packets advertising the atalkd seeded network! (other routers can still seed from atalkd) The only time this bug manifested is if you are using atalkd in '-router' mode and had no other routers on the network.

NJRoadfan:
Also, I noticed that the `-d` switch for atalkd is completely broken. The service hangs on start. Suggest reverting all changes in #597 bringing back "quiet" and "chatty" mode.

rdmark:
That is some impressive sleuthing work. I’m impressed. The default behavior of atalkd is routerless, IIRC? Meaning only users who explicitly enables the router and has no other AppleTalk routers on the network would have experienced the bug?

rdmark:
> Also, I noticed that the `-d` switch for atalkd is completely broken. The service hangs on start. Suggest reverting all changes in #597 bringing back "quiet" and "chatty" mode.

I’m good with reverting the logic changes, if it makes debug mode work again. I want to keep the improvements to the atalkd manual page I made in that PR, though. 

I couldn’t see quiet and chatty being referenced anywhere in the code, so I assumed they were vestigial, from a pre-CVS era. But maybe they’re part of some C macro black magic?

rdmark:
@NJRoadfan May I ask you to update the atalkd manual page xml, removing the -q option paragraph, in this PR? I want to aim for a clean commit history where each increment is self contained. Also protects against us forgetting to updating it later. :)

No need to regenerate the man page with xsltproc though. We do that as a batch job before tagging the next release.

rdmark:
The Fedora Rawhide runner is failing because of a pointer type incompatibility. I don't think it's related to this change, so raised https://github.com/Netatalk/netatalk/issues/682 for fixing later.

rdmark (approved):

PR #683, opened 2024-01-19
[2.3] libatalk: Appease Fedora gcc incompatible pointer error

rdmark:


rdmark:
The Sonar job failure is because we're using the deprecated v1 of the sonarscanner runner. Bumping to v2 in https://github.com/Netatalk/netatalk/pull/685

dgsga (approved):

PR #685, opened 2024-01-19
[2.3] Workflow: Bump SonarScanner to v2

rdmark:


rdmark:
Bypassing the CR process to unblock the CI pipeline.

PR #686, opened 2024-01-19
[2.3] Remove obsoleted Kerberos v4 UAM

rdmark:


rdmark:
@NJroadfan Please note that this removes Kerberos v4 auth for papd. I don't think it was functional anymore anyways, but I could be wrong. Let me know if you object to this change.

dgsga (approved):

NJRoadfan:
Fine with me. PAP has other authentication methods that likely work fine.

PR #696, opened 2024-01-23
[2.3] papd: update cups_get_printer_status()

NJRoadfan:
More code cleanup. Switch to using solely the CUPS Destination API to speak with the server instead of doing IPP queries ourselves. If we have a DNS-SD discovered printer, we query it directly for the status.

### DIFF
--- a/etc/papd/print_cups.c
+++ b/etc/papd/print_cups.c
@@ -96,12 +96,13 @@ cups_printername_ok(char *name)         /* I - Name of printer */
 	cups_dest_t	*dest = NULL;	/* Destination */
         ipp_t           *request,       /* IPP Request */
                         *response;      /* IPP Response */
+        //char            uri[HTTP_MAX_URI]; /* printer-uri attribute */
 
        /*
         * Make sure we don't ask for passwords...
         */
 
-        cupsSetPasswordCB2(cups_passwd_cb, NULL);
+        cupsSetPasswordCB(cups_passwd_cb, NULL);
 
 	/*
 	 * Try to connect to the requested printer...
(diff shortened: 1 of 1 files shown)


etc/papd/print_cups.c
```diff
@@ -531,16 +505,9 @@ int cups_print_job ( char * name, char *filename, char *job, char *username, cha
...
-	{
-		LOG(log_error, logtype_papd,
-		    "Unable to connect to destination \"%s\": %s", dest->name, cupsLastErrorString());
-		cupsFreeDests(1,dest);
-		return (0);
-	}
 
 	info = cupsCopyDestInfo(CUPS_HTTP_DEFAULT, dest);
-
+
```
rdmark, line 510: Nitpick: Seems unnecessary to add indentation here. ;)

NJRoadfan: Blame Microsoft and their editor that doesn't show tab stops.

NJRoadfan:
I will be testing this PR with an actual IPP Everywhere printer this weekend. Code seems to work, but I might refactor it into something a bit cleaner.

rdmark:
Sounds good. I’ll test it with my printer tomorrow.

rdmark:
@NJRoadfan I assume you would like to have this changeset in the upcoming 2.3.1 release?

NJRoadfan:
Yes, if possible. Only reason why I'm holding off on this one is testing. Hitting the printer directly for status has a latency hit in some cases. Don't want to introduce any unintended crashes on the client side if it gets something it doesn't like.

rdmark:
@NJRoadfan If there is a potential performance impact concern, that could be an argument for waiting until the next point release. We have a good changeset lined up already. Up to you though. I would like to tag 2.3.1 this coming weekend.

NJRoadfan:
I've given this a through shakedown. I simulated the worst possible conditions possible with the equipment I have. The setup was minivmac running on WiFi via LToUDP, printing to a IPP Everywhere printer also via WiFi. I can tell you its pretty slow, but the status updates worked fine. No crashing or odd behavior from the LaserWriter 8 driver on System 7.5.

rdmark (approved):

PR #700, opened 2024-01-26
[2.3] papd: Refactor and cleanup cups_get_printer_ppd()

NJRoadfan:
Clean up code to query a printer for creating a temporary PPD file. We no longer exploit cupsConnectDest() to create a temporary queue for DNS-SD discovered printers to query printer data. Instead we check for what type of printer the destination is and query it direct.

In addition indentation of code was cleaned up for readability.


NJRoadfan:
Gave this code a good shakedown with two different makes of printers. Should be ready to go.

rdmark (approved):
Looks reasonable, thanks

PR #701, opened 2024-01-31
[2.3] manual: Remove obsoleted html-upload target, and update css URL

rdmark:
Small updates to accommodate the move away from sourceforge hosting


rdmark:
Bypassing code review to unblock release

PR #704, opened 2024-02-04
[2.3] Create Dockerfile and entry point

rdmark:
Dockerfile and entry point script for running an AFP server + AppleTalk services.

Based on the original POC by eharmon

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,9 @@
+.git
+.github
+.gitignore
+Dockerfile
+docker-compose.yml
+sonar-project.properties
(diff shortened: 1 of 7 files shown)


rdmark:
I think the two security hotspots reported by Sonar are unavoidable by present constraints. All operations require root access, while we need all files in the tarball/git repo to perform a build...

rdmark:
pulling back to draft... there's a bug when connecting via Avahi Zeroconf

dgsga (approved):

PR #706, opened 2024-02-07
Fix 'MAXPHYS' macro redefinition warning in macOS

dgsga:


rdmark:
@dgsga any chance we can check for capabilities and not platform? What does this macro do in the first place anyways, and is it still needed?

rdmark:
According to ChatGPT:

> In C, `MAXPHYS` is a macro that defines the maximum size of a single I/O transfer in bytes. It is often used in operating system kernel development to set limits on the size of data transfers. The value of `MAXPHYS` can vary depending on the operating system and hardware architecture. It's typically set to a value that optimizes performance and system stability.

dgsga:
> According to ChatGPT:
> 
> > In C, `MAXPHYS` is a macro that defines the maximum size of a single I/O transfer in bytes. It is often used in operating system kernel development to set limits on the size of data transfers. **The value of `MAXPHYS` can vary depending on the operating system** and hardware architecture. It's typically set to a value that optimizes performance and system stability.

If this is the case it makes sense to check for platform...

rdmark:
Fair point.

rdmark (approved):

PR #711, opened 2024-02-07
Fix failure of pkg update in DragonflyBSD vmactions VM

dgsga:


rdmark:
@dgsga I assume you're working towards fixing the failing DragonflyBSD job with this PR?

dgsga:
> @dgsga I assume you're working towards fixing the failing DragonflyBSD job with this PR?

Thought it best to suspend it until a fix is found. My bare metal Dragonfly VM works fine so it must be an issue with the GitHub vacations vm. Strange as the same error occurs in all versions of the VM

PR #713, opened 2024-02-10
Docker configuration for netatalk v3

rdmark:
Ported from branch-netatalk-2-3 with modification

Using oldstable Debian bullseye-slim as base image because of OpenSSL 1.1 -- we get a working DHX UAM

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@ build-aux/**
 aclocal.m4
 config.*
 configure
+docker-compose.yml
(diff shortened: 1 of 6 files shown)


rdmark:
Next step would be to build and public docker containers via Github CI. One example to get inspiration from: https://github.com/nrwiersma/docker-afpd/blob/master/.github/workflows/build.yml

dgsga (approved):

PR #720, opened 2024-02-16
[2.3] atalkd: Bail out when attempting to -seed a single interface

rdmark:


rdmark:
@NJRoadfan do you think it’d be a good idea to also check for and bail out early when two or more interfaces are used with `-router`? (The inverted case.)

dgsga (approved):
Just fixed a small typo :)

rdmark:
Nice catch :)

PR #730, opened 2024-03-04
Make macOS initscripts prefix sensitive

dgsga:
This  commit fixes the failure of the macOS initscripts if Netatalk is installed in a non-standard location


rdmark (approved):
This reminds me that I should backport these macos initscripts to the v2 branch

PR #744, opened 2024-03-17
docs: Improve docbook templates and man page summaries

rdmark:
This forward-ports recent improvements to 2.x docs


rdmark:
@dgsga The one drawback with this changeset is that we get two copies of the GPL license text in the repo: The plain text one in COPYING, and this newly added one in doc/manual/gpl.xml – do you think this is acceptable, or do you have an idea how to reduce the duplication?

dgsga:
> @dgsga The one drawback with this changeset is that we get two copies of the GPL license text in the repo: The plain text one in COPYING, and this newly added one in doc/manual/gpl.xml – do you think this is acceptable, or do you have an idea how to reduce the duplication?

I think this is acceptable as most users will be concerned with installing  and configuring Netatalk successfully, rather than building the documentation. It is, however, important that they can access the terms of the licence easily, hence the need for COPYING.

dgsga (approved):

PR #749, opened 2024-03-24
Regenerate manpages

dgsga:


rdmark (approved):
Thanks for doing this chore!

FWIW, regenerating the man pages is a step of the Release Process, so my approach has been to leave them be until that moment. No harm in refreshing them now of course.

https://netatalk.io/docs/Release-Process

PR #750, opened 2024-03-24
Create meson build system

dgsga:
Notes:

- This commit allows Netatalk to be compiled using meson as an alternative to GNU autotools.
- Installation instructions added for both build systems
- Readme updated to reflect these changes
- No changes to netatalk codebase
- Less user configuration required as more options are autodetected
- Faster compile times
- Supports all operating systems currently tested with autotools build
- Solaris meson workflow currently disabled until Solaris vmactions VM is updated


.github/workflows/build.yml
```diff
@@ -80,12 +82,23 @@ jobs:
...
+        run: cd build && meson test
+      - name: Meson distribution tests
+        run: cd build && meson dist
 
   build-debian:
     name: Debian
     runs-on: ubuntu-22.04
     container:
-      image: debian:oldstable
+      image: debian:bookworm
```
rdmark, line 101: For the record, the main reason to hold this on oldstable (bullseye) is for building against OpenSSL 1.1 for the sake of the DHX UAM.

But you were saying that the version of Meson on bullseye is too ancient to be usable, didn't you?

.gitignore
```diff
@@ -17,3 +17,4 @@ stamp-h*
 *.tar.xz
 .deps
 .libs
+**/.DS_Store
```
rdmark, line 20: May I suggest that for OS or IDE specific gitignore filters, we encourage the use of local filters? One way to achieve this is to put a .gitignore in the home dir and do:

```
git config --global core.excludesfile ~/.gitignore
```

If we agree on this policy, we should probably remove `*~` from gitignore since that's a vi temp file pattern.

INSTALL
```diff
@@ -0,0 +1,169 @@
+           Netatalk Installation
+           =====================
+
+Quick start
+===========
+
+Netatalk can be built by using Meson or GNU Autotools.
+The meson build system is the recommended build system for the master
```
rdmark, line 8: Terminology nitpick, `main` branch instead of `master`

INSTALL
```diff
@@ -0,0 +1,169 @@
...
+branch, autotools support may be be removed at some point in the future.
```
rdmark, line 9: Shall we aim for a deprecation policy and window? E.g. if no major breakages or loss of compatibility with a tier 1 OS, we deprecate autotools in v3.3

INSTALL
```diff
@@ -0,0 +1,169 @@
...
+Optional:
+
+  - Kerberos V              (for krbV UAM support)
+  - libressl or OpenSSL@1.1 (for Randnum and DHX UAM support)
+  - xsltproc                (for manpages and mananual documentation)
+  - Docbook XSL stylesheets (for manpages & manual documentation)
+  - GLib 2                  (for Spotlight and afpstats support)
+  - dbus-glib               (for afpstats support)
+  - cracklib                (for cracklib support)
+  - avahi                   (for Zeroconf support)
```
rdmark, line 86: How about mDNSresponder for Zeroconf support?

INSTALL
```diff
@@ -0,0 +1,169 @@
...
+External software dependencies
+==============================
+
+Required:
+
+  - libevent 2
+  - libgcrypt version 1.2.3 or later (for DHX2 UAM support)
+  - Berkeley DB version 4.6.0 or later
+  - bison
+  - flex
```
rdmark, line 74: How about adding another section for "Required (Spotlight)" and list tracker plus these two there? If you don't care about Spotlight, you don't need 'em. :)

rdmark (approved):
It's finally happening! Tentatively approving, but please have a look at my comments.

I only glanced through the actual build config files.

.gitignore
dgsga: Done

INSTALL
dgsga: Fixed

INSTALL
dgsga: Yes, I agree. The dbus devel team window for withdrawing autotools support was about 1 year. Meson was introduced in version 1.15.0 (22/9/22) and autotools was removed in version 1.15.8 (21/8/23). Deprecation warnings were posted in NEWS.

INSTALL
dgsga: Added

INSTALL
dgsga: Done

.github/workflows/build.yml
rdmark: From what I've seen, you can indeed compile the DHX library with the warning, but it will crash when you attempt to authenticate from a classic Mac OS client. Because OpenSSL will reject the weak crypto. It's a quite insidious fail state.

dgsga:
I'll keep the dgsga-meson-build branch in case we encounter any regressions...

PR #767, opened 2024-03-29
[2.4] Swap out the old ldap_init for the wonderful ldap_initialize

rdmark:
- Use the more modern ldap_initialize to init LDAP, which enables the use of advanced server URI schemes
- **BREAKING** The `ldap_server` options has been replaced by `ldap_uri` in afpd.conf and has a new syntax
- **BREAKING** Make the build system check for the ldap_initialize symbol, which requires the system ldap library to supply this symbol
- Update the documentation

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -875,7 +875,7 @@ if test x"$with_ldap" != x"no" ; then
         fi
 		with_ldap=no
         ])
-	AC_CHECK_LIB(ldap, ldap_init, with_ldap=yes,
+	AC_CHECK_LIB(ldap, ldap_initialize, with_ldap=yes,
         [ if test x"$with_ldap" = x"yes" ; then
             AC_MSG_ERROR([Missing LDAP library])
         fi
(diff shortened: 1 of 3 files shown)


rdmark:
@dgsga @NJRoadfan I think this is a change that should be made, but it is also a breaking change (afpd.conf syntax, required library), so I'm inclined to hold off on merging this to the stable 2.3 branch.

Rather, I am thinking of planning ahead for a 2.4 feature release in the not too distant future. The mini-roadmap for 2.4 would look something like:

- Meson build system
- WolfSSL to bring the DHX UAM back to life
- This breaking change to LDAP
- Whatever else that comes up in the meantime

Thoughts?

dgsga:
I thinks it's best to target the meson build system for the 2.4 release as there will no doubt be some snagging to do in the early stages of transition. I also agree with not merging this PR until the regressions are fixed. Now that the meson implementation is nearly done my next job is replacement of OpenSSL with WolfSSL for both platforms

dgsga (approved):

rdmark:
@dgsga I have released 2.3.2 now, and subsequently branched branch-netatalk-2-4 to open up for breaking changes.

Please retarget your Meson PR and other active PRs when it's good timing for you.

PR #768, opened 2024-03-29
Swap out the old ldap_init for the wonderful ldap_initialize

rdmark:
- Use the more modern ldap_initialize to init LDAP, which enables the use of advanced server URI schemes
- **BREAKING** The `ldap server` options has been replaced by `ldap uri` in afp.conf and has a new syntax
- **BREAKING** Make both build systems check for the ldap_initialize symbol, which requires the system ldap library to supply this symbol
- Update the documentation
- Update the OmniOS CI job to look for libldap in /opt/local
- Fix error handling for vmactions jobs
- Disable failing integration tests for FreeBSD and OpenBSD (existing issue; not caused by this changeset)
- Various syntax fixes in the CI job yml file

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -280,6 +280,7 @@ jobs:
               talloc \
               tracker3
           run: |
+            set -e
             ./bootstrap
             ./configure \
               --enable-krbV-uam \
@@ -327,6 +328,7 @@ jobs:
               pkgconf \
               tracker3
           run: |
+            set -e
             export AUTOCONF_VERSION=2.71
             export AUTOMAKE_VERSION=1.16
             autoreconf -fi
@@ -373,6 +375,7 @@ jobs:
               pkg-config \
               talloc
           run: |
+            set -e
             ./bootstrap
             ./configure \
               --enable-krbV-uam \
@@ -420,6 +423,7 @@ jobs:
               talloc \
               tracker3
           run: |
+            set -e
             ./bootstrap
             ./configure \
               --with-ssl-dir=/usr/local \
@@ -467,6 +471,7 @@ jobs:
             make install
             cd ..
           run: |
+            set -e
             ./bootstrap
             ./configure \
               --enable-krbV-uam \
@@ -505,12 +510,14 @@ jobs:
               meson \
               talloc
           run: |
+            set -e
             export PATH=/opt/local/sbin:/opt/local/bin:/usr/gnu/bin:/usr/bin:/usr/sbin:/sbin:$PATH
             ./bootstrap
             ./configure \
               --enable-pgp-uam \
-              --with-init-style=solaris \
               --with-bdb=/opt/local \
+              --with-init-style=solaris \
+              --with-ldap=no \
               --with-libgcrypt-dir=/opt/local \
               --with-tracker-pkgconfig-version=3.0 \
               MAKE=gmake \
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -291,10 +291,8 @@ jobs:
               PKG_CONFIG_PATH=/usr/local/libdata/pkgconfig
             gmake -j $(nproc)
             meson setup build \
-              -Dpkg_config_path=/usr/local/libdata/pkgconfig \
-              -Dbuild-tests=true
+              -Dpkg_config_path=/usr/local/libdata/pkgconfig
             ninja -C build
-            cd build && meson test
 
   build-openbsd:
     name: OpenBSD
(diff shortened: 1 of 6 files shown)


rdmark:
OmniOS ships with a too old LDAP library:

```
  FAILED: libatalk/libatalk.so.18 
  gcc  -o libatalk/libatalk.so.18  -z defs -shared -fPIC -Wl,--start-group -Wl,-soname,libatalk.so.18 -Wl,--whole-archive libatalk/acl/libacl.a libatalk/adouble/libadouble.a libatalk/compat/libcompat.a libatalk/talloc/libtalloc.a libatalk/bstring/libbstring.a libatalk/cnid/libcnid.a libatalk/dsi/libdsi.a libatalk/iniparser/libiniparser.a libatalk/unicode/libunicode.a libatalk/util/libutil.a libatalk/vfs/libvfs.a -Wl,--no-whole-archive -lsocket -lsendfile -lnsl libatalk/unicode/charsets/libcharsets.a -lsec -lsec -lldap -Wl,-rpath,/opt/local/lib -Wl,-R/opt/local/lib /opt/local/lib/libtalloc.so -lwrap -Wl,--end-group
  Undefined			first referenced
   symbol  			    in file
  ldap_initialize                     libatalk/acl/libacl.a(ldap.c.o)
  ld: fatal: symbol referencing errors. No output written to libatalk/libatalk.so.18
  collect2: error: ld returned 1 exit status
  [97/235] Compiling C object bin/ad/ad.p/ad_mv.c.o
  [98/235] Compiling C object bin/ad/ad.p/ad_ls.c.o
```

rdmark:
OmniOS comes with openldap-client-2.6.7 whereas ldap_initialize is an ldap 3 symbol I think.

rdmark:
@dgsga Is there a particular syntax for disabling a feature in Meson? Equivalent to autotools `--with-ldap=no`?

rdmark:
For the record: We're getting a nodejs 16 deprecation warning in the static analysis job. This is caused by upstream sonar action. They're fixing it in https://github.com/SonarSource/sonarcloud-github-c-cpp/pull/48

rdmark:
@dgsga This led me down a bit of a rabbit hole with the GitHub CI jobs. Happily, the OmniOS issue was a relatively easy fix. By default the build system was finding libldap in /usr/lib which is too old to have the `ldap_initialize` symbol. The modern library installed by pkgin lives in /opt/local ...

All in all, this change will break compatibility with very old openldap versions as evident by the OmniOS situation. But we have now confirmed that the platforms we support can all get a compatible library so I think we should go ahead with this change. 

The openldap project isn't great at maintaining changelogs for historical releases, but according to this documentation by OpenSUSE, v2.3 is when ldap_initialize was introduced, or at least made a first-class citizen.

https://tr.opensuse.org/OpenLDAP_2.3_libldap_upgrade_howto

To my reading of the docs, the ldap_initialize function is more capable than ldap_init, and allows for connecting to granular server URIs. In addition, ldap_init is flagged as deprecated.

> At this time, ldap_open() and ldap_init() are deprecated in favor of ldap_initialize(), essentially because the latter allows to specify a schema in the URI and it explicitly returns an error code. 

https://linux.die.net/man/3/ldap_initialize

Thoughts?

dgsga:
> @dgsga This led me down a bit of a rabbit hole with the GitHub CI jobs. Happily, the OmniOS issue was a relatively easy fix. By default the build system was finding libldap in /usr/lib which is too old to have the `ldap_initialize` symbol. The modern library installed by pkgin lives in /opt/local ... so doing `-Dwith-ldap=/opt/local` did the trick.
> 
> All in all, this change will break compatibility with very old openldap versions as evident by the OmniOS situation. But we have now confirmed that the platforms we support can all get a compatible library so I think we should go ahead with this change.
> 
> The openldap project isn't great at maintaining changelogs for historical releases, but according to this documentation by OpenSUSE, v2.3 is when ldap_initialize was introduced, or at least made a first-class citizen. And v2.3 was released in 2005, so...
> 
> https://tr.opensuse.org/OpenLDAP_2.3_libldap_upgrade_howto
> 
> To my reading of the docs, the ldap_initialize function is more capable than ldap_init, and allows for connecting to granular server URIs. In addition, ldap_init is flagged as deprecated.
> 
> > At this time, ldap_open() and ldap_init() are deprecated in favor of ldap_initialize(), essentially because the latter allows to specify a schema in the URI and it explicitly returns an error code.
> 
> https://linux.die.net/man/3/ldap_initialize
> 
> Thoughts?

I agree entirely with your conclusions. To disable a feature in meson you use -Dwith-feature=false for boolean options or -Dwith-feature=disabled for feature options. See here:
https://mesonbuild.com/Build-options.html

dgsga (approved):

PR #776, opened 2024-03-30
Create meson build system for [2.4]

dgsga:


dgsga:
@rdmark, here it is! Still a draft PR as there are a few final tweaks to make. One question I have is about the Solaris kernel module, the macro for which appears in configure.ac. Is this legacy code as there doesn't appear to be a makefile for it anywhere? Is it related to the code in sys/solaris (which contains no makefile)?

rdmark:
@dgsga the Solaris kernel module is indeed legacy. It does have a Makefile, but it's not using autoconf but rather this hard-coded Makefile.in

https://github.com/Netatalk/netatalk/blob/branch-netatalk-2-3/sys/solaris/Makefile.in

It might be time to retire this code. In the past I have tried and failed to build it on modern Solaris and Illumos. The current code is SPARC architecture only, so this involved hacking it to compile on amd64. Someone with a long history in the community advised me that this code has been rotted for a long time.

If a Solaris expert volunteers their time in the future we can always bring it back from the attic then!

What do you think?

INSTALL
```diff
@@ -0,0 +1,173 @@
...
+==============================
+
+Required:
+
+  - libevent 2
+  - libgcrypt version 1.2.3 or later (for DHX2 UAM support)
+  - Berkeley DB version 4.6.0 or later
+  - perl
+
+Required for Spotlight support:
```
NJRoadfan, line 75: Spotlight not supported by Netatalk 2.x.

INSTALL
```diff
@@ -0,0 +1,173 @@
...
+Optional:
+
+  - Kerberos V              (for krbV UAM support)
+  - libressl or OpenSSL@1.1 (for Randnum and DHX UAM support)
+  - xsltproc                (for manpages and mananual documentation)
+  - Docbook XSL stylesheets (for manpages & manual documentation)
+  - GLib 2 and dbus-glib    (for afpstats support)
+  - cracklib                (for cracklib support)
+  - avahi or mDNSresponder  (for Zeroconf support)
+  - tcpwrap                 (for TCP wrapper support)
```
NJRoadfan, line 90: CUPS integration support is optional for papd.

INSTALL
dgsga: Thanks @NJRoadfan, well-spotted!

INSTALL
dgsga: Added.

dgsga:
> @dgsga the Solaris kernel module is indeed legacy. It does have a Makefile, but it's not using autoconf but rather this hard-coded Makefile.in
> 
> https://github.com/Netatalk/netatalk/blob/branch-netatalk-2-3/sys/solaris/Makefile.in
> 
> It might be time to retire this code. In the past I have tried and failed to build it on modern Solaris and Illumos. The current code is SPARC architecture only, so this involved hacking it to compile on amd64. Someone with a long history in the community advised me that this code has been rotted for a long time.
> 
> If a Solaris expert volunteers their time in the future we can always bring it back from the attic then!
> 
> What do you think?

The code was written for Solaris 2 so it would need to be completely re-written for current versions of solaris or illumos. I have removed the macro code I started writing for it in the meson implementation.

.github/workflows/build.yml
```diff
@@ -133,7 +179,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - name: Install dependencies
-        run: brew install automake libressl pkg-config krb5 berkeley-db
+        run: brew install automake libressl pkg-config krb5 berkeley-db meson
```
rdmark, line 182: Do we still need to install pkg-config here? I think you removed it from the main branch job.

INSTALL
```diff
@@ -0,0 +1,168 @@
+           Netatalk Installation
+           =====================
+
+Quick start
+===========
+
+Netatalk can be built by using Meson or GNU Autotools.
+The meson build system is the recommended build system for the main
```
rdmark, line 8: Once we’re done here I would recommend using meson for the 2.x branch too

INSTALL
```diff
@@ -0,0 +1,168 @@
...
+  - libevent 2
+  - libgcrypt version 1.2.3 or later (for DHX2 UAM support)
+  - Berkeley DB version 4.6.0 or later
+  - perl
+
+Optional:
+
+  - avahi or mDNSresponder  (for Zeroconf support)
+  - cracklib                (for cracklib support)
+  - cups                    (for papd support)
```
rdmark, line 79: This should say something like “enables papd to send jobs to CUPS printers”

INSTALL
```diff
@@ -0,0 +1,168 @@
...
+  - Docbook XSL stylesheets (for manpages & manual documentation)
+  - GLib 2 and dbus-glib    (for afpstats support)
+  - Kerberos V              (for krbV UAM support)
+  - libressl or OpenSSL@1.1 (for Randnum and DHX UAM support)
+  - tcpwrap                 (for TCP wrapper support)
+  - xsltproc                (for manpages and mananual documentation)
```
rdmark, line 85: Maybe list this together with “docbook” above? You need both to achieve anything. 

BTW, I found recently that you can run `dblatex` on manual.xml to get a nicely formatted PDF manual via the latex formatter.

configure.ac
```diff
@@ -685,56 +685,55 @@ if test x"$this_os" = "xsolaris"; then
...
+        if test "$COMPILE_64BIT_KMODULE" = yes; then
+            # use Sun CC (for a 64-bit kernel, uncomment " -xarch=v9 -xregs=no%appl ")
+            KCFLAGS="-xarch=v9 -xregs=no%appl"
+            KLDFLAGS="-64"
+        else
+            KCFLAGS=""
+            KLDFLAGS=""
+        fi
+        KCFLAGS="-D_KERNEL $KCFLAGS -mno-app-regs -munaligned-doubles -fpcc-struct-return"
+    fi
```
rdmark, line 736: It's hard to tell from the diff, but I assume this is only an indentation change?

As a nitpick, it seems unnecessary to do stylistic changes to the autotools files in this changeset.

distrib/initscripts/meson.build
```diff
@@ -0,0 +1,168 @@
+if get_option('enable-redhat-sysv')
```
rdmark, line 1: How much of an effort would it be to align the initscripts parameter syntax with v3 / main? I.e. `--with-init-style=redhat-sysv`

Since we're redoing the build system anyways, it would be helpful for maintainability in the long run to consolidate behavior between the two branches where it makes sense.

rdmark (changes requested):
Two general pieces of feedback:
- There's an errant etc/.DS_Store file that should be removed
- Specifying init style for each meson build would be good for testing the build logic of those init files. Might be a moot point soon if you figure out autodetection, but I suggest we add it in the meantime.

configure.ac
dgsga: Yes, this was just to aid me deciphering the opaque and arcane logic of the autoconf syntax. Will remove from PR.

INSTALL
```diff
@@ -0,0 +1,167 @@
+in           Netatalk Installation
```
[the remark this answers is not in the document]
dgsga: Removed!

meson_options.txt
```diff
@@ -0,0 +1,252 @@
...
+    type: 'boolean',
+    value: false,
+    description: 'Disable TCP wrappers support',
+)
+option(
+    'enable-cups',
+    type: 'feature',
+    value: 'auto',
+    description: 'Turn on CUPS support',
+)
```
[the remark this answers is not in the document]
dgsga: Done.

distrib/initscripts/meson.build
dgsga: Not too difficult but then we'd end up with one init flag syntax for autotools and a different one for meson. This could cause confusion for the user during the transition. I suggest we keep the option naming syntax for autotools and meson the same until autotools is in the rear-view mirror. That day cannot come a day too soon ;)

.github/workflows/build.yml
```diff
@@ -149,6 +195,16 @@ jobs:
         run: make -j $(nproc) all
       - name: Run tests
         run: make check
+      - name: Meson configure
+        run: |
+          meson setup build \
```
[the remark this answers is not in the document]
dgsga: Done

.github/workflows/build.yml
```diff
@@ -108,21 +137,38 @@ jobs:
             avahi-devel \
             cracklib-devel \
             cups-devel \
+            file \
```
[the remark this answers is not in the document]
dgsga: Because file is needed for the 64-bit compiler mode check and it is not installed by default on the Fedora container.

dgsga:
> Two general pieces of feedback:
> 
> * There's an errant etc/.DS_Store file that should be removed
> * Specifying init style for each meson build would be good for testing the build logic of those init files. Might be a moot point soon if you figure out autodetection, but I suggest we add it in the meantime.

I have removed the errant file. I have added the appropriate init flag to all meson workflows apart from Ubuntu. The autotools distribution test runs make install with privileged hooks to get around systemctl requiring user input when running daemon-reload. Meson does not allow this for security reasons so the meson distribution test fails because interactive authentication is requested on the Ubuntu runner.

distrib/initscripts/meson.build
rdmark: Fair enough. We can do another breaking change in the future when retiring autotools.

rdmark (approved):
I think this is ready to be shipped.

BTW, any tips for getting a new enough version of meson on Solaris 11.4?

```
meson.build:1:0: ERROR: Meson version is 0.59.2 but project requires >=0.61.2
```

dgsga:
> I think this is ready to be shipped.
> 
> BTW, any tips for getting a new enough version of meson on Solaris 11.4?
> 
> ```
> meson.build:1:0: ERROR: Meson version is 0.59.2 but project requires >=0.61.2
> ```

```
sudo pkg update --accept
sudo pip install meson
```

rdmark:
@dgsga What is your solaris version? Mine is:

> Oracle Solaris 11.4.42.111.0

```
# pkg update --accept
No updates available for this image.
# pip install meson
Requirement already satisfied: meson in /usr/lib/python3.7/vendor-packages (0.59.2)
```

For some reason I'm not getting a more recent meson this way.

dgsga:
Will keep the branch so we can fix any snags during the transition period.

rdmark:
> > @dgsga What is your solaris version? Mine is:
> > > Oracle Solaris 11.4.42.111.0
> > 
> > 
> > ```
> > # pkg update --accept
> > No updates available for this image.
> > # pip install meson
> > Requirement already satisfied: meson in /usr/lib/python3.7/vendor-packages (0.59.2)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > For some reason I'm not getting a more recent meson this way.
> 
> Morning! I/m on the same version. Uninstall the old meson version first with sudo pkg uninstall meson.

This did the trick, thanks!

PR #780, opened 2024-03-31
meson: Refactor tcpwrap macro and widen gssapi/kerberos header searches

dgsga:
This commit fixes errors in the meson tcpwrap macro and widens the search for gssapi/kerberos headers so they are picked up on more platforms


rdmark (approved):
I don’t have the knowledge to evaluate the meson syntax. But I defer to your expertise.

PR #785, opened 2024-04-04
meson: Add init-style to configuration summary and fix missing gentoo-systemd style

dgsga:


dgsga:
Proposed auto detection of initscripts should be easy to setup for oses with just one option. For those with two, eg debian, redhat, gentoo, we could do an initial search for systemctl. If systemctl is found the we can install the systemd initscript else install the other. Thoughts?

rdmark:
@dgsga I think this would be a usability improvement. There may be corner cases so let's make sure that a manual override is always effective.

rdmark (approved):

PR #790, opened 2024-04-06
zeroconf: Register mDNS services with the correct ports

rdmark:
Use port 9 for _adisk._tcp and port 0 for _device-info._tcp as per mDNS spec:

https://developer.apple.com/documentation/dnssd/1804733-dnsserviceregister

### DIFF
--- a/etc/netatalk/afp_mdns.c
+++ b/etc/netatalk/afp_mdns.c
@@ -262,7 +262,7 @@ static void register_stuff(const AFPObj *obj) {
                                    ADISK_SERVICE_TYPE,
                                    "",            // default domains
                                    NULL,            // default host name
-                                   htons(port),
+                                   htons(9),      // port number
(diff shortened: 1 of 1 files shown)


rdmark:
For the record: I have yet to observe a difference in behavior before and after this change.

The `_device-info._tcp` service can be seen run when inspecting from other devices on the LAN. But the `mimic model` attribute is not taking effect in the Network tray, before or after the patch. So something's broken or stuck somewhere else, potentially. I've found zeroconf to be a little flaky in general in recent months. macOS sometimes doesn't see zeroconf services and has to be restarted to get back to normal.

This is the command I use to inspect zeroconf services: `dns-sd -B _device-info._tcp`

rdmark:
Testing on Solaris 11 and macOS Sonoma, I can only confirm that the patch breaks device info for mounted AFP shares. Discarding.

PR #792, opened 2024-04-06
docs: Generate compilation manual page from build.yml

rdmark:
- Add a python script for generating a docbook XML manual page derived from the GitHub workflow build.yml
- Add the new manual page as an appendix to the html manual
- Remove autotools specific human written steps

To run the python script, the `xmltodict` package is required. The other libraries are built in. On debian, you can get this package by installing `python3-xmltodict` with apt. Elsewhere you may have to install it with pip (system or virtual env.)

### DIFF
--- a/doc/manual/compile.xml
+++ b/doc/manual/compile.xml
@@ -1,0 +1,556 @@
+<?xml version="1.0" encoding="utf-8"?>
+<appendix id="compile">
+	<title>Compile Netatalk from Source</title>
+	<sect1 id="compile-overview">
+		<title>Overview</title>
+	</sect1>
+	<sect1>
+		<para>
+            This section describes how to compile Netatalk from source for specific operating systems.
+            </para>
+		<para>
+            The Netatalk project is in the process of transitioning the build system
+            from GNU Autotools to Meson.
+            During the transitional period,
+            the documentation will contain both Autotools and Meson build instructions.
+            </para>
+		<para>
+            Please note that this document is automatically generated from the GitHub workflow YAML file.
+            This may or may not be the most optimal way to build Netatalk for your system.
+            </para>
+	</sect1>
+	<sect1 id="compile-os">
+		<title>Operating Systems</title>
+	</sect1>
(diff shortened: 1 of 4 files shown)


dgsga:
@rdmark, I have separated out the meson hot fixes from the workflow updates to keep changes more transparent. Please see PR #800

dgsga:
> @dgsga I'm on the fence whether to attempt to generate this new XML through the build system or not. It would be clunky to add a dependency on Python and these particular python libraries to the build system.
> 
> I've tested that the current workflow fails gracefully when the new XML file isn't available.
@rdmark, IMHO it would be better just to add a new xml page to the manual detailing the installation process on different operating systems.

rdmark:
@dgsga my reasoning for doing it this way is:

1. We already maintain the build.yml file as part of the development process
2. build.yml is guaranteed to always correct and complete since it is executed dynamically
3. Human written documentation can have mistakes and shortcomings
4. Human written documentation has to be kept up to date, meaning double effort for us

When I joined the project, there was nothing but outdated installation documentation, in XML and wiki. I've spent a lot of time and effort to update, add new, and maintain the docs. If it can be partially automated I would be very happy. :)

I'm curious to see what format and contents you had in mind though.

rdmark:
Broke out the improvements to the github workflow yml into a separate PR: https://github.com/Netatalk/netatalk/pull/802

dgsga:
On macOS the shebang fails to find python. Just like the Debian fix for macusers it might be better to replace python with python3.

dgsga (approved):
The python script just needs the shebang tweak:
#!/usr/bin/env python3

rdmark:
Good call, thanks.

Let me do some further refinement of the manual page before merging this...

rdmark:
@dgsga would you be very upset if I commit a snapshot of the generated compile.xml? It kind of goes against the push to remove generated code, but it’s arguably a bit cumbersome to generate right now… we can make it an improvement area later to automate this.

PR #800, opened 2024-04-07
Meson hotfixes

dgsga:
- Fixes for PAM detection on Solaris and Omnios: The preferred mechanism for PAM configuration on Solaris is now using per-service files in /etc/pam.d. Some solaris-based systems still use the /etc/pam.conf configuration file
- Fix typos in root meson.build and meson options file
- Fix error in redhat_sysv custom target

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1879,7 +1879,7 @@ summary({'Netatalk version': version}, section: 'Configuration Summary:')
 summary_info = {
     '  Initscript style': get_option('with-init-style'),
 }
-summary(summary_info, bool_yn: true, section: '  INIT STYLE:')
+summary(summary_info, bool_yn: true, section: '  Init Style:')
 
 summary_info = {
     '  Extended Attributes': netatalk_ea,
--- a/meson.build
+++ b/meson.build
@@ -1397,14 +1397,17 @@ pam_paths = [
     '/usr/local',
 ]
 
-if host_os == 'sunos'
-    pam_dir += '/etc/pam.conf'
-else
-    foreach path : pam_paths
-        if fs.is_dir(path / 'etc/pam.d')
-            pam_dir += path
-        endif
-    endforeach
+foreach path : pam_paths
+    if fs.is_dir(path / 'etc/pam.d')
+        pam_dir += path
+    endif
+    break
+endforeach
+
+if pam_dir == '' and host_os == 'sunos'
+    warning(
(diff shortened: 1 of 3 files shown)


meson.build
```diff
@@ -1396,15 +1396,16 @@ pam_paths = [
     '/usr',
     '/usr/local',
 ]
-
-if host_os == 'sunos'
-    pam_dir += '/etc/pam.conf'
+message('atalk_libname =', atalk_libname)
+if run_command('uname', '-a', check: true).stdout().strip().contains('omnios')
```
rdmark, line 1400: Any idea how to make this more portable? The are others in the OpenSolaris family, for instance, which we don't know for sure yet how they function, e.g. SmartOS, OpenIndiana... And there might be particular revisions of said OSes which are generally compatible but has the old PAM structure.

rdmark: For instance keep "if sunos" as a fallback branch in the decision tree, in the case no "pam.d" dir is found.

Besides, autotools throws this warning in the configure summary when no "pam.d" is found:

```
configure: WARNING:  PAM support was configured for your system, but the netatalk PAM configuration file
configure: WARNING:  cannot be installed. Please install the config/netatalk.pamd file manually.
configure: WARNING:  If you're running Solaris or BSD you'll have to edit /etc/pam.conf to get PAM working.
configure: WARNING:  You can also re-run configure and specify --without-pam to disable PAM support.
```

I don't know if the latter point is still valid for *all* Solaris and BSD variants these days, however. They are becoming more POSIX/GNU-ish over time. :)

And finally, I don't think we have an equivalent to `--without-pam` in the Meson build system now, do we?

meson.build
```diff
@@ -1477,7 +1482,7 @@ if have_pam
     endif
 elif have_pam and (pam_dir == '')
     warning(
-        'PAM support can be compiled, but the install location for the netatalk.pamd file could not be determined. Either install this file by hand or specify the install path.',
+        'PAM support can be compiled, but the install location for the netatalk.pamd file could not be determined. Please install this file by manually. If you are running a Solaris-based host which still relies on /etc/pam.conf you will have to edit this file to get PAM working',
```
rdmark, line 1485: Nitpick: "by manually" is poor grammar. :)

meson.build
dgsga: Added option to disable PAM support. If pam.d dir is not found and the host_os is sunos (all solaris-based systems) then a message advises that /etc/pam.conf needs to be edited to enable PAM support

PR #801, opened 2024-04-07
Build with spotlight in the Solaris workflow

dgsga:


.github/workflows/build.yml
```diff
@@ -465,6 +468,11 @@ jobs:
             ./configure --prefix=/usr
             make
             make install
+            cd ../talloc-2.4.1
+            CC=gcc ./configure --prefix=/usr/local
+            make
+            make install
+            ln -s /usr/local/lib/pkgconfig/talloc.pc /usr/lib/64/talloc.pc
```
rdmark, line 475: I think this should be `ln -s /usr/local/lib/pkgconfig/talloc.pc /usr/lib/64/pkgconfig/talloc.pc` no?

.github/workflows/build.yml
```diff
@@ -473,8 +481,9 @@ jobs:
               --enable-krbV-uam \
               --enable-pgp-uam \
               --without-afpstats \
+              --with-tracker-pkgconfig-version=1.0
```
rdmark, line 484: FWIW a fully updated Solaris 10.4 has v2.0 of the tracker tools.

dgsga: The existing VM is so old it only has Tracker 1.0

dgsga:
Closing this PR as it cannot be applied until the VM is updated. Will keep the branch in case the upstream repo is fixed.

PR #809, opened 2024-04-09
docs: Japanese localization of the manual

rdmark:
This puts Japanese localized manpage and manual XML files into doc/ja and has the build system generate the manual.

The first commit is all the docbook XML files from https://github.com/Netatalk/netatalk-doc-jp/tree/branch-3.1.11/doc

The second commit adopts all changes between 3.1.11 and 3.1.18 in the branch-netatalk-3-1 branch that I translated to Japanese

Note that this only adds Meson build rules for the Japanese manual. I would suggest that we don't bother with autotools for this. 

A weakness of the current solution is that we get mild code duplication, e.g. the html.xsl stylesheet


dgsga (approved):
Have to take your word on this one!

Issue #793, opened 2024-04-07
meson: netatalk executables not linking correctly with libatalk shared library


rdmark:
The latest main branch when building with the Meson build system: On Debian 12, the netatalk daemon cannot find the libatalk shared library:

```
Apr 07 09:21:43 alien netatalk[838084]: /usr/local/sbin/netatalk: error while loading shared libraries: libatalk.so.18: cannot open shared object file: No such file or directory
```

Built with `meson setup build -Dwith-init-style=debian-systemd` and launched with systemd.

I haven't done much troubleshooting yet, but filing this to not forget about following up.

PR #811, opened 2024-04-09
meson: Fix executables not linking correctly with libatalk shared library

dgsga:
Fixes issue #793

### DIFF
--- a/bin/ad/meson.build
+++ b/bin/ad/meson.build
@@ -18,4 +18,5 @@ executable(
     link_with: libatalk,
     c_args: [ad, dversion],
     install: true,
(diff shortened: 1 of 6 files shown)


rdmark (approved):

rdmark:
SGTM... if all the jobs pass.

dgsga:
> SGTM... if all the jobs pass.

Bit strange this one as the linking errors does not occur when compiling with clang (eg macOS). The linking errors are due to a missing rpath to the libatalk shared library.I have tested the compiled executables on Debian 12.5 and they appear to work fine.

PR #815, opened 2024-04-10
meson: Fix incorrect symlinking of uams modules and refactor uams meson.build

dgsga:


rdmark (approved):
At a glance this looks mostly like code moving around. What was the concrete issue you fix here?

dgsga:
> At a glance this looks mostly like code moving around. What was the concrete issue you fix here?

Missing uams_clrtxt.so symlink and logic of symlinking according to which dependencies are present.

PR #819, opened 2024-04-10
[2.4] Remove superfluous env variable from io.netatalk.daemon.plist.tmpl

rdmark:
Same change that was made in main


rdmark:
@dgsga vmactions strike again. Now the NetBSD VM has lost its cups package, apparently. Ugly workaround in this PR...

dgsga (approved):

Issue #820, opened 2024-04-10
Meson: mysql headers incorrectly detected


rdmark:
Meson seems to enable mysql CNID backend even when mysql.h doesn't exist.

This is from building bleeding edge main with a Dockerfile on bullseye-slim

```
[12/227] Linking static target libatalk/acl/libacl.a
[13/227] Compiling C object libatalk/adouble/libadouble.a.p/ad_lock.c.o
[14/227] Compiling C object libatalk/adouble/libadouble.a.p/ad_conv.c.o
[15/227] Compiling C object libatalk/cnid/mysql/libcnid_mysql.a.p/cnid_mysql.c.o
FAILED: libatalk/cnid/mysql/libcnid_mysql.a.p/cnid_mysql.c.o
cc -Ilibatalk/cnid/mysql/libcnid_mysql.a.p -Ilibatalk/cnid/mysql -I../libatalk/cnid/mysql -I. -I.. -Iinclude -I../include -Ietc/afpd -I../etc/afpd -I/opt/homebrew/Cellar/mysql/8.3.0_1/include/mysql -fdiagnostics-color=always -Wall -Winvalid-pch -Wextra -Wpedantic -std=c11 -O0 -g -DHAVE_CONFIG_H '-D_U_=__attribute__((unused))' -Wno-pedantic -Wno-extra -Wno-all -Wno-deprecated-declarations -MD -MQ libatalk/cnid/mysql/libcnid_mysql.a.p/cnid_mysql.c.o -MF libatalk/cnid/mysql/libcnid_mysql.a.p/cnid_mysql.c.o.d -o libatalk/cnid/mysql/libcnid_mysql.a.p/cnid_mysql.c.o -c ../libatalk/cnid/mysql/cnid_mysql.c
../libatalk/cnid/mysql/cnid_mysql.c:30:10: fatal error: mysql.h: No such file or directory
30 | #include <mysql.h>
| ^~~~~~~~~
compilation terminated.
```

PR #821, opened 2024-04-10
docker: Build with Meson

rdmark:


rdmark:
Currently blocked by https://github.com/Netatalk/netatalk/issues/820

dgsga:
> Currently blocked by #820

@rdmark, I have reworked the mysql macro so the backend is only enabled if mysqlclient  libraries and headers are present

rdmark:
Now the blocker is https://github.com/Netatalk/netatalk/issues/825

We're still using Debian Bullseye as the base image here...

rdmark:
I took the opportunity to bump the base image to bookworm.

The main reason for staying at bullseye was for the sake of openssl1.1

I'm anticipating that we can make something happen with wolfssl in this release cycle...

dgsga (approved):

PR #822, opened 2024-04-10
meson: Ensure mysql backend is only enabled when  mysqlclient library and headers are present

dgsga:


rdmark:
@dgsga The logic doesn't quite work yet, when mysql_config exists but the dev headers don't.

It fails at this part
```
    use_mysql_backend = (
        mysqlclient.found()
        and mysql_cflags
        and mysql_libs
    )
```
...when mysql_cflags or mysql_libs aren't defined.

dgsga:
> @dgsga The logic doesn't quite work yet, when mysql_config exists but the dev headers don't.
> 
> It fails at this part
> 
> ```
>     use_mysql_backend = (
>         mysqlclient.found()
>         and mysql_cflags
>         and mysql_libs
>     )
> ```
> 
> ...when mysql_cflags or mysql_libs aren't defined.

@rdmark, Try the latest force-push. The logic now depends on the 'mysql.h' header being present.

rdmark (approved):
This looks better! I will tentatively approve the change.

However the Docker build runs into a new issue immediately afterwards:

```
FileNotFoundError: [Errno 2] No such file or directory: '/opt/homebrew/bin/pkg-config'
```

I think this is caused by building the image on macOS, when the image OS is Debian. the pkg-config path of macOS is passed to Debian, seemingly, which is another issue in itself probably.

dgsga:
> This looks better! I will tentatively approve the change.
> 
> However the Docker build runs into a new issue immediately afterwards:
> 
> ```
> FileNotFoundError: [Errno 2] No such file or directory: '/opt/homebrew/bin/pkg-config'
> ```
> 
> I think this is caused by building the image on macOS, when the image OS is Debian. the pkg-config path of macOS is passed to Debian, seemingly, which is another issue in itself probably.

How do we best approach these issues with Docker? Our main goal is to ensure that compilation works successfully on native host platforms, I'm not sure how we can determine which errors lie with Docker and which lie with the meson syntax...

rdmark:
Let me open up an issue ticket for posterity. I agree that it looks like a corner case or bug with Meson behavior in Docker.

I agree that it is arguably a narrow use case to build a Linux based docker image from scratch on macOS. But this can be done with autotools, so it _should_ be possible.

PR #849, opened 2024-04-16
afpd: Deal better with large numbers of FCE events

rdmark:
Two measures to allow for better handling of very large numbers of simultaneous FCE events

- If fce holdfmod is 0 the modification events must be thrown directly
- Added configurable wait between fce event sending to avoid udp event loss

Both patches by @seiman


dgsga (approved):

rdmark:
Pulling back to draft in order to address static analysis defect: usage of deprecated `usleep` function.

PR #851, opened 2024-04-16
meson: Fix dbus-session.conf.tmpl substitution

dgsga:


dgsga:
Once we migrate to meson fully we can dispense with all this sed nonsense and have meson do it automatically.

rdmark (approved):

Issue #853, opened 2024-04-17
More init script install hooks


rdmark:
We’re currently missing init script install hooks for NetBSD/Debian-sysv/openrc

the two latter even have commended out hooks in the autotools Makefile

rdmark:
A bit of git archaeology revealed these commits that disabled the init script hooks for Gentoo and Debian:

https://github.com/Netatalk/netatalk/commit/4c16a8e1f590cb133ef1984302ca2c831311fe62
https://github.com/Netatalk/netatalk/commit/ac18bd634e3a9f7153a2eb489281452effa8bfa8

If I read this correctly, both were done per request by the downstream package manager.

rdmark:
Reactivated the hooks, because we now have a compile time option to disable them

Issue #796, opened 2024-04-07
Privileged hooks mandatory for systemd init style installation


rdmark:
>  ... The autotools distribution test runs make install with privileged hooks to get around systemctl requiring user input when running daemon-reload. Meson does not allow this for security reasons so the meson distribution test fails because interactive authentication is requested on the Ubuntu runner.

_Originally posted by @dgsga in https://github.com/Netatalk/netatalk/issues/776#issuecomment-2040395408_

rdmark:
I made a workaround in the 2.x branch for this: 

Issue: https://github.com/Netatalk/netatalk/issues/405

PR: https://github.com/Netatalk/netatalk/pull/419

Do we want to port this over, or have a different solution? How is it working in Meson?

PR #854, opened 2024-04-19
Build system improvements for initscripts

rdmark:
Introduce `disable-init-hooks` build system parameter which skips the hook that launches an OS specific initscript helper app. https://github.com/Netatalk/netatalk/issues/796

Also: Restore the hooks for Debian-sysv and Gentoo-openrc. https://github.com/Netatalk/netatalk/issues/853
And: Check for the old launchd service when installing the new one. https://github.com/Netatalk/netatalk/issues/856
As well as: Add rc init scripts for FreeBSD and OpenBSD. https://github.com/Netatalk/netatalk/issues/870


rdmark:
@dgsga Do you have a suggestions for a better name of the parameter?

dgsga:
> @dgsga Do you have a suggestions for a better name of the parameter?

I think it's a good name, does what it says on the tin.

PR #857, opened 2024-04-19
NEWS for 3.2.0

rdmark:


rdmark:
@dgsga Started drafting this early since I know from experience it can be time consuming to get right...

dgsga (approved):
@rdmark, nice work. Good to see all those 'NEW' words in there ;)

rdmark:
@dgsga Cheers! Yes I think this will be a good one. 

Unless we find any last minute bugs, I think we’re ready to ship this. Do you have a preference / suggestions how we go about it? 

My suggestion would be to run a few more tests and then tag 3.2.0 on Saturday (Japan time). Unless any severe bugs are found. 

Thoughts?

PR #860, opened 2024-04-21
GitHub CI job with openSUSE Tumbleweed

rdmark:
- Because openSUSE Tumbleweed doesn't have a insserv or insserv-compat package available, cannot use `suse-sysv`
- Tried using openSUSE Leap (each 15.x release image) for a more stable base OS but all of them fails in GitHub Actions (cannot find `tar` which points to PATH being bad)
- The `zypper` command is split up into two because it bugs out when installing too many packages at once 👎

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -201,6 +201,7 @@ jobs:
             libtool \
             meson \
             ninja \
+            openrc \
             openssl-dev \
             pkgconfig &&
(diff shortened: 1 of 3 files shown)


rdmark:
The CI failures seem to be a Microsoft infrastructure fail: https://github.com/orgs/community/discussions/120966

dgsga (commented):
@rdmark, I have fixed the header issue in my latest PR so we can install the db devel package now :)

rdmark:
The PR now includes a few hotfixes for the Alpine job and openrc that I noticed after the fact. Sorry about the mess.

dgsga (approved):

PR #875, opened 2024-04-24
meson: Fix BerkeleyDB header detection on openSUSE

dgsga:


dgsga:
Tested on Tumbleweed VM

rdmark (approved):
elegant

PR #877, opened 2024-04-25
Adopt rc script from FreeBSD ports.

rdmark:
Takes in the FreeBSD rc script from FreeBSD ports, by Joe Marcus Clarke, with minor modification.

- Added a copyright notice with the BSD 2-clause license
- Changed substitution deliminator from %%%% to :: to match our build system

Implementation note: `avahi_daemon` and `mdnsd` is what the respective zeroconf daemon is named in FreeBSD. We need one of these to be injected into the rc script for the startup ordering to work properly.


rdmark:
Note: Substitution works only in the Meson build system right now.

rdmark:
@dgsga Substitution achieved in autotools as well. Can you think of a better way to do this?

PR #878, opened 2024-04-25
meson: Fix false positive for DHX support in Alpine when SSL headers absent

dgsga:
This commit also enables all options within the Alpine Linux job


rdmark (approved):
FWIW I originally imagined the Alpine CI job to be minimal, as a CI check for the disabling of options too. But I'm ok either way really.

PR #885, opened 2024-04-25
GitHub CI: Use libdb package for Alpine

rdmark:


rdmark:
@dgsga Just salvaged your idea of using db-dev instead of building it from source.

dgsga (approved):

PR #889, opened 2024-04-26
docker: Alpine based container

rdmark:
Use the latest stable alpine 3.19 docker image instead of bookworm-slim.

Pros:
- About half the size, ~1GB -> ~500MB
- Zero vulnerabilities (down from almost 200)
- Builds so, so quickly


rdmark:
For the record, avahi has be launched manually on Alpine: https://wiki.alpinelinux.org/wiki/MDNS

dgsga (approved):

PR #892, opened 2024-04-26
Remove hard dependency on perl and grep

rdmark:
When perl and grep are not detected, don't generate or install the apple_dump/asip-status/macusers scripts and man pages.


dgsga (approved):
@rdmark, I have reformatted some of the meson.build files for the sake of consistency throughout the repository. I use muon fmt to format all the meson.build files as it applies the meson syntax recommendations to all files.
https://github.com/annacrombie/muon

Install muon and save the following script as 'muonfmt' in your local bin directory (chmod +x it too):
```
#!/bin/bash

find . -type f -name "meson.build" -exec muon fmt -i {} \;
```
Then all you need to do is:
```
cd netatalk
muonfmt
```

rdmark:
Pushed a few final bugfixes and updated the manual to indicate that Perl is now an optional dependency.

rdmark:
@dgsga What do you think about documenting the muonfmt trick on the Developer Notes wiki page? https://github.com/Netatalk/netatalk/wiki/Developer-Notes

rdmark:
Created a new section on that page https://github.com/Netatalk/netatalk/wiki/Developer-Notes#build-system

PR #895, opened 2024-04-26
Rewrite afpstats in Perl

rdmark:
This removes python3 as a de-facto runtime dependency for the netatalk package (not that we actually checked for it, but...)

As an added boon, we eliminate the python/python3 shebang cross-platform conundrum. The Debian deb package can now remove a downstream patch for this, for instance.


dgsga:
I like it...

dgsga (approved):

PR #897, opened 2024-04-26
meson: Fix configuration data path to tracker, tracker3 and tracker-control

dgsga:


dgsga:
@rdmark, the TRACKER_MANAGING_COMMAND defines in config.h (autoconf) and build/config.h (meson) are now identical. Well spotted!

rdmark (approved):

PR #912, opened 2024-04-28
[2.4] meson: Ensure compatibility with linux systems using musl

dgsga:


dgsga:
@rdmark, this is equivalent to the following code snippet in configure.ac:
```
dnl --------------------- GNU source
case "$this_os" in
	linux)	AC_DEFINE(_GNU_SOURCE, 1, [Whether to use GNU libc extensions])
        ;;
     kfreebsd-gnu) AC_DEFINE(_GNU_SOURCE, 1, [Whether to use GNU libc extensions])
        ;;
esac
```

rdmark (approved):

rdmark:
That explains why autotools was happily compiling this thing!

PR #920, opened 2024-04-29
[2.4] Fix missing rresvport function on systems using musl C library

dgsga:
This commit:

- Fixes the missing rresvport function on linux systems using musl, credit @daimh

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1066,6 +1066,12 @@ if test x"$neta_cv_REALPATH_TAKES_NULL" = x"yes"; then
     AC_DEFINE(REALPATH_TAKES_NULL,1,[Whether the realpath function allows NULL])
 fi
 
+dnl --------------------- Check for musl C library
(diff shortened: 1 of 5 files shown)


rdmark:
@dgsga Great job! I will run some tests later today.

The MUSL detection logic is neat. Maybe in the long run when our local rresvport implementation has been "proven" we can use that across the board rather than the likely-to-be-deprecated glibc one.

dgsga:
> @dgsga Great job! I will run some tests later today.
> 
> The MUSL detection logic is neat. Maybe in the long run when our local rresvport implementation has been "proven" we can use that across the board rather than the likely-to-be-deprecated glibc one.

@rdmark, I have separated out the different issues so they each have a separate PR (for easier tracking). So this PR now only contains the rresvport fix

rdmark (approved):

PR #921, opened 2024-04-29
[2.4] meson: Fix linux detection macro on Alpine Linux

dgsga:


NJRoadfan:
I'm seeing a pattern here. Alpine Linux and meson insist on using the correct `__linux__` defintion in macros!

rdmark (approved):

PR #926, opened 2024-04-30
Add archlinux GitHub actions job

dgsga:
Notes:

- The archlinux docker container is not booted with systemd set up as the init system so  the initscript for systemd (the default init system on archlinux) fails with both autotools and meson.
- For some reason the meson /autotools compilation of afpd fails with an undeclared glib-related function when spotlight support is enabled. This does not occur on my archlinux vm.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -239,6 +239,52 @@ jobs:
       - name: Meson - Uninstall
         run: ninja -C build uninstall
 
+  build-archlinux:
+    name: Archlinux
+    runs-on: ubuntu-22.04
+    container:
+      image: archlinux:latest
+    steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -239,6 +239,52 @@ jobs:
       - name: Meson - Uninstall
         run: ninja -C build uninstall
 
+  build-archlinux:
+    name: Archlinux
+    runs-on: ubuntu-22.04
+    container:
+      image: archlinux:latest
```
rdmark, line 246: Have you tried using an earlier "stable" image? I wonder if they broke something with tracker in "latest"?

rdmark (approved):

PR #927, opened 2024-04-30
[2.4] Update rc.openrc.tmpl: a2boot and timelord depend on atalkd

NJRoadfan:
We should only attempt to load timelord and a2boot if atalkd is running. Also register NBP entries on port 4 to be consistent with other init scripts.

### DIFF
--- a/distrib/initscripts/rc.openrc.tmpl
+++ b/distrib/initscripts/rc.openrc.tmpl
@@ -33,6 +33,18 @@ atalk_startup () {
 			eend $?
 		fi
 
+		if [ "${TIMELORD_RUN}" = "yes" ]; then
+			ebegin "  Starting timelord"
+			start-stop-daemon --start --quiet --exec :SBINDIR:/timelord
+			eend $?
+		fi
+
+		if [ "${A2BOOT_RUN}" = "yes" ]; then
+			ebegin "  Starting a2boot"
+			start-stop-daemon --start --quiet --exec :SBINDIR:/a2boot
+			eend $?
+		fi
+
 	fi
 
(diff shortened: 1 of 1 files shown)


distrib/initscripts/rc.openrc.tmpl
```diff
@@ -84,25 +84,25 @@ stop () {
...
 		fi
 
+		if [ "${TIMELORD_RUN}" = "yes" ]; then
+			ebegin "Stopping timelord"
+			start-stop-daemon --stop --quiet --exec :SBINDIR:/timelord
+			eend $?
+		fi
+
+		if [ "${A2BOOT_RUN}" = "yes" ]; then
+			ebegin "Stopping a2boot"
```
rdmark, line 101: Minor thing, but it would look nicer to indent the "Stopping" messages like you did with the "Starting" messages.

NJRoadfan:
The terminal output (at least on Alpine Linux) wouldn't look correct since the services stop in the reverse order.

Currently:
```
 * WARNING: you are stopping a boot service
 * Stopping afpd ...                                                                                              [ ok ]
 * Stopping papd ...                                                                                              [ ok ]
 * Stopping timelord ...                                                                                          [ ok ]
 * Stopping a2boot ...                                                                                            [ ok ]
 * Unregistering alpinetest:Workstation ...                                                                       [ ok ]
 * Unregistering alpinetest:netatalk ...                                                                          [ ok ]
 * Stopping atalkd ...                                                                                            [ ok ]
 * Stopping cnid_metad ...                                                                                        [ ok ]
```

There would be an awkward looking indent in the middle.

rdmark (approved):
Good point

PR #942, opened 2024-05-02
[2.4] afppasswd: Correct setting of <8 length passwords with -w option

rdmark:
Turns out my initial go at this only worked for passwords of length = 8...
It is required to null out the padding for the rest of the password scrambling logic to work.


dgsga (approved):

rdmark:
Remaining code issues are pre-existing...

PR #952, opened 2024-05-05
Forward-port netatalk-classic patches

rdmark:
Originally from the netatalk-classic port, ported to my 2.x port, then to mainline 2.3

- Remove unused code
- Remove ancient bug workarounds and standard library shortcomings
- Refactor functions
- More granular typecasting
- Initialize variables
- Declare variables separately
- Handle system call return values


rdmark:
To double check that the logic in etc/uams/uams_gss.c is correct after this.

dgsga:
Great job! 👌

dgsga (approved):

rdmark:
I'm keeping this branch for the time being in case we run into corner case bugs later. It's handy to be able to do binary search for breakages.

PR #954, opened 2024-05-06
meson: Add WolfSSL as DHX provider

dgsga:


dgsga:
@rdmark, the more time I spend on this the more I'm coming to the conclusion that adding WolfSSL as an alternative provider of DHX support offers no real advantage. This is because:

- Not all WolfSSL packages have been compiled with the OpenSSL compatibility layer (eg Alpine, Archlinux) so affected distros have to use OpenSSL/libressl anyway unless the user installs WolfSSL from source.
- We still need OpenSSL or libressl for RANDNUM support
- We have to import a third-party header ('cast.h) to provide cast functions. This file is not licensed with GPL2

Thoughts?

rdmark:
@dgsga Understood, the landscape for distributed WolfSSL packages is truly complex.

Two question...

Can we WolfSSL as an external provider library *as well as* an embedded library, electable at configure/compile time? As I mentioned some time ago, at least Debian has a policy against software packages embedding libraries that are provided by another Debian package.

If we embed WolfSSL, can it be customized to provide the DES_* symbols required for RandNum? Or will this still rely on OpenSSL/LibreSSL?

For the record, all of the DES_crypt functions are slated for deprecation in OpenSSL3, so we may lose them too in future releases. https://www.openssl.org/docs/man3.2/man3/DES_crypt.html

dgsga:
> @dgsga Understood, the landscape for distributed WolfSSL packages is truly complex.
> 
> Two questions...
> 
> Can we WolfSSL as an external provider library _as well as_ an embedded library, electable at configure/compile time? As I mentioned some time ago, at least Debian has a policy against software packages embedding libraries that are provided by another Debian package.
> 
> If we embed WolfSSL, can it be customized to provide the DES_* symbols required for RandNum? Or will this still rely on OpenSSL/LibreSSL?
> 
> For the record, all of the DES_crypt functions are slated for deprecation in OpenSSL3, so we may lose them too in future releases. https://www.openssl.org/docs/man3.2/man3/DES_crypt.html

Yes and yes!

rdmark:
Lovely!

rdmark:
@dgsga If I’m not mistaken you’re doing (minor) modifications to the bundled wolfssl source code? If I can ask a favor, can you please devise a way to keep track of those local modifications? For the benefit of future contributors who may have to merge with major upstream wolfssl versions etc. 

I remember the headaches we faced when upgrading the iniparser code

dgsga:
> @dgsga If I’m not mistaken you’re doing (minor) modifications to the bundled wolfssl source code? If I can ask a favor, can you please devise a way to keep track of those local modifications? For the benefit of future contributors who may have to merge with major upstream wolfssl versions etc. 
> 
> 
> 
> I remember the headaches we faced when upgrading the iniparser code

👍 Just trying different approaches ATM

PR #959, opened 2024-05-07
Start netatalk in GitHub CI jobs

rdmark:
Introducing "start netatalk" and "stop netatalk" steps for the jobs. This constitutes a very simple sanity test that the netatalk/afpd deamon starts up and serves an AFP service on port 548

A 2s delay is inserted after starting netatalk until attempting to ping it with asip-status. This is because it always takes a moment for afpd to start up fully.

Note: There are slight differences between the "stop netatalk" steps in autotools and meson. This is because meson do not have support for post uninstall scripts. See https://github.com/mesonbuild/meson/issues/1974

Additional changes:
- Move Fedora from rawhide to latest (stable OS)
- Don't install soft requirements in the Fedora job
- Fix and enable the DragonflyBSD job (required `usesh: true` to enable the shell)
- Make the Alpine job full-featured rather than minimal
- Sort the jobs pseudo-alphabetically

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -59,60 +59,133 @@ env:
     xsltproc
 
 jobs:
-  build-ubuntu:
-    name: Ubuntu
+  build-alpine:
+    name: Alpine Linux
     runs-on: ubuntu-22.04
+    container:
+      image: alpine:latest
     steps:
       - uses: actions/checkout@v4
       - name: Install dependencies
         run: |
-          sudo apt-get update
-          sudo apt-get install --assume-yes --no-install-recommends ${{ env.APT_PACKAGES }}
+          apk add \
+            acl-dev \
+            autoconf \
+            automake \
+            avahi-dev \
+            bison \
+            build-base \
+            cracklib \
+            cracklib-dev \
+            cracklib-words \
+            curl \
+            db-dev \
+            dbus-dev \
+            dbus-glib-dev \
+            flex \
+            gcc \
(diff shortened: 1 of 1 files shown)


rdmark:
This turned out to be easier said than done for one critical reason: Most Linux distros ship their docker images sans init system. And installing it (systemd or openrc) on the fly doesn't work. It requires a system reboot for the init system bus to start up, which isn't feasible here.

Less of a problem with the vmactions jobs.

rdmark:
The NetBSD failure is a potential bug. https://github.com/Netatalk/netatalk/issues/961

Start/Stop is not implemented for Alpine, Arch, Fedora, or openSUSE. These are docker containers that don't come with an init system, and launching netatalk stand-alone results in a non-functioning AFP server for some reason (more investigation required.)

dgsga (approved):

rdmark:
@dgsga FYI I'm holding off on merging this until the two failures have been investigated. If they're due to vmactions quirks I'll just disable the start/stop steps for those jobs.

rdmark:
@dgsga Heads up that I fixed the DragonflyBSD job in this changeset, plus sorted the jobs pseudo-alphabetically.

rdmark:
Blocker is this failure on Debian with meson https://github.com/Netatalk/netatalk/issues/968 ...

rdmark:
Disabled the start/stop steps for Debian+meson for now so that this can be merged. Will reactivate once the underlying issue has been resolved.

PR #965, opened 2024-05-10
meson: Add WolfSSL as DHX provider 

dgsga:
Notes:

- The built-in ssl library option is the minimal WolfSSL file set required to provide dhx/randnum support.
- New -Dwith-embedded-ssl configure option added (defaults to false)
- SSL provider added to configuration summary
- WolfSSL directory structure is preserved so updating is easy
- A third party atalk/cast.h header is required to provide cast functions for dhx support. The original version used is from SSLeay before it evolved into OpenSSL, and has its own licence.  This file was eventually incorporated into OpenSSL and hence into libreSSL
- False positive SonarCloud result on WolfSSL code. WolfSSL is highly modular and configurable so there are many identical #ifdef and #ifndef conditionals, defines and includes throughout the codebase. SonarCloud algorithm just sees duplicated code and cannot distinguish whether it is actually needed or not. There are no new security hotspots in the added code and the security review rating is A. Can we fine-tune what is covered by the SonarCloud check?

Thought for the day:
"In machine learning the term stochastic parrot is a metaphor to describe the theory that large language models, though able to generate plausible language, do not understand the meaning of the language they process."


rdmark:
@dgsga 320k lined added. Encryption is serious coding. 🔥 

Did you consider making "with-embedded-ssl" enabled by default? I can see a few pros and cons myself, but am curious about your reasoning here.

BTW, there are a few errant .DS_Store files in the changeset. Please remove...

I will do a proper code review shortly!

rdmark:
> False positive SonarCloud result on WolfSSL code. WolfSSL is highly modular and configurable so there are many identical #ifdef and #ifndef conditionals, defines and includes throughout the codebase. SonarCloud algorithm just sees duplicated code and cannot distinguish whether it is actually needed or not. There are no new security hotspots in the added code and the security review rating is A. Can we fine-tune what is covered by the SonarCloud check?

@dgsga How this for a solution: Disable code duplication scanning for `include/wolfssl` entirely. Like you say, the wolfssl code is designed in this way, and since it's not our own code we don't really care about duplication in the first place.

A quick googling suggests that you can add something like this to the sonar-project.properties file: (I haven't tested that the syntax is 100% correct.)

```
sonar.cpd.exclusions=include/wolfssl/**
```

https://github.com/Netatalk/netatalk/blob/main/sonar-project.properties

Perhaps you can add this

dgsga:
> .DS_Store

Removed

dgsga:
> > False positive SonarCloud result on WolfSSL code. WolfSSL is highly modular and configurable so there are many identical #ifdef and #ifndef conditionals, defines and includes throughout the codebase. SonarCloud algorithm just sees duplicated code and cannot distinguish whether it is actually needed or not. There are no new security hotspots in the added code and the security review rating is A. Can we fine-tune what is covered by the SonarCloud check?
> 
> @dgsga How this for a solution: Disable code duplication scanning for `include/wolfssl` entirely. Like you say, the wolfssl code is designed in this way, and since it's not our own code we don't really care about duplication in the first place.
> 
> A quick googling suggests that you can add something like this to the sonar-project.properties file: (I haven't tested that the syntax is 100% correct.)
> 
> ```
> sonar.cpd.exclusions=include/wolfssl/**
> ```
> 
> https://github.com/Netatalk/netatalk/blob/main/sonar-project.properties
> 
> Perhaps you can add this to your changeset and see if it has the intended effect?

We have to exclude all WolfSSL code from SonarCloud scanning (sonar.exclusions). If we just do sonar.cpd.exclusions then the scan fails with:
[E Reliability Rating on New Code](https://sonarcloud.io/dashboard?id=Netatalk_netatalk&pullRequest=965) (required ≥ A)
 [D Security Rating on New Code](https://sonarcloud.io/dashboard?id=Netatalk_netatalk&pullRequest=965) (required ≥ A)
This is plainly nonsense.

dgsga:
> Did you consider making "with-embedded-ssl" enabled by default? I can see a few pros and cons myself, but am curious about your reasoning here.
I would prefer to enable it by default, what stopped me was the Debian issue. I have now enabled it by default, let me know if you wish to revert this.

rdmark:
> I would prefer to enable it by default, what stopped me was the Debian issue. I have now enabled it by default, let me know if you wish to revert this.

This can be patched when doing the deb packaging, in an appropriate manner. Not to worry.

etc/uams/uams_dhx_pam.c
```diff
@@ -25,19 +25,19 @@
...
 #include <openssl/dh.h>
 #include <openssl/cast.h>
 #include <openssl/err.h>
 #include "openssl_compat.h"
-#else /* OPENSSL_DHX */
-#include <bn.h>
-#include <dh.h>
-#include <cast.h>
-#include <err.h>
 #endif /* OPENSSL_DHX */
```
rdmark, line 41: Nitpick: Perhaps label this as the endif for WOLFSSL_DHX

etc/uams/uams_dhx_passwd.c
```diff
@@ -26,17 +26,18 @@
...
 #elif defined(OPENSSL_DHX)
 #include <openssl/bn.h>
 #include <openssl/dh.h>
 #include <openssl/cast.h>
 #include "openssl_compat.h"
-#else /* OPENSSL_DHX */
-#include <bn.h>
-#include <dh.h>
-#include <cast.h>
 #endif /* OPENSSL_DHX */
```
rdmark, line 41: Same as the above.

etc/uams/uams_pgp.c
```diff
@@ -23,16 +23,17 @@
...
+#include <atalk/cast.h>
 #elif defined(OPENSSL_DHX)
 #include <openssl/bn.h>
 #include <openssl/dh.h>
 #include <openssl/cast.h>
-#else /* OPENSSL_DHX */
-#include <bn.h>
-#include <dh.h>
-#include <cast.h>
 #endif /* OPENSSL_DHX */
```
rdmark, line 37: Same as the above.

meson.build
```diff
@@ -526,15 +533,37 @@ crypto = declare_dependency(
...
     cdata.set('HAVE_LIBCRYPTO', 1)
-    cdata.set('OPENSSL_DHX', 1)
     cdata.set('UAM_DHX', 1)
+endif
+
+ssl_deps = []
+ssl_type = ''
+
+#if have_embedded_ssl and wolfssl.found()
+#    error('embedded SSL cannot be compiled when WolfSSL is installed')
```
rdmark, line 546: Do you plan to use this commented out code at some point?

meson.build
```diff
@@ -151,6 +151,11 @@ if dl.found()
     netatalk_common_link_args += '-ldl'
 endif
 
+math = cc.find_library('m', required: true)
```
rdmark, line 154: What is this library for?

meson.build
```diff
@@ -526,15 +533,37 @@ crypto = declare_dependency(
...
+#if have_embedded_ssl and wolfssl.found()
+#    error('embedded SSL cannot be compiled when WolfSSL is installed')
+if have_embedded_ssl
+    ssl_type += 'built-in'
+    cdata.set('WOLFSSL_DHX', 1)
+    cdata.set('EMBEDDED_SSL', 1)
+elif wolfssl.found() and run_command(
+    '/bin/cat',
+    wolfssl.get_variable(pkgconfig: 'includedir') / 'wolfssl/options.h',
+    check: true,
```
rdmark, line 554: This seems like a somewhat crude way to detect the wolfssl library. I wonder if there is a more portable method?

include/wolfssl/meson.build
```diff
@@ -0,0 +1,8 @@
+if 'aarch64'.contains(host_machine.cpu_family())
+
+    wolfssl_options = configure_file(input: 'options_aarch64.in', output: 'options.h', configuration: cdata)
```
rdmark, line 3: Any chance we can support 32bit ARM? Running netatalk on f.e. a RPi Zero is a fairly common usecase.

rdmark (approved):
Impressive work!

meson.build
dgsga: The dh.c file in WolfSSL is linked to the external math library.

When you use the math library with all recent versions of GCC on GNU/Linux systems you have to explicitly link to it. It is not automatically linked to along with the rest of the standard C library.

etc/uams/uams_dhx_pam.c
dgsga: Done

etc/uams/uams_dhx_passwd.c
dgsga: Done

etc/uams/uams_pgp.c
dgsga: Done

include/wolfssl/meson.build
dgsga: Working on it...

dgsga: Done

meson.build
dgsga: This is not detecting the WolfSSL library, it is checking if the WolfSSL package installed on the host OS has been compiled with the OpenSSL compatibility layer.

meson.build
dgsga: No!

dgsga:
> Impressive work!

Thanks! :)

PR #967, opened 2024-05-11
meson: Fix 'libatalk not found' error on NetBSD

dgsga:
This commit ensures that all executables are correctly linked to libatalk at build-time


dgsga:
@rdmark, this fixes a regression introduced with commit a0cfb48, which wrecked the libatalk linking for all executables. Please can you approve and commit ASAP. Correct linking can be confirmed by running the following after install:
`ldd 'full_path_to_executable'`

rdmark (approved):

PR #975, opened 2024-05-13
meson: Fix rpaths for all executables

dgsga:


rdmark:
@dgsga Approved. Will you bring it out of Draft state?

PR #976, opened 2024-05-13
Embedded SSL: Comment out _GNU_SOURCE redefinitions

dgsga:


rdmark (approved):
Nitpick, but I think `#if 0` is the more common style in this codebase for disabling blocks of code.

dgsga:
> Nitpick, but I think `#if 0` is the more common style in this codebase for disabling blocks of code.

#if 0 still generates warning...

PR #984, opened 2024-05-14
Optimize sort order of all headers

dgsga:
Header includes are now in correct sort order (global to local) and arranged alphabetically within each group.
#ifdefs now have a space either side so they can be read more easily.

This has been bugging me for ages.


etc/afpd/afprun.c
```diff
@@ -24,20 +24,20 @@
...
-/* #define __USE_GNU 1 */
+#include <sys/wait.h>
 #include <unistd.h>
-#include <grp.h>
 
-#include <errno.h>
-#include <sys/wait.h>
-#include <sys/param.h>
-#include <string.h>
+/* #define __USE_GNU 1 */
```
rdmark, line 37: Maybe remove this?

etc/afpd/afprun.c
```diff
@@ -24,20 +24,20 @@
...
 
 /* FIXME */
```
rdmark, line 39: Do you know what this FIXME refers to?

etc/cnid_dbd/cnid_metad.c
```diff
@@ -33,20 +33,20 @@
 #include <unistd.h>
 #undef __USE_GNU
 
-#include <stdlib.h>
-#include <sys/param.h>
+// #define _XPG4_2 1
```
rdmark, line 36: What does this commented out macro mean?

etc/cnid_dbd/dbd.h
```diff
@@ -11,6 +11,8 @@
 
 #include <atalk/cnid_bdb_private.h>
 
+#include "dbif.h"
```
rdmark, line 14: How did you realize that this header needed to be imported?

etc/uams/uams_dhx2_passwd.c
```diff
@@ -11,34 +11,32 @@
...
+#include <time.h>
 #include <unistd.h>
 
 #ifdef HAVE_CRYPT_H
 #include <crypt.h>
 #endif
 
-#include <sys/time.h>
-#include <time.h>
+#ifdef HAVE_LIBGCRYPT
```
rdmark, line 28: FWIW libgcrypt is a mandatory dependency so this macro is arguably redundant.

rdmark (approved):
In addition to avoid triggering your OCD, this will help preventing duplicate header inclusion in the future. I’ve run into more than one instance of duplication in this code base in the past.

etc/afpd/afprun.c
dgsga: Removed

etc/cnid_dbd/dbd.h
dgsga: Because it provides the definition of the type DBD which is called upon by multiple files in etc/cnid_dbd. If not present then there are multiple DBD undefined expression errors when the headers are correctly ordered.

etc/cnid_dbd/cnid_metad.c
dgsga: Refers to ancient SunOS version so removed

etc/afpd/afprun.c
dgsga: I have no idea, it's always been there. From the linux man pages:

setresuid() sets the real user ID, the effective user ID, and the saved set-user-ID of the calling process.
seteuid() sets the effective user ID of the calling process. Unprivileged processes may only set the effective user ID to the real user ID, the effective user ID or the saved set-user-ID.

Hmm...

etc/uams/uams_dhx2_passwd.c
dgsga: Best done in a separate commit, I think.

etc/afpd/afprun.c
rdmark: Surely something to do with the safety of doing this. Looking at git blame, this fixme goes back to 2005 when the afprun feature was added initially. Something for another changeset, so never mind!

etc/uams/uams_dhx2_passwd.c
rdmark: Fair enough

PR #986, opened 2024-05-15
[2.4] meson: Add WolfSSL as an external or built-in SSL provider

dgsga:


rdmark (approved):
That was fast! The patch must have applied pretty cleanly on 2.x?

dgsga:
> That was fast! The patch must have applied pretty cleanly on 2.x?

It did, only a few conflicts needed to be resolved :)

NJRoadfan:
Guess I'm going to have to switch over to using meson from now on.

rdmark:
It's totally worth the effort!

PR #988, opened 2024-05-15
Fix regression from previous commit

dgsga:


rdmark (approved):
It seems main has the same typo?

PR #1000, opened 2024-05-18
meson: Refactor SSL macro

dgsga:


dgsga:
@rdmark, whilst working on enhancing the SSL provider in the configuration summary it suddenly occurred to me that we could simplify the SSL macro greatly by utilising the fact that WolfSSL, LibreSSL and OpenSSL all use pkg-config. So in this PR there are 2 commits. The first updates the macro, the second removes WolfSSL as an external provider of SSL. This is because none of the prebuilt packages I have tested so far have the right configuration to provide DHX and RANDNUM support. The wolfssl/options.h file **must** contain OPENSSL_ALL and DES_ECB defines for it to work, so any user will have to compile WolfSSL from source anyway. What do you think?

rdmark:
Agreed. I had no luck with Debian's WolfSSL library either.

rdmark (approved):
Congrats on scoring #1000 !

PR #1003, opened 2024-05-19
meson: Remove unneeded embedded SSL configure flag from workflow

dgsga:


rdmark (approved):
I'm fine either way, but there's some value in being explicit with certain options, in case we change the default behavior in the future.

PR #1008, opened 2024-05-19
Revisit CVE-2022-23121, CVE-2022-23123 regression fixes by @andychen-syno and @anodos325

dgsga:
- Added guard check before access ad_entry()
- Allow zero length entry in compliance with AppleDouble specification
- Avoid setting adouble entries on symlinks

### DIFF
--- a/libatalk/adouble/ad_attr.c
+++ b/libatalk/adouble/ad_attr.c
@@ -23,15 +23,14 @@
 int ad_getattr(const struct adouble *ad, uint16_t *attr)
 {
     uint16_t fflags;
-    char *adp = NULL;
     *attr = 0;
 
-    if (ad_getentryoff(ad, ADEID_AFPFILEI) && (adp = ad_entry(ad, ADEID_AFPFILEI)) != NULL) {
-        memcpy(attr, adp + AFPFILEIOFF_ATTR, 2);
+    if (ad_getentryoff(ad, ADEID_AFPFILEI) && ad_entry(ad, ADEID_AFPFILEI)) {
+        memcpy(attr, ad_entry(ad, ADEID_AFPFILEI) + AFPFILEIOFF_ATTR, 2);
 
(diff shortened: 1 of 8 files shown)


dgsga:
@rdmark, this commit implements @andychen-syno's original 2022 CVE patches along with @anodos325's patch to avoid setting double entries on symlinks. This gives the best of both worlds with a much cleaner log, especially on classic Mac clients.

rdmark:
@dgsga Sorry it's taking a while to wrap my head around this changeset. I trust your judgement of course, but I also want to internalize the logic changes.

dgsga:
No worries, if you find any flaws please feed back. This PR is definitely optional!

libatalk/adouble/ad_attr.c
```diff
@@ -98,71 +96,59 @@ int ad_setattr(const struct adouble *ad, const uint16_t attribute)
...
-    if (ade == NULL) {
-        LOG(log_warning, logtype_ad, "ad_setid(%s): failed to set ADEID_PRIVINO", adp->ad_name);
+    if (!ad_entry(adp, ADEID_PRIVINO)) {
+        LOG(log_debug, logtype_default, "ad_setid(%s): invalid PRIVINO", adp->ad_name);
         EC_FAIL;
     }
-    memcpy(ade, &ino, sizeof(ino_t));
+    memcpy(ad_entry( adp, ADEID_PRIVINO ), &ino, sizeof(ino_t));
 
-    if (adp->ad_vers != AD_VERSION_EA) {
```
rdmark, line 146: Interesting that the check for EA style metadata was removed here. Otherwise the block’s logic seems unchanged. Is DID a valid concept for EA metadata? (I’m asking out of curiosity)

rdmark (approved):

dgsga:
Reverted this PR to draft as want to do more research...

PR #1015, opened 2024-05-22
[2.4] meson: Fix linking of UAM modules using embedded SSL

dgsga:


dgsga:
@rdmark, please can you test loading of uams_dhx.so and uams_randnum.so on your Debian system with this PR?

rdmark:
@dgsga Yep this does away with the symbol error when loading the DHX library in my Debian 12 VM. I'm traveling for work right now and can't do any more thorough testing than this. :)

I'm getting `main: atp_open: Cannot assign requested address` errors when starting afpd. Assuming it's something in my configuration, but mentioning here just in case it's a weird breakage of some sort.

As a side note, it's kind of unfortunate that we have to bundle so much code that originates with the OpenSSL project, encumbered with that iffy license.

dgsga:
> @dgsga Yep this does away with the symbol error when loading the DHX library in my Debian 12 VM. I'm traveling for work right now and can't do any more thorough testing than this. :)
> 
> I'm getting `main: atp_open: Cannot assign requested address` errors when starting afpd. Assuming it's something in my configuration, but mentioning here just in case it's a weird breakage of some sort.
> 
> As a side note, it's kind of unfortunate that we have to bundle so much code that originates with the OpenSSL project, encumbered with that iffy license.

Yes, I agree. There is no way around it if we want to maintain DHX authentication. All of the cast library files are taken from SSLeay 0.9 which was the precursor of OpenSSL. These files do not use the OpenSSL Apache license, just the original SSLeay licence (which is co-incidentally also used by LibreSSL)

Randnum and DHX UAM's now tested as loading correctly with v2 and v3 :)

include/atalk/cast.h
```diff
@@ -71,36 +71,16 @@ extern "C" {
...
 #define CAST_KEY_LENGTH	16
 
-typedef struct cast_key_st
-	{
+typedef struct cast_key_st {
 	CAST_LONG data[32];
-	} CAST_KEY;
+} CAST_KEY;
 
-#ifndef NOPROTO
```
rdmark, line 79: What’s the reason for this local modification?

rdmark (approved):

include/atalk/cast.h
dgsga: It's an irrelevant conditional

PR #1017, opened 2024-05-24
[2.4] docs: Move documentation from conf file templates to man pages

rdmark:
- Move docs and examples from conf files to man pages, leaving only a file description and reference to the man page in the former
- Move docs concerning the conf file from the daemon man pages to the conf file man pages
- Grammar and organization touchups
- Format with XMLmind


rdmark:
@NJRoadfan FYI I'm making minor adjustments to the papd documentation here. Please let me know if you have any feedback.

PR #1025, opened 2024-05-24
[2.4] GitHub CI: Add Alpine and openSUSE; run netatalk after installing

rdmark:
- Add Alpine and openSUSE jobs
- Sort all jobs alphabetically
- Bump OS versions to the latest stable across the board
- Run netatalk after installing where feasible

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -5,8 +5,13 @@ on:
       - main
       - branch-*
     paths-ignore:
-      - 'CONTRIBUTORS'
-      - 'NEWS'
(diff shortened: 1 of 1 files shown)


rdmark:
Current failures will be unblocked with https://github.com/Netatalk/netatalk/pull/1026

Additionally, a few constraints:
- the Debian docker image doesn't have the appletalk kernel module (even when you try to manually load it)
- can't start netatalk on Solaris because the init script attempt to start atalkd as well

dgsga (approved):

Issue #1028, opened 2024-05-25
Bring back quota in 3.x


rdmark:
This is to discuss bringing back quota support before the 3.2.0 release.

The reasoning is:

At least two users brought up specific present and past use cases, e.g.
- https://github.com/Netatalk/netatalk/issues/697
- https://sourceforge.net/p/netatalk/mailman/message/38268026/

We have kept quota alive and working fine in the 2.x branch, and supporting it with the meson build system has turned out to be fairly low-effort.

rdmark:
The revert applied 99% cleanly, so I'll attempt to port over the Meson build rules from 2.x

rdmark:
While working on this, I found a critical patch for NetBSD quota that was merged to the 2.x branch in 2017 but not brought to the 3.x main branch: bf1bbf20b105c69712d786465ce3452c3617af9f "Use the netbsd-6 quota API, not the prerelease stuff from 5.99.x."

Without this, compilation fails on NetBSD, even when quota is disabled (tries to compile part of the quota code.)

PR #1032, opened 2024-05-25
Revert "Remove quota support from Netatalk 3"

rdmark:
This reverts commit 5e05f3b912619eed2730b4b6bc06e3ff88ace432.

Bring back Quota support, GitHub #1028

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -89,6 +89,7 @@ jobs:
             krb5-dev \
             libevent-dev \
             libgcrypt-dev \
+            libtirpc-dev \
             libtool \
             libtracker \
             linux-pam-dev \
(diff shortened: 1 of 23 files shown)


rdmark:
To remember to follow up: Right now, autotools detects and builds quota for OmniOS, but the meson build scripts do not.

rdmark:
@dgsga This works on Linux now, but I think quota support in 3.x was a bit broken on BSD platforms with https://github.com/Netatalk/netatalk/commit/450114e8c52574c5444d856e8a0208f2bbaab80c (and potentially other places). Unless you have a better idea, I'll try my hands on ifdeffing.

rdmark:
Found that 2.x has a major patch for NetBSD style quota that was never ported to 3.x: https://github.com/Netatalk/netatalk/commit/bf1bbf20b105c69712d786465ce3452c3617af9f

Issue #493, opened 2023-09-18
Remove quota support from Netatalk 3


dgsga:
No longer needed as storage capacity is not an issue on current SSD's and HDD's

rdmark:
A thought: When adding or removing "major" features, we might want to consider a minor version bump of this software. I.e. 3.2.0

The other major thing in the pipeline is the whole WolfSSL thing to get DHX working again.

Thoughts? Any other features (adding or removing) on the roadmap that you had in mind?

rdmark:
Also, I think we should ask on the mailing lists if anyone in the current install base are relying on this feature.

rdmark:
Asked the community on the admins mailing list. Let's see if anyone raises concerns.

cdevers-es:
This would have been a big problem for us a couple years ago.

At my employer, we make heavy use of the quota feature for shared storage solutions, where different user groups get different space allocations. If quota support wasn’t available for AFP mounts, then this would have led to a lot of unexpected behavior & tech support calls with convoluted workarounds, because the quota limits would still be enforced at the server’s filesystem layer, but anyone using AFP would falsely be led to believe that each of their mounts has access to the full capacity of the overall system. We’d basically have to tell AFP users to ignore what Finder is telling them, and to instead look in other interfaces to understand how their space allocation & availability information. 

For better or worse, the proposed change won’t matter to us nowadays …but that’s because we deprecated & removed Netatalk/AFP from our product, so our customers aren't using it anymore. When Apple started downplaying AFP support in modern Mac versions, our customers started switching to alternatives, so by the time we removed Netatalk, few if any of our customers were still using AFP anyway. 

(If Samba were to drop quota support, that would be a bigger headache for us. But that's a discussion for elsewhere.)

There may well be other shared-storage products out there that are still making similar use of quotas, and might be caught out by this change. It certainly seems worth putting in something more than a dot-dot-patch release where people would normally only expect minor bug fixes, not the removal of a once-significant feature. Hopefully, any affected users/companies see this discussion before the September 25th deadline (three days from now) and can weigh in if this is going to be a problem for them. 

I certainly understand the desire to pay down technical debt, and phase out support for features that aren't widely used anymore, in order to make the slimmed-down software easier to support & extend. Hopefully it's true that the population of people using quotas with Netatalk is small enough now that this is the right course of action. Good luck!

Michael-Wohlstadter:
[...]

PR #1034, opened 2024-05-28
Reintroduce quota support, modernized with meson build system

rdmark:
As per discussion in https://github.com/Netatalk/netatalk/issues/1028, bring back quota which was removed within this release cycle.

- Revert https://github.com/Netatalk/netatalk/issues/493
- Forward-port NetBSD patch that was only in 2.x https://github.com/Netatalk/netatalk/commit/bf1bbf20b105c69712d786465ce3452c3617af9f
- Modernize quota code, removing ancient platform workarounds
- Linux ifdefs for code introduced with https://github.com/Netatalk/netatalk/commit/450114e8c52574c5444d856e8a0208f2bbaab80c (caused build errors for DragonflyBSD; might be a cleaner way to do this...)
- Forward-port Meson build system scripts from 2.x with modifications

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -47,12 +47,14 @@ env:
     libpam0g-dev \
     libssl-dev \
     libtalloc-dev \
+    libtirpc-dev \
     libtool \
     libtool-bin \
     libtracker-sparql-3.0-dev \
     libwrap0-dev \
     meson \
     ninja-build \
+    quota \
     systemtap-sdt-dev \
     tcpd \
     tracker \
@@ -87,6 +89,7 @@ jobs:
             krb5-dev \
             libevent-dev \
             libgcrypt-dev \
+            libtirpc-dev \
             libtool \
             libtracker \
             linux-pam-dev \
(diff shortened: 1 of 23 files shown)


dgsga (approved):
@rdmark, Great job and detective work! 🕵️‍♂️
At some point in the near future I would like to rationalise all the dependencies and link args in the meson build system so the netatalk executables and libraries are only linked against those required to provide requested features. So, for example, quota_link_args  would not be added to the link_args kwarg if quota support was disabled. At the moment this approach is not consistently applied across the code base.

rdmark:
@dgsga Thanks for reviewing. It was surprisingly quite the journey to get to a state that's on par with 2.x.

Do you remember the purpose of this line in the root meson.build scripts? I brought it over verbatim from 2.x and it doesn't seem to have an adverse effect (can still enable quota for many Linux flavors) but don't quite understand it.

```
elif host_os.contains('linux')
...
cdata.set('NO_QUOTA_SUPPORT', 1)
```

It's obvious why we do this for Darwin (disable quota support), but not sure why it's needed for Linux here.

rdmark:
Answering my own question: Without setting NO_QUOTA_SUPPORT for Linux, compilation fails in nfsquota.c across all Linux distros. It's a dirty hack, but we need it for now...

PR #1036, opened 2024-05-28
Embedded SSL: Remove OpenSSL compatibility layer and unused headers

dgsga:
This commit:
Reduces the size of libssl from 2.6 to 1.5 MB
Removes 36 unused headers from the codebase
Maintains code compatibility with external SSL providers


dgsga:
@rdmark, we finally have the minimal embedded SSL library I have been after. The behaviour of the WolfSSL codebase is now entirely controlled by the addition of EMBEDDED_SSL conditionals. If we decide in the future to move away from external SSL providers altogether then we can migrate the code entirely to WolfSSL functions. I have already tested this in v3 and it works great.

rdmark (approved):
Nicely done! Getting it down to the bare minimum is quite the feat.

PR #1040, opened 2024-05-29
GitHub CI: quota and tracker on Fedora

rdmark:
configuring with quota and libtirpc -- turned out that we need a capability check to make this work on Fedora

tracker isn't a dependency of tracker-devel in Fedora, so let's install it explicitly

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -307,8 +307,10 @@ jobs:
             openssl-devel \
             pam-devel \
             perl \
+            quota-devel \
             systemd \
             systemtap-sdt-devel \
+            tracker \
             tracker-devel
       - name: Autotools - Bootstrap
         run: ./bootstrap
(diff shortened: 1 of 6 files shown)


rdmark:
Funnily, Fedora's libtirpc is kinda different from libtirpc in other Linux distros. If I hack the enum data type name, the code otherwise builds fine on Fedora. Some clever compatibility abstraction is required...

from rpcsvc/rquota.h

Fedora does

```
enum qr_status {
	Q_OK = 1,
	Q_NOQUOTA = 2,
	Q_EPERM = 3,
};
```

Debian (and Solaris etc.) does

```
enum gqr_status {
	Q_OK = 1,
	Q_NOQUOTA = 2,
	Q_EPERM = 3,
};
```

dgsga (approved):

rdmark:
@dgsga This is a lot of added build system complexity for a single enum having a slightly different name on Red Hat...

But a good learning experience for me. :)

PR #1042, opened 2024-05-31
docs: Restructure some sources, and localize Japanese header navbar

rdmark:
- Rename `netatalk.html` to the more accurate `pageheader.txt` -- it is not a stand-alone html page, but rather a fragment that is injected into the manual pages.
- Localize pageheader.txt to Japanese.
- Move the python script for generating the 'compile' page to `doc/`, since it also compiles the Japanese translation.
- Add a publishing date stamp to the compile page
- Touch up the wordings in the compile page
- Correct documentation and NEWS (most line changes are XMLmind reformatting afp.conf.5.xml)
- Don't build with libressl on macOS in GitHub CI


rdmark:
@dgsga A few last touch-ups as I was preparing the documentation for release today.

dgsga (approved):

PR #1047, opened 2024-06-01
[2.4] GitHub CI: More OSes: Arch, DragonflyBSD, FreeBSD, OpenBSD, OmniOS

rdmark:
This gets us to parity between 2.x and 3.x CI configurations.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -123,6 +123,56 @@ jobs:
       - name: Meson - Uninstall
         run: ninja -C build uninstall
 
+  build-archlinux:
+    name: Arch Linux
+    runs-on: ubuntu-22.04
+    container:
+      image: archlinux:latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Install dependencies
+        run: |
+          pacman -Sy --noconfirm \
+            autoconf \
+            automake \
+            cracklib \
+            gcc \
+            libtool \
+            make \
+            meson \
+            ninja \
+            pkgconfig
+      - name: Autotools - Bootstrap
+        run: ./bootstrap
+      - name: Autotools - Configure
+        run: |
+          ./configure \
+            --disable-install-privileged \
+            --enable-systemd \
+            --with-cracklib
+      - name: Autotools - Build
+        run: make -j $(nproc)
+      - name: Autotools - Run tests
+        run: make check
+      - name: Autotools - Install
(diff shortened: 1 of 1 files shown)


dgsga:
> OmniOS blocked by #1048 and #1049

@rdmark I have committed the ldap_initialize PR so #1048 is resolved. As well as issue #1049 there appear to be two causes for the Omnios VM failure:
The network/openldap package is not installed
There is a failure in the libsearch_dir macro for Omnios as meson does not recognise 'omnios' as a host_os
1 is easily fixable
2 is being fixed as a separate issue

rdmark:
@dgsga Curiously, we do not install `openldap` for omnios in the 3.x workflow, but it seems to be pulled in (at least the client library) by some other dependency. Which is why I concluded it isn't mandatory. Won't hurt to install it though, certainly.

dgsga:
@rdmark, this should now be unblocked by PR's #1052 and #1054

rdmark:
@dgsga Cheers, thanks for working that out. I'll review them momentarily. If all checks out for me too, I will consider branch-netatalk-2-4 release candidate code, and start testing/release prep later today.

rdmark:
@dgsga FYI, Meson is behaving a bit weirdly on the OmniOS runner... even though it has privileges to run the normal install script, it does not get permissions to run `'/usr/gnu/bin/rm -f /etc/rc2.d/S90netatalk'` (as part of the install hook.)

I chalk this up to vmactions weirdness and/or a bug in the particular software versions on this VM, since it works correctly in my local VM with sudo privileges.

Proceeding with a workaround...

rdmark:
@dgsga I think this is good to go now.

OmniOS required a few workarounds...

- Workaround no.1: Don't run the Solaris init script hooks in the Autotools job. That seems to mess up the file system permissions somehow, and prevent Meson from doing the same later.
- Workaround no.2: After installing the Solaris init scripts with Meson, explicity do chmod 755 on them to set the executable bit. This isn't needed in my local VM...
- Workaround no.3: Install the openldap package. It is needed in the vmactions environment for 2.x for some reason for the right ldap.h to be available to Meson...

dgsga:
Just one small tweak: Updated the Omnios pkgsrc repo

dgsga (approved):

rdmark:
Good call, thanks!

PR #1056, opened 2024-06-02
meson: Fix library search macro on Omnios hosts

dgsga:


rdmark (approved):

rdmark:
Using uname to check for omnios seems a bit hackish and breaking the host_os detection model in Meson... Let me merge this fix now to unblock the release, but it might be worth revisiting later.

PR #1060, opened 2024-06-02
[2.4] GitHub CI: Build with libtirpc on Alpine

rdmark:
Enables the quota feature


dgsga (approved):
@rdmark, should we enable quota support automatically if libtirpc is found? One less configure flag :) It's easily done.

rdmark:
Sure, i think that would be pretty harmless. 

I’m not super confident in the quality of our quota code but it’s a good idea to get early warning when it starts failing somewhere. :)

PR #1072, opened 2024-06-04
[2.4] meson: set setuid bit to allow user afppasswd changing

rdmark:


dgsga (approved):
Fixed mini typo!

rdmark:
...and the same notation is represented in the shell (Debian below), if you need corroborating evidence! The stark text highlighting is actually the reason I noticed this in the first place...

Screenshot 2024-06-05 at 23 59 30

rdmark:
All that to say: Please let me know if you object to the reverting of your commit before merging this. I may have overlooked something. :)

dgsga:
You learn something everyday! Quite happy for you to revert my fix 👍

PR #1073, opened 2024-06-05
[2.4] meson: Fix papd crash when cupsautoadd is enabled

dgsga:
This commit:

- Fixes the logic of the UCS-2-INTERNAL macro that was causing the crash (credit: @NJRoadfan)
- Creates the required empty spooldir on install
- Fixes the linking of a2boot and timelord with libatalk

### DIFF
--- a/contrib/a2boot/meson.build
+++ b/contrib/a2boot/meson.build
@@ -14,4 +14,6 @@ executable(
     ],
     install: true,
     install_dir: sbindir,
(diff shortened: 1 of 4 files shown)


NJRoadfan:
Note: Updates to UCS-2-INTERNAL macro test should be ported to the 3.x branch as well.

rdmark (approved):
Beautiful!

Issue #1068, opened 2024-06-04
I can't backup after updating to 3.2


marcosvrs:
**Describe the bug**
I can't backup after updating to 3.2.
It says, "Verification may have been interrupted by a network problem. Try again after the network problems are resolved."
<img width="262" alt="Screenshot 2024-06-04 at 16 53 33" src="https://github.com/Netatalk/netatalk/assets/10147574/521ac4c0-d824-446c-804a-de66c3d33233">

Just tested with 3.1.18 and it works with the same settings.

**To Reproduce**
```yaml
  netatalk:
    image: netatalk/netatalk3
    environment:
      AFP_PASS: [secret]
      AFP_USER: marcosvr
      LANG: C.UTF-8
    cap_add:
      - NET_ADMIN
    network_mode: host
    volumes:
      - /mnt/afpbackup:/mnt/afpbackup
      - /var/run/dbus:/var/run/dbus
```

**Expected behavior**
It should work as version 3.1.18.

**Environment**
 - Server OS: Ubuntu 24.04
 - Client OS: macOS Sonoma 14.5
 - Netatalk Version: 3.2

**Logs**
https://pastebin.com/XHUZU1cY

rdmark:
Thanks for reporting the issue. This is a strange one. I cannot see anything out of the ordinary in those logs...

Does the same thing happen if you start a fresh backup on a fresh volume? And then subsequently run an update to the backup onto the backup volume?

The main difference between the 3.1.18 and 3.2.0 Docker images is that the former uses Debian 11 and the latter Alpine 3.10. So perhaps some minute difference in the file system or AFP metadata on the volume, between the two Linux variants...?

marcosvrs:
<details>

<summary>Default log level running 3.2</summary>

```
netatalk-1  | *** Setting up users and groups
netatalk-1  | chpasswd: password for 'marcosvr' changed
netatalk-1  | afppasswd: max password length is 8.
netatalk-1  | NOTE: Use a password of 8 chars or less to authenticate with Mac OS 8 or earlier clients
netatalk-1  | *** Configuring shared volume
netatalk-1  | *** Fixing permissions
netatalk-1  | *** Configuring Netatalk
netatalk-1  | *** Starting AFP server
netatalk-1  | Found user 'avahi' (UID 86) and group 'avahi' (GID 86).
netatalk-1  | Successfully dropped root privileges.
netatalk-1  | avahi-daemon 0.8 starting up.
[...]

PR #1076, opened 2024-06-07
docker: Revert to Debian Bookworm as base image

rdmark:
For testing https://github.com/Netatalk/netatalk/issues/1068


rdmark:
Discarding.

PR #1079, opened 2024-06-07
Alpine workflow: Add missing tracker-miners package

dgsga:
The tracker-miners package adds the missing daemon command to tracker3 on Alpine Linux


rdmark (approved):

rdmark:
Ah, good catch. I now recall doing the same "fix" in the 2.4 branch. :)

PR #1082, opened 2024-06-09
meson: Address various issues with the meson build system

dgsga:
This changeset refines the meson build system:

- Enables quota support on all flavours of linux and BSD (including macOS)
- Adds the quota provider to the configuration summary
- Adds a user option to disable LDAP support
- Sets dependencies according to user configuration
- Improves the syntax of the ACL macro

### DIFF
--- a/etc/afpd/unix.h
+++ b/etc/afpd/unix.h
@@ -27,12 +27,17 @@
 #include <sys/mnttab.h>
 #endif /* __svr4__ || HAVE_SYS_MNTTAB_H */
 
+#if defined(__DragonFly__)
+#define dqblk ufs_dqblk
+#endif
+
+#if defined(HAVE_SYS_MOUNT_H) || defined(BSD4_4) || defined(__linux__)
 #include <sys/mount.h>
+#endif /* HAVE_SYS_MOUNT_H || BSD4_4 || __linux__ */
 
-#if defined(HAVE_MNTENT_H)
+#if defined(linux) || defined(HAVE_MNTENT_H)
 #include <mntent.h>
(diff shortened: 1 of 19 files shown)


dgsga:
@rdmark, I will backport this PR to 2.4 on approval.

etc/afpd/unix.h
```diff
@@ -202,4 +210,8 @@ extern int setfilunixmode   (const struct vol *, struct path*, const mode_t);
 extern int setfilowner      (const struct vol *, const uid_t, const gid_t, struct path*);
 extern void accessmode      (const AFPObj *obj, const struct vol *, char *, struct maccess *, struct dir *, struct stat *);
 
+#ifdef AFS	
+    #define accessmode afsmode
+#endif
```
rdmark, line 215: I thought we got rid of all traces of Andrew File System... but I assume this is something else?

etc/cnid_dbd/meson.build
```diff
@@ -1,73 +1,81 @@
...
-    'dbd_lookup.c',
-    'dbd_rebuild_add.c',
-    'dbd_resolve.c',
-    'dbd_search.c',
-    'dbd_update.c',
-    'dbif.c',
-    'main.c',
-    'pack.c',
-]
+if use_dbd_backend
```
rdmark, line 1: Is there an argument to be made to always enable the dbd backend?

The argument to always enable: It is the one single stable and full-featured CNID backend presently.

The argument to make it optional: The potential of creating a new CNID backend in the future (e.g. sqlite) which will become the default.

rdmark (approved):
Huge improvement!

PR #1083, opened 2024-06-10
Fix regression from commit 9fad81a

dgsga:


rdmark:
@dgsga May I be super anal for a moment here? With the goal of preserving a readable commit log for this project, I would like to suggest a more descriptive description of the commit here. Removing any context from other commits, what is it that these code changes fix? In this particular case, it would be something like "Use the more portable __linux__ platform macro" ...

And if you want to call out that it is a regression, it can be mentioned on a second line of the commit message, for instance.

What do you think?

rdmark (approved):
Cheers!

PR #1084, opened 2024-06-11
[2.4] meson: Address various issues with the meson build system

dgsga:
This changeset refines the meson build system:

- Enables quota support on all flavours of linux and BSD (including macOS)
- Adds the quota provider to the configuration summary
- Adds user options to disable gssapi, cracklib and LDAP support
- Sets dependencies according to user configuration
- Improves the syntax of the ACL macro
- Automates Berkeley DB library detection on macOS

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -142,7 +142,8 @@ jobs:
             make \
             meson \
             ninja \
-            pkgconfig
+            pkgconfig \
+            rpcsvc-proto
       - name: Autotools - Bootstrap
         run: ./bootstrap
       - name: Autotools - Configure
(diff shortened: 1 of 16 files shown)


dgsga:
> Beautiful...
> 
> I think it was a fair approach to start out by converting the Autotools logic as-is, and then iterate to mesonify it. This is looking better and better.
> 
> The obvious remaining low hanging fruit off the top of my head are... Standardize the setup options, and a cleaner approach to substitution. Anything else?
> 
> I personally think both can wait until 3.3.0 (or whatever the next feature release will be.)

Thanks! I have just added missing options to disable cracklib and gssapi support to this changeset. I agree that the next step is standardising setup options. Fixing substitutions (which is dead easy on meson) will have to wait until we drop autotools support as it requires making simple changes to some .tmpl and .xml files which would break autotools. The great thing is we will finally be able to restore conf file paths to the man and manual files.

PR #1085, opened 2024-06-12
[2.4] meson: Convert missing library messages to warnings

dgsga:


meson.build
```diff
@@ -788,7 +792,7 @@ else
 endif
 
 if with_quota != 'false' and not have_quota
-    message('Ouota support requested but libtirpc or libquota libraries not found')
+    warning('Ouota support requested but libtirpc or libquota libraries not found')
```
rdmark, line 795: Typo: The first word of the string should be "Quota" with a Q

rdmark (approved):
Just the one typo fix and this is good to go.

meson.build
dgsga: @rdmark, Oops missed this one! I'll correct it in the options semantics PR I'm preparing for v2...

PR #1086, opened 2024-06-12
meson: Further refinements to meson build system

dgsga:
This changeset:

Adds further refinements to quota macro
Adds user options to disable cracklib and GSSAPI support
Automates Berkeley DB library detection on macOS (via Homebrew or MacPorts)

### DIFF
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -65,10 +65,10 @@ option(
     description: 'Enable build of PGP UAM module',
 )
 option(
-    'enable-quota',
+    'with-quota',
     type: 'feature',
     value: 'auto',
-    description: 'Turn on quota support',
+    description: 'Turn on quota support (default: true)',
 )
 option(
     'enable-rpath',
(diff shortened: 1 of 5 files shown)


meson_options.txt
```diff
@@ -65,10 +65,10 @@ option(
...
 )
 option(
-    'enable-quota',
-    type: 'feature',
-    value: 'auto',
-    description: 'Turn on quota support',
+    'with-quota',
+    type: 'string',
+    value: '',
+    description: 'Turn on quota support (default: true)',
```
rdmark, line 71: How will users know that they can pass a path string to this option?

dgsga: Good point. Will revert this change until we deal with option semantics as I need this one to be merged first

rdmark (approved):

meson_options.txt
rdmark: SGTM

PR #1092, opened 2024-06-14
meson: Fix issues with quota support on linux and macOS hosts

dgsga:
This changeset:

- Fixes quota support on macOS hosts
- Restores missing configuration option for linux hosts (taken from autotools netatalk.m4 macro)
- Removes obsolete quota configuration data for linux and macOS hosts

macOS quota support tested on Sonoma host with Mac OS 9 client


meson_config.h
```diff
@@ -109,6 +109,9 @@
 /* Define to 1 if you have the `backtrace_symbols' function. */
 #mesondefine HAVE_BACKTRACE_SYMBOLS
 
+/* Define to 1 if dbtob impllementation is broken. */
```
rdmark, line 112: Small typo in code comment: _implementation_ ;)

rdmark (approved):
Is the "broken dbtob implementation" common knowledge? How did you figure that you?

meson_config.h
dgsga: Fixed!

dgsga:
> Is the "broken dbtob implementation" common knowledge? How did you figure that you?

I referred back to the quota macro in the autotools netatalk.m4 kitchen sink macro and noticed that this config flag had not been passed over to meson.

```
dnl ----- Linux specific -----
if test x"$this_os" = "xlinux"; then
	AC_MSG_RESULT([ * Linux specific configuration])
    AC_DEFINE(LINUX, 1, [OS is Linux])	
	dnl ----- check if we need the quotactl wrapper
    AC_CHECK_HEADERS(linux/dqblk_xfs.h,,
		[AC_CHECK_HEADERS(linux/xqm.h linux/xfs_fs.h)
        	AC_CHECK_HEADERS(xfs/libxfs.h xfs/xqm.h xfs/xfs_fs.h)]
	)


	dnl ----- as far as I can tell, dbtob always does the wrong thing
	dnl ----- on every single version of linux I've ever played with.
	dnl ----- see etc/afpd/quota.c
	AC_DEFINE(HAVE_BROKEN_DBTOB, 1, [Define if dbtob is broken])
fi
```

PR #1093, opened 2024-06-14
[2.4] meson: Fix issues with quota support on macOS and linux hosts

dgsga:
This changeset:

- Fixes quota support on macOS hosts
- Restores missing configuration option for linux hosts (taken from autotools netatalk.m4 macro)
- Removes obsolete quota configuration data for linux and macOS hosts

macOS quota support tested on Sonoma host with Mac OS 9 client

### DIFF
--- a/meson_config.h
+++ b/meson_config.h
@@ -119,6 +119,9 @@
 /* Define to 1 if you have the 'backtrace_symbols' function. */
 #mesondefine HAVE_BACKTRACE_SYMBOLS
 
+/* Define to 1 if dbtob implementation is broken. */
+#mesondefine HAVE_BROKEN_DBTOB
+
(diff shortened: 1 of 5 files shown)


meson_config.h
```diff
@@ -119,6 +119,9 @@
 /* Define to 1 if you have the 'backtrace_symbols' function. */
 #mesondefine HAVE_BACKTRACE_SYMBOLS
 
+/* Define to 1 if dbtob impllementation is broken. */
```
rdmark, line 122: Typo in comment: implementation

rdmark (approved):

meson_config.h
dgsga: Sorted!

PR #1099, opened 2024-06-16
meson: Simplify options semantics and standardize feature macros

dgsga:
This changeset:

- Simplifies the syntax of the build options to make them more user-friendly
- Standardises the syntax of the feature macros so they all share the same logic
-  Fixes the logic of the largefile support macro
- Disables kerberos and gssapi support if the Kerberos V UAM is not required
- Uses the recommended syntax for compiling and installing (using meson compile and meson install rather than their ninja equivalents)
- Converts the install guide to markdown format for easy reading


dgsga:
@rdmark, I will backport this to 2.4 once you have approved and/or suggested any changes...

rdmark:
@dgsga Can you think of any particular pros/cons to using `meson compile -C` vs. `ninja -C`? In most cases the compile tool is still ninja anyways, right?

I guess one thing is that `meson compile` allows for other compile backends transparently, which has some cross-platform / portability benefits. The one other plus off the top of my head is that using the `meson` command for everything is a bit more intuitive to first-time users of the build system.

The main drawback I can think of is that you have to type more characters to do the same thing.

rdmark:
@dgsga One high level remark: I notice you did away with the use of "feature" options altogether. Did you conclude that the semantics and [object oriented facilities of "feature"](https://mesonbuild.com/Build-options.html#features) doesn't bring much benefit to the way that you implemented Meson for netatalk? (I assume so, but am curious about your reasoning.)

I guess all netatalk features are effectively in "auto" mode at this point, and the only drawback I can think of myself right now with the boolean options is that even if you explicitly set it to "true" the setup will still pass with a warning, whereas "feature=enabled" will fail the build if a dependency can't be resolved (unless I'm mistaken).

Any other considerations that counted towards your design here?

INSTALL.md
```diff
@@ -0,0 +1,233 @@
...
+with-pgp-uam           Compile PGP UAM                             false
+with-quota             Enable Quota support                        true
+with-sendfile          Enable sendfile support                     true
+with-shadow            Enable shadow password support              true
+with-overwrite         Overwrite existing configuration files      false
+with-spotlight         Enable Spotlight support                    true
+with-tcp-wrappers      Enable TCP wrappers                         true
+with-tests             Compile tests                               false
+with-zeroconf          Enable Zeroconf (Bonjour) support           true
+```
```
rdmark, line 101: My own personal preference is to not overdocument something that is already self-documented (i.e. in meson_options.txt) – keeps maintenance overhead lower when options get changed.

I can see the argument for overdocumentation here though: To help onboard users with the new build system. Maybe we keep it like this for now, and downsize the documentation in future major releases. :)

INSTALL.md
```diff
@@ -0,0 +1,233 @@
...
+options.
+
+## External software dependencies
+
+Required:
+
+  - libevent 2
+  - libgcrypt version 1.2.3 or later (for DHX2 UAM support)
+  - Berkeley DB version 4.6.0 or later
+  - perl
```
rdmark, line 131: Perl is an optional dependency now...

INSTALL.md
```diff
@@ -0,0 +1,233 @@
...
+docbook
+libiconv
+pam
+```
+
+Please see `meson_options.txt` for full details of all Netatalk-specific options,
+and https://mesonbuild.com/Builtin-options.html for details of generic Meson
+options.
+
+## External software dependencies
```
rdmark, line 124: On the topic of overdocumentation, the external dependencies are also [listed up in the manual](https://github.com/Netatalk/netatalk/blob/main/doc/manual/install.xml). The purpose is slightly different in that the manual explains in a bit more depth what each dependency does. Having both is probably a good idea.

INSTALL.md
```diff
@@ -0,0 +1,233 @@
...
+Required for Spotlight support:
+  - talloc
+  - tracker version 0.12 or later
+  - bison
+  - flex
+
+Optional:
+
+  - Kerberos V              (for krbV UAM support)
+  - libressl or OpenSSL@1.1 (for Randnum and DHX UAM support)
```
rdmark, line 142: Maybe a note here that a bundled SSL provider is available?

INSTALL.md
```diff
@@ -0,0 +1,233 @@
...
+  - Berkeley DB version 4.6.0 or later
+  - perl
+
+Required for Spotlight support:
+  - talloc
+  - tracker version 0.12 or later
+  - bison
+  - flex
+
+Optional:
```
rdmark, line 139: This list is missing libacl, libldap, libpam, libquota/libtirpc ...

meson_options.txt
```diff
@@ -191,98 +131,171 @@ option(
...
+    'with-krbV-uam',
+    type: 'boolean',
+    value: true,
+    description: 'Enable build of Kerberos V UAM module',
+)
+option(
+    'with-largefile',
+    type: 'boolean',
+    value: true,
+    description: 'Enable support for large files',
```
rdmark, line 155: I think it's good to mention that this is mandatory for AFP 3.x

meson_options.txt
```diff
@@ -191,98 +131,171 @@ option(
...
-    description: 'Kerberos V support',
+    type: 'boolean',
+    value: true,
+    description: 'Enable Kerberos V API support',
+)
+option(
+    'with-krbV-uam',
+    type: 'boolean',
+    value: true,
+    description: 'Enable build of Kerberos V UAM module',
```
rdmark, line 149: Is there a scenario when with-kerberos and with-krbV-uam are enabled separately? In other words, would you ever enable the Kerberos V API without also enabling the Kerberos V UAM?

meson_options.txt
```diff
@@ -191,98 +131,171 @@ option(
...
 )
 option(
     'with-mysql-config',
     type: 'string',
     value: 'mysql_config',
-    description: 'Path to mysql-config binary',
+    description: 'Specify name of mysql-config binary',
+)
+option(
+    'with-overwrite',
```
rdmark, line 206: Minor thing, but while adjusting option names, why not be more specific with this one and call it something like "with-overwrite-conf"? (As opposed to binaries or libraries etc.)

meson.build
```diff
@@ -885,60 +897,66 @@ if rpc_headers_ok
...
+        quota_provider += 'libtirpc'
+        if cc.compiles(
+            '''
+                    #include <rpcsvc/rquota.h>
+                    int main(void) {
+                        enum qr_status foo;
+                        foo = Q_OK;
+                        return 0;
+                    }
+                ''',
```
rdmark, line 917: Nitpick: Close of multiline string has uneven indentation compared to the beginning.

meson.build
```diff
@@ -1116,65 +1134,65 @@ endif
...
-    enable_pgp_uam = have_ssl
-    if get_option('enable-pgp-uam').enabled() and not have_ssl
-        error('Need SSL support to build PGP UAM')
+    have_pgp_uam = have_ssl
+    if have_pgp_uam
+        cdata.set('UAM_PGP', 1)
+    else
+        have_pgp_uam = false
+        warning(
+            'Embedded SSL, LibreSSL or OpenSSL veresion 1.1 is required to build PGP UAM',
```
rdmark, line 1176: Nitpick: Typo in "version"

meson.build
```diff
@@ -1320,73 +1341,83 @@ endif
...
-if iconv.found() or libiconv.found()
-    if cc.compiles(
-        '''
+if not enable_iconv
+    have_iconv = false
+else
+    have_iconv = (
+        (iconv.found() or libiconv.found())
+        and cc.compiles(
+            '''
```
rdmark, line 1369: Uneven indentation with the closing `'''` again – but maybe this is deliberate?

meson.build
```diff
@@ -1763,25 +1810,24 @@ dtrace_command = [
...
-if host_os == 'freebsd' and cpu == 'aarch64'
-    with_dtrace = false
-    message('The version of dtrace in aarch64 FreeBSD is currently unsupported')
-elif not get_option('with-dtrace')
-    with_dtrace = false
+if not enable_dtrace
+    have_dtrace = false
+elif host_os == 'freebsd' and cpu == 'aarch64'
+    have_dtrace = false
+    warning('The version of dtrace in aarch64 FreeBSD is currently unsupported')
```
rdmark, line 1817: How about we leave a reference to https://wiki.freebsd.org/arm64#Known_Limitations in either a code comment, or the message?

INSTALL.md
```diff
@@ -0,0 +1,233 @@
...
+
+  - Kerberos V              (for krbV UAM support)
+  - libressl or OpenSSL@1.1 (for Randnum and DHX UAM support)
+  - xsltproc                (for manpages and mananual documentation)
+  - Docbook XSL stylesheets (for manpages & manual documentation)
+  - GLib 2 and dbus-glib    (for afpstats support)
+  - cracklib                (for cracklib support)
+  - avahi or mDNSresponder  (for Zeroconf support)
+  - tcpwrap                 (for TCP wrapper support)
+
```
rdmark, line 149: Maybe tell the reader that more details can be found in the html manual at https://netatalk.io/stable/htmldocs/installation

rdmark (approved):
Huge improvement, with a few minor improvement suggestions inline...

One thought kept bubbling up in my brain: Isn't this too big a changeset for a mere 0.0.1 release version bump? It's effectively a breaking change that introduces new functionality (granularity) to the build system. But then again, it is more of a "do-over" of the main new feature of 3.2.0, so there is arguably leeway to roll forward since I positioned 3.2.0 as an "early adopter" release.

In either case, I'll make sure to put a before-after matrix of the options that changed names and syntax in the release notes.

I'm on the fence here frankly and would like to hear your opinion.

dgsga:
> @dgsga Can you think of any particular pros/cons to using `meson compile -C` vs. `ninja -C`? In most cases the compile tool is still ninja anyways, right?
> 
> I guess one thing is that `meson compile` allows for other compile backends transparently, which has some cross-platform / portability benefits. The one other plus off the top of my head is that using the `meson` command for everything is a bit more intuitive to first-time users of the build system.
> 
> The main drawback I can think of is that you have to type more characters to do the same thing.

Yes, I agree with the last comment. To my mind it makes sense to use the meson-recommended syntax now as it may become mandatory at some point in the future. Good to get into good habits right from the start.

dgsga:
> @dgsga One high level remark: I notice you did away with the use of "feature" options altogether. Did you conclude that the semantics and [object oriented facilities of "feature"](https://mesonbuild.com/Build-options.html#features) doesn't bring much benefit to the way that you implemented Meson for netatalk? (I assume so, but am curious about your reasoning.)
> 
> I guess all netatalk features are effectively in "auto" mode at this point, and the only drawback I can think of myself right now with the boolean options is that even if you explicitly set it to "true" the setup will still pass with a warning, whereas "feature=enabled" will fail the build if a dependency can't be resolved (unless I'm mistaken).
> 
> Any other considerations that counted towards your design here?

OK, here's my take. The reason I discarded feature options was purely for the sake of simplicity for the user. 

My own personal preference would be to change all the -Dwith-feature= true/false boolean-type options to -Denable-feature=auto/enabled/disabled feature-type ones. This would mean that the user would have to input two different option types and values according to whether they are enabling/disabling a feature, or setting a string library path/init-style.

I will revert this PR to draft until we have reached consensus. Let me know what you think...

INSTALL.md
dgsga: I agree.

INSTALL.md
dgsga: Fixed

meson.build
dgsga: All these formatting issues will be fixed once meson 1.5.0 is released. At this stage I will meson format all the meson.build files so their formatting is entirely consistent with meson guidelines.

meson.build
dgsga: As above

meson.build
dgsga: Fixed

INSTALL.md
dgsga: Fixed

INSTALL.md
dgsga: Done

INSTALL.md
dgsga: OK

meson.build
dgsga: Done

INSTALL.md
dgsga: Fixed

PR #1100, opened 2024-06-16
[2.4] a2boot - Update meson.build file paths

### LINKED ISSUE #1095: [2.4] a2boot: Not working with meson builds.
System is running Debian 11 x64.

Steps to reproduce:
-Build netatalk 2,4 using meson
-Attempt to netboot an Apple IIgs.

When attempting to netboot the machine ether locks up or crashes at the first stage boot loader.

The following is in the debug log:
`a2boot open error on /usr/local/etc/netatalk/a2boot/ProDOS16\ Boot\ Blocks`

The listed file is present in that directory. Builds using autotools are working correctly.

NJRoadfan:
When built with meson, a2boot wasn't able to find the boot block files due to invalid path. Fixes #1095

### DIFF
--- a/contrib/a2boot/meson.build
+++ b/contrib/a2boot/meson.build
@@ -6,11 +6,11 @@ executable(
     c_args: [
         '-D_PATH_A_GS_BLOCKS="'
         + pkgconfdir
-        + '/a2boot/ProDOS16\ Boot\ Blocks"',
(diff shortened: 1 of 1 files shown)


rdmark (approved):

rdmark:
Thank you for the patch!

PR #1107, opened 2024-06-19
Update the SonarCloud scanner command

rdmark:
Also removes the shallow clone rule. The job is fast enough without it.


dgsga (approved):

rdmark:
Oops, I misunderstood what `fetch-depth:0` does. Apparently, it fetches all commit history for all tags and branches, which is the opposite of what it sounds like...

https://github.com/actions/checkout?tab=readme-ov-file#fetch-all-history-for-all-tags-and-branches

Will revert that part.

PR #1114, opened 2024-06-21
[2.4] meson: Set executable flags when installing scripts

rdmark:
- Use the install_mode property to set executable flags on installed scripts
- Rename lp2pap.sh to lp2pap, to align with the naming pattern of other installed scripts
- Set file system executable flag on add_netatalk_printer, which is a stand-alone script


dgsga (commented):
@rdmark, shouldn't we be setting the install_mode kwarg for these scripts?

https://nettools.club/chmod_calc

NJRoadfan:
There is no guarantee that file permissions will be retained with source code. Metadata like file permissions can be easily lost in transit.

rdmark:
Fair enough, let me apply 755 permissions in the meson script.

NJRoadfan (approved):
Tested meson build and install. Scripts now have executable bit set on install and can be run from the command line.

rdmark:
Thanks for testing (and reporting)!

dgsga (approved):

PR #1121, opened 2024-06-22
GitHub CI: Shouldn't need avahi libraries to build zeroconf with mDNS

rdmark:


rdmark:
Blocked by https://github.com/Netatalk/netatalk/issues/1122

PR #1130, opened 2024-06-23
meson: with-rpath syntax is invalid

rdmark:
If you try to do `-Dwith-rpath=true` you get error

```
../meson.build:414:7: ERROR: Tried to access unknown option 'enable-rpath'.
```

I think that part should just be removed, no?

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -353,7 +353,7 @@ enable_rpath = get_option('with-rpath')
 
 if not get_option('with-rpath')
     enable_rpath = false
-elif host_os == 'sunos' or host_os == 'netbsd' or get_option('enable-rpath')
+elif host_os == 'sunos' or host_os == 'netbsd'
     enable_rpath = true
 endif
 


rdmark:
BTW, the reason this came up was when attempting to package 2.4.0 for Debian... lintian complains that rpath violates policy. This was not a problem with Autotools.

```
E: netatalk2: custom-library-search-path RUNPATH /usr/lib/x86_64-linux-gnu [usr/bin/ad]
N: 
N:   The binary or shared library sets RPATH or RUNPATH. This overrides the
N:   normal library search path, possibly interfering with local policy and
N:   causing problems for multilib, among other issues.
N:   
N:   The only time a binary or shared library in a Debian package should set
N:   RPATH or RUNPATH is if it is linked to private shared libraries in the
N:   same package. In that case, place those private shared libraries in
N:   /usr/lib/*package*. Libraries used by binaries in other packages should be
N:   placed in /lib or /usr/lib as appropriate, with a proper SONAME, in which
N:   case RPATH/RUNPATH is unnecessary.
N:   
N:   To fix this problem, look for link lines like: gcc test.o -o test
N:   -Wl,--rpath,/usr/local/lib or gcc test.o -o test -R/usr/local/lib and
N:   remove the -Wl,--rpath or -R argument. You can also use the chrpath
N:   utility to remove the RPATH.
N: 
N:   Please refer to https://wiki.debian.org/RpathIssue for details.
N: 
N:   Visibility: error
N:   Show-Always: no
N:   Check: binaries/rpath
N:   Renamed from: binary-or-shlib-defines-rpath
```

dgsga (approved):

PR #1132, opened 2024-06-23
autotools and meson: Use pkg-config to find libgcrypt

dgsga:


dgsga:
@rdmark, I have disabled the 2018 Solaris vmactions VM until it is fixed upstream as it cannot use meson and it gives a false negative for libgcrypt with this PR. I have tested it on my 11.4.42 Solaris VM and it finds libgcrypt using both meson and autotools build systems. The up to date version of Solaris does not need the PKG_CONFIG_PATH configuration flag either (with either build system)

rdmark:
Got it, let's keep Solaris on ice for now, then. We can add tentative Meson steps so that they can be reflected in the generated manual appendix.

rdmark (approved):
The Meson change is so clean compared to Autotools... I'm sorry I had to make you do both!

PR #1133, opened 2024-06-23
[2.4] autotools and meson: Use pkg-config to find libgcrypt

dgsga:


.github/workflows/build.yml
```diff
@@ -714,6 +712,7 @@ jobs:
             ninja -C build uninstall
 
   build-solaris:
+    if: false
```
rdmark, line 715: Is this a temporary disabling of the Solaris job?

Does the the libgcrypt detection work on an up to date Solaris system?

dgsga: @rdmark, I have disabled the 2018 Solaris vmactions VM until it is fixed upstream as it cannot use meson and it gives a false negative for libgcrypt with this PR. I have tested the PR on my 11.4.42 Solaris VM and it finds libgcrypt using both meson and autotools build systems. It is safe to remove the CAST5 API check as this only applied to a version shipped with OpenSolaris in 2008!

Another user and myself have requested that the vmactions VM be updated but no luck so far...

https://github.com/vmactions/solaris-vm/issues/38

rdmark (approved):

PR #1137, opened 2024-06-24
Add copyright for mac_roman.h

rdmark:
All code in this file is lifted verbatim from lib/mac_roman.h in libiconv.

Therefore, adding the same copyright year and author as in libiconv 1.9.2 which was contemporaneous to when mac_roman.h was added to Netatalk.

https://ftp.gnu.org/pub/gnu/libiconv/libiconv-1.9.2.tar.gz


rdmark:
@dgsga FYI, this was something that Debian's copyright scanner caught as a source file without author and copyright year.

dgsga (approved):

PR #1141, opened 2024-06-24
[2.4] Don't attempt to restart atalkd if NBP registrations fail

NJRoadfan:
In some configurations, the NBP registrations of "workstation" and "netatalk" may fail. Don't attempt to restart the atalkd service as these are not critical failures. Note, this fixes the issue that lead to these changes being reverted in #473.


rdmark (approved):
Thanks for the improvement!

PR #1152, opened 2024-06-25
Bump to version 1.0.0 of bstrlib

rdmark:
https://github.com/websnarf/bstrlib/releases/tag/v1.0.0

### DIFF
--- a/include/atalk/bstrlib.h
+++ b/include/atalk/bstrlib.h
@@ -1,13 +1,14 @@
 /*
  * This source file is part of the bstring string library.  This code was
- * written by Paul Hsieh in 2002-2008, and is covered by the BSD open source
+ * written by Paul Hsieh in 2002-2015, and is covered by the BSD open source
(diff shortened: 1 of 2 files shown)


dgsga (approved):
Quality.

I see the bounds check in bstrlib.c
```
	/* Did the operation complete successfully within bounds? */
	if (n >= (l = b->slen + (int) (strlen) ((char *) b->data + b->slen))) {
```
satisfies the Sonarcloud security warning  'make sure use of "strlen" is safe here'.

rdmark:
@dgsga This is my read, too!

As a side note, the same line existed before this changeset, except a const qualifier was dropped.

rdmark:
Reverting to draft. I want to spend a little more time analyzing the Sonar reported issues to make sure we don't introduce new bugs.

rdmark:
I confirmed that none of the Sonar reported bugs are new, or significant. There is some layout changes and moving code around that triggers scanning.

PR #1155, opened 2024-06-26
GitHub Cl: Fix zeroconf support on Alpine Linux

dgsga:


.github/workflows/build.yml
```diff
@@ -75,6 +75,7 @@ jobs:
             acl-dev \
             autoconf \
             automake \
+            avahi-compat-libdns_sd \
             avahi-dev \
```
rdmark, line 79: I think we got our wires crossed here. I meant the opposite change. :)

But I may very well be mistaken and these are needed for Zeroconf?

dgsga: Paradoxically both deps are both needed for Zeroconf support (via mDNS) on Alpine:
Just avahi-compat-libdns_sd = No zeroconf support
Both avahi-compat-libdns_sd and avahi-dev = Zeroconf support using mDNS
Just avahi-dev: No can do as avahi-compat-libdns_sd  is installed as a dependency of avahi-dev.

In other words, just using Avahi to provide Zeroconf support on Alpine is not an option...

rdmark (approved):
I trust your judgment on this one...

PR #1157, opened 2024-06-26
[2.4] GitHub Cl: Add zeroconf support to Alpine workflow

dgsga:


.github/workflows/build.yml
```diff
@@ -69,6 +69,8 @@ jobs:
             acl-dev \
             autoconf \
             automake \
+            avahi-compat-libdns_sd \
+            avahi-dev \
```
rdmark, line 73: Do we really need Avahi for Zeroconf support on Alpine? It seems to come with mDNS libraries out of the box.

dgsga: Paradoxically both deps are both needed for Zeroconf support (via mDNS) on Alpine:
Neither avahi-compat-libdns_sd or avahi-dev = No zeroconf support
Just avahi-compat-libdns_sd = No zeroconf support
Both avahi-compat-libdns_sd and avahi-dev = Zeroconf support using mDNS
Just avahi-dev: No can do as avahi-compat-libdns_sd is installed as a dependency of avahi-dev.

In other words, just using Avahi to provide Zeroconf support on Alpine is not an option...

rdmark: That’s odd! I set up the Dockerfile for 2.4 with Alpine and without the Avahi packages, and am sure that I tested that it worked with service discovery. Let me double check that docker setup.

dgsga: See here for the origin of libdns_sd:

https://pkgs.alpinelinux.org/contents?branch=edge&name=avahi-compat-libdns_sd&arch=x86&repo=main

rdmark (approved):

PR #1160, opened 2024-06-28
meson: Prioritize native WolfSSL package over built-in SSL library if it has the correct configuration

dgsga:
This commit enables meson to detect if a native WolfSSL library with the correct configuration is present and use it in preference to built-in SSL support

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -522,13 +522,44 @@ if enable_dtags
     ssl_link_args += '-Wl,--enable-new-dtags'
 endif
 
-if have_embedded_ssl or crypto.found()
+wolfssl = dependency('wolfssl', required: false)
+
+if wolfssl.found()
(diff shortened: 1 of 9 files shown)


meson.build
```diff
@@ -542,7 +573,7 @@ elif crypto.found()
 else
     have_ssl = false
     warning(
-        'Either LibreSSL or OpenSSL version 1.1 is required for DHX and RANDNUM support',
+        'Built-in SSL, LibreSSL or OpenSSL version 1.1 is required for DHX and RANDNUM support',
```
rdmark, line 576: You mean "Built-in WolfSSL"?

dgsga: Fixed

rdmark (approved):

PR #1161, opened 2024-06-28
[2.4] meson: Prioritize native WolfSSL package over built-in SSL library if it has the correct configuration

dgsga:


dgsga:
@rdmark, only Debian and Ubuntu WolfSSL packages appear to have the correct config...

meson.build
```diff
@@ -582,7 +613,7 @@ elif crypto.found()
 else
     have_ssl = false
     warning(
-        'Either LibreSSL or OpenSSL version 1.1 is required for DHX and RANDNUM support',
+        'Built-in SSL, LibreSSL or OpenSSL version 1.1 is required for DHX and RANDNUM support',
```
rdmark, line 616: Did you mean to say "Built-in WolfSSL"?

.github/workflows/build.yml
```diff
@@ -91,7 +92,8 @@ jobs:
             openrc \
             openssl-dev \
             pkgconfig \
-            rpcsvc-proto-dev
+            rpcsvc-proto-dev \
+            wolfssl-dev
```
rdmark, line 96: Isn't it misleading to install the wolfssl-dev for this OS if it won't be usable by netatalk?

rdmark: (Same comment for other OSes except Debian and Ubuntu below)

dgsga: I thought it was a good way to test the logic of the macro. Will remove now I know it works

meson.build
dgsga: Fixed

rdmark (approved):

PR #1164, opened 2024-06-29
docs: Rephrase tarball section of manual

rdmark:
This is to avoid talking about specific compression techniques (we currently don't distribute bzip2)


rdmark:
Merging trivial documentation update without review.

PR #1167, opened 2024-06-29
meson: netatalk version not propagated properly

rdmark:
These were overlooked in a recent PR


rdmark:
Merging without review to expedite release.

PR #1175, opened 2024-06-30
[2.4] meson: Add option to override system WolfSSL with embedded WolfSSL

dgsga:


meson.build
```diff
@@ -548,6 +548,7 @@ message('BDB version: ', bdb_version)
 #
 
 have_embedded_ssl = get_option('with-embedded-ssl')
+force_embedded_ssl = get_option('with-ssl-override')
```
rdmark, line 551: Sorry if I'm overlooking an obvious constraint, but can't `with-embedded-ssl` be used as the "forcing" option?

Or conversely, is there a reason to keep `with-embedded-ssl` when introducing `with-ssl-override`?

dgsga: with-embedded-ssl is enabled by default. If a correctly configured wolfssl package is detected then with-embedded-ssl has to be disabled to prevent linking issues. So if we wish to override native wolfssl we have to have a separate option which overrides both...

meson_options.txt
```diff
@@ -165,6 +165,12 @@ option(
     value: true,
     description: 'Enable SLP (srvloc) support',
 )
+option(
+    'with-ssl-override',
+    type: 'boolean',
+    value: false,
+    description: 'Force built-in WolfSSL asSSL provider',
```
rdmark, line 172: Typo "asSSL provider" -> "as SSL provider"

meson.build
rdmark: I think get it now. The `with-embedded-ssl` option effectively functions as a flag for whether a functional system library was found or not.

rdmark (approved):

meson_options.txt
dgsga: Fixed

PR #1178, opened 2024-06-30
meson: Fix errors in PAM support macro

dgsga:
Restore missing option to set PAM config path
Install PAM config file to correct location
Fix PAM_SESSION entry in PAM config file for Debian/SuSE


dgsga:
@rdmark, I'm glad you spotted the missing option as while testing the fix I discovered two important bugs in the PAM macro that have now been fixed in this PR

rdmark (approved):
Thanks for the quick fix!

PR #1182, opened 2024-06-30
meson: Fix operation of D-Bus path macros

dgsga:
This commit fixes errors in setting the correct dbus-daemon and dbus-sysconf path variables. The variables now match those set by autotools.

### DIFF
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -204,7 +204,7 @@ option(
 option(
     'with-dbus-sysconf-path',
     type: 'string',
-    value: '/etc/dbus-1/system.d',
+    value: '',
     description: 'Set path to D-Bus system bus security configuration directory',
 )
 option(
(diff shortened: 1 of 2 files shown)


meson_options.txt
```diff
@@ -198,13 +198,13 @@ option(
...
     type: 'string',
-    value: '/bin/dbus-daemon',
+    value: '',
     description: 'Set path to D-Bus daemon',
 )
 option(
     'with-dbus-sysconf-path',
     type: 'string',
-    value: '/etc/dbus-1/system.d',
+    value: '',
```
rdmark, line 207: It's a bit unfortunate that the default value isn't defined here anymore, since we lose the transparency to users. I assume you did this because you want a fallback for macOS?

rdmark (approved):

meson_options.txt
dgsga: The value: argument in meson_options.txt may only be a string, int, bool, or array of those. The default dbus sysconf path is a concatenation of the `prefix` variable and `'etc/dbus-1/system.d'` so we have to leave it as an empty string. We can keep the default bus-daemon path though as it is just a pure string.

PR #1186, opened 2024-07-01
Use external SSL dependency to provide cast header

dgsga:


meson.build
```diff
@@ -1156,12 +1156,16 @@ endif
 #
 
 dbus_daemon_path = get_option('with-dbus-daemon-path')
+dbus_daemonpath = ''
```
rdmark, line 1159: All the dbus lines below seem to be from a different changeset?

dgsga:
@rdmark, this provides a workaround until we have a better alternative. Working on it...

rdmark (approved):
LGTM

dgsga:
Are you happy with the Sonarcloud static analysis warning?

rdmark:
It seems fine. It's just warning that CAST functions are deprecated in OpenSSL3, from what I can see.

rdmark:
@dgsga I can't get this to work in my Debian 12 VM. The resulting DHX UAM doesn't work for auth with either OS9 or macOS... I may be missing something obvious though!

NJRoadfan:
A longer term solution might be to use the CAST functions in libgcrypt. The license is compatible and its already a dependency.

dgsga:
> A longer term solution might be to use the CAST functions in libgcrypt. The license is compatible and its already a dependency.

Great minds think alike...

dgsga:
> @dgsga I can't get this to work in my Debian 12 VM. The resulting DHX UAM doesn't work for auth with either OS9 or macOS... I may be missing something obvious though!

@rdmark, tested login and file transfer on my Bookworm VM with OS9 client and all is well. Have you pulled the latest force commit for the branch?

rdmark:
Hmm, it seems like it's my OS9 setup that's broken. I'm certain it used to work last week.

image

I'm traveling for work right now so have limited means of testing. Please go ahead and merge at your convenience.

rdmark:
My problem was fixed by installing a fresh AppleShare Client 3.8.8 from the Mac Garden. Not sure what happened to the stock extension.

rdmark:
@dgsga With this solution, I think we need a check in Meson for openssl/libressl when building with bundled SSL. Otherwise you get compile errors like

```
14.73 ../etc/uams/uams_dhx_pam.c:33:10: fatal error: openssl/cast.h: No such file or directory
14.73    33 | #include <openssl/cast.h>
14.73       |          ^~~~~~~~~~~~~~~~
```

Also, the Dockerfile needs to be updated as well to add openssl and openssl-dev.

As a side note: For some reason Alpine's libressl ships with libcrypto.so.3 while netatalk expects libcrypto.so.52 at runtime ... so while you can link with it, the library cannot be found so DHX won't work.

dgsga:
> @dgsga With this solution, I think we need a check in Meson for openssl/libressl when building with bundled SSL. Otherwise you get compile errors like
> 
> ```
> 14.73 ../etc/uams/uams_dhx_pam.c:33:10: fatal error: openssl/cast.h: No such file or directory
> 14.73    33 | #include <openssl/cast.h>
> 14.73       |          ^~~~~~~~~~~~~~~~
> ```
> 
> Also, the Dockerfile needs to be updated as well to add openssl and openssl-dev.
> 
> As a side note: For some reason Alpine's libressl ships with libcrypto.so.3 while netatalk expects libcrypto.so.52 at runtime ... so while you can link with it, the library cannot be found so DHX won't work.

Good point, fixed.

PR #1198, opened 2024-07-06
[2.4] Fix Zeroconf on Alpine Linux, and in the Docker configuration

rdmark:
This restores functionality of the Docker container in 2.4:

- Missing fix for https://github.com/Netatalk/netatalk/pull/1155 in 2.4
- openssl needs to be a LIB_DEV, otherwise it gets purged before exporting the layers
- avahi compatibility libraries are required
- Disabling quota for now, as I was seeing errors in the afpd logs
- More explicit configuration flags

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,18 +2,20 @@ FROM alpine:3.20
 
 ENV LIB_DEPS \
     acl \
+    avahi \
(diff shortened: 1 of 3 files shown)


rdmark:
Merging without review to unblock release

PR #1205, opened 2024-07-07
meson: Don't install Japanese manual

rdmark:
The Japanese html manual is meant for publishing to the web, so let's not install it with Meson.

This is akin to how we treat the Japanese manual pages.


dgsga (approved):

rdmark:
Discarding. Turns out that this leads to the html pages not even being generated in the build dir.

PR #1206, opened 2024-07-07
meson: Replace OpenSSL with nettle as low-level crypto provider, convert built-in WolfSSL library to a submodule

dgsga:
This commit:

- Replaces OpenSSL with GNU nettle as the provider of CAST128 and DES functions.
- Converts the built-in WolfSSL library to a subproject to make package management easier for those distros that forbid bundled libraries eg Debian (see https://mesonbuild.com/Subprojects.html)
- Upgrades the Dockerfile to use nettle
- Upgrades the GitHub Cl workflow to use nettle

Legacy OpenSSL support is maintained for autotools builds, will be removed when transition to meson is complete.


rdmark:
This is reminder to test if `etc/uams/openssl_compat.h` can be removed with this changeset.

rdmark (approved):

Issue #262, opened 2023-04-02
ABI checks have not been run since 3.1.8


rdmark:
The ABI check step of the release QA process has not been run since the 3.1.8 release https://github.com/Netatalk/netatalk/wiki/Developer-Notes#abi-checking

rdmark:
@slowfranklin Since you were the last person committing an ABI file, can you comment on whether this QA mechanism is still valid today? I would assume so, and it seems to be a good idea to make sure we don't inadvertently change the behavior of the libatalk interface.

slowfranklin:
Yes, it's a good idea to do semantic versioning of public libraries and this was initially added to allow external consumers of libatalk safe usage, but no, as afaict there are no external consumers of libatalk I just dropped maintaing the ABI stuff at some point and we can drop this stuff. As an additional change we can then make libatalk a private library.

rdmark:
Now I understand the original intent. Agreed that there is little use in keeping this around. For starters I removed the ABI checking steps from the developer notes wiki page.

A few things to clean up:
- Remove the DEVELOPER flag from makefiles
- Remove VERSION_INFO from makefiles
- Remove the --enable-developer configure option
- Remove all stored abi files
- Remove any remaining documentation

Anything else?

rdmark:
@slowfranklin Please see if this covers all the bases https://github.com/Netatalk/netatalk/pull/269

PR #1261, opened 2024-07-18
autotools: Restore ABI versioning of libatalk, set to 18.0.0

rdmark:
The libatalk binary API versioning scheme was removed after discussion in GitHub #262 and effective with v3.1.16

However, since then, external consumers of libatalk have been found, such as macipgw. Additionally, ABI versioning allows distinguishing 3.x libatalk (18.0.0) from 2.x libatalk (0.0.0) in the wild.

This changeset also includes a modification to the OpenBSD CI job, due to bug https://github.com/Netatalk/netatalk/issues/1262

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -699,17 +699,12 @@ jobs:
             export LDFLAGS=-L/usr/local/lib
             autoreconf -fi
             ./configure \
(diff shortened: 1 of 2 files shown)


rdmark:
Also, Meson kept the v18 ABI versioning from its initial introduction in 3.2.0

PR #1294, opened 2024-07-23
meson: Better way to handle rpath executable targets

rdmark:
Instead of repeating each executable target with and without rpath, this defines a global variable that controls the same.


rdmark:
This seems to have the intended effect. On NetBSD, we get the RPATH set in the ELF library, but not on Debian where we don't want it.

```
# objdump -x build/libatalk/libatalk.so.18.0.0 | grep 'R.*PATH'
  RPATH                /usr/pkg/lib
```

PR #1303, opened 2024-07-24
Reintroduce AppleTalk transport layer

rdmark:
Bring back the modules, daemons and utility programs used for AppleTalk networking.


rdmark:
The failing Sonar quality gate is code duplication of 3.3% in new code. The biggest offender is PAP, specifically UAM auth for PAP. I think this is acceptable, given the boiler plate nature.

rdmark:
All the relevant code has now been brought back, and build system scaffolding set up. AppleTalk/DDP is presently disabled by default in the build system, because it doesn't fully compile yet, e.g. with the modified daemon child process model. We we will work through these areas of improvement one by one in subsequent PRs.

PR #1327, opened 2024-07-27
Explicitly import inttypes.h to build tests on *BSD

rdmark:


cheesestraws (approved):
same comment applies as on #1329 - assuming it's intentional for the build.yml changes to be in this PR, it looks sensible

PR #1328, opened 2024-07-27
[3.2] Avoid using reserved keyword to build the tests on NetBSD

rdmark:
Using `type` as the data type at the same time as `type` as the variable name causes the compiler on NetBSD to bail out.

This PR also includes changes to the GitHub CI workflow to build and run the tests on NetBSD, in addition to some general cleanup of the test steps elsewhere.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -128,8 +128,8 @@ jobs:
       - name: Meson - Configure
         run: |
           meson setup build \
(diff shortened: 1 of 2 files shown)


cheesestraws (approved):
same comment applies as on #1329 - assuming it's intentional for the build.yml changes to be in this PR, it looks sensible

PR #1329, opened 2024-07-27
[2.4] Avoid using reserved keyword to build the tests on NetBSD

rdmark:
Using `type` as the data type at the same time as `type` as the variable name causes the compiler on NetBSD to bail out.

This PR also includes changes to the GitHub CI workflow to build and run the tests on NetBSD, in addition to some general cleanup of the test steps elsewhere.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -114,9 +114,9 @@ jobs:
       - name: Meson - Configure
         run: |
           meson setup build \
-            -Dwith-tests=true \
             -Dwith-init-style=openrc \
-            -Dwith-quota=true
+            -Dwith-quota=true \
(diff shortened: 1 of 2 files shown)


cheesestraws:
Looks sensible to me, though is it intentional that the changes to build/run the tests are also in this PR?

cheesestraws (approved):

rdmark:
@cheesestraws Thank you for the review! Yes, the GitHub workflow commit was intentional, because it serves as a test that the code change works as intended. I did some other cleanup of the test related lines for the other jobs, which was part of testing if the same change had an impact on the other OSes (which it didn't). I've updated the PR description to indicate this.

Issue #1333, opened 2024-07-28
Restore AppleTalk support in afpd


rdmark:
afpd in Netatalk 2.x supports both ATP and DSI transport layers, whereas in 3.x it was refactored to only speak DSI.

Additionally, support for AFP revisions 1.1, 2.0 and 2.1 should be brought back.

NJRoadfan:
Some notes from my codebase comparisons:

`afp_avahi.c`
Add logic to disable entries if server is ATP only.

`auth.h`
Re-add old AFP revisions

`auth.c`
Re-add DSI and ATP logic in `status_versions()`
Maybe re-add ASP logic in `login()`? This looks like part of a feature that was removed in later revisions unrelated to AppleTalk.
Figure out if `print_groups()` and `set_groups()` is needed. Maybe removed due to lack of use?

`desktop.c`
New code was added to 3.x, verify it won't conflict with older AFP revisions.
Re-add DSI and ATP logic in `afp_addicon()`.

`file.c`
Re-add GS/OS date fix to `setfilparams()`.

`fork.c`
Check if support for nlmask and nlchar needs to be added back into `read_fork()`. This function saw significant refactoring in 3.x.
Re-add DSI and ATP logic in `read_fork()` and `write_fork()`

`ofork.c`
Re-add `of_pforkdesc()`.

`messages.c`
Re-add logic for `maxmsgsize` depending on DSI or ATP in `readmessage()`.

`unix.c`
See why `setdirmode()` and code using it in `directory.c` was removed.

`volume.c`
Re-add logic for limiting volume freespace for prodos and limitsize options in `getvolspace()`.
Quite a bit of code was reshuffled and refactored in here (volume creation was moved to libatalk under `netatalk_conf.c`), so there will be other changes yet to be discovered.

'netatalk_conf.c`
Re-add DSI and ATP specific logic in `volxlate()`.
Re-add DSI specific check in `hostaccessvol()`.

rdmark:
@NJRoadfan One question I don't think we have discussed yet, is controlling which transport layer to activate.

In 2.x afpd.conf has its `-ddp -tcp -transall` options that allows you to enable and disable either or both transport layers at runtime.

We either ought to bring this back as a dynamic afp.conf option, *or* always enable the transport layers that were compiled into afpd. I.e. always enable DSI, and enable DDP when netatalk was compiled with `-Dwith-appletalk=true`.

What do you think?

I'm leaning towards the latter to keep the logic simple.

[...]

PR #1339, opened 2024-07-28
Initial AppleTalk fixes

NJRoadfan:
-Add pre-AFP 2.2 login strings.
-Add back GS/OS date preservation fix.
-Get the classic AppleTalk daemons working.
-Get the classic AppleTalk commandline tools working.

Fixes part of #1333


rdmark:
@NJRoadfan Thank you for working on this. A minor housekeeping note: For PRs that target the main branch, let's not add the version number tag like [4.0]. They are only for distinguishing PRs against stable branches. Let's keep the tag soup manageable. ;)

I updated https://github.com/Netatalk/netatalk/wiki/Developer-Notes#pull-requests to make note of this.

NJRoadfan:
I've got a2boot and timelord compiling. Right now working on getting atalkd and papd working. After that, all that is left is getting afpd back to supporting ASP transports.

NJRoadfan:
If anyone wants to test, you need to disable ASP from libatalk for now in order to compile with AppleTalk turned on. In `libatalk/meson.build`, remove line 30 `libasp,`. ASP is only used by afpd, which has not been hacked apart to support AppleTalk yet.

etc/papd/uam.c
```diff
@@ -60,8 +60,10 @@ struct uam_mod *uam_load(const char *path, const char *name)
 
   /* version check would go here */
 
+  /* Temporary hack below to get this to compile with 3.x, we really should be passing an AFPObj as the first parameter */
```
rdmark, line 63: Please prefix this with a `TODO:` so that it's easy to find later.

libatalk/atp/atp_open.c
```diff
@@ -50,9 +53,9 @@ ATP atp_open(uint8_t port, const struct at_addr *saddr)
     struct timeval	tv;
     int			pid;
 
-#ifdef DEBUG
+#ifdef EBUG
```
rdmark, line 56: Why do you not want to use the DEBUG label here?

etc/meson.build
```diff
@@ -5,3 +5,8 @@ subdir('afpd')
 subdir('cnid_dbd')
 subdir('netatalk')
 subdir('uams')
+
+if have_appletalk
+    subdir('atalkd')
+    subdir('papd')
+endif
```
rdmark, line 12: Nitpick: Let's add a final newline for consistency. :)

rdmark:
@NJRoadfan Looks good overall. I left a few minor comments.

Can you please rebase this branch now so that it's a single commit, or a series of commits with atomic changes that compile independently? (reminder: `git rebase -i HEAD~n` where n is the number of commits that you want to squash.)

libatalk/atp/atp_open.c
NJRoadfan: Oddly, DEBUG isn't defined in the 3.x tree. It also might have been an orphan. All the other ATP source files are using EBUG in the same places for debug messages.

rdmark: Interesting. A deep dive into git blame will probably give a hint to what happened. I agree that we should stick with the current code standard.

rdmark (approved):
I’m good to merge this now. Please let me know if you are working on some final touches.

NJRoadfan:
Nothing further here.

PR #1341, opened 2024-07-29
[3.2] meson: Control the MySQL CNID backend, and support MariaDB

rdmark:
- When MySQL dependencies aren't available, fall back to MariaDB
- Introduce a `with-cnid-mysql-backend' option
- Remove obsolete `with-mysql-config' option
- Install *SQL dependencies in GitHub CI workflows

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -43,7 +43,7 @@ env:
     libglib2.0-dev \
     libkrb5-dev \
     libldap2-dev \
-    libmysqlclient-dev \
+    libmariadb-dev \
     libpam0g-dev \
     libssl-dev \
     libtalloc-dev \
@@ -96,6 +96,7 @@ jobs:
             libtool \
             libtracker \
             linux-pam-dev \
+            mariadb-dev \
             meson \
             ninja \
             openldap-dev \
@@ -155,6 +156,7 @@ jobs:
             gcc \
             libtool \
             make \
+            mariadb-clients \
             meson \
             ninja \
             pkgconfig \
@@ -300,6 +302,7 @@ jobs:
             libtalloc-devel \
             libtool \
             meson \
+            mysql-devel \
             ninja-build \
             openldap-devel \
             openssl-devel \
@@ -504,6 +507,7 @@ jobs:
               libgcrypt \
               libtool \
               meson \
+              mysql80-client \
               openldap26-client \
               perl5 \
               pkgconf \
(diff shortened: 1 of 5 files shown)


knight-of-ni:
So this check for mariadb doesn't quote work in Fedora.  Instead of mardiab-devel package, we must use mariadb-connector-c-devel. 

This isn't picked up with the change in this PR:
'Run-time dependency mariadb found: NO (tried pkgconfig and cmake)'

If use mariadb-devel instead, this check does work, but eventually a bug report gets filed against the package stating to no use this.

From below, mariadb-connector-c-devel places header files under /usr/include/mysql, rather than /usr/include/mysql/server

Doesn't meson have a find_library function, to just find mysql.h wherever it may be? Perhaps that would be simpler.

```
$ rpm -ql mariadb-connector-c-devel
/usr/bin/mariadb_config
/usr/bin/mysql_config
/usr/include/mysql
/usr/include/mysql/errmsg.h
/usr/include/mysql/ma_list.h
/usr/include/mysql/ma_pvio.h
/usr/include/mysql/ma_tls.h
/usr/include/mysql/mariadb
/usr/include/mysql/mariadb/ma_io.h
/usr/include/mysql/mariadb_com.h
/usr/include/mysql/mariadb_ctype.h
/usr/include/mysql/mariadb_dyncol.h
/usr/include/mysql/mariadb_rpl.h
/usr/include/mysql/mariadb_stmt.h
/usr/include/mysql/mariadb_version-x86_64.h
/usr/include/mysql/mariadb_version.h
/usr/include/mysql/mysql
/usr/include/mysql/mysql.h
/usr/include/mysql/mysql/client_plugin.h
/usr/include/mysql/mysql/plugin_auth.h
/usr/include/mysql/mysql/plugin_auth_common.h
/usr/include/mysql/mysql_version.h
/usr/include/mysql/mysqld_error.h
/usr/lib/.build-id
/usr/lib/.build-id/5a
/usr/lib/.build-id/5a/ddd9b3d0ad5717af2802268841d37042ad5984
/usr/lib64/libmariadb.so
/usr/lib64/libmysqlclient.so
/usr/lib64/libmysqlclient_r.so
/usr/lib64/pkgconfig/libmariadb.pc

```

knight-of-ni:
UPDATE:
Changing the dependency line to the follow allows Netatalk to find mariadb, tested againsts mariadb-connector-c-devel and mariadb-devel:

mariadb = dependency('**lib**mariadb', required: false)

rdmark:
Nice catch! The CI workflows confirms that this dependency works on all other platforms (in addition to Fedora, Alpine, Arch and OpenBSD are the ones that solely rely on the mariadb dependency at present.)

https://github.com/Netatalk/netatalk/pull/1347
https://github.com/Netatalk/netatalk/pull/1348

PR #1351, opened 2024-07-30
Reintroduce AppleTalk related manual pages

rdmark:
Brings back and updates the metadata for all AppleTalk related man pages, including building them for the html manual.

One change from 2.x, is that the man page aliases are achieved through straight up symlinking, rather than the static "link" man pages under man/man* ... since we did away with the intermediate step. The drawback is that you cannot install 4.x on top of 2.x, since a symlink cannot overwrite a file. Not sure if there's a better solution for this.

I *think* this is acceptable, since the 2.x->4.x migration path involves manual steps anyways (compare the [2.x->3.x migration path](https://netatalk.io/3.2/htmldocs/upgrade)) so adding one for deleting these fake man pages shouldn't be too bad...

OTOH, maybe we can just ditch those man page aliases for simplicity.


rdmark:
For the record, the part that generates the symlinks is

```
nbp_symlinks = [
    'nbplkup.1',
    'nbprgstr.1',
    'nbpunrgstr.1',
]

foreach symlink : nbp_symlinks
    install_symlink(symlink, install_dir: mandir / 'man1', pointing_to: 'nbp.1')
endforeach
```

PR #1355, opened 2024-07-31
Fix ASP services to work with 3.x codebase.

NJRoadfan:
There were some changes to `server_child.h` in the 3.x codebase. This updates the ASP routines to work with the newer code. Please note that this is untested until AppleTalk support is re-added into afpd.

With this PR, one can now enable AppleTalk and build main successfully.


libatalk/asp/asp_getsess.c
```diff
@@ -45,7 +45,7 @@
 #endif /* ! MIN */
 
 static ASP server_asp;
-static struct server_child *children = NULL;
+static server_child_t *children = NULL;
```
rdmark, line 48: Isn’t server_child_t also a struct, or am I mistaken?

rdmark:
This might be a good time to turn on -Dwith-appletalk=true for the GitHub CI workflow jobs to start continuously testing the code. Do you want to make that change in this PR?

libatalk/asp/asp_getsess.c
NJRoadfan: Might be a bug perhaps? 

Looking at  2.x, its declared as:
```

typedef struct server_child {

} server_child;
```

While 3.x its:

```
typedef struct {

} server_child_t;
```

Everything else that creates a `server_child_t` in the code base (including afpd) doesn't include `struct` in front of it... except for ASP, which was removed after the above change was made. Either way, it should be equivalent. I guess it depends on how fussy the compiler is.

rdmark: A quick dive into stack overflow teaches me that a struct type can be aliased in the global namespace if you double declare it, so to speak, in which case you don’t have to use the struct keyword when declaring variables that use it as a data type. 

This is the effect that the trailing type declaration in the 3.x construct has, as I understand it. But I think it would be even better to explicitly declare the initial struct type on the first line… but if unless some compiler complains I’m fine with the current state.

NJRoadfan:
Added -Dwith-appletalk=true to CI actions. Several platforms are failing with weird errors. Alpine is failing on values that are coming from built in libraries.

.github/workflows/build.yml
```diff
@@ -120,6 +120,7 @@ jobs:
           meson setup build \
             -Dwith-init-style=openrc \
             -Dwith-pgp-uam=true \
+            -Dwith-appletalk=true \
```
rdmark, line 123: In the future, please list options in alphabetical order.

This might seem like an onerous requirement, however there is a practical reason. When these lists of options get longer, it can be very easy to accidentally list the same option more than once.

This is a real thing that's happened frequently in this codebase, sometimes discovered years later. ;)

No need to take action right now, but please keep it mind the next time you touch this file.

rdmark:
> Added -Dwith-appletalk=true to CI actions. Several platforms are failing with weird errors. Alpine is failing on values that are coming from built in libraries.

Alpine and OmniOS are failing in a similar way with missing `O_RDWR` and similar symbols.

All of the BSDs are failing in papd with a missing `WNOHANG` symbol.

Off the top of my head: probably due to some missing platform or capability macro.

BTW, OpenBSD is passing just because there's a hard coded `cdata.set('NO_DDP', 1)` for this OS in the main meson.build. This originates with a [hotfix from 2011](https://github.com/Netatalk/netatalk/commit/6dd1fb33fedf945027b2b69ec180af5135ce16c8#diff-49473dca262eeab3b4a43002adb08b4db31020d190caaad1594b47f1d5daa810) and should probably be removed because it is better controlled at configure time, in my opinion.

NJRoadfan:
Looks like there are a few obsolete platform macros remaining in the 2.x code base causing the failures. The two in question are `HAVE_SYS_WAIT_H` and `HAVE_FCNTL_H` . I also ran into `HAVE_ERRNO_H` with the earlier commit. The 3.x codebase assumes all these headers are included, so it is likely safe to remove these macros.

rdmark:
@NJRoadfan This looks good to me now. I think it would be cleaner to rebase this changeset into a single commit.

rdmark (approved):

PR #1369, opened 2024-08-02
AppleTalk daemon support in init scripts

rdmark:
- Forward port NetBSD and systemd init scripts from 2.4
- Adapt the openrc and debian rc scripts in 2.4 into atomic init scripts
- Rename all script templates to a unified naming scheme


rdmark:
Tested that the atomic systemd/openrc/netbsd/debian init scripts are working. This should be good to go.

PR #1372, opened 2024-08-02
meson: Detect Linux distro and choose an appropriate init style

rdmark:
Use `/etc/os-release` to detect which Linux distro we are running on.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -119,7 +119,6 @@ jobs:
         run: |
           meson setup build \
             -Dwith-appletalk=true \
-            -Dwith-init-style=openrc \
             -Dwith-tests=true
       - name: Build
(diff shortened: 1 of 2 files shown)


rdmark:
There are surely other Linux distros that are worth detecting, but I used the ones that I have handy to actually test.

rdmark:
The distro detection here is fairly loose. I tried doing exact matches with e.g. `ID=alpine` but found that some distros put the id in quotation marks and some don't, while Meson does not support regular expression matching. So for now I do a global match for the distro keywords, which means that `ID_LIKE` fields will get matched too. For instance, Mint will be detected as Ubuntu since I don't have an explicit check for Mint:

```
ID=linuxmint
ID_LIKE=ubuntu
```

PR #1376, opened 2024-08-04
meson: Introduce option to control which manual l10n to build

rdmark:
- New meson option `-Dwith-manual-l10n` which takes an array of locales (currently only one, `ja`). Default is empty array, which leads to only the English l10n being built.
- Brings over the netatalk.css stylesheet from the netatalk-homepage repo and uses it when building the manual for local consumption.

### DIFF
--- a/doc/html.xsl.in
+++ b/doc/html.xsl.in
@@ -6,4 +6,5 @@
 	<xsl:param name="chunk.section.depth" select="0"/>
 	<xsl:param name="chunk.separate.lots" select="1"/>
 	<xsl:param name="toc.max.depth" select="2"/>
+	<xsl:param name="html.stylesheet" select="'netatalk.css'"/>
 </xsl:stylesheet>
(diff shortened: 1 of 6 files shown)


knight-of-ni:
Patch doesn't apply cleanly in my environment because I'm tracking netatalk-3-2 branch, rather than main, but this pr does look to be good.  

I do wonder if we ought to make the target folder path configurable as well. I get it, though. Having three different options just for the documentation, seems excessive.

rdmark:
Right, sorry I'm only doing clear-cut bugfixing for the stable 3.2 branch, and this is arguably more of an improvement.

At this moment I'm leaning towards not adding an option for customizing the subdir for the html manual. Installing a local html manual is already something very few users care about (I think) that I cannot really justify it in my head. For now, please continue using sed to patch the build files if you don't like the `htmldocs` subdir. ;)

rdmark:
I hear no objection, so merging this now. :)

PR #1377, opened 2024-08-04
Restore AppleTalk functionality in afpd

rdmark:
Very much a work in progress, but putting up this PR since the code compiles at least. (Although the tests are failing on some platforms.)

A few memos...

I chose not to port over srvloc, authprintdir or icon features. Maybe for later.

Mostly not touched desktop.c, volume.c, and uam.c yet because code has moved around alot.

The big commit where all the AppleTalk stuff got removed first is https://github.com/Netatalk/netatalk/commit/be96d276348da0a7e66eeec844b306e8a5fa8fac which served as a good reference.

The AFPobj struct got rejigged a lot, removing AppleTalk functionality when the ini parsing was rewritten https://github.com/Netatalk/netatalk/commit/df7560dfdb12b06090dc4b2c6e88d0858930b591#diff-ad0729adb892a8c3518ab325a752d53b59c6778afe97b76a32207d2c99bbb0d3

This was renamed: `afp_options->ipaddr` -> `afp_options->listen`

Two commits dealing with old charsets that we should check if they should be reverted: https://github.com/Netatalk/netatalk/commit/778588234ea038cae2ae0ca3cce1c61773154df4 https://github.com/Netatalk/netatalk/commit/f6f22820647c9654afa07c30a4d05441609b1f82

### DIFF
--- a/etc/afpd/afp_asp.c
+++ b/etc/afpd/afp_asp.c
@@ -1,0 +1,314 @@
+/*
+ * Copyright (c) 1997 Adrian Sun (asun@zoology.washington.edu)
+ * Copyright (c) 1990,1993 Regents of The University of Michigan.
+ * All Rights Reserved.  See COPYRIGHT.
+ *
+ * modified from main.c. this handles afp over asp.
+ */
+
+#ifdef HAVE_CONFIG_H
+#include "config.h"
+#endif /* HAVE_CONFIG_H */
+
+#ifndef NO_DDP
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <signal.h>
+#include <atalk/logger.h>
+#include <errno.h>
+#ifdef HAVE_SYS_TIME_H
+#include <sys/time.h>
+#endif /* HAVE_SYS_TIME_H */
+#ifdef HAVE_SYS_STAT_H
+#include <sys/stat.h>
+#endif /* HAVE_SYS_STAT_H */
+
+#include <netatalk/endian.h>
+#include <atalk/atp.h>
+#include <atalk/asp.h>
+#include <atalk/compat.h>
+#include <atalk/util.h>
+#include <atalk/globals.h>
+
+#include "switch.h"
+#include "auth.h"
+#include "fork.h"
+#include "dircache.h"
+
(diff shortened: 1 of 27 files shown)


NJRoadfan:
Originally, afpd supported multiple servers and multiple protocols using an AFPConfig struct for each server and protocol that wrapped around an AFPObj struct. When AppleTalk was removed, the code was simplified to directly call an AFPObj and the protocol handle was hardcoded to a DSI handle. Quite a bit of refactoring in `afp_config.c` and `main.c` simplified things as well.

First off, the `afp_options_parse_cmdline()` function now directly packs an AFPObj with command line options. This was changed to fix a bug here: https://github.com/Netatalk/netatalk/commit/7db833b1021f3daa8e30966e6f025b2a894348dc 

A few new features appeared after these changes that need to be maintained:
-The GSS UAM saw changes, which requires passing an AFPObj to `uam_load()` in `uam.c`. See: https://github.com/Netatalk/netatalk/commit/53e7f8a164ca820d5dfe41b7d39ffba8fb056a72 and https://github.com/Netatalk/netatalk/blob/main/etc/uams/uams_gss.c#L619
-An "afp interfaces" configuration option was introduced to allow listening on multiple network interfaces: https://github.com/Netatalk/netatalk/commit/21058d44aabd9f97e2edfd245770584412c30f2d

NJRoadfan:
Regarding charsets, I wouldn't worry about it as afpd configures them directly from settings in afpd.conf now. ATALK_MAC_CHARSET and ATALK_UNIX_CHARSET aren't being set in the modern init scripts anyway.

rdmark:
> Originally, afpd supported multiple "virtual" servers and multiple protocols using an AFPConfig struct for each server and protocol that wrapped around an AFPObj struct. When AppleTalk was removed, the code was simplified to directly call an AFPObj and the protocol handle was hard coded to a DSI handle, with only one server available.

This explains a lot. I think it will suffice for afpd to support exactly one DSI and one ASP server. No need to bring back the ability to spawn an arbitrary number of servers.

rdmark:
Handled in https://github.com/Netatalk/netatalk/pull/1435

PR #1404, opened 2024-08-14
[2.4] meson: Introduce a with-lockfile-path option

rdmark:
Use with-lockfile-path to set a lockfile path that is not defined in paths.h

This leverages the special `-D_PATH_LOCKDIR` compiler flag that is only in 2.x (refactored away in 3.x and later)

### DIFF
--- a/include/atalk/paths.h
+++ b/include/atalk/paths.h
@@ -13,13 +13,13 @@
 /* lock file path. this should be re-organized a bit. */
 #if ! defined (_PATH_LOCKDIR)
 #  if defined (FHS_COMPATIBILITY) || defined (__NetBSD__) || defined (__OpenBSD__) || defined (__APPLE__)
-#    define _PATH_LOCKDIR	"/var/run/"
+#    define _PATH_LOCKDIR	"/var/run"
 #  elif defined (BSD4_4)
-#      define _PATH_LOCKDIR	"/var/spool/lock/"
+#      define _PATH_LOCKDIR	"/var/spool/lock"
 #  elif defined (__linux__)
-#    define _PATH_LOCKDIR	"/var/lock/"
+#    define _PATH_LOCKDIR	"/var/lock"
 #  else
-#    define _PATH_LOCKDIR	"/var/spool/locks/"
+#    define _PATH_LOCKDIR	"/var/spool/locks"
 #  endif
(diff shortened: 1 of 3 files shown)


rdmark:
@VorpalBlade Would you be able to test if this works to set the lockfile path in your environment?

VorpalBlade:
I should be able to take a look on Friday evening (Europe time)

VorpalBlade:
It almost works:

`-Dwith-lockfile-path=/run/lock/netatalk` results in files like `/run/lock/netatalkatalkd` rather than `/run/lock/netatalk/atalkd`. Given that the doc is "Set path to the Netatalk daemon lockfile *directory*" (emphasis mine), I didn't expect that.

rdmark:
@VorpalBlade You just need to append one slash and it should work, i.e. `-Dwith-lockfile-path=/run/lock/netatalk/`

You can observe the mechanism in: https://github.com/Netatalk/netatalk/blob/048471636f083900aea0cfffc20a22fdadbc2f14/include/atalk/paths.h#L16

VorpalBlade:
> @VorpalBlade You just need to append one slash and it should work, i.e. `-Dwith-lockfile-path=/run/lock/netatalk/`
> 
> You can observe the mechanism in:
> 
> https://github.com/Netatalk/netatalk/blob/048471636f083900aea0cfffc20a22fdadbc2f14/include/atalk/paths.h#L16

Of course, but it is inconsistent with how other path variables (such as `pkgconfdir-path` and `pam-config-path`) are handled. Those *don't* seem to need a trailing slash.

I do sympathise though: doing this correctly (e.g. not just string concatenation) in C is a bit annoying.

rdmark:
Fair point. This is the only spot where path string concatenation is done directly in C, rather than through the build system, IINM. It's using this macro `ATALKPATHCAT` which concatenates exactly two strings, so in order to avoid broader refactoring, I pushed a commit now that simply moves the final slash over to the beginning of the file name string. Does this behave correctly for you now?

VorpalBlade:
This should work, I haven't had time to test it (and I will be out hiking today).

rdmark:
Enjoy your hike!

I'll go ahead and merge then, and we can roll another patch later if needed.

PR #1405, opened 2024-08-14
Add Back Legacy Options

### LINKED ISSUE #1332: Obsolete volume options (prodos, size limit, etc.) need to be added back in
Obsolete volume options (prodos, size limit, etc.) need to be added back in and reworked for the new INI configuration system. These options were removed in this commit https://github.com/Netatalk/netatalk/commit/863882cb7af7d0bd2651f957b155fd5aa9e2d9d6

Only some of these are actually important. The automatic CR/LF handling can be ignored.

_Originally posted by @NJRoadfan in https://github.com/Netatalk/netatalk/issues/220#issuecomment-2248848276_

NJRoadfan:
Adds back the limit volume size and ProDOS options from Netatalk 2.x. Update man page. Fixes #1332

### DIFF
--- a/doc/manpages/man5/afp.conf.5.xml
+++ b/doc/manpages/man5/afp.conf.5.xml
@@ -2198,6 +2198,17 @@
         </varlistentry>
 
         <varlistentry>
+          <term>legacy volume size = <replaceable>BOOLEAN</replaceable> (default:
+          <emphasis>no</emphasis>) <type>(V)</type></term>
+
+          <listitem>
+            <para>Limit disk size reporting to 2GB for legacy clients. This can
+            be used for older Macintoshes running System 7.1 or earlier and
+            using newer AppleShare clients.</para>
+          </listitem>
+        </varlistentry>
+
+        <varlistentry>
           <term>network ids = <replaceable>BOOLEAN</replaceable> (default:
           <emphasis>yes</emphasis>) <type>(V)</type></term>
 
@@ -2216,6 +2227,18 @@
             <para>A non-zero return code from preexec close the volume being
             immediately, preventing clients to mount/see the volume in
             question.</para>
+          </listitem>
+        </varlistentry>
+
+        <varlistentry>
+          <term>prodos = <replaceable>BOOLEAN</replaceable> (default:
+          <emphasis>no</emphasis>) <type>(V)</type></term>
+
+          <listitem>
+            <para>Enable ProDOS support. This option should only be enabled
+            for volumes you expect to netboot an Apple II from. In addition to
+            setting the boot flag on the volume, it restricts the volume free
+            space shown to 32MB.</para>
           </listitem>
         </varlistentry>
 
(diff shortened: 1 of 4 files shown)


NJRoadfan:
Note: I have tested this only with a AFP3.4 clients. I can see the AFPGetSrvrParms changing for the ProDOS option, but are unable to test the size limiting options as 3.x clients make an extended disk size call and always report the correct disk sizes. I need to get a TCP/IP AFP2.2 client working for that.

doc/manpages/man5/afp.conf.5.xml
```diff
@@ -2197,6 +2197,16 @@
           </listitem>
         </varlistentry>
 
+        <varlistentry>
+          <term>limit size = <replaceable>BOOLEAN</replaceable> (default:
+          <emphasis>no</emphasis>) <type>(V)</type></term>
+
+          <listitem>
+            <para>Limit disk size reporting to 2GB. This can be used for older
+            Macintoshes using newer AppleShare clients.</para>
```
rdmark, line 2206: What do you think about calling out this should not be used at the same time as `vol size limit`?

Users enabling both has come up a few times in the community.

Or perhaps worth considering: Can we merge the two into a single option?

rdmark:
> Note: I have tested this only with a AFP3.4 clients. I can see the AFPGetSrvrParms changing for the ProDOS option, but are unable to test the size limiting options as 3.x clients make an extended disk size call and always report the correct disk sizes. I need to get a TCP/IP AFP2.2 client working for that.

One tip: the SheepShaver emulator running OS9 will give you exactly AFP2.2 over TCP.

I can run some tests on my end in the evening today.

doc/manpages/man5/afp.conf.5.xml
NJRoadfan: They technically don't conflict if both are used, but they can't be merged. This setting only caps the legacy 32-bit VOLPBIT_BTOTAL and VOLPBIT_BFREE values to 2GB (older Finders only handle up to 2GB and overflows with anything larger). The `vol size limit` option only limits the AFP2.2+ only 64-bit VOLPBIT_BTOTAL and VOLPBIT_XBFREE values. Its a confusing naming issue as `vol size limit` was added much later in Netatalk's life.

I did some testing with AFPBridge on the IIgs, which runs AFP2.0 (and can fake AFP2.2 if needed) over TCP/IP, and these options seem to be working properly.

rdmark: We could take this opportunity to rename `limit size` to something that makes more sense in today's context. Say `legacy vol size` or similar. Thoughts?

NJRoadfan:
Changed the name to `legacy volume size` and revised the documentation to clarify what the option is used for. Its certainly an edge case as Netatalk already limits volumes to 2GB if the AFP client version is below 2.2. Unofficially one can run an AFP2.2 client on System 7.1, which may cause issues with Finder crashing since the free space and volume size values overflow above 2GB.

rdmark:
This does a much better job at describing the option to users! Thank you. I’ll go ahead and merge this now.

rdmark (approved):

PR #1406, opened 2024-08-15
meson: Allow ldconfig to run unprivileged during setup

rdmark:
Should fix the issue of ldconfig not being run as an install hook on Debian Bookworm.


rdmark:
@NJRoadfan Can you please test if this resolves your issue?

NJRoadfan:
Looks like its working here now, no errors on runtime. I `sudo` the install command FWIW.

NJRoadfan (approved):

Issue #1421, opened 2024-08-21
Unable to build with wolfssl 5.7.2


rdmark:
When building netatalk against wolfssl 5.7.2, compilation fails with the error in the title.

This is a spin-off of the upstream bug https://github.com/wolfSSL/wolfssl/issues/7827

The proposed fix is to include `wolfssl/wolfcrypt/settings.h` in each netatalk source file that uses wolfssl symbols.

knight-of-ni:
I've been placing most of my notes over on the wolfssl site. This is a wierd one that has me completely stumped at the moment. Despite including [settings.h](https://github.com/wolfSSL/wolfssl/blob/master/wolfssl/wolfcrypt/settings.h), the build still fails exactly the same as it did before. 

The build fails, stating the FIPS_VERSION3_GE macro is not defined, but it is defined right here: 
https://github.com/wolfSSL/wolfssl/blob/master/wolfssl/wolfcrypt/settings.h#L382

I'm hoping someone else will come along with the same problem and possibly uncover something I have not realized.

knight-of-ni:
ok, we had a eureka moment over at https://github.com/wolfSSL/wolfssl/issues/7827.
They identified that, despite meson detecting the external wolfssl libraries and headers, the compiler was finding the bundled settings.h first, which was based off wolfssl 5.7.0, rather than 5.7.2... hence the compiler complaining the macro in question was not defined (the macro in question is new for 5.7.2).

What this means for me, is that I need to delete the wolfssl bundled with netatalk during the rpm build process. I'm not sure if there is anything the Netatalk team should or can do here. 

After making that change, however, I ran into a new error:
```
FAILED: bin/afppasswd/afppasswd.p/afppasswd.c.o 
[...]

PR #1423, opened 2024-08-29
define DES_KEY_SZ if its not defined

knight-of-ni:
supplements #1421

### DIFF
--- a/bin/afppasswd/afppasswd.c
+++ b/bin/afppasswd/afppasswd.c
@@ -66,6 +66,10 @@
 #define HEXPASSWDLEN 16
 #define PASSWDLEN 8
 
+#ifndef DES_KEY_SZ
+#define DES_KEY_SZ (sizeof(DES_cblock))
+#endif
+
 static char buf[MAXPATHLEN + 1];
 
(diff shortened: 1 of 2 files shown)


rdmark (approved):
LGTM. Do we still need https://github.com/Netatalk/netatalk/pull/1422 ?

knight-of-ni:
> LGTM. Do we still need #1422 ?

Depends. I was not sure of your preferred workflow. I got the sense you might prefer merge #1422 into main and then cherry pick it into netatalk-3-2 branch?  If not, then I can turn the [equivalent commit](https://github.com/knight-of-ni/Netatalk/commit/847036e55604baff0bcb202ec5a6a5da99108cd7) I made in my own netatalk-3-2 branch into a PR, which you can merge.

rdmark:
Good question! My preferred workflow is to get PRs against stable branches, after which I can cherry-pick to `main` later. If a patch applies to both 3.x and 2.x, I like to see a PR for both. The main reason for doing it this direction, is because each merged PR gets picked up by GitHub and added to auto-generated changelog which is mighty handy for reducing the manual labor when rolling new releases.

Let me update the developer guidelines page with these details.

PR #1424, opened 2024-08-30
[3.2] include wolfssl/settings.h

knight-of-ni:
same pr as #1422, but targets branch-netatalk-3-2 instead of main.


rdmark:
FWIW https://github.com/Netatalk/netatalk/pull/1422 was actually for 3.2 ;)

rdmark (approved):

PR #1425, opened 2024-09-03
upgrade to wolfssl 2.7.2

knight-of-ni:
This upgrades the bundled wolfssl to version wolfssl 2.7.2 in branch-netatalk-3-2.

Since the main branch moved wolfssl to a meson subproject, the main branch would need a slightly different PR.

NOTE: I kept the same structure by copying the entire contents of the [wolfssl subfolder](https://github.com/wolfSSL/wolfssl/tree/v5.7.2-stable/wolfssl) from the upstream project. This doesn't sit well with me, because README, AUTHORS, and particularly LICENSING files are not part of this folder structure. It's my understanding these files should be included when distributing a project.


rdmark:
Done in https://github.com/Netatalk/netatalk/pull/1426

Issue #1442, opened 2024-09-07
Reintroduce remaining legacy options


rdmark:
This is to bring back any 2.x afpd options that are still relevant, e.g.

`auth print dir`, `ddp addr`

rdmark:
@NJRoadfan Created this for tracking. Please feel free to flesh out.

NJRoadfan:
Custom DDP addresses can now be configured again. Tested and working with behavior exactly like Netatalk 2.x. Another option I might add in the future is setting a custom AppleTalk zone to register the server in as opposed to the default zone. Netatalk 2.x did this by using a full AppleTalk name for the server name(ex: `MyServer@MyZone`) in `afpd.conf`. This can still be done, but it may override the value of `hostname` over on the IP side if I'm reading the code right.

As for `authprintdir`, I need to setup papd to properly support CAP style authentication. Not treating it as a priority since this feature is rarely used. The LaserWriter 8 driver has support for a newer authentication system that papd supports that should be working.

rdmark:
Agreed, supporting CAP style auth is very low priority. I would go so far as to say let's wait until we get feedback from a user who absolutely need it for their setup.

Anything else here that we should target for the 4.0 release?

rdmark:
Filed https://github.com/Netatalk/netatalk/issues/1462 for Zones support.

rdmark:
@NJRoadfan Do you think it'd be worth to reintroduce the `-proxy` option?

> -proxy
>
> Runs an AppleTalk proxy server for the specified AFP-over-TCP server. If the address and port aren't given, then the first IP address of the system and port 548 will be used. If you don't want the proxy server to act as a DDP server as well, set -uamlist "".

I have never used this and can't think of a usecase off the top of my head, but maybe you have some experience?

NJRoadfan:
I have never used it, but the proxy option appears to be used to setup a NBP registration for an AFP-over-TCP server on a different subnet. That is, it was strictly a service discovery proxy, not an actual network transport proxy. Since we no longer support multiple configurations, the feature can't be supported anymore.

rdmark:
By multiple configurations, do you mean multiple AFP-over-ASP sessions?

PR #1450, opened 2024-09-11
Add back "ddp address" option.

NJRoadfan:
Allow user to configure DDP address the server listens on by default if they have multiple interfaces running AppleTalk. Addresses part of #1442. Tested in multi-interface configuration and confirmed working.


rdmark (approved):
LGTM, cheers

PR #1465, opened 2024-09-12
uams: Fix DHCAST128 key alignment problem.

demonfoo:
Fixes: when Mb (the server's "public" key part in the Diffie-Hellman exchange) or K (the "private" key) are shorter than KEYSIZE, the UAM code does not verify that they are aligned correctly to the memory regions they're being placed in.

### DIFF
--- a/etc/uams/uams_dhx_passwd.c
+++ b/etc/uams/uams_dhx_passwd.c
@@ -10,6 +10,7 @@
 
 #include <arpa/inet.h>
 #include <pwd.h>
+#include <errno.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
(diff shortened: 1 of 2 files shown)


etc/uams/uams_dhx_passwd.c
```diff
@@ -274,8 +285,13 @@ static int passwd_logincont(void *obj, struct passwd **uam_pwd,
 
     /* check for session id */
     memcpy(&sessid, ibuf, sizeof(sessid));
-    if (sessid != dhxhash(obj))
+    if (sessid != dhxhash(obj)) {
+    /* Log Entry */
+           LOG(log_info, logtype_uams, "uams_dhx_passwd.c :passwd Session ID - DHXHash Mismatch -- %s",
```
rdmark, line 290: Isn’t “error” log level more appropriate for this message? I think users would want to have some context to why their auth is failing. Thoughts?

rdmark: Agreed that there should be parity across the two modules. On second thought, let’s keep it as is.

rdmark (approved):

PR #1477, opened 2024-09-14
wolfssl: Disable AES crypto to avoid symbol conflict with Nettle

rdmark:
We don't use AES crypto in the first place. It was just a matter of finding the right combination of configuration flags.


rdmark:
This interferes with the system wolfssl library on Debian/Ubuntu. More work is needed.

PR #1479, opened 2024-09-15
Fix crash in uam.c when connecting via ASP

NJRoadfan:
Attempts to connect with the "Cleartxt" UAM over an AppleTalk connection were crashing afpd.


rdmark (approved):

rdmark:
Good catch. I’ll merge in an hour.

PR #1488, opened 2024-09-18
Switch Randnum UAM to use libgcrypt

NJRoadfan:
Switches the RandNum UAM and afppasswd utility from using WolfSSL+Nettle to libgcrypt for DES encryption functions.

Adopted from : https://sourceforge.net/p/netatalk/feature-requests/60/

### DIFF
--- a/bin/afppasswd/afppasswd.c
+++ b/bin/afppasswd/afppasswd.c
@@ -35,18 +35,15 @@
 #include <sys/types.h>
 #include <unistd.h>
 
+#include <gcrypt.h>
+
+#ifndef DES_KEY_SZ
+#define DES_KEY_SZ 8
+#endif
+
 #ifdef USE_CRACKLIB
 #include <crack.h>
 #endif /* USE_CRACKLIB */
-
-#if defined(EMBEDDED_SSL) || defined(WOLFSSL_DHX)
-#if defined(WOLFSSL_DHX)
-#include <wolfssl/options.h>
-#include <wolfssl/wolfcrypt/settings.h>
-#endif
-#include <nettle/des.h>
-#include <wolfssl/openssl/des.h>
-#endif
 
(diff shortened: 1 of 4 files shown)


bin/afppasswd/afppasswd.c
```diff
@@ -35,19 +35,13 @@
 #include <sys/types.h>
 #include <unistd.h>
 
+#include <gcrypt.h>
+#define DES_KEY_SZ 8
```
rdmark, line 39: Do you want to put this in a `#ifndef` block just in case this gets defined in a header somewhere in the future?

I also tried to find out if libgcrypt had an equivalent to `DES_cblock` constant that we could use, but couldn't find traces of one. From my understanding DES block size is always 8 anyways so it's a minor concern.

NJRoadfan: Done.

rdmark: Cheers.

NJRoadfan:
Tested with AFP2.0 RandNum client (GS/OS) : PASS
Tested with AFP2.2 2-Way RandNum client (MacOS 8.1): PASS
Tested password generation with afppasswd: PASS

Note: Original code accidentally swapped the values of the received key and response key in the 2-way response. This has been fixed.

rdmark (approved):
So much nicer! Better 14 years late than never. :)

rdmark:
We also need to update the documentation: INSTALL.md and the Install chapter in the manual.

Optional whether to do it in this PR or not.

NJRoadfan:
Tomorrow's project, along with back porting to 3.x and 2.x.

rdmark:
Alright! Let me merge this one for starters.

What is your argument for backporting this to the stable branches? Changing the dependent library is arguably a major change.

PR #1489, opened 2024-09-18
docs: RandNum UAM now relies on Libgcrypt

rdmark:


rdmark:
@NJRoadfan FYI since I need to do the Japanese translation anyways, I'm doing this update here.

PR #1490, opened 2024-09-18
Migrate afpstats from dbus-glib to GDBus

rdmark:
Use GDBus (part of glib 2.0) as the interface with D-Bus, rather than the deprecated dbus-glib.


rdmark:
This is not working yet, not least because the `handle_method_call()` callback handler is a stub.

When attempting to use the `afpstats` script afpd throws this error:

```
GLib-GIO: Using cross-namespace EXTERNAL authentication (this will deadlock if server is GDBus < 2.73.3)
```

smcv:
> GLib-GIO: Using cross-namespace EXTERNAL authentication (this will deadlock if server is GDBus < 2.73.3)

This is not an error, this is a debugging message (`g_debug()`). If it is appearing by default, then your `GLogFunc` or `GLogWriterFunc` is more verbose than it needs to be.

etc/afpd/afpstats.c
```diff
@@ -22,31 +22,74 @@
...
+                     GError          **error,
+                     gpointer          user_data)
+{
+}
+
+static const GDBusInterfaceVTable interface_vtable =
+{
+  handle_method_call,
+  handle_get_property,
+  handle_set_property,
```
smcv, line 80: If your interface doesn't have any properties, you can (and should) leave the get-property and set-property callbacks set to `NULL`. You don't need to provide a stub implementation.

If you do provide a stub implementation, it needs to set the error and return `NULL` or `FALSE` as appropriate. A minimal implementation would be something like:

```
static GVariant handle_get_property (..., GError **error, gpointer user_data)
{
  g_set_error (error, G_DBUS_ERROR, G_DBUS_ERROR_NOT_SUPPORTED, "Not implemented");
  return NULL;
}

static gboolean handle_set_property (..., GError **error, gpointer user_data)
{
  g_set_error (error, G_DBUS_ERROR, G_DBUS_ERROR_NOT_SUPPORTED, "Not implemented");
  return FALSE;
}
```

But, again, setting both `get_property` and `set_property` to NULL is easier.

etc/afpd/afpstats.c
```diff
@@ -56,34 +99,75 @@ static gpointer afpstats_thread(gpointer _data _U_)
...
 
     AFPStatsObj *obj = g_object_new(AFPSTATS_TYPE_OBJECT, NULL);
-    dbus_g_connection_register_g_object(bus, "/org/netatalk/AFPStats", G_OBJECT(obj));
+
+    registration_id = g_dbus_connection_register_object(bus,
+                                      "/org/netatalk/AFPStats",
+                                      introspection_data->interfaces[0],
+                                      &interface_vtable,
+                                      G_OBJECT(obj),
+                                      NULL,
```
smcv, line 153: It's a good idea to explicitly keep the object "alive" for as long as it's registered:

```suggestion
                                      g_object_ref (obj),
                                      g_object_unref,
```

etc/afpd/afpstats.c
```diff
@@ -56,34 +99,75 @@ static gpointer afpstats_thread(gpointer _data _U_)
...
+                                         NULL,
+                                         NULL,
+                                         NULL,
+                                         NULL,
+                                         NULL);
 
     g_main_loop_run(thread_loop);
+
+    g_bus_unown_name(request_name_result);
+    g_dbus_node_info_unref(introspection_data);
```
smcv, line 169: Explicitly undo the registration when you no longer want that object to be available on the bus:

```suggestion
    g_dbus_connection_unregister_object (bus, registration_id);
    g_dbus_node_info_unref(introspection_data);
```

You'll probably also want to free the `AFPStatsObj` with `g_object_unref`?

etc/afpd/afpstats.c
```diff
@@ -22,31 +22,74 @@
...
+static void handle_method_call (GDBusConnection *connection,
+                    const gchar           *sender,
+                    const gchar           *object_path,
+                    const gchar           *interface_name,
+                    const gchar           *method_name,
+                    GVariant              *parameters,
+                    GDBusMethodInvocation *invocation,
+                    gpointer               user_data)
+{
+}
```
smcv, line 52: You need an implementation here, which should look a bit like the one in https://gitlab.gnome.org/GNOME/glib/-/blob/main/gio/tests/gdbus-example-server.c?ref_type=heads (but it can be a lot simpler, because it seems like your interface only has one method).

Probably something like this (untested!):

```suggestion
static void handle_method_call (GDBusConnection *connection,
                    const gchar           *sender,
                    const gchar           *object_path,
                    const gchar           *interface_name,
                    const gchar           *method_name,
                    GVariant              *parameters,
                    GDBusMethodInvocation *invocation,
                    gpointer               user_data)
{
  AFPStatsObj *obj = AFPSTATS_OBJECT (user_data);

  if (g_strcmp0 (method_name, "GetUsers") == 0)
    {
      GError *local_error = NULL;
      gchar **users = NULL;
      
      if (afpstats_obj_get_users (obj, &users, &local_error))
        g_dbus_method_invocation_return_value (invocation, g_variant_new ("(^as)", users)));
      else
        g_dbus_method_invocation_return_gerror (invocation, local_error);
        
      g_strfreev (users);
      g_clear_error (&error);
    }
}
```

smcv:
> > GLib-GIO: Using cross-namespace EXTERNAL authentication (this will deadlock if server is GDBus < 2.73.3)
> 
> This is not an error, this is a debugging message (`g_debug()`). If it is appearing by default, then your `GLogFunc` or `GLogWriterFunc` is more verbose than it needs to be.

Specifically, it looks as though `my_glib_log()` is ignoring the `log_level` and logging everything as if it was an error:

```
static void my_glib_log(const gchar *log_domain,
                        GLogLevelFlags log_level _U_,
                        const gchar *message,
                        gpointer user_data _U_)
{
    LOG(log_error, logtype_afpd, "%s: %s", log_domain, message);
}
```

If you don't override GLib's ordinary logging, it will log something reasonable on stderr (with debug messages suppressed by default, see [Running GLib Applications](https://docs.gtk.org/glib/running.html)).

Or, if you do want to override GLib's ordinary logging, you should look at the `GLogLevelFlags` to find out how severe the message is, and convert it to your own logging levels (or perhaps just silence it instead of logging it at all, in the case of debug messages).

You might find this conversion from GLogLevelFlags to `syslog()` levels useful as a starting point to adapt to your own concept of log levels:

```
static int
get_level_priority (GLogLevelFlags log_level)
{
  if (log_level & (G_LOG_FLAG_RECURSION
                   | G_LOG_FLAG_FATAL
                   | G_LOG_LEVEL_ERROR
                   | G_LOG_LEVEL_CRITICAL))
    return LOG_ERR;

  if (log_level & G_LOG_LEVEL_WARNING)
    return LOG_WARNING;

  if (log_level & G_LOG_LEVEL_MESSAGE)
    return LOG_NOTICE;

  if (log_level & G_LOG_LEVEL_INFO)
    return LOG_INFO;

  if (log_level & G_LOG_LEVEL_DEBUG)
    return LOG_DEBUG;

  return LOG_NOTICE;
}
```

(copyright 2022-2024 Collabora Ltd., MIT license, originally from [here](https://gitlab.steamos.cloud/steamrt/steam-runtime-tools/-/blob/main/steam-runtime-tools/log.c?ref_type=heads))

etc/afpd/afpstats.c
rdmark: Good call. Pushing a commit that frees `obj`.

etc/afpd/afpstats.c
rdmark: The only use for this interface right now is a one-way blanket request for server stats. I went ahead and NULLed out the callbacks.

etc/afpd/afpstats.c
```diff
@@ -111,7 +193,9 @@ int afpstats_init(server_child_t *childs_in)
     GThread *thread;
 
     childs = childs_in;
+#if 0
     dbus_g_thread_init();
```
rdmark, line 197: @smcv Is there a GDBus equivalent to `dbus_g_thread_init()` or is it not needed anymore?

This is the one authoritative source I could find but it's very brief. https://dbus.freedesktop.org/doc/dbus-glib/dbus-glib-DBusGConnection.html#dbus-g-thread-init

smcv: There is no equivalent in GDBus, just delete it. GDBus is always multi-threaded and thread-safe.

With modern versions of libdbus, `dbus_g_thread_init()` doesn't do anything anyway.

etc/afpd/afpstats.c
```diff
@@ -56,34 +94,78 @@ static gpointer afpstats_thread(gpointer _data _U_)
...
+                                      "/org/netatalk/AFPStats",
+                                      introspection_data->interfaces[0],
+                                      &interface_vtable,
+                                      g_object_ref (obj),
+                                      g_object_unref,
+                                      &error);
+    g_assert (registration_id > 0);
+
+    request_name_result = g_bus_own_name(G_BUS_TYPE_SYSTEM,
+                                         "org.freedesktop.DBus",
```
smcv, line 153: Er, this is saying "please give me the name `org.freedesktop.DBus`", which is certainly never going to work.

I think you mean:

```suggestion
                                         "org.netatalk.AFPStats",
```

etc/afpd/afpstats.c
```diff
@@ -56,34 +94,78 @@ static gpointer afpstats_thread(gpointer _data _U_)
...
+                                         G_DBUS_PROXY_FLAGS_NONE,
```
smcv, line 154: `g_bus_own_name()` doesn't take GDBusProxyFlags, it takes GBusNameOwnerFlags. I think you mean

```suggestion
                                         G_DBUS_NAME_OWNER_FLAGS_NONE,
```

(No practical change, both are numerically zero.)

etc/afpd/afpstats.c
```diff
@@ -56,34 +94,78 @@ static gpointer afpstats_thread(gpointer _data _U_)
...
+                                         NULL,
+                                         NULL,
+                                         NULL,
```
smcv, line 157: This is going to be a lot easier to debug if you set non-null handlers for all three of these events. If you don't need to run any "business logic" when these things happen, they can just do a `g_debug()` or your local equivalent (I think that's `LOG(log_debug, ...)` or something?) so that you have a chance to see what's going on.

etc/afpd/afpstats.c
```diff
@@ -56,34 +94,78 @@ static gpointer afpstats_thread(gpointer _data _U_)
     ctxt = g_main_context_new();
```
smcv, line 94: Is this running in a non-main thread? If yes, you'll probably need to set the main-context to be the thread-default main context:

```
    ctxt = g_main_context_new();
    g_main_context_push_thread_default(ctxt);

    ... do all the other stuff that happens in this thread ...

    g_main_context_pop_thread_default(ctxt);
    g_main_context_unref(ctxt);
}
```

Otherwise, `handle_method()` will get called from a different thread that is not this one (most likely your main thread). Or is that what you were assuming? I don't know what your multi-threading design in this project looks like.

Reference: https://developer.gnome.org/documentation/tutorials/main-contexts.html

With dbus-glib, this might have worked accidentally because dbus-glib was not thread-safe, so it would have called methods/callbacks on some arbitrary thread that is not the one you expected...

etc/afpd/afpstats.c
rdmark: Cheers, then I understand what the docs was trying to say.

etc/afpd/afpstats.c
rdmark: That's a great idea. Let me tinker with the LOG macro. I definitely need some "insights" to get this over the finish line. :)

etc/afpd/afpstats.c
smcv: It would probably be useful to add some

```
g_debug ("called %s in thread %p", G_STRFUNC, g_thread_self ());
```

(or the equivalent with your `LOG` macro) in strategic places, to check that everything is being called in the thread where you expected it to be called.

rdmark: This is definitely running in a non-main thread!

I'm pleased to report that this was the missing piece that made it work.

```
$ afpstats 
Connected user   PID      Login time        State          Mounted volumes
dmark            573288   Sep 20 10:29:45   active         My Share
```

rdmark:
Remaining tasks:

- Try to get this working on Solaris. Oracle docs covers gdbus but need to investigate if their implementation is compatible.
- Refactor the build system. We no longer need the dbus-glib checks. Question is if glib 2.0 is sufficient, or if we need to check for a particular symbol.

smcv:
> Try to get this working on Solaris. Oracle docs covers gdbus but need to investigate if their implementation is compatible.

GDBus is part of GLib, https://gitlab.gnome.org/GNOME/glib, and to the best of my knowledge there is no other implementation under a similar name. Solaris might have an older version of GLib than typical Linux distributions, but basic GDBus functionality is pretty old (most of it was already available 10 years ago).

> Refactor the build system. We no longer need the dbus-glib checks. Question is if glib 2.0 is sufficient, or if we need to check for a particular capability.

I see you use Meson. `dependency('gio-2.0')` includes GDBus.

Any version from the last 10 years is probably sufficient (I've successfully used GDBus in GLib 2.32.x, from 2012). If in doubt, install a VM or a real machine with the oldest OS you care about supporting (probably an "enterprise" distribution like Debian (E)LTS, Ubuntu oldoldLTS, something from the RHEL/CentOS family, or Solaris), and build and test there.

rdmark:
For the record: Arch Linux and Solaris are being obnoxious and seemingly ship glib-2.0 that lack the GDBus functions. I'm trying to figure out a sophisticated way to detect this, but am running into challenges getting Meson to link with glib in test programs...

rdmark:
Regarding Solaris, this issue is isolated to the GitHub runner environment. In my local Solaris 11 VM the compilation succeeds. Disabling afpstats for the Solaris job for now. Also disabling for Arch, but I will escalate this to the Arch team for assistance.

rdmark:
This has been tested working on:

- Fedora
- Debian
- Ubuntu
- FreeBSD

On Alpine Linux, it can be built but the afpstats clients don't work (different issue: no Perl Net::DBus module)

On NetBSD, the compilation succeeds, but there's some kind of error when launching:

```
Sep 21 22:29:52.324723 afpd[8011] {afpstats.c:80} (debug:AFPDaemon): [afpstats] on_bus_acquired(): org.netatalk.AFPStats
Sep 21 22:29:52.324989 afpd[8011] {afpstats.c:94} (debug:AFPDaemon): [afpstats] on_name_lost(): org.netatalk.AFPStats
```

More logging will be needed to further troubleshoot.

On OpenSUSE Leap, there's a configuration issue that is probably easy to fix.

```
Sep 21 22:43:23 localhost afpd[11328]: Failed to acquire DBUS name: Connection ":1.34" is not allowed to own the service "org.netatalk.AFPStats" due to security policies in the configuration file
```

rdmark:
Apart from the Arch issue, I think this is good enough to ship now. It works as well or better than before this change.

rdmark:
On the SonarCloud bug issues: All 9 have to do with how the GDBus API is designed, and can't be addressed. For instance, if I change the indicated parameters to pointer-to-const, compilation fails with strict compilers because it doesn't match the non-const counterpart in GLib.

rdmark:
@smcv Huge thanks for all your assistance! Your advice and code snippets were invaluable to get this across the finishing line.

smcv:
> Arch Linux and Solaris are being obnoxious and seemingly ship glib-2.0 shared libraries that lack the GDBus symbols

Solaris, maybe. Arch, I would be very surprised - in my experience Arch's GLib libraries are "normal".

The various GDBus symbols are expected to be found in `libgio-2.0.so.0`, not in `libglib-2.0.so.0` - but you must be getting that right already, or it wouldn't have worked on Debian and Fedora.

> `Failed to acquire DBUS name: Connection ":1.34" is not allowed to own the service "org.netatalk.AFPStats" due to security policies in the configuration file`

This is the same for GDBus as for any other D-Bus client library: it's about configuring the instance of dbus-daemon or dbus-broker that is used to implement the system bus, rather than about configuring your client library. For a user (even root!) to be allowed to own a name on the system bus, you need a configuration snippet (in your case this seems to be https://github.com/Netatalk/netatalk/blob/main/config/netatalk-dbus.conf) installed in a directory that the system bus will read.

rdmark:
@smcv Yes the Arch situation is puzzling. I'm working off of the latest archlinux:latest Docker image, and what I observe is that libgio.so* lives next to libglib.so* and have the right symbols that afpstats.c needs, e.g.

```
# ls -al /usr/lib/libgio*
-rw-r--r-- 1 root root 4919718 Aug 26 10:39 /usr/lib/libgio-2.0.a
lrwxrwxrwx 1 root root      15 Aug 26 10:39 /usr/lib/libgio-2.0.so -> libgio-2.0.so.0
lrwxrwxrwx 1 root root      22 Aug 26 10:39 /usr/lib/libgio-2.0.so.0 -> libgio-2.0.so.0.8000.5
-rwxr-xr-x 1 root root 1879352 Aug 26 10:39 /usr/lib/libgio-2.0.so.0.8000.5
# ls -al /usr/lib/libglib*
-rw-r--r-- 1 root root 2412806 Aug 26 10:39 /usr/lib/libglib-2.0.a
lrwxrwxrwx 1 root root      16 Aug 26 10:39 /usr/lib/libglib-2.0.so -> libglib-2.0.so.0
lrwxrwxrwx 1 root root      23 Aug 26 10:39 /usr/lib/libglib-2.0.so.0 -> libglib-2.0.so.0.8000.5
-rwxr-xr-x 1 root root 1363288 Aug 26 10:39 /usr/lib/libglib-2.0.so.0.8000.5
```

```
# readelf -Ws --dyn-syms /usr/lib/libgio-2.0.so.0.8000.5 |grep g_object_new
   314: 0000000000000000     0 FUNC    GLOBAL DEFAULT  UND g_object_new_valist
   353: 0000000000000000     0 FUNC    GLOBAL DEFAULT  UND g_object_new
   504: 0000000000000000     0 FUNC    GLOBAL DEFAULT  UND g_object_newv
```

But, somehow the linker cannot find the libgio symbols, but *can* find the libglib symbols. I'm a bit at a loss right now. Let me ask and see if the Arch community has an idea.

smcv:
> `   353: 0000000000000000     0 FUNC    GLOBAL DEFAULT  UND g_object_new`

This is an undefined symbol (you can tell by the `UND`), which is required but not provided by `libgio-2.0.so.0`.

In fact I know from out-of-band knowledge that it's `libgobject-2.0.so.0` that is meant to provide this particular symbol.

rdmark:
@smcv Thanks for the hints. This did the trick: https://github.com/Netatalk/netatalk/pull/1542

PR #1506, opened 2024-09-21
Switch DHCAST128 UAM to use libgcrypt

NJRoadfan:
Switches the DHCAST128 UAM from WolfSSL+Nettle to libgcrypt for CAST encryption functions.

### DIFF
--- a/etc/uams/meson.build
+++ b/etc/uams/meson.build
@@ -29,6 +29,7 @@ library(
     'uams_passwd',
     uams_passwd_sources,
     include_directories: root_includes,
+    dependencies: crypt,
     name_prefix: '',
     name_suffix: lib_suffix,
     override_options: 'b_lundef=false',
@@ -42,15 +43,16 @@ if have_embedded_ssl
     ssl_links += libatalk
 endif
 
-if have_ssl
+if have_libgcrypt
+    uams_randnum_sources = ['uams_randnum.c']
+    uams_dhx2_passwd_sources = ['uams_dhx2_passwd.c']
     uams_dhx_passwd_sources = ['uams_dhx_passwd.c']
 
     library(
-        'uams_dhx_passwd',
-        uams_dhx_passwd_sources,
+        'uams_randnum',
+        uams_randnum_sources,
         include_directories: root_includes,
-        dependencies: ssl_deps,
-        link_with: ssl_links,
+        dependencies: [crack, libgcrypt],
         name_prefix: '',
         name_suffix: lib_suffix,
         override_options: 'b_lundef=false',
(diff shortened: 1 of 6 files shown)


etc/uams/meson.build
```diff
@@ -42,15 +42,16 @@ if have_embedded_ssl
     ssl_links += libatalk
 endif
 
-if have_ssl
+if have_libgcrypt
```
rdmark, line 45: I think we can remove the `have_libgcrypt` conditional and assume that this library is always available. Which we can safely do, since setup will fail early in the Meson script if libgcrypt cannot be detected.

rdmark:
@demonfoo Heads-up that you may want to test compatibility with your AFP client with this change.

demonfoo:
@rdmark I'd been considering reimplementing that based on `libgcrypt`, actually. I'll build it In a VM and try it out.

demonfoo:
@rdmark I was able (with a bit of messing about) to get it building undef my FreeBSD 13.4 VM. Forcing the `DHCAST128` UAM succeeds (I've confirmed the DSO it installed is the new gcrypt-using one). The code appears to do the right thing in all the relevant places.

NJRoadfan:
Found some nagging meson issues when not using PAM since `crypt.h` wasn't being linked. Tested DHCAST128 with and without PAM and its working.

NJRoadfan:
Fixed the change password function. Now a user can change their password.

demonfoo:
I don't recall Apple's server implementations returning `kFPAccessDenied` when the password change was complete, but _does_ change the password.

rdmark:
@demonfoo Thanks for testing!

NJRoadfan:
@demonfoo I'm not seeing any PAM errors here.

demonfoo:
> @demonfoo I'm not seeing any PAM errors here.

It looks like this may be some kind of bug in FreeBSD's `libutil`. Running `truss` on the process shows its invocation of `pwd_mkdb` does work, but it looks like it's not waiting for the process correctly, and thus `pw_mkdb()` is returning -1 even though it shouldn't.

rdmark (approved):

coffnix:
working like a charm on my macOS 9.2.2. Thanks!

new spec to create netatalk-4.4.1 RPM on openSUSE 15.6: https://area31.net.br/downloads/netatalk-libgcrypt.spec



Picture 9

rdmark:
@coffnix very cool, the spec looks good

do you plan to contribute the spec to the openSUSE project so that it can get distributed in their repos?

coffnix:
@rdmark I will send this spec to a friend who maintains a SUSE repository. Thanks for the feedback!

Issue #1507, opened 2024-09-21
Proposal: Remove the PGP UAM.


NJRoadfan:
This UAM relies on a 3rd party plug-in and is limited to classic MacOS 8.5+. It appears that it was a prototype version of what became the DHCAST128 UAM. The link to download the plug-in is long since dead and the DHCAST128 UAM superseded its functionality. This code also has not been maintained in a very long time and might not even work! Very few if anyone is actually using this today and removing it will remove another piece of code that relies on OpenSSL/WolfSSL/Nettle.

rdmark:
In theory I like the idea of the PGP UAM, as an arguably more secure way to authenticate with OS8/9 compared to DHCAST128, for folks who want to harden their Netatalk setup. Last year I fixed some minor bugs and got the UAM nominally working, but in the end failed to actually get the Mac OS client to recognize Netatalk as a PGP capable AppleShare server. See my notes in https://github.com/Netatalk/netatalk/issues/548

The present situation is that the Mac OS UAM is a prototype that's only known to work with a very narrow combination of Mac OS and PGP versions. The author lists several known bugs and seem to admit to hackish use of Mac OS APIs. The I scoured the mailing list archives but could not find direct evidence that the Netatalk PGP UAM was ever confirmed working in the field.

On the positive side, the full C codebase and MetroWerks Codewarrior project are available (albeit not under a permissive license.) So in theory, a better Mac OS client could be written.

Ultimately, I approve of removing the Netatalk PGP UAM at this point. If someone wants to attempt to revive it in the future, the historical code is available!

NJRoadfan:
It can be ported to libgcrypt, but I'm not going to bother if it isn't being used or its broken (it looks like its incomplete). If one is going to write a client for classic MacOS, it should be for DHX2 at this point.

rdmark:
Definitely not worth the effort to adapt it for libgcrypt.

PR #1522, opened 2024-09-22
Remove PGP UAM

NJRoadfan:
Addresses #1507


rdmark:
@NJRoadfan Please resolve the merge conflict!

Also, it should be safe to remove the entire bundled wolfssl library now, right?

rdmark:
Also note for self: The Dockerfile needs to be updated to remove the PGP option (and the nettle dependencies.)

NJRoadfan:
Yes, it should be safe to remove. Also, there may be artifacts left in the `meson.build` file in the uams directory.

rdmark:
Pro tip: Rather than a branch merge, if you do something like `git rebase main` you get a branch with clean commit history that you can force-push as is.

rdmark (changes requested):
Looks good, but please squash the two commits into one by any means.

PR #1525, opened 2024-09-22
Remove PGP UAM

NJRoadfan:
Addresses #1507


rdmark:
Note to self: Japanese manual needs to be updated too.

rdmark (approved):

PR #1527, opened 2024-09-23
Remove the bundled wolfSSL library

rdmark:
DHCAST128 and RandNum have been migrated to Libgcrypt. PGP has been removed. Hence, we no longer need the bundled wolfSSL library.


rdmark:
Tested that DHX, RandNum and afppasswd are still working.

NJRoadfan:
Are these rpath statements still needed for the Randnum and DHX UAM sources in meson.build?

https://github.com/Netatalk/netatalk/blob/main/etc/uams/meson.build#L61-L62

rdmark:
They were probably never needed, given that the UAMs aren't normal shared libraries. Let's test.

PR #1532, opened 2024-09-23
Bundle, configure and install the Webmin module

rdmark:
Bundles the netatalk-webmin source code previously managed at https://github.com/Netatalk/netatalk-webmin

The Webmin module code is distributed under the same GPLv2 license as Netatalk.

Integration with the Netatalk Meson build system allows you to configure, package, and install the module on the fly.

Meson is used in a somewhat hackish way here. One known bug is that `meson dist` doesn't work when building the webmin module. There may be a cleaner way to achieve the module tarball creation without relying on specific build dir locations..

The workflow is:

- Setup the build with `meson setup build -Dwith-webmin` (disabled by default)
- Compile as usual
- Install as usual. At this step, the module tarball will first be created via an install script in `build/contrib/webmin_module/netatalk_webmin-4.0.0.wbm.gz`
- If Webmin itself was detected during setup, yet another install script will attempt to install the module

When preparing Netatalk releases, the release manager would grab the tarball under `build/contrib/webmin_module/` and upload it with the other Netatalk tarballs.

~~TODO: The build system configures binary and config file paths (see the `config.in` file). It does not configure the init script commands yet, which are currently hard coded to systemd commands.~~


rdmark:
@NJRoadfan Doing it this way will have a few upsides:

- Easier to keep the Webmin module in sync with Netatalk's feature set (added/removed apf.conf options etc.)
- Streamline release management: netatalk and webmin module tarballs are built through the same build system
- Users who build netatalk from source have a handy option for configuring and installing the module

I can't think of any downsides, apart from making this repo slightly more monolithic.

rdmark:
Automatic configuration of init commands per OS has been achieved now. I think this is good to ship.

PR #1539, opened 2024-09-23
Update systemd.netatalk.service.in to support atalkd

NJRoadfan:
Allow for loading atalkd before starting the AFP server, like 2.x did.


NJRoadfan:
netatalk.service starts up without an issue when atalkd isn't running. I just tested a 4.0.0 build and clean install without AppleTalk and atalkd installed and it worked.

rdmark (approved):

PR #1540, opened 2024-09-23
Cleanup libgcrypt initialization

NJRoadfan:
Fix for libgcrypt libraries that weren't consistently being initialized depending on how a user logged in.


rdmark (approved):
This reminds me of a very old but still relevant Debian bug. :)

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=568601

PR #1543, opened 2024-09-24
Integrate macipgw into the project

rdmark:
macipgw originally by Stefan Bethke with modifications by Jason King. The netatalk integration was cherry-picked from the now-defunct netatalk-classic fork, and converted to Meson.

The macipgw module is built with the `-Dwith-appletalk=true` build system flag.

Since the tunnel implementation is hard coded to either Linux or *BSD, we will only build the module on those systems.


rdmark:
The security hot spots called out by Sonarcloud are all of the strcpy/strlen potential buffer overflow kind. The netatalk codebase are crawling with these anyways given the age of the code. I will leave them as-is for now, and address them in a future push for hardening these across netatalk.

PR #1563, opened 2024-09-29
errno.h should always be available

rdmark:
<sys/errno.h> is an alias for <errno.h> on all supported OSes

Also, removes obsoleted capability flags in config.h


rdmark:
@NJRoadfan FYI, last minute change. The `HAVE_SYS_ERRNO_H` flag wasn't even defined so I think this should be safe.

PR #1570, opened 2024-09-30
SQLite CNID backend

rdmark:
Based on the MySQL CNID backend by Ralph Boehme, and proof-of-concept code by Christopher Kobayashi originally from the now-defunct netatalk-classic project

The meson build system will build the sqlite backend when the sqlite3 dependency is found, but this can also be controlled with the -Dwith-cnid-backends option

Put 'cnid scheme = sqlite' in the afp.conf volume section where you want to use the new backend

Similarly to dbd (BerkeleyDB) this backend is zero-configuration that stores its database in a file under /var/netatalk/CNID/[volumedir]

Please note that this is considered EXPERIMENTAL code at this point, so please proceed with caution and keep backups of any sensitive data on your volumes

Worth noting is a fundamental limitation of SQLite is that is supports a single database write at a time, so we rely entirely on the library to [serialize simultaneous writes](https://www.sqlite.org/lang_transaction.html), which may behave slightly different depending on the library version used

I have tested it with two simultaneous users, which worked flawlessly, but very heavy access by dozens or hundreds of users has not been tested

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -73,6 +73,7 @@ jobs:
             perl \
             pkgconfig \
             rpcsvc-proto-dev \
+            sqlite-dev \
             talloc-dev \
             tinysparql-dev
       - name: Configure
@@ -130,6 +131,7 @@ jobs:
             perl \
             pkgconfig \
             rpcsvc-proto \
+            sqlite \
             talloc \
             tinysparql
       - name: Configure
@@ -191,6 +193,7 @@ jobs:
             libldap2-dev \
             libmariadb-dev \
             libpam0g-dev \
+            libsqlite3-dev \
             libtalloc-dev \
             libtirpc-dev \
             libtracker-sparql-3.0-dev \
@@ -266,6 +269,7 @@ jobs:
             perl \
             perl-Net-DBus \
             quota-devel \
+            sqlite-devel \
             systemd \
             systemtap-sdt-devel \
             tracker \
@@ -391,7 +395,8 @@ jobs:
             iniparser \
             mariadb \
             meson \
-            openldap
+            openldap \
+            sqlite
       - name: Configure
         run: |
           meson setup build \
@@ -448,6 +453,7 @@ jobs:
               py39-gdbm \
               py39-sqlite3 \
               py39-tkinter \
+              sqlite \
               talloc \
               tracker3
           run: |
@@ -491,6 +497,7 @@ jobs:
               p5-Net-DBus \
               perl5 \
               pkgconf \
+              sqlite3 \
               talloc
           run: |
             set -e
@@ -541,6 +548,7 @@ jobs:
               p5-Net-DBus \
               perl \
               pkg-config \
+              sqlite3 \
               talloc
           run: |
             set -e
@@ -597,6 +605,7 @@ jobs:
               openldap-client-2.6.9p0v0 \
               p5-Net-DBus \
               pkgconf \
+              sqlite \
               tinysparql-3.8.2
           run: |
             set -e
@@ -646,6 +655,7 @@ jobs:
               libgcrypt \
               meson \
               mysql-client \
+              sqlite3 \
               talloc
           run: |
             set -e
@@ -692,7 +702,8 @@ jobs:
               libgcrypt \
               meson \
               ninja \
-              pkg-config
+              pkg-config \
+              sqlite-3
             curl --location -o cmark.tar.gz \
               https://github.com/commonmark/cmark/archive/refs/tags/0.31.1.tar.gz
             curl --location -o iniparser.tar.gz \
--- a/.github/workflows/containers.yml
+++ b/.github/workflows/containers.yml
@@ -230,7 +230,6 @@ jobs:
                     -e AFP_HOST=afp_server \
                     ${{ env.REGISTRY }}/netatalk/netatalk-testsuite:${{ github.sha }}
 
-
     afp-spectest-mysql-afp34-prod:
         name: AFP spec test (cnid:mysql) AFP 3.4 - Alpine (Prod)
         needs:
@@ -290,7 +289,6 @@ jobs:
                     -e AFP_CNID_SQL_PASS=${{ secrets.MARIADB_ROOT_PASSWORD }} \
                     ${{ env.REGISTRY }}/netatalk/netatalk-testsuite:${{ github.sha }}
 
-
     afp-spectest-mysql-afp34-debian:
         name: AFP spec test (cnid:mysql) AFP 3.4 - Debian
         needs: build-container-testsuite-debian
@@ -331,6 +329,52 @@ jobs:
                     -e AFP_CNID_SQL_HOST=mariadb \
                     -e AFP_CNID_SQL_PASS=${{ secrets.MARIADB_ROOT_PASSWORD }} \
                     ${{ env.REGISTRY }}/netatalk/netatalk-testsuite-debian:${{ github.sha }}
+
+    afp-spectest-sqlite-afp34:
+        name: AFP spec test (cnid:sqlite) AFP 3.4 - Alpine
+        needs: build-container-testsuite
+        runs-on: ubuntu-latest
+        timeout-minutes: 5
+        env:
+            AFP_USER: atalk1
+            AFP_USER2: atalk2
+            AFP_PASS: ${{ secrets.AFP_PASSWD }}
+            AFP_PASS2: ${{ secrets.AFP_PASSWD }}
+            AFP_GROUP: afpusers
+            AFP_CNID_BACKEND: sqlite
+            SHARE_NAME: test1
+            SHARE_NAME2: test2
+            INSECURE_AUTH: 1
+            DISABLE_TIMEMACHINE: 1
+            VERBOSE: 1
+            TESTSUITE: spectest
+            AFP_VERSION: 7
+            AFP_EXTMAP: 1
+        steps:
+            - uses: docker://ghcr.io/netatalk/netatalk-testsuite:latest
+
+    afp-spectest-sqlite-afp34-debian:
+        name: AFP spec test (cnid:sqlite) AFP 3.4 - Debian
+        needs: build-container-testsuite-debian
+        runs-on: ubuntu-latest
+        timeout-minutes: 5
+        env:
+            AFP_USER: atalk1
+            AFP_USER2: atalk2
+            AFP_PASS: ${{ secrets.AFP_PASSWD }}
+            AFP_PASS2: ${{ secrets.AFP_PASSWD }}
+            AFP_GROUP: afpusers
+            AFP_CNID_BACKEND: sqlite
+            SHARE_NAME: test1
+            SHARE_NAME2: test2
+            INSECURE_AUTH: 1
+            DISABLE_TIMEMACHINE: 1
+            VERBOSE: 1
+            TESTSUITE: spectest
+            AFP_VERSION: 7
+            AFP_EXTMAP: 1
+        steps:
+            - uses: docker://ghcr.io/netatalk/netatalk-testsuite-debian:latest
 
 
     afp-spectest-afp34:
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,6 +13,7 @@ ARG RUN_DEPS="\
     mariadb-client \
     mariadb-connector-c \
     openldap \
+    sqlite \
     talloc \
     tinysparql \
     tzdata"
@@ -36,6 +37,7 @@ ARG BUILD_DEPS="\
     ninja \
     openldap-dev \
     pkgconfig \
+    sqlite-dev \
     talloc-dev \
     tinysparql-dev"
 
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -77,8 +77,10 @@ documentation for more details.
 | libldap                    | For LDAP support |
 | libpam                     | For PAM support |
 | libtirpc **OR** libquota   | For Quota support |
+| mysql-client **OR** mariadb-client | For *mysql* CNID backend support |
 | Perl                       | For admin scripts |
 | po4a                       | For localization of documentation |
+| sqlite3                    | for *sqlite* CNID backend support |
 | tcpwrap                    | For TCP wrapper support |
 | [UnicodeData.txt](https://www.unicode.org/Public/UNIDATA/UnicodeData.txt) | For regenerating Unicode lookup tables |
 
--- a/contrib/webmin_module/edit_vol_section.cgi
+++ b/contrib/webmin_module/edit_vol_section.cgi
@@ -245,6 +245,20 @@ if ($subject ne 'homes') {
 }
 
 print &ui_table_row(
+                    $text{'edit_vol_section_cnid_scheme'},
+                    &build_select(
+                                  $afpconfRef, $sectionRef, \%in, 'cnid scheme', $text{'edit_undefined'},
+                                  'dbd',
+                                  'dbd',
+                                  'last',
+                                  'last',
+                                  'mysql',
+                                  'mysql',
+                                  'sqlite',
+                                  'sqlite'
+                    )
+);
+print &ui_table_row(
                     $text{'edit_vol_section_ea'},
                     &build_select(
                                   $afpconfRef, $sectionRef, \%in, 'ea', $text{'edit_undefined'}, 'sys',
@@ -457,15 +471,6 @@ if ($subject ne 'homes') {
                         &build_select(
                                       $afpconfRef, $sectionRef, \%in, 'convert appledouble', $text{'edit_undefined'},
                                       'yes',       'yes',       'no', 'no'
-                        )
-    );
-
-    @values = get_parameter_of_section($afpconfRef, $sectionRef, 'cnid scheme', \%in);
-    print &ui_table_row(
-                        $text{'edit_vol_section_cnid_scheme'},
-                        &build_select(
-                                      $afpconfRef, $sectionRef, \%in,   'cnid scheme', $text{'edit_undefined'}, 'dbd',
-                                      'dbd',       'last',      'last', 'mysql',       'mysql'
                         )
     );
 
--- a/doc/manpages/man5/afp.conf.5.md
+++ b/doc/manpages/man5/afp.conf.5.md
@@ -1077,14 +1077,26 @@ hosts deny = *IP host address/IP netmask bits* [ ... ] **(V)**
 cnid scheme = *backend* **(V)**
 
 > set the CNID backend to be used for the volume.
-The "dbd" and "last" backends are always available. Default is **dbd**.
-Run *afpd -v* to see a list of available backends.
+The *dbd* and *last* backends are always available. Default is *dbd*.
+Run **afpd -v** to see a list of available backends.
 >
-> The "last" backend is read only, used to mount CD-ROMs and similar media.
+> *dbd* is a zero-configuration, full-featured, and reliable backend
+using Berkeley DB, where database access is managed through
+the **cnid_dbd** daemon.
 >
-> The optional "mysql" backend requires the system administrator to provision
-a MySQL database instance for use with Netatalk, as well as setting the
+> The *last* backend uses a read only Trivial Database,
+commonly used to mount CD-ROMs and similar media.
+>
+> The optional *mysql* backend requires the system administrator to provision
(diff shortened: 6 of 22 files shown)


rdmark:
@NJRoadfan This is close to working now, FYI.

Two outstanding issues:

The directory that houses the sqlite cnid databases (e.g. `/var/lib/netatalk/CNID` needs to have 777 permissions since the database client is embedded in `afpd` which runs as a normal user (obviously). We may need a helper daemon like `cnid_dbd` to run the database client as root.

But more importantly, there's a crash in `dircache_add()` when it tries to enumerate the volume, so something isn't lining up quite yet.

Nonetheless, the backend builds and starts up now, and can generate a little sqlite database file that you can query, which is neat.

NJRoadfan:
Running `dbd -f` to force rebuild an existing volume spits out the following error:
```
Nov 14 18:35:29.225001 dbd[2978] {cnid_sqlite.c:100} (error:Default): sqlite3_bind_int64(db->cnid_add_stmt, 1, stmt_param_id) failed: No such file or directory
Segmentation fault
```

Here is a dump of the database at that point, its not getting very far:
```
# sqlite3 /var/local/netatalk/CNID/A2FILES/cnid.sqlite .dump
PRAGMA foreign_keys=OFF;
BEGIN TRANSACTION;
CREATE TABLE volumes ( VolUUID CHAR(32) PRIMARY KEY,VolPath TEXT(4096),Stamp BINARY(8),Depleted INT);

INSERT INTO volumes VALUES('830A28BBB084683384019DD280C51FA8','/srv/A2SERVER/A2FILES','A�6g',0);

CREATE TABLE '830A28BBB084683384019DD280C51FA8'(Id INTEGER PRIMARY KEY AUTOINCREMENT,Name VARCHAR(255) NOT NULL,Did INTEGER NOT NULL,DevNo INTEGER NOT NULL,InodeNo INTEGER NOT NULL,UNIQUE (Did, Name), UNIQUE (DevNo, InodeNo));

DELETE FROM sqlite_sequence;
INSERT INTO sqlite_sequence VALUES('830A28BBB084683384019DD280C51FA8',16);
CREATE INDEX idx_volpath ON volumes(VolPath);
COMMIT;
```

NJRoadfan:
Here is what comes up when trying to rebuild an existing volume with `dbd -f`:
`
Nov 14 18:53:18.614566 dbd[5231] {cnid_sqlite.c:100} (error:Default): sqlite3_bind_text(db->cnid_add_stmt, 1, stmt_param_name, strlen(stmt_param_name), SQLITE_STATIC) failed: No such file or directory`

Here is a dump of the database, it doesn't get very far:

```
PRAGMA foreign_keys=OFF;
BEGIN TRANSACTION;
CREATE TABLE volumes ( VolUUID CHAR(32) PRIMARY KEY,VolPath TEXT(4096),Stamp BINARY(8),Depleted INT);
INSERT INTO volumes VALUES('830A28BBB084683384019DD280C51FA8','/srv/A2SERVER/A2FILES','n�6g',0);
CREATE TABLE `830A28BBB084683384019DD280C51FA8`(Id INTEGER PRIMARY KEY AUTOINCREMENT,Name VARCHAR(255) NOT NULL,Did INTEGER NOT NULL,DevNo INTEGER NOT NULL,InodeNo INTEGER NOT NULL,UNIQUE (Did, Name), UNIQUE (DevNo, InodeNo));
DELETE FROM sqlite_sequence;
INSERT INTO sqlite_sequence VALUES('`830A28BBB084683384019DD280C51FA8`',16);
CREATE INDEX idx_volpath ON volumes(VolPath);
COMMIT;
```

rdmark:
This is on hold while I clarify the license grant for the mysql CNID module that this is based on.

rdmark:
For the record, the Arch failure is due to their dependency tree being broken at the moment.

https://bbs.archlinux.org/viewtopic.php?id=306080

rdmark:
The `cnid_sqlite_add` operation works now. But there are still plenty of bugs to work through.

```
PRAGMA foreign_keys=OFF;
BEGIN TRANSACTION;
CREATE TABLE volumes ( VolUUID CHAR(32) PRIMARY KEY,VolPath TEXT(4096),Stamp BINARY(8),Depleted INT);
INSERT INTO volumes VALUES('56027931E1405C87F7C97E56D21DBA8F','/opt/afplite','*ICh',0);
CREATE TABLE IF NOT EXISTS "56027931E1405C87F7C97E56D21DBA8F"(Id INTEGER PRIMARY KEY AUTOINCREMENT,Name VARCHAR(255) NOT NULL,Did INTEGER NOT NULL,DevNo INTEGER NOT NULL,InodeNo INTEGER NOT NULL,UNIQUE (Did, Name), UNIQUE (DevNo, InodeNo));
INSERT INTO "56027931E1405C87F7C97E56D21DBA8F" VALUES(17,'TheVolumeSettingsFolder',2,16777241,70839489);
INSERT INTO "56027931E1405C87F7C97E56D21DBA8F" VALUES(18,'Network Trash Folder',2,16777241,70839491);
INSERT INTO "56027931E1405C87F7C97E56D21DBA8F" VALUES(19,'Trash Can Usage Map',18,16777241,70839493);
INSERT INTO "56027931E1405C87F7C97E56D21DBA8F" VALUES(20,'Trash Can #2',18,16777241,70839495);
INSERT INTO "56027931E1405C87F7C97E56D21DBA8F" VALUES(21,'AppleShare 3.sit.1',2,16777241,70839552);
DELETE FROM sqlite_sequence;
INSERT INTO sqlite_sequence VALUES('56027931E1405C87F7C97E56D21DBA8F',21);
CREATE INDEX idx_volpath ON volumes(VolPath);
COMMIT;
```

rdmark:
@NJRoadfan @cheesestraws This is ready for initial review. Basic file operations work for me, but there are several failing spectests that need to be worked through.

The sqlite backend is built by default when the sqlite3 dependency is satisfied. To enable the backend for a volume, set `cnid scheme = sqlite`.

A known limitation is that the dir that houses the sqlite db needs to be writable by all users. The meson install script sets this up for you.

I'm looking forward to you pointing out all of the obvious mistakes I've made. :-D

rdmark:
@NJRoadfan dbd -f does indeed not work still, but now with errors like

```
Jun 07 19:01:15.813359 dbd[61676] {cnid_sqlite.c:616} (error:CNID): cnid_sqlite_add: sqlite query error: datatype mismatch
```

NJRoadfan:
`dbd -f` is kind of important to migrate from one CNID backend to another.

rdmark:
Agreed, a migration path would be crucial for deprecating BDB.

BTW, I don't think that the mysql backend supports a migration path either, does it?

Anyways, I consider the migration path a fast-follow feature. I want to stabilize the base functionality and get this very long-running PR merged sooner rather than later!

NJRoadfan:
What `dbd -f` does is calls `cnid_wipe()` then scans/rebuilds the CNID from scratch.

rdmark:
I touched up `cnid_sqlite_wipe()` a bit and can confirm that the db gets wiped when invoked through dbd, but there obviously some piece missing to restore the db to a good state afterwards. I'll keep working on it later tonight.

rdmark:
@NJRoadfan the `dbd -f` db wipe should be working now, although I have only tested a few trivial cases so far.

rdmark:
There’s a tricky bug right now where when you copy multiple files in one batch into a subdirectory, only the first file gets written into the database. This does not happen when copying multiple files into the root directory.  

~~I suspect some kind of concurrency situation, so I will attempt to refactor the code to use manually controlled SQLite transactions rather than autocommit that I’m relying on at the moment.~~

The actual bug was that the cnid_sqlite_resolve() function used a transient statement that caused some kind of conflict or race condition with the prepared statements used by other functions. The solution was to consistently use prepared statements for all CNID API functions, save for the initial setup of the sqlite database.

rdmark:
This practically works now.

The big question to figure out right now is how to make the database writable by any AFP user while avoiding 666 permissions on the sqlite db file. The one idea I have right now is to call become_root() before every SQL write query but this feels like a crude solution too.

rdmark:
Pulling this back into draft again. The current solution works great for a single threaded scenario, when a single afpd process accesses the sqlite database.

However, when multiple afpd forks opens the database, one of them will put a lock on it and prevent the others from writing to it.

The [SQLite docs](https://www.sqlite.org/isolation.html) has this to say

> ... SQLite implements serializable transactions by actually serializing the writes. There can only be a single writer at a time to an SQLite database. There can be multiple database connections open at the same time, and all of those database connections can write to the database file, but they have to take turns. SQLite uses locks to serialize the writes automatically; this is not something that the applications using SQLite need to worry about.

But somehow the automatic serialization is not taking effect in our CNID implementation, and we encounter fatal `cnid_sqlite_add: sqlite query error: database is locked` errors.

More research will be required.

NJRoadfan:
The joys of concurrency. What command is actually failing? Multiple SELECTs should work as long as its coming from separate database connections. Users shouldn't be calling CREATE or DROP tables either. I haven't taking an in-depth peek, but is any code using the COMMIT command?

Ref: https://www2.sqlite.org/cvstrac/wiki?p=DatabaseIsLocked

Maybe give WAL mode a try? https://til.simonwillison.net/sqlite/enabling-wal-mode
Also: https://www.reddit.com/r/golang/comments/1988ch8/database_is_locked_when_writing_sqlite3_interface/mvxgmmy/

rdmark:
The failing command is an INSERT (the _cnid_put_stmt_ prepared statement). This is a sample log if you're curious. It is curious to note that the logs suggest that .DS_Store doesn't exist in the database at the moment of error, which it actually seems to do when you dump the database.

```
Jun 16 21:57:35.774255 afpd[25653] {cnid_sqlite.c:756} (maxdebug:CNID): cnid_sqlite_add(id: 34, did: 2, name: "manual.pdf", hint: 0): END
Jun 16 21:57:35.781940 afpd[25653] {cnid_sqlite.c:609} (maxdebug:CNID): cnid_sqlite_add(did: 2, name: "manual.pdf", hint: 34): BEGIN
Jun 16 21:57:35.782715 afpd[25653] {cnid_sqlite.c:415} (maxdebug:CNID): cnid_sqlite_lookup(did: 2, name: "manual.pdf"): BEGIN
Jun 16 21:57:35.783324 afpd[25653] {cnid_sqlite.c:476} (maxdebug:CNID): cnid_sqlite_lookup: id=34, did=2, name=manual.pdf, dev=16777239, ino=72929441
Jun 16 21:57:35.783884 afpd[25653] {cnid_sqlite.c:580} (maxdebug:CNID): cnid_sqlite_lookup(id: 34, did: 2, name: "manual.pdf"): END
Jun 16 21:57:35.784459 afpd[25653] {cnid_sqlite.c:632} (debug:CNID): cnid_sqlite_add: lookup returned id=34, errno=93
Jun 16 21:57:35.785035 afpd[25653] {cnid_sqlite.c:756} (maxdebug:CNID): cnid_sqlite_add(id: 34, did: 2, name: "manual.pdf", hint: 34): END
Jun 16 21:57:36.384288 afpd[25653] {cnid_sqlite.c:609} (maxdebug:CNID): cnid_sqlite_add(did: 2, name: "manual.pdf", hint: 34): BEGIN
Jun 16 21:57:36.385143 afpd[25653] {cnid_sqlite.c:415} (maxdebug:CNID): cnid_sqlite_lookup(did: 2, name: "manual.pdf"): BEGIN
Jun 16 21:57:36.386338 afpd[25653] {cnid_sqlite.c:476} (maxdebug:CNID): cnid_sqlite_lookup: id=34, did=2, name=manual.pdf, dev=16777239, ino=72929441
Jun 16 21:57:36.386896 afpd[25653] {cnid_sqlite.c:580} (maxdebug:CNID): cnid_sqlite_lookup(id: 34, did: 2, name: "manual.pdf"): END
Jun 16 21:57:36.387818 afpd[25653] {cnid_sqlite.c:632} (debug:CNID): cnid_sqlite_add: lookup returned id=34, errno=93
Jun 16 21:57:36.388336 afpd[25653] {cnid_sqlite.c:756} (maxdebug:CNID): cnid_sqlite_add(id: 34, did: 2, name: "manual.pdf", hint: 34): END
Jun 16 21:57:36.973773 afpd[25081] {cnid_sqlite.c:609} (maxdebug:CNID): cnid_sqlite_add(did: 2, name: ".DS_Store", hint: 33): BEGIN
Jun 16 21:57:36.975327 afpd[25081] {cnid_sqlite.c:415} (maxdebug:CNID): cnid_sqlite_lookup(did: 2, name: ".DS_Store"): BEGIN
Jun 16 21:57:36.975928 afpd[25081] {cnid_sqlite.c:449} (debug:CNID): cnid_sqlite_lookup: name: '.DS_Store', did: 2 is not in the CNID database
Jun 16 21:57:36.976524 afpd[25081] {cnid_sqlite.c:580} (maxdebug:CNID): cnid_sqlite_lookup(id: 0, did: 2, name: ".DS_Store"): END
Jun 16 21:57:36.976962 afpd[25081] {cnid_sqlite.c:632} (debug:CNID): cnid_sqlite_add: lookup returned id=0, errno=0
Jun 16 21:57:36.977211 afpd[25081] {cnid_sqlite.c:648} (debug:CNID): cnid_sqlite_add: binding name='.DS_Store', did=2, dev=16777239, ino=72929430
Jun 16 21:57:36.977786 afpd[25081] {cnid_sqlite.c:705} (error:CNID): cnid_sqlite_add: sqlite query error: database is locked
Jun 16 21:57:36.978256 afpd[25081] {cnid_sqlite.c:756} (maxdebug:CNID): cnid_sqlite_add(id: 0, did: 2, name: ".DS_Store", hint: 33): END
Jun 16 21:57:36.978731 afpd[25081] {dircache.c:442} (error:AFPDaemon): dircache_add(): did:0 is less than the allowed 17. Try rebuilding the CNID database for: ".DS_Store"
```

And no, I hardly use the COMMIT command at all, but rather rely on SQLite's autocommit.

And yes, WAL journaling mode is activated. See the *cnid_sqlite_open()* function right after opening the database.

As a side note, modifying the cnid_dbd daemon to support SQLite might be the safest choice. It was designed literally to solve this concurrency problem.

rdmark:
@NJRoadfan I've refactored the code to use BEGIN + COMMIT / ROLLBACK consistently for destructive transactions, rather than relying on SQLite's automatic transaction handling. This doesn't resolve the database lock issue unfortunately, but it's probably an improvement nonetheless.

I also added some timeout and retry logic for good measure.

rdmark:
@NJRoadfan Read/write access by concurrent users works now. The key change was to explicitly call sqlite3_reset() on the prepared statement in the cleanup step of every CNID function call that uses a prepared statement. Not entirely intuitively, but it turns out that bookending a query with BEGIN / COMMIT (or ROLLBACK) doesn't necessarily end the transaction.

This documentation page was helpful in understanding the behavior: https://www.sqlite.org/lang_transaction.html

rdmark:
~~There's a bug where moving a file from one folder to another folder leads to an undefined error in the logs. This doesn't have any visible impact on the AFP client however. The CNID record gets updated correctly.~~

It was an easy fix. I was using the wrong prepared statement in cnid_sqlite_get().

```
Jun 21 19:01:03.756417 afpd[45368] {cnid_sqlite.c:877} (maxdebug:CNID): cnid_sqlite_get(did: 21, name: "mar-0.2.0-tc6.68k"): BEGIN
Jun 21 19:01:03.757512 afpd[45368] {cnid_sqlite.c:897} (error:Default): sqlite3_bind_int64(db->cnid_resolve_stmt, 2, ntohl(did)) failed: Undefined error: 0
Jun 21 19:01:03.758526 afpd[45368] {cnid_sqlite.c:911} (maxdebug:CNID): cnid_sqlite_get(id: 0, did: 21, name: "mar-0.2.0-tc6.68k"): END
Jun 21 19:01:03.767506 afpd[45368] {cnid_sqlite.c:926} (maxdebug:CNID): cnid_sqlite_resolve(id: 29): BEGIN
Jun 21 19:01:03.768353 afpd[45368] {cnid_sqlite.c:953} (debug:CNID): cnid_sqlite_resolve: resolved did: 2, name: "asdf"
Jun 21 19:01:03.769116 afpd[45368] {cnid_sqlite.c:955} (maxdebug:CNID): cnid_sqlite_resolve(): END
Jun 21 19:01:03.773238 afpd[45368] {cnid_sqlite.c:926} (maxdebug:CNID): cnid_sqlite_resolve(id: 21): BEGIN
Jun 21 19:01:03.774480 afpd[45368] {cnid_sqlite.c:953} (debug:CNID): cnid_sqlite_resolve: resolved did: 2, name: "mar-0.2.0 Folder"
Jun 21 19:01:03.775249 afpd[45368] {cnid_sqlite.c:955} (maxdebug:CNID): cnid_sqlite_resolve(): END
```

rdmark:
~~The spectest suite runs immensely better now, but we still have 9 failing tests out of the set that passes with the mysql backend.~~

These were fixed by improving error handling, notably using the CNID_DBD_RES_NOTFOUND error code in the correct spots.

  Failed tests:
    FPGetFileDirParms:test58: folder 1 (DIRDID_ROOT_PARENT)
    FPGetFileDirParms:test132: GetFilDirParams errors
    FPGetSrvrParms:test316: GetSrvrParms for a volume with option nostat set
    FPOpenDir:test57: OpenDir call
    FPSetDirParms:test82: test set dir parameters
    FPSetDirParms:test351: change root folder unix perm
    FPSetDirParms:test356: change root folder unix perm
    FPSetFileDirParms:test350: change root folder perm

  Not tested tests (setup step failed):
    FPGetFileDirParms:test396: dir root attribute

rdmark:
The full spectest is now running against the sqlite backend in CI jobs, and all tests are passing. Even those that are consistently failing with the mysql backend.

~~The final thing I want to test before considering this shippable, is that the CNID depletion logic works correctly. Namely that when hitting CNID=0xffffffff (32 bit integer) then we reset the CNID back to 17 (0-16 are reserved) and start incrementing from there again. I can't think of a way to comfortably copy over 4 billion files over AFP, so temporarily using a smaller depletion trigger is probably good enough. :)~~

Ran tests with the threshold of CNID > 128 and it works as expected, with no disruption to the AFP clients. It functions akin to a wipe, but unlike "dbd -f" the new CNIDs are assigned lazily, and the CNID_SQLITE_FLAG_DEPLETED set so that existing appledouble CNID hints are ignored. It's probably advisable to run a wipe with "dbd -f" as soon as possible after a depletion to get a complete set of CNID assignments plus restore appledouble hints.

NJRoadfan:
I have been tied up with some other matters the past few weeks. I'm going to try this out when I have a chance, namely migrating an existing share from BDB to SQLite along with exercising the `ad` tool suite. I know some of the tools have checks to only run with bdb shares, but I have been successful in using the tools with those checks disabled and the MySQL backend.

rdmark:
@NJRoadfan Would it make sense to merge this now, and roll forward with any fixes for `ad` compatibility that we discover later?

NJRoadfan:
I wouldn't worry about the `ad` tool suite right now. There is a check in the volume open function that checks for BDB only. The priority is making sure volume migration from one of the other CNID back ends to SQLite works properly and safely. I would strongly suggest that the SQLite backend be labelled experimental at first given its importance to afpd working correctly.

rdmark:
I was thinking along the same lines. But leaning towards a "beta" label with a notice about taking backups of data before using it. And perhaps also throw a Warning level log message when initializing the backend.

rdmark:
@NJRoadfan Pushed changes to libatalk to print a big warning message when loading the sqlite module, plus notes on the experimental state in the docs.

rdmark:
@NJRoadfan I'm aiming to merge this before the end of the week, unless you see any major blocker.

My plan is to get this experimental backend into the wild with v4.3 and get some data and feedback from users how it performs. Then in the meantime in the main branch start rejigging the bdb backend to swap out BerkeleyDB for SQLite. Once we feel comfortable about the reliability of the rejigged bdb backend, put out v5.0 that removes the BerkeleyDB dependency.

Thoughts?

rdmark:
Further tweaked the reliability of the module:

- Explicit ROLLBACK on all failed cnid_sqlite_execute()
- cnid_sqlite_execute() now takes static SQL queries only, to remove a SQL injection attack vector
- Throw info level logs when CNID has depleted, encouraging rebuilding CNID with dbd -f

This should cover the basics, and I'm confident enough to merge this now and ask the broader community to start beta testing it. :)

PR #1591, opened 2024-10-04
meson: Detect native Avahi before mDNS

singular0:
As Avahi includes `dns_sd.h`, currently build will always detect mDNS as Zeroconf provider. Patch changes the logic to first detect and use Avahi and if not found fallback to mDNS.

### DIFF
--- a/etc/netatalk/meson.build
+++ b/etc/netatalk/meson.build
@@ -5,10 +5,10 @@ if use_mysql_backend
     netatalk_deps += mysqlclient
 endif
 
-if have_dns
-    netatalk_deps += dns_sd_libs
-elif avahi.found()
+if avahi.found()
     netatalk_deps += avahi
+elif have_dns
+    netatalk_deps += dns_sd_libs
 endif
 
 executable(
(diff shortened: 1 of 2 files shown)


rdmark:
@singular0 Thank you for the PR! I understand the nature of the fix. Empirically, this makes the builds on Arch / Alpine / NetBSD / OmniOS use native Avahi instead of mDNS compatibility. Only macOS and Solaris uses native mDNS.

I have one request: Can you please amend the commit message to be a little bit more descriptive? These commit messages are very valuable when generating changelogs or traversing the commit log in the future.

See the project guidelines here: https://github.com/Netatalk/netatalk/wiki/Developer-Notes#commit-messages

In your case, something like "meson: Detect native Avahi before mDNS"

If you need a refresher, the git commands would look like:

```
git commit --amend
git push --force
```

singular0:
@rdmark Updated commit as requested

rdmark (approved):

rdmark:
Merged. Thanks for your contribution!

PR #1599, opened 2024-10-06
Update ad manual page to cover 'ad set' utility.

NJRoadfan:


rdmark:
@NJRoadfan may I ask you to do the Japanese translation as well? :)

Google translate does a very good job with snippets of XML. I can touch it up later if needed.

rdmark:
The only caveat is that Google translate flattens the XML indentation. But I always put it through XMLmind afterwards to autoformat the XML.

rdmark (approved):

Issue #1604, opened 2024-10-09
`cnid_dbd` coredump with `db` 6.2.32


singular0:
**Describe the bug**
`cnid_dbd` built with `db` 6.2.32 on Arch Linux dumps core upon any request.

**To Reproduce**
Build using the following script:

```
  arch-meson . build \
    --localstatedir /var/lib \
    -Dwith-appletalk=true \
    -Dwith-dbus-sysconf-path=/usr/share/dbus-1/system.d \
    -Dwith-debug=true \
    -Dwith-docbook-path=/usr/share/xml/docbook/xsl-stylesheets-${docbookver} \
    -Dwith-dtrace=false \
    -Dwith-init-hooks=false \
    -Dwith-init-style=systemd \
    -Dwith-lockfile-path=/run/netatalk.pid \
    -Dwith-pam-config-path=/etc/pam.d \
    -Dwith-readmes=false \
    -Dwith-spooldir=/var/spool/netatalk
  meson compile -C build
```

Then either connect via AFP or run `dbd -c`.

Same version works fine if built with `db5.3` 5.3.28.

**Expected behavior**
Works without errors.

**Environment**
 - Server OS: Arch Linux
 - Client OS: macOS 15.0.1
 - Netatalk Version: 4.0.1

**Logs**
[build.log](https://github.com/user-attachments/files/17317481/build.log)

**Additional context**
Coredump backtrace:
```
> $ sudo coredumpctl debug 39024
           PID: 39024 (cnid_dbd)
           UID: 0 (root)
           GID: 0 (root)
        Signal: 11 (SEGV)
     Timestamp: Wed 2024-10-09 17:33:12 BST (1h 58min ago)
  Command Line: /usr/bin/cnid_dbd -F /etc/afp.conf -p /home/denis -t 7 -l 5 -u denis
    Executable: /usr/bin/cnid_dbd
 Control Group: /system.slice/netatalk.service
          Unit: netatalk.service
         Slice: system.slice
       Boot ID: b4c48847759f4321a090cc881b9dcc29
    Machine ID: 77e04aa897bb4fb18b0ebc46fe629daf
      Hostname: arch
       Storage: /var/lib/systemd/coredump/core.cnid_dbd.0.b4c48847759f4321a090cc881b9dcc29.39024.1728491592000000.zst (present)
  Size on Disk: 192.3K
       Message: Process 39024 (cnid_dbd) of user 0 dumped core.
                
                Stack trace of thread 39024:
[...]

Issue #1234, opened 2024-07-13
invalid metadata EA


Yanson:
**Describe the bug**
I am seeing tones of log entries like the following:

> 2024-07-13T09:40:13.462563+00:00 smithers afpd[1343023]: ad_header_read_ea("/mnt/space/Downloads/."): invalid metadata EA this is now being treated as a fatal error. if you see this log entry, please file a bug ticket with your upstream vendor and attach the generated core file.

**To Reproduce**
I moved a disk (ZFS format) from a FreeBSD machine to Ubuntu. After installing Netatalk on the new machine, I see these errors for what looks like every file on the disk when I connect to the server from macOS.

**Expected behavior**
No fatal errors being logged

**Environment**
```
iain@smithers:~$ lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 24.04 LTS
Release:	24.04
Codename:	noble
iain@smithers:~$ uname -a
Linux smithers 6.8.0-36-generic #36-Ubuntu SMP PREEMPT_DYNAMIC Mon Jun 10 10:49:14 UTC 2024 x86_64 x86_64 x86_64 GNU/Linux
iain@smithers:~$ netatalk -v
netatalk 3.1.18 - Netatalk AFP server service controller daemon

This program is free software; you can redistribute it and/or modify it under
the terms of the GNU General Public License as published by the Free Software
Foundation; either version 2 of the License, or (at your option) any later
version. Please see the file COPYING for further information and details.

netatalk has been compiled with support for these features:

      Zeroconf support:	Avahi
     Spotlight support:	Yes

                  afpd:	/usr/sbin/afpd
            cnid_metad:	/usr/sbin/cnid_metad
       tracker manager:	/usr/bin/tracker3 daemon
           dbus-daemon:	/usr/bin/dbus-daemon
              afp.conf:	/etc/netatalk/afp.conf
     dbus-session.conf:	/etc/netatalk/dbus-session.conf
    netatalk lock file:	/var/lock/netatalk

```
 - Client OS: macOS Sonoma 14.5

**Logs**
As above

**Additional context**
```
grep -v ';' /etc/netatalk/afp.conf | grep .
[Global]
afp listen = 192.168.1.101
[Homes]
basedir regex = /home
[Waylon's Disk]
path = /mnt/space
[Time Machine on Smithers]
path = /mnt/time-machine/
time machine = yes
```

```
[...]

PR #1610, opened 2024-10-10
meson: Rewrite Berkley DB detection

singular0:
Rewrite of Berkley DB detection in meson aimed to handle issues when multiple versions of library are installed on the system (such as #1604):
* Added `with-bdb-version` option that allows to specify particular Berkley DB version to link against
* Leave option to link against lib file that does not have version in its name as a last resort only (as it could lead to mismatch between header and lib versions)
* Check against minimum supported library version did not work previously
* Extend and rearrange list of potential `db` installation subdirectories for searching from older to newer version

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -400,6 +400,9 @@ endif
 #
 
 bdb_path = get_option('with-bdb-path')
+bdb_req_version = get_option('with-bdb-version')
+
+have_bdb = false
 
 bdb_header = ''
 bdb_includes = []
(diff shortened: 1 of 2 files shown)


meson.build
```diff
@@ -409,107 +412,116 @@ bdb_major_version = ''
...
+    'db50',
     'db5.1',
+    'db51',
     'db5.2',
     'db5.3',
     'db5',
-    'db50',
-    'db51',
     'db6.1',
     '',
```
rdmark, line 443: Do you want to append `db6.2` for good measure, since this is the later version you have?

rdmark:
In case you didn't spot it, the OpenBSD job is failing with

```
meson.build:460:32: ERROR: Command `/usr/bin/grep DB_VERSION_MAJOR /usr/include/db.h` failed with status 1.
```

singular0:
> In case you didn't spot it, the OpenBSD job is failing with
> 
> ```
> meson.build:460:32: ERROR: Command `/usr/bin/grep DB_VERSION_MAJOR /usr/include/db.h` failed with status 1.
> ```

Yes, this looks to be exactly the case of lib without version in its name. I will see what we could do here.

rdmark:
CI jobs look good now!

Can you please squash and force push the two commits? Also, please append the GitHub ticket number to the commit string in this format:

> meson: I did something, GitHub #1234

meson.build
```diff
@@ -409,107 +412,122 @@ bdb_major_version = ''
...
+        
+            bdb_version = bdb_major_version + '.' + bdb_minor_version
+
+            if not bdb_version.version_compare('>=4.6')
+                continue
+            endif
+
+            if bdb_req_version != '' and not bdb_version.version_compare('~' + bdb_req_version)
+                continue
+            endif
```
rdmark, line 487: I would like to throw a warning message here, along the lines of "bdb version X requested via -Dwith-bdb-version but not found"

This gives the user more context to the error message that follows.

What do you think?

singular0: As we search for requested `db` version in cycle over all the dirs and subdirs this would emit a multiple warnings if there are multiple non-matching versions installed. I would rather add this explanation to final error in case particular version was requested but not found.

rdmark: Much better, thanks. This achieves what I wanted to see.

rdmark (approved):

rdmark:
@singular0 FYI I ran into another corner case failure, with the same error

```
  meson.build:482:32: ERROR: Command `/usr/bin/grep DB_VERSION_MAJOR /usr/include/db.h` failed with status 1.
```

https://github.com/Netatalk/netatalk/actions/runs/11311236806/job/31457230668#step:3:2141

This was while tinkering with the OpenBSD CI job. It seems that when you install the "ambiguous" `db` package, this happens. I need to set up a local OpenBSD VM to figure out what the ambiguous package actually installs. It might be the unsupported 3.1 package.

```
Ambiguous: db could be db-4.6.21p7v0 db-3.1.17p3v0
```

PR #1611, opened 2024-10-11
meson: spooldir FHS compliance; add statedir override option

### LINKED ISSUE #1608: FHS compliance
This ticket is for discussing whether Netatalk should go all-in on [FHS 3.0 compliance](https://refspecs.linuxfoundation.org/FHS_3.0/fhs/index.html), or keep some of its non-compliant file system choices.

The one case in point that led to me starting this thread is the localstatedir, namely where CNID databases and other runtime data files are stored. The current default location is `/var/netatalk`. This location was introduced with [3.0-alpha1](https://netatalk.io/3.0/ReleaseNotes-3.0-alpha1) as a centralized place for these, as opposed to individual AFP volume dirs.

FHS 3.0 has this to say about the localstatedir:

> 5.8.1. Purpose
>
>This hierarchy holds state information pertaining to an application or the system. State information is data that programs modify while they run, and that pertains to one specific host. Users must never need to modify files in /var/lib to configure a package's operation, and the specific file hierarchy used to store the data must not be exposed to regular users. [[40]](https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch05s08.html#ftn.idm236090988912)
>
>State information is generally used to preserve the condition of an application (or a group of inter-related applications) between invocations and between different instances of the same application. State information should generally remain valid after a reboot, should not be logging output, and should not be spooled data.
>
>An application (or a group of inter-related applications) must use a subdirectory of /var/lib for its data. There is one required subdirectory, /var/lib/misc, which is intended for state files that don't need a subdirectory; the other subdirectories should only be present if the application in question is included in the distribution. [[41]](https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch05s08.html#ftn.idm236090984800)
>
>/var/lib/<name> is the location that must be used for all distribution packaging support. Different distributions may use different names, of course.

I read this as: Netatalk should put its local state data under `/var/lib/netatalk` rather than `/var/netatalk`

Two questions that I want to answer: 1. Are there other instances like this when netatalk uses file system locations that are non-FHS compliant, and 2. do we want to be FHS compliant in this project, or are the arguments for keeping local idiosyncrasies.

rdmark:
Introduce override option for statedir `with-statedir-path`, while refactoring for future FHS compliance of the CNID dir.

At the same time, make the print spool dir FHS compliant.

Also updates the documentation with the current de-facto statedir defaults.

### DIFF
--- a/doc/ja/manual/configuration.xml
+++ b/doc/ja/manual/configuration.xml
@@ -159,8 +159,8 @@ basedir regex = /usr/home</programlisting></para>
       </indexterm> オプションで選択が可能である。CNID バックエンドは基本的には、ストアされた ID &lt;-&gt;
     名前を一対一対応させるデータベースである。</para>
 
-    <para>The CNID データベースはデフォルトで <filename>/var/netatalk/CNID</filename>
-    に置かれる。コンパイル時に <command>localstatedir=PATH</command> オプションで場所を変えられる。</para>
+    <para>CNID データベースはデフォルトで <filename>$prefix/var/netatalk/CNID</filename>
+    に置かれる。コンパイル時に <command>-Dwith-statedir-path=PATH</command> オプションで場所を変えられる。</para>
 
     <para>CNID データベースの検証・修復・再構築のために用いることができる <command>dbd</command>
     という名のコマンドが用意されていている。</para>
--- a/doc/ja/manual/upgrade.xml
+++ b/doc/ja/manual/upgrade.xml
@@ -169,7 +169,7 @@
           </listitem>
 
           <listitem>
-            <para>CNID データベースはデフォルトで <filename>/var/netatalk/CNID/</filename>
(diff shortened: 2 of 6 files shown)


quentinmit:
With this PR and not manually overriding the spooldir, I get

```
Print spool directory    : var/spool/netatalk
```

which is a problem because it's missing the leading slash. `papd` first `chdir`'s into the spool directory, and then tries to access a spool file with its full path, resulting in an attempt to use `/var/spool/netatalk/var/spool/netatalk/df...`.

rdmark:
@quentinmit good catch, did you test in the latest stable version as well as the main branch? if you file an issue ticket for tracking I can draft a fix before the next release.

quentinmit:
> @quentinmit good catch, did you test in the latest stable version as well as the main branch? if you file an issue ticket for tracking I can draft a fix before the next release.

I opened #2951. Thanks!

PR #1613, opened 2024-10-11
AFP over ASP: Cleanup unused code

singular0:
* Completely remove `afp_options.authprintdir` in `include/atalk/globals.h`
* Remove `afp_authprint_remove` from `etc/afpd/afp_asp.c`

### DIFF
--- a/etc/afpd/afp_asp.c
+++ b/etc/afpd/afp_asp.c
@@ -40,8 +40,6 @@
 
 static AFPObj *child;
 
-static void afp_authprint_remove(AFPObj *);
-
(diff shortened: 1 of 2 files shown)


rdmark:
For the record: We briefly discussed in #1442 whether to bring back CAP style auth or not, but put off the decision.

@NJRoadfan Do you think it's time to remove these remnants? I'm personally ok to clean this up now. We can always go back to old revisions if we want to revive this feature in the future.

rdmark:
@NJRoadfan if you disagree with this change, please file a PR later. :)

PR #1617, opened 2024-10-11
Spotlight: Support TinySPARQL/LocalSearch

singular0:
As of version 3.8 Tracker has [split into TinySPARQL and LocalSearch](https://blogs.gnome.org/carlosg/2024/07/14/goodbye-tracker-hello-tinysparql-and-localsearch/) with command line tool renamed to `localsearch`.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -781,27 +781,32 @@ endif
 
 talloc = dependency('talloc', required: false)
 
-tracker_manager = ''
 tracker_prefix = get_option('with-tracker-prefix')
 
 if not get_option('with-spotlight')
     have_spotlight = false
 else
-    # Check for tracker SPARQL
+    # Check for SPARQL
 
-    tracker_sparql_version = ['3.0', '2.0', '1.0']
+    sparql_versions = [
+        'tracker-sparql-3.0',
+        'tracker-sparql-2.0',
+        'tracker-sparql-1.0'
+    ]
 
-    foreach version : tracker_sparql_version
-        tracker_sparql = dependency('tracker-sparql-' + version, required: false)
-        if tracker_sparql.found()
+    foreach sparql_version : sparql_versions
+        sparql = dependency(sparql_version, required: false)
+        if sparql.found()
             break
         endif
     endforeach
 
-    if not tracker_sparql.found()
-        warning('tracker SPARQL not found (required for Spotlight support)')
+    if not sparql.found()
+        warning('Tracker SPARQL or TinySPARQL not found (required for Spotlight support)')
     else
-        # Check for tracker
+        sparql_prefix = sparql.get_variable(pkgconfig: 'prefix')
+
+        # Check for indexer
 
         tracker = find_program(
             'tracker',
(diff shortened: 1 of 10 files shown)


rdmark:
Thanks for taking action on this one! It's been brewing for a while...

So has Arch already migrated to the renamed versions? Or are you using something else as a test bed?

Some additional changes that I think should be made now, either in this PR or as a fast follow. Please let me know if you want to work on these.

- Update ./meson.build summary block for Spotlight to be Tracker-agnostic. Say, rather than "tracker prefix" call it "indexer prefix" (other suggestions welcome)
- In the same vein, update netatalk.c `show_netatalk_paths()` function to not print the word "tracker" to the user
- Update INSTALL.md
- Update the manual: Install chapter, Configure chapter, afp.conf man page

Changes that should be made eventually:

- afp.conf has an option named `start tracker`. We probably shouldn't rename the option in a bugfix release, but flag it for changing in the future
- meson setup options `with-tracker-install-prefix` and `with-tracker-prefix` ... likewise, not advisable to make breaking changes in a bugfix release, but should be changed eventually
- Refactor internal data structures named after tracker
- references to tracker* / libtracker* headers and symbols ... I assume there will be backwards compatibility headers available for a good long while, but eventually go away

singular0:
Yes, Arch switched to tinysparql/localsearch since 19th of September.

I will update the PR shortly.

meson.build
```diff
@@ -818,49 +823,36 @@ else
...
-        if not tracker_found
-            warning('tracker not found (required for Spotlight support)')
+        indexer_found = (
+            tracker.found()
+            or tracker3.found()
+            or tracker_control.found()
+            or localsearch.found()
+        )
+        if not indexer_found
+            warning('indexer not found (required for Spotlight support)')
```
rdmark, line 852: Might be better to explicitly call out that the build system looks for either Tracker or LocalSearch here, for clarity.

rdmark:
Two things I forgot to mention regarding docs: First, it'd be great if you could update the Japanese localization in `docs/ja`. Google Translate does a really good job with technical docs these days (and I can touch up any awkward language later.) You can throw snippets of XML with embedded text at it, and it accurately translates the text while retaining the XML structure.

The one drawback is that is strips XML indentation completely, which brings me to the second point: We use the XMLmind editor to auto-format XML. Instructions are in: https://github.com/Netatalk/netatalk/wiki/Work-with-Documentation

Sorry about the for the busywork, but doing this work up front really saves me time later.

singular0:
Given my current setup this will take time.

rdmark (approved):
It looks good now apart from XML formatting and the Japanese localization. The latter two can be a fast follow.

rdmark:
@singular0 FYI we got a regression in the Debian buildd environment following this change.

Before the change:

```
tracker manager           : /usr/bin/tracker3 daemon
```

https://buildd.debian.org/status/fetch.php?pkg=netatalk&arch=all&ver=4.0.1%7Eds-1&stamp=1728759567&raw=0

After the change:

```
Indexer command          : "/bin/tracker3 daemon"
```

https://buildd.debian.org/status/fetch.php?pkg=netatalk&arch=amd64&ver=4.0.2%7Eds-1&stamp=1729465547&raw=0

The latter is an invalid location of tracker3 on Debian.

rdmark:
One closer inspection, it might be caused by an unexpected change in the buildd environment. The logic for actually detecting the tracker3 program itself hasn't changed on our end. But despite this, the detected program path has changed in buildd:

`Program tracker3 found: YES (/usr/bin/tracker3)` -> `Program tracker3 found: YES (/bin/tracker3)`

PR #1619, opened 2024-10-12
docker: Use equal sign syntax with ENV

rdmark:
In order to conform to new Docker syntax standards.


rdmark:
Fixed in https://github.com/Netatalk/netatalk/pull/1620

PR #1620, opened 2024-10-12
docker: use multistage build to isolate build from final image

cy384:
For those not familiar, Docker multistage builds reduce the size of the final image and keep a cleaner separation between build artifacts vs. what actually needs to be deployed.  The general idea is to build and then copy the artifacts into a fresh base.

In this case, it looks like this cuts the image size from ~300MB to ~120MB.

I'll note that I only did basic testing (seems fine), but if something's wrong, I would guess it's a missing dependency that wasn't declared explicitly before I changed the Dockerfile.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,4 @@
-FROM alpine:3.20
-
-ENV LIB_DEPS \
+ARG RUN_DEPS="\
     acl \
     avahi \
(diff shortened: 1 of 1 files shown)


rdmark (approved):
Much more elegant solution, thank you!

PR #1626, opened 2024-10-13
Bring back custom icon option for Classic Mac OS

rdmark:
In addition to the ability to customize the icon, we bring back the logic that gives you a different default icon depending on whether the DSI or DDP transport layer is used.

String option `legacy icon' allows you to enable the custom icon by setting it to one of the below strings. Default value is a blank string, which will fall back to either the DSI or DDP icon.

The icon styles are brought back from the 2.x era icon.h. In the past the icon style was auto detected depending on your host OS, but now you are able to pick and choose. Available icon styles are:

- daemon (previously the only option)
- hagar
- declogo
- sunlogo
- globe


rdmark:
@NJRoadfan When you have the opportunity, may I ask you to test the AppleTalk transport layer icon? I don't have any AppleTalk systems handy at the moment (in the middle of being shipped across the world.)

NJRoadfan:
It does work. Note that if AppleTalk is enabled, it always overrides the ASIP icon with the classic AppleTalk style no matter how you connect.

rdmark:
@NJRoadfan Thanks for testing. Good to know that nothing is broken, although it was not exactly what I wanted to achieve. Will make another attempt in the future when I have AppleTalk machines again to test with.

NJRoadfan:
No need for an actual machine. You can test it with `asip-status -i`

rdmark:
asip-status only makes TCP requests though, doesn't it? So I cannot test the DDP-only scenario this way.

NJRoadfan:
The FPGetSrvrInfo packet is the same for both transports. That might need to change in the future though as some older clients are choking if a FPGetSrvrInfo packet exceeds 512 bytes. This is mostly because of the added AFP versions and UAMs reported by Netatalk.

PR #1627, opened 2024-10-13
meson: More robust Berkeley DB version detection

singular0:
In case unable to grep Berkeley DB version from header file, show warning and continue to next header file location


rdmark (approved):

rdmark:
I can see that the "ambiguous package" case on OpenBSD leads to a more helpful message with this change.

PR #1630, opened 2024-10-14
Add sdcard legacy icon style, and an ICN# data conversion script

rdmark:
Brand new `sdcard` legacy icon style, drawn by myself in ResEdit.

Also adding a utility Python script that takes a string of hexadecimal numbers (i.e. copy-pasted from ResEdit's hex editor) and spits out C code with the corresponding array of bytes for use in icon.h

image


rdmark:
> The format used in the Netatalk source is similar to XBM. See: https://thegogglesdonothing.com/archives/2012/02/netatalk_appleshare_finder_icons.shtml

TIL. The format for storing the bytes is truly identical. I doubt the bytes are the same though. ICN# is encoding a suite of differently sized bitmaps and masks.

PR #1634, opened 2024-10-15
Bundle the afptest test suite

rdmark:
To build the test suite, setup meson with `-Dwith-testsuite=true` (default:false). Then after compiling, enter the `build/test/testsuite` dir and execute the test binaries with parameters as described in the readme. Requires a running AFP server to test against, with users and shared volumes configured appropriately.

DONE

- Import the HEAD of https://github.com/Netatalk/afptest
- Write meson build scripts
- Clean up code
- benchmark tests working (lantest, speedtest)
- spectest works with AFP 2.x

TODO

- Refactor away local copies of afp.h / dsi.h / ea.h / exattr.c etc.
- Get spectest to work with AFP 3.x


rdmark:
@NJRoadfan FYI I got this working as well as before, namely functional `lantest` and `speedtest`.

However the most important `spectest` is constantly either segfaulting afpd or getting bad DSI responses. I tested with a 3.1 server as well so it's nothing we broke recently.

rdmark:
Progress. After cleaning up the code (resolving type safety warnings) the spectest started working. I had to disable two test modules that crashes the test suite. Main problem right now is that only AFP 2.x works. When using AFP 3.x there's an immediate `dsi_stream_read(0): EOF` error.

But I'm pleased to see 100+ passing tests, with a handful of gracefully failing ones.

rdmark:
Disabled 11 failing tests with https://github.com/Netatalk/netatalk/pull/1634/commits/80eed07de697cc77769f7d4d24c42161fcb6f916

Tests failing for unknown reasons were in categories: icon validation, dir params, Mac code page, forks.

Two tests were failing because of missing extension to type/creator mapping (i.e. via extmap.conf)

rdmark:
Attempted to set up GitHub workflow jobs to run the spectest as-is within a runner or Docker container. However this wasn't feasible, because there appears to be safeguards against running useradd. I think the proper approach will be to craft a netatalk-testsuite Docker container specifically for running spectest.

PR #1638, opened 2024-10-16
NULL pointer checks before calling bstring macros

rdmark:


rdmark:
These additional NULL checks don't actually silence the compiler warnings I set out to address.

Compare https://stackoverflow.com/questions/26508000/how-to-avoid-null-argument-where-non-null-required-compiler-warning

Issue #1576, opened 2024-10-01
Debian Unstable / gcc 14.2.0-5 memory safety warnings


rdmark:
gcc 14.2.0-5 in Debian Unstable throws warnings related to memory safety

```
../libatalk/adouble/ad_conv.c: In function ‘ad_conv_dehex’:
../libatalk/adouble/ad_conv.c:260:5: warning: argument 2 null where non-null expected [-Wnonnull]
  260 |     strlcpy(buf, bdata(newpath), sizeof(buf));
      |     ^~~~~~~
In file included from ../libatalk/adouble/ad_conv.c:29:
/usr/include/string.h:506:15: note: in a call to function ‘strlcpy’ declared ‘nonnull’
  506 | extern size_t strlcpy (char *__restrict __dest,
      |               ^~~~~~~
```

```
../etc/afpd/uam.c: In function ‘uam_getname’:
../etc/afpd/uam.c:213:17: warning: argument 1 null where non-null expected [-Wnonnull]
  213 |         pwent = getpwnam(bdata(princ));
      |                 ^~~~~~~~
In file included from ../include/atalk/uam.h:8,
                 from ../include/atalk/globals.h:23,
                 from ../include/atalk/dsi.h:17,
                 from ../etc/afpd/uam.c:30:
/usr/include/pwd.h:116:23: note: in a call to function ‘getpwnam’ declared ‘nonnull’
  116 | extern struct passwd *getpwnam (const char *__name) __nonnull ((1));
      |                       ^~~~~~~~
```

```
../etc/afpd/afp_asp.c: In function ‘afp_authprint_remove’:
../etc/afpd/afp_asp.c:76:30: warning: format ‘%s’ expects argument of type ‘char *’, but argument 3 has type ‘int’ [-Wformat=]
   76 |     sprintf(addr_filename, "%s/net%d.%dnode%d", obj->options.authprintdir,
      |                             ~^                  ~~~~~~~~~~~~~~~~~~~~~~~~~
      |                              |                              |
      |                              char *                         int
      |                             %d
```

```
./etc/papd/main.c: In function ‘getstatus’:
../etc/papd/main.c:570:36: warning: ‘snprintf’ output may be truncated before the last format character [-Wformat-truncation=]
  570 |         snprintf(buf->buf, 254, "%s", pr->p_status);
      |                                    ^
../etc/papd/main.c:570:9: note: ‘snprintf’ output between 1 and 255 bytes into a destination of size 254
[...]

PR #1639, opened 2024-10-16
Change PAPStatus buffer to 254 characters

NJRoadfan:
The maximum length of a PAPStatus buffer is 254 characters as the first character is reserved for storing the string length (This is a Pascal style string). Should fix one of the compiler warnings raised in #1576

### DIFF
--- a/etc/papd/print_cups.c
+++ b/etc/papd/print_cups.c
@@ -351,7 +351,7 @@ cups_get_printer_status (struct printer *pr)
 	{
 		LOG(log_error, logtype_papd,
 			"Unable to get destination \"%s\": %s", pr->p_printer, cupsLastErrorString());
-		snprintf(pr->p_status, 255, "status: busy; info: \"%s\" appears to be offline.", pr->p_printer);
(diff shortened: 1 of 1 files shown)


rdmark:
@NJRoadfan While I think this is an improvement, it doesn't address the specific gcc compiler warning that we were seeing. It was raised for a line that was already set to 254 byte buffer size.

```
[256/369] Compiling C object etc/papd/papd.p/main.c.o
../etc/papd/main.c: In function ‘getstatus’:
../etc/papd/main.c:570:36: warning: ‘snprintf’ output may be truncated before the last format character [-Wformat-truncation=]
  570 |         snprintf(buf->buf, 254, "%s", pr->p_status);
      |                                    ^
../etc/papd/main.c:570:9: note: ‘snprintf’ output between 1 and 255 bytes into a destination of size 254
  570 |         snprintf(buf->buf, 254, "%s", pr->p_status);
      |         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

PR #1645, opened 2024-10-17
testsuite: Workaround for enabling AFP 3.x tests

rdmark:
The extended AFP login method is broken, so use the simple one even for AFP3 tests for now.

Also, throw error when AFP login fails, and disable broken AFP3 tests.


rdmark:
With this, the spectest suite (spectest+T2_spectest) is in a passing state, with 261 successful tests and takes 53s to run when testing a local afpd instance.

PR #1647, opened 2024-10-17
Fix papd PAPStatus string copy

NJRoadfan:
As per "Inside AppleTalk", the PAPStatus string can be a max of 255 characters, not 254 characters. The packed struct already accommodates the prefixed byte (`buf->buf_len`) containing the string length. This should silence the compiler error in papd encountered in: #1576


rdmark (approved):
This pacified the warnings, thanks!

Issue #1661, opened 2024-10-21
Vintage AppleShare clients not connecting to server due to FPGetSvrInfo packet being too large.


NJRoadfan:
One common issue encountered with vintage clients (basically anything older then AppleShare Client 3.8) is that they are unable to connect to a Netatalk server if the size of a FPGetSvrInfo response packet exceeds 512 bytes (they error out claiming no response from server). This limit can easily be reached with Netatalk 4.0 if one enables all UAMs and/or has a long host name. This is due to Netatalk 4.0 reporting many versions of the AFP protocol being available in addition to things like the UTF8 server name.

One possible fix is to limit reporting of AFP server versions above 2.1 to AppleTalk clients. This can be easily done as the current code path can be altered to deliver unique response to ASP and DSI clients.

NJRoadfan:
I have an easy potential fix for this. It involves not sending an icon at all if a custom icon isn't configured. The icon+mask takes up 256 bytes of the reply. Otherwise the alternative is to build another status packet in `status_init()` just for AppleTalk connections. That would free up 42 bytes by not reporting AFP versions 2.2 to 3.4 to AppleTalk clients.

rdmark:
I'm supportive of the easy solution. The default icons have little value anyways.

My only regret is that this situation makes it hard to have a single "universal" server for the newest and oldest Apple machines, with all of the bells and whistles activated at the same time: the fancy custom icon, all of the UAMs, and so on. It's a minor detail that few people will actually care about beyond myself.

NJRoadfan:
Most of the additions to the FPGetSrvrInfo packet are exclusive to AFP3.0 and newer thus require TCP/IP. AppleTalk clients do not use the UTF8 server name or directory name.

rdmark:
Right, and this is one argument for introducing a new status packet for AppleTalk like you say. Then you can configure your netatalk server in any way you please and have the older clients only get the FPGetSrvrInfo data that they need, without the need to compromise for newer clients.

The drawback is the added code complexity and maintenance, I suppose.

NJRoadfan:
Fix is done. Of note, the code that builds the status string really needs a refactor. In addition it technically violates the AFP spec by not putting the icon data at the end of the response. Clients don't seem to care though as they just read the offset to find it.

PR #1671, opened 2024-10-23
Limit FPGetSrvrInfo packet for AppleTalk clients

NJRoadfan:
Early AppleTalk clients struggle with FPGetSrvrInfo packets that have a payload larger then 512 bytes. In many cases they claim that they received no response from the server. In some cases they may even crash.

To resolve this, craft a response just for AppleTalk clients without most of the AFP3.x only options (UTF8 server name, Directory Name, etc.). In addition, unless a custom icon is sent, the server will not send an icon image to AppleTalk clients by default.

Fixes: #1661

### DIFF
--- a/etc/afpd/auth.c
+++ b/etc/afpd/auth.c
@@ -75,7 +75,7 @@ void status_versions( char *data,
 #ifndef NO_DDP
         if (!asp && (afp_versions[i].av_number <= 21)) continue;
 #endif /* ! NO_DDP */
-        if ( !dsi && (afp_versions[ i ].av_number >= 22)) continue;
+        if ( !dsi && (afp_versions[ i ].av_number > 22)) continue;
         count++;
     }
     data += ntohs( status );
@@ -85,7 +85,7 @@ void status_versions( char *data,
 #ifndef NO_DDP
         if (!asp && (afp_versions[i].av_number <= 21)) continue;
 #endif /* ! NO_DDP */
-        if ( !dsi && (afp_versions[ i ].av_number >= 22)) continue;
+        if ( !dsi && (afp_versions[ i ].av_number > 22)) continue;
         len = strlen( afp_versions[ i ].av_name );
         *data++ = len;
         memcpy( data, afp_versions[ i ].av_name , len );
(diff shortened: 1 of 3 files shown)


etc/afpd/status.c
```diff
@@ -548,16 +550,82 @@ void status_init(AFPObj *dsi_obj, AFPObj* asp_obj, DSI *dsi)
...
+    status_machine(aspstatus);
+    status_versions(aspstatus,
+        asp,
+        NULL);
+    status_uams(aspstatus, options->uamlist);
+
+    if (strcmp(options->legacyicon, "daemon") == 0) {
+        icon = daemon_icon;
+        iconlen = sizeof(daemon_icon);
+    }
```
rdmark, line 579: We get duplication of the icon lookup code here. Better would be to have a shared utility function.

etc/afpd/status.c
```diff
@@ -520,12 +528,6 @@ void status_init(AFPObj *dsi_obj, AFPObj* asp_obj, DSI *dsi)
         icon = sunlogo_icon;
         iconlen = sizeof(sunlogo_icon);
     }
-#ifndef NO_DDP
-    else if (asp) {
-        icon = apple_atalk_icon;
-        iconlen = sizeof(apple_atalk_icon);
-    }
```
rdmark, line 527: I want to turn this into a manual option instead, add it to the selection of icons that you can choose. Optional for this PR, though. I can do it later.

etc/afpd/auth.c
```diff
@@ -75,7 +75,7 @@ void status_versions( char *data,
 #ifndef NO_DDP
         if (!asp && (afp_versions[i].av_number <= 21)) continue;
 #endif /* ! NO_DDP */
-        if ( !dsi && (afp_versions[ i ].av_number >= 22)) continue;
+        if ( !dsi && (afp_versions[ i ].av_number > 22)) continue;
```
rdmark, line 78: What's the reason for this change? AFP 2.2 is still capable of DSI, isn't it?

rdmark (approved):

rdmark:
As a side note, since the testsuite is operational now, adding validation of FPGetSrvInfo responses would be a valuable area to explore. AFAICT we only validate that a non-error is returned in `test1` (literally the first test in the suite).

PR #1673, opened 2024-10-23
meson: Fix indexer path detection on meson 1.6

singular0:


rdmark (approved):
Seems like a safe fix, thanks.

PR #1679, opened 2024-10-24
Fix regression in get_vol_uuid()

NJRoadfan:
This fixes a regression introduced in commit: https://github.com/Netatalk/netatalk/commit/f4ae924843e376e96ea46de5e23c176dbe959682

The file containing the volume UUID should be opened read-only because in some contexts, the user does not have read-write privileges (ex: running the `ad` tool suite).


rdmark (approved):
Interesting, I see that the signature file is opened with "a" elsewhere when creating it from scratch. I wonder if there's any code path now that expects this file to be writable when existing (which obviously would be a bug, since UUIDs shouldn't be arbitrarily changed)...

PR #1681, opened 2024-10-24
meson: Fix PAM config directory detection

singular0:
In case `-Dwith-pam-config-path` was not explicitly specified, even if actual `pam.d` directory was located on the system, `{prefix}/etc/pam.d` was used instead to install `netatalk.pamd`

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1594,6 +1594,7 @@ pam_path = get_option('with-pam-path')
 pam_conf_path = get_option('with-pam-config-path')
 
 pam_dir = ''
+pam_d_dir = ''
 pam_confdir = ''
 pam_includes = []
 pam_link_args = []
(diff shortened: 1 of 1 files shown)


rdmark:
@singular0 Would you mind writing a sentence or two in the initial comment with a summary of what you changed, and why? This is because the release notes changelog will have a link to this PR, and people clicking through to see the details of the change deserve a brief summary...

rdmark:
Inspecting the CI job logs, Solaris and OmniOS now get empty PAM paths by default. This should be the correct behavior on Solaris-likes since they historically rely on manually updating `/etc/pam.conf` after installing Netatalk... unless these distros changed recently.

All the other OSes are now detecting the PAM dir as `/etc/pam.d` which is *probably* correct across the board.

PR #1684, opened 2024-10-24
meson: Consistent formatting in setup summary

rdmark:
Add the quotations to the indexer command only at the time of substitution, to give a consistent look of the Meson setup summary.


rdmark:
@singular0 FYI, a small tweak to your contribution.

PR #1687, opened 2024-10-24
Fix regression in get_vol_uuid() 2nd attempt

NJRoadfan:
We need the file to be in  read-write mode after all. The function later writes to the file to append data.


rdmark:
Bummer, it seems branches originating from forks don't have permissions to publish to the registry. Which means that only project members will be able to run the testsuite in CI. :(

Like with the Sonarcloud static analysis job, we may have to put them behind a fork check...

rdmark (approved):

PR #1689, opened 2024-10-25
Test container builds from forks

rdmark:


rdmark:
The experiment is a success. Container jobs will be cleanly skipped for fork-originating PRs.

PR #1701, opened 2024-10-27
GitHub CI: Use NetBSD 10.0 vmactions profile

rdmark:


rdmark:
Kerberos is still broken unfortunately...

```
  In file included from ../etc/afpd/status.c:16:
  /usr/pkg/include/krb5/krb5.h:42:10: fatal error: krb5-types.h: No such file or directory
     42 | #include <krb5-types.h>
        |          ^~~~~~~~~~~~~~
  compilation terminated.
```

PR #1704, opened 2024-10-27
Revert "Fix loss of FinderInfo on resource fork creation with AppleDo…

NJRoadfan:
…uble EA backend."

If a resource fork is written after FinderInfo, we could lose the Filetype/Creator data for the file.


rdmark (approved):
Only project members and outside collaborators have the privileges to run the Containers jobs... so the AFP test jobs will probably run once I merge to main...

rdmark:
Indeed. Are you happy with the assumptions of the tests? We can certainly modify the tests if you disagree with them.

PR #1717, opened 2024-10-31
testsuite: Break down login testsuite into atomic tests

rdmark:
Take parts of `main()` and move it into four new test cases with setup, teardown, and results reporting.


rdmark:
For the record, I considered moving the Login tests into spectest as well, but it does non-standard things with sessions and auth that may mess with other tests.

PR #1719, opened 2024-11-01
meson: Fix build fail with -Dwith-spotlight=false

singular0:
Fix build fail with `-Dwith-spotlight=false` due to `sparql` being undefined, as suggested by @APCCV, plus logic simplification


rdmark (approved):
Thanks for the fix!

PR #1732, opened 2024-11-05
Fix NBP Lookup Reply duplication.

### LINKED ISSUE #1731: atalkd: NBP service sending multiple NBP Reply packets
Currently when responding to NBP Lookups, the atalkd service sends back multiple reply packets to the client depending on the number of network interfaces the machine has plus the loopback interface. On earlier Linux kernels (pre-6.9), these packets all had an source address of the interface that originally received the packet. Starting with Linux 6.9, a long time AppleTalk broadcast bug was fixed. The result is now each of those responses contains a source address for each of the network interfaces on the machine, including the loopback device which uses an invalid address of `0.0`.

The logic in `nbp.c` needs to be reworked to only send one NBP Reply back to the client. The initial PR does this, but will break NBP Lookups if one has multiple interfaces setup with the `-dontroute` option with services (ex: afpd) listening on different networks. Even now this logic is broken with the current code as atalkd will advertise services on non-routed networks that clients on those networks can't actually access.

NJRoadfan:
Only send one reply packet back for the interface that originally requested it. Mostly fixes #1731

### DIFF
--- a/etc/atalkd/nbp.c
+++ b/etc/atalkd/nbp.c
@@ -506,8 +506,7 @@ Can't find route's interface!" );
 
         for ( ntab = nbptab; ntab; ntab = ntab->nt_next ) {
             /* don't send out entries if we don't want to route. */
-            if ((ap->ap_iface != ntab->nt_iface) &&
-                (ntab->nt_iface->i_flags & IFACE_ISROUTER) == 0) {
+            if (ap->ap_iface != ntab->nt_iface) {
                 continue;
             }
 


NJRoadfan:
This is going to need more work. Currently this breaks things if you have different zones on different interfaces.

NJRoadfan:
Closing for now. Further investigation is needed here. Finding the source of the problem of anything in atalkd is an adventure.

rdmark:
are you still certain that the problem lies in atalkd at this point?

NJRoadfan:
Pretty sure. It appears to have something to do with zones and how atalkd's NBP responder determines what interface to send a reply from. This only affects services running on the same machine (ex: afpd and papd) as atalkd and when atalkd is acting as a router with multiple interfaces. Even standalone with the `-router` switch it happens (you get the extra "response" coming back from the loopback interface). It does not happen if atalkd is run without any seeding and running standalone (adhoc self configured in the startup range).

NJRoadfan:
I just tested this under NetBSD (old macipgw VM), and the duplicate responses exist there as well. So its definitely Netatalk.

PR #1734, opened 2024-11-07
atalkd: Don't send NBP Reply packets from the loopback interface

### LINKED ISSUE #1731: atalkd: NBP service sending multiple NBP Reply packets

NJRoadfan:
In Linux 6.9+, NBP Reply packets can appear to come from the loopback interface, which has an address of `0.0`. This address is invalid on Phase 2 AppleTalk networks and should never appear. One minor benefit is less traffic is generated when doing NBP Lookups. Fixes #1731 without breaking anything additional.

### DIFF
--- a/etc/atalkd/nbp.c
+++ b/etc/atalkd/nbp.c
@@ -499,6 +499,11 @@ Can't find route's interface!" );
         /*FALL THROUGH*/
 
     case NBPOP_LKUP :
+        /* do not send replies from the loopback interface */
+        if (ap->ap_iface->i_flags & IFACE_LOOPBACK)
+        {
+            return 0;
+        }
         /* search our data */
         n = i = 0;
         data = packet + 1 + SZ_NBPHDR;


NJRoadfan:
OK. New patch just filters out packets from the loopback adapter. One can't suppress the duplicate packets. It is by design if multiple interfaces are on the same zone. Some sort of split horizon logic would need to be done, but its not worth the effort over the minimum additional traffic. This patch fixes the biggest problem, which is packets from the loopback interface leaking onto the network.

Note, this wasn't tested on NetBSD or pre-6.9 Linux kernels. There should be no regressions though. At the worst, its still sending the extra reply from the loopback, but with the address of the server's interface as before.

rdmark:
@NJRoadfan Is this ready for review now?

rdmark (approved):

PR #1747, opened 2024-11-11
Default to no init scripts if service management command not found

singular0:
In case system is assumed to have particular init style, but corresponding service management command (such as `systemctl`, `rc-service`, `rcctl`, etc.) is not found, default to no init scripts. In case `-Dwith-init-style` was explicitly specified, fail the build.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1900,26 +1900,43 @@ endif
 
 init_style = get_option('with-init-style')
 init_dirs = []
+init_style_auto = false
+init_cmd = ''
 
 if 'auto' in init_style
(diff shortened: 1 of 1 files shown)


rdmark:
@singular0 Thanks for the the PR! I wonder if it would be worth to do the same for OpenRC? It is our default for Alpine and Gentoo for a "normal" OS setup, but there's no guarantee it is installed under all circumstances (such as in a Docker container). A minor consideration, but worth a thought.

rdmark (approved):
good idea, cheers

PR #1749, opened 2024-11-12
Limit versions reported by FPGetSrvrInfo Over DSI

NJRoadfan:
Refer to #1661. Turns out that AppleShare Client 3.7.4 suffers from the 512 byte packet limit bug, including when receiving responses over DSI. Being that this is the only officially supported version with AFP-over-TCP for System 7 clients, it is important that this is fixed.

This change limits Netatalk to replying with AFP versions 2.2 to 3.4 to DSI clients, which poses no compatibility issues as there are technically no TCP/IP clients supporting the older versions.

The only exception to this is the AFPBridge control panel for GS/OS, which identifies as AFPVersion2.0 by default. This TCP/IP client has been tested and still works fine even with this change.

### DIFF
--- a/etc/afpd/status.c
+++ b/etc/afpd/status.c
@@ -499,7 +499,7 @@ void status_init(AFPObj *dsi_obj, AFPObj* asp_obj, DSI *dsi)
     status_machine(status);
     status_versions(status,
 #ifndef NO_DDP
-                    asp,
+                    NULL,
 #endif
                     dsi);
     status_uams(status, options->uamlist);


NJRoadfan:
Most people don't use long host names and/or enable all the UAMs. Netatalk 3.0 added AFP3.4 support (another 7 bytes) which tightened how long a hostname could be before crossing the 512 byte boundary, particularly if you have all the UAMs turned on. Things are even worse if you are using LDAP or the server is advertising multiple listening addresses.

etc/afpd/status.c
```diff
@@ -499,7 +499,7 @@ void status_init(AFPObj *dsi_obj, AFPObj* asp_obj, DSI *dsi)
     status_machine(status);
     status_versions(status,
 #ifndef NO_DDP
-                    asp,
+                    NULL,
 #endif
```
rdmark, line 503: Can we remove the ifdef and always send the NULL parameter now? That would allow us to simplify `status_versions()` as well, unless I'm mistaken.

NJRoadfan: Not easily. The ASP struct that is defined in `status_versions()` doesn't exist if NO_DDP is set.

rdmark: Ah ok, maybe a rafactoring for another time. It would be nice to reduce the `#ifndef NO_DDP` spaghetti and use normal logic.

rdmark (approved):

PR #1751, opened 2024-11-13
Fix "ad find" utility regression.

NJRoadfan:
Sometime after the switch to the INI based configuration in Netatalk 3.0, the `ad find` utility broke. Currently it only works if one enters only lowercase letters and strings of two characters or more. Otherwise it either returns no results or errors out.

### DIFF
--- a/bin/ad/ad_find.c
+++ b/bin/ad/ad_find.c
@@ -117,8 +117,8 @@ int ad_find(int argc, char **argv, AFPObj *obj)
 
     uint16_t flags = CONV_TOLOWER;
     char namebuf[MAXPATHLEN + 1];
-    if (convert_charset(vol.vol->v_volcharset,
-                        vol.vol->v_volcharset,
+    if (convert_charset(CH_UNIX,
+                        CH_UNIX,
                         vol.vol->v_maccharset,
                         argv[optind],
                         strlen(argv[optind]),


bin/ad/ad_find.c
```diff
@@ -117,8 +117,8 @@ int ad_find(int argc, char **argv, AFPObj *obj)
 
     uint16_t flags = CONV_TOLOWER;
     char namebuf[MAXPATHLEN + 1];
-    if (convert_charset(vol.vol->v_volcharset,
-                        vol.vol->v_volcharset,
```
rdmark, line 121: ad_util.c is also using the vol->v_volcharset structure. Does it need updating too?

NJRoadfan: Shouldn't be an issue. In this particular case `convert_charset()` was being used to convert case, not character sets. Calling `vol->v_volcharset` should be returning CH_UNIX (alias, almost always set to UTF8), but for some reason it was having problems. This code worked in Netatalk 2.x, so I don't know exactly where things broke.

I tested the ad find utility with non-Latin characters and it still works. Explicitly using CH_UNIX as the source and destination type is safe here as we aren't actually converting character sets and this is solely a server side operation where everything should be Unicode.

rdmark (approved):

PR #1754, opened 2024-11-15
only build appletalk docs when enabled

### LINKED ISSUE #1753: Only build appletalk documentation when with-appletalk=true
Because the appletalk kernel module is not available on rhel, I created a netalk-appletalk subplackage, that only gets built on fedora.

However, I noticed that the manpages and html documentation for the Appletalk features were still being built, even when the with-appletalk flag was off. I needed a way to stop this, so I made the appropriate meson.build changes to make this happen.

PR to resolve this is incoming...

knight-of-ni:
fixes #1753 

tested to work properly on el9 and f41

### DIFF
--- a/doc/ja/manual/meson.build
+++ b/doc/ja/manual/meson.build
@@ -11,7 +11,6 @@ configure_file(
 )
 
 manual_pages = [
-    'appletalk.xml',
     'compile.xml',
     'configuration.xml',
     'gpl.xml',
@@ -19,6 +18,12 @@ manual_pages = [
     'intro.xml',
     'upgrade.xml',
 ]
+
+if have_appletalk
+    manual_pages += [
+        'appletalk.xml',
+    ]
+endif
 
 foreach page : manual_pages
     configure_file(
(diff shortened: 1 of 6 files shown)


rdmark:
@knight-of-ni Can you please do these changes for the Japanese localization as well? The localization is disabled by default, and can be enabled with `-Dwith-manual-l10n=ja` if needed for testing.

knight-of-ni:
I reran the build on el9 and f41 while setting `-Dwith-manual-l10n=ja` and the build succeeded in both cases.

If any of the filenames of the html or man files had changed during the build, the rpm build process would have caught that and failed. I'm sure the contents are different.

knight-of-ni:
Done in [ee366dc](https://github.com/Netatalk/netatalk/pull/1754/commits/ee366dcc88d4372828f08b219a81665dda994857)

rdmark:
Cheers! Can you also please squash them (e.g. `rebase -i HEAD~2`) and force push the resulting commit? We like to keep the commit log clean like this.

rdmark (approved):

PR #1755, opened 2024-11-15
atalkd: Remove obsolete NO_DATA_LINK_PASSTHROUGH config flag

rdmark:


NJRoadfan:
It can be removed. The code setting the macro (for FreeBSD) was removed from Netatalk 2.3 as part of removing obsolete platform support.

NJRoadfan (approved):

PR #1756, opened 2024-11-15
GitHub CI: Use openSUSE Leap instead of Tumbleweed base image

rdmark:


rdmark:
A year later, and it's still not possible to use Leap in a GitHub workflow...

PR #1757, opened 2024-11-15
docs: Document afppasswd -w option in the man page synopsis

rdmark:
Also contains an overhaul of the command argument markup in the synopsis, to reduce unnecessary nesting and superfluous styles.


rdmark:
@NJRoadfan FYI, this changes the default styling of the arguments in the synopses a little bit. I think it's worth the improved simplicity of the markup.

PR #1768, opened 2024-11-21
meson: Correct capability check for SunRPC on *BSD

rdmark:
For the SunRPC / libquota check, look for `quota_open()` instead of the obsolete `getfsquota()`

Don't compile the bundled nfsquota module when using libquota.

### DIFF
--- a/etc/afpd/meson.build
+++ b/etc/afpd/meson.build
@@ -73,9 +73,13 @@ endif
 
 if have_quota
     afpd_sources += [
-        'nfsquota.c',
         'quota.c',
     ]
+    if not have_libquota
+        afpd_sources += [
+            'nfsquota.c',
+        ]
+    endif
     afpd_external_deps += quota_deps
 endif
 
--- a/etc/afpd/quota.c
+++ b/etc/afpd/quota.c
@@ -32,6 +32,7 @@
 #include <atalk/util.h>
 
 #include "auth.h"
+#include "directory.h"
 #include "volume.h"
 #include "unix.h"
 
--- a/meson.build
+++ b/meson.build
@@ -971,6 +971,7 @@ tirpc = dependency('libtirpc', required: false)
 
 quota_deps = []
 quota_provider = ''
+have_libquota = false
 
 rpc_headers = [
     'rpc/rpc.h',
@@ -1017,7 +1018,8 @@ else
         if have_quota
             quota_deps += rpcsvc
             quota_provider += 'SunRPC'
-            if quota.found() and cc.has_function('getfsquota', dependencies: [quota, prop, rpcsvc])
+            if quota.found() and cc.has_function('quota_open', dependencies: [quota, prop, rpcsvc])
+                have_libquota = true
                 quota_deps += [quota, prop]
                 cdata.set('HAVE_LIBQUOTA', 1)
             endif
--- a/test/afpd/meson.build
+++ b/test/afpd/meson.build
@@ -56,9 +56,13 @@ endif
 
 if have_quota
     test_sources += [
-        meson.project_source_root() / 'etc/afpd/nfsquota.c',
         meson.project_source_root() / 'etc/afpd/quota.c',
     ]
+    if not have_libquota
+        test_sources += [
+            meson.project_source_root() / 'etc/afpd/nfsquota.c',
+        ]
+    endif
     test_external_deps += quota_deps
 endif
 


hfath:
Two comments:

I wonder if there should be a have_libquota meson selector, capturing support for the NetBSD libquota(3) / FreeBSD quotafile(3) API, instead of conditionalizing on the OS.

The quota.c #ifdef __NetBSD__ should not be necessary, because that code is #ifdef'ed out with libquota support.

rdmark:
> Two comments:
> 
> I wonder if there should be a have_libquota meson selector, capturing support for the NetBSD libquota(3) / FreeBSD quotafile(3) API, instead of conditionalizing on the OS.
> 

Easily done! I actually had that in an earlier revision, if you look at the force push diff.

The way you do this in Meson, is to define a boolean variable somewhere in the global scope, i.e. `have_libquota = false`, then set the flag to `true` in the branch that check for that dependency.

It's recommended to define the flag early and in the global scope first to avoid errors for certain branches.

> The quota.c `#ifdef __NetBSD__` should not be necessary, because that code is #ifdef'ed out with libquota support.

I think we need a new capability check, either to replace or supplement the current libquota check. The current libquota check is false positive for NetBSD. Do you have an idea?

rdmark:
@hfath Scratch the above. You were right.

It builds cleanly on all OSes now.

rdmark:
Thanks for the patch! Do you have any interest whatsoever to file a GitHub PR for this?

A PR would give you proper credits in the commit log and the next release notes. :)

rdmark:
> I have a related patch which
> 
>     1. checks for libquota first - should work both on Net- and FreeBSD
> 
>     2. sets quota_provider accordingly (is this functional, or mainly eye candy?)
> 

Eye candy primarily, with the added benefit of informing ignorant netatalk maintainers which library was detected on which OS.

hfath:
@PR -- I should get some practice, I guess... give me a couple of days, too many balls in the air right now

rdmark:
BTW, you can define a code block delineated by a pair of three backticks like this:

```
$NetBSD$

--- meson.build.orig	2024-11-24 17:34:19.000000000 +0000
+++ meson.build
@@ -1012,40 +1012,43 @@ endif
 if not enable_quota
     have_quota = false
     cdata.set('NO_QUOTA_SUPPORT', 1)
-else
-    have_quota = tirpc.found() and cc.has_header('rpcsvc/rquota.h')
-    if have_quota
-        cdata.set('NEED_RQUOTA', 1)
-        quota_deps += tirpc
-        quota_provider += 'libtirpc'
-        if cc.compiles(
-            '''
-                    #include <rpcsvc/rquota.h>
-                    int main(void) {
-                        enum qr_status foo;
-                        foo = Q_OK;
-                        return 0;
-                    }
-                ''',
-            dependencies: tirpc,
+elif quota.found() and cc.has_function('quota_open')
+    # FreeBSD provides libquota API in libc
+    have_libquota = true
+    have_quota = true
+    quota_provider += 'libquota'
+    cdata.set('HAVE_LIBQUOTA', 1)
+elif quota.found() and cc.has_function('quota_open', dependencies: [quota, prop])
+    have_libquota = true
+    have_quota = true
+    quota_provider += 'libquota'
+    quota_deps += [quota, prop]
+    cdata.set('HAVE_LIBQUOTA', 1)
+elif tirpc.found() and cc.has_header('rpcsvc/rquota.h')
+    cdata.set('NEED_RQUOTA', 1)
+    quota_deps += tirpc
+    quota_provider += 'libtirpc'
+    if cc.compiles(
+        '''
+        #include <rpcsvc/rquota.h>
+        int main(void) {
+        enum qr_status foo;
+        foo = Q_OK;
+        return 0;
+        }
+        ''',
+        dependencies: tirpc,
         )
-            cdata.set('HAVE_RQUOTA_H_QR_STATUS', 1)
-        endif
-    else
-        have_quota = rpcsvc.found() and rpc_headers_ok
-        if have_quota
-            quota_deps += rpcsvc
-            quota_provider += 'SunRPC'
-            if quota.found() and cc.has_function('quota_open', dependencies: [quota, prop, rpcsvc])
-                have_libquota = true
-                quota_deps += [quota, prop]
-                cdata.set('HAVE_LIBQUOTA', 1)
-            endif
-        else
-            have_quota = false
-            cdata.set('NO_QUOTA_SUPPORT', 1)
-        endif
+        cdata.set('HAVE_RQUOTA_H_QR_STATUS', 1)
     endif
+    have_quota = true
+elif rpcsvc.found() and rpc_headers_ok
+    quota_deps += rpcsvc
+    quota_provider += 'SunRPC'
+    have_quota = true
+else
+    have_quota = false
+    cdata.set('NO_QUOTA_SUPPORT', 1)
 endif
 
 if enable_quota and not have_quota
 ```

Just in case you weren’t aware. It’s a handy way to share code snippets or patches.

PR #1769, opened 2024-11-22
Remove bitrotted inline prototype for a libc function

rdmark:
When a compiler complains about a missing prototype for a libc function, just inserting one for what you think the interface might be is probably not a good idea.

See also <https://github.com/mvp/uhubctl/issues/586>


rdmark:
This is actually a regression, as this was previously patched for 3.x in https://github.com/Netatalk/netatalk/commit/b84db222fd98f18b612fc64e0460a88e78a2f070

I broke it again by bumping to bstrlib 1.0 code in https://github.com/Netatalk/netatalk/pull/1152

PR #1779, opened 2024-11-23
meson: Option to specify path to perl runtime

rdmark:
Introduces `-Dwith-perl-path` which takes the path to the containing dir of perl. Passes to absolute path to perl to the build system.


rdmark:
@hfath Please let me know if this allows you achieve what you wanted to do.

PR #1780, opened 2024-11-23
meson: Configure dbus paths and config files only if dbus exists

rdmark:
Remove spurious hard coded defaults. Leverage the existence of dirs or dbus libs.


rdmark:
@hfath This should rid you of defined dbus paths and either dbus config file when the library doesn't exist on your system. Please let me know if there are other remnants elsewhere.

PR #1783, opened 2024-11-23
meson: Generate appendix doc and allow custom manual install path

rdmark:
This adds build system functionality for project admins specifically.

First, when building with `-Dwith-manual=www` the python script for refreshing the compile manual XML sources is run
Secondly, with the new `-Dwith-manual-install-path` option you can set a custom install path, e.g. the path to a checked out copy of netatalk-homepage

These automates currently manual parts of the release process.

Note that the build system won't check for required python libraries. Only python3 itself. This is up to the user to prepare.

### DIFF
--- a/doc/ja/manual/meson.build
+++ b/doc/ja/manual/meson.build
@@ -98,5 +98,5 @@ custom_target(
         '@INPUT@',
     ],
     install: true,
-    install_dir: datadir / 'doc/netatalk/htmldocs/ja',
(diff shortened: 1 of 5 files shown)


rdmark:
The OmniOs failure seems unrelated. The vmactions backend is in bad shape.

rdmark:
The release process docs has been updated https://github.com/Netatalk/netatalk/wiki/Release-Process#making-a-stable-release

PR #1785, opened 2024-11-24
webmin: Split up index page into three tabs

rdmark:
Rather than a monolithic index page for the webmin module, this introduces a tabbed UI with three tabs: One for the AFP file server (the default), one for DDP (AppleTalk) services, and one for other global settings.

This brings two benefits: Better at-a-glance understanding of the state of the AFP file server when you enter the module, which is what most people are interested in. And secondly, more logical grouping of settings and service control.

The main drawback is that operations on secondary tabs gives you a roundtrip to the default tab, which can be a chore especially when starting/stopping the AppleTalk services.

Screenshot 2024-11-24 at 8 56 38

Screenshot 2024-11-24 at 8 57 12

Screenshot 2024-11-24 at 8 57 28


rdmark:
For the record, I'm holding off on merging this until 4.0 is finally branched off and we start work on 4.1. This change is too drastic for a bugfix release.

contrib/webmin_module/delete_atalk.cgi
```diff
@@ -44,7 +44,7 @@ eval {
 		}
 	}
 
-	redirect("show_atalk.cgi");
+	redirect("index.cgi");
```
rdmark, line 47:
```suggestion
	redirect("index.cgi");
	redirect("index.cgi");
```

PR #1789, opened 2024-11-25
On macOS, enable binary section workaround for ObjC fork policy

rdmark:
Rather than setting the `OBJC_DISABLE_INITIALIZE_FORK_SAFETY` env variable, use the linker to inject a section with `__DATA,__objc_fork_ok` into the resulting macOS binary. This allows for a more persistent way to bypass Apple's anti-forking policy.

### DIFF
--- a/distrib/initscripts/macos.netatalk.in
+++ b/distrib/initscripts/macos.netatalk.in
@@ -3,7 +3,6 @@
 # # Startup daemon for netatalk on macOS
 
 # Prepare the environment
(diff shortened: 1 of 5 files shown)


rdmark:
Pulling back to draft. I need to make sure this workaround is still required. I cannot reproduce the fork assertion in macOS Sequoia, with or without workarounds.

rdmark:
Superseded by https://github.com/Netatalk/netatalk/pull/1792

PR #1796, opened 2024-12-01
Set resource max limit to 10240 on macOS

rdmark:
Earlier versions of macOS such as 10.15 Catalina will throw an error when ulimit is set to too high a number. This PR uses a compliant value when calling `setrlimit()` on macOS specifically.


rdmark:
@NJRoadfan Would you be able to test this on your Catalina Mac? The Macs I have at hand are all running too new versions...

rdmark:
Good idea. Let's apply this workaround for afpd too. Please test the updated PR when you have the opportunity!

Issue #1800, opened 2024-12-04
docs: man pages are not rendered correctly


singular0:
**Describe the bug**
All man pages are not correctly formatted when displayed, see attached screenshot.

<img width="1273" alt="Screenshot 2024-12-04 at 19 46 07" src="https://github.com/user-attachments/assets/b4fea8e2-27ba-4939-b8b1-70d36edd5fce">

**To Reproduce**
Build and install netatalk, open any man page, say `man afp.conf`.

**Expected behavior**
Formatting should be correctly applied.

**Environment**
 - Server OS: Arch Linux
 - `netatalk` 4.0.7
 - `docbook-xsl` 1.79.2
 - `libxslt` 1.1.42
 - `man-db` 2.13.0

**Additional context**
It looks like the reason is that macros in man page are space indented after XSLT processing.

rdmark:
Thanks for the report! This is curious, because I haven't seen this bug. I'm testing on bleeding edge Debian Testing and macOS with Homebrew libraries. The library versions are literally the exactly the same as you have on Arch. (Except xsltproc is 1.1.35 on Debian).

I wonder if this is caused by system locale? What locale are you using?

singular0:
Ok, after additional research, the problem seems to be Arch specific. Here `docbook-xsl` installs two sets of stylesheets, `xsl-stylesheets` and `xsl-stylesheets-nons`. When compiled with the latter one, indentation spaces are omitted and correct man pages are produced. I could not find any information on this `nons` version right away, but I would probably just switch to it in the AUR package.

rdmark:
I'm pretty sure `nons` stands for "no namespace". Debian has the flipped dichotomy in its packages: `docbook-xsl` is non-namespaced, and `docbook-xsl-ns` is namespaced.

The DocBook maintainers made a bit of a mess when they moved hosting from SourceForge to GitHub, between release versions 1.79.1 and 1.79.2, which introduced namespaced XML and effectively broke compatibility with all legacy DocBook code. So we're of course stuck with having to distribute both downstream...

See https://github.com/Netatalk/netatalk/issues/318 and the linked bug in the xslt10-stylesheets project.

singular0:
I get mixed results on different Arch configurations, so I'm reopening this one until issue is localized properly

rdmark:
@singular0 Any update on your troubleshooting of this issue?

singular0:
[...]

PR #1801, opened 2024-12-04
docs: Trim indentation spaces from generated man pages

singular0:
Indentation spaces from XML were transferred to man pages which lead to incorrect rendering (#1800)


rdmark:
Do you think there could be a risk of side effects for this? The rule is applied to literally all elements if I understand it correctly, so the scope is big.

singular0:
After additional research, the problem seems to be Arch Linux specific.

PR #1806, opened 2024-12-07
meson: Refactor DocBook detection logic to stop when found

rdmark:
- `-Dwith-docbook-path` now overrides all default search paths
- Break when getting a first hit on DocBook stylesheets
- Detect `xsl-stylesheets-nons` with higher priority than `xsl-stylesheets` (Arch Linux compatibility)
- Report xsltproc and docbook-xsl results separately

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -135,7 +135,6 @@ jobs:
             -Dbuildtype=release \
             -Dwith-appletalk=true \
             -Dwith-dbus-sysconf-path=/usr/share/dbus-1/system.d \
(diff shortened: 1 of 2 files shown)


rdmark:
@singular0 Would you be able to test if docbook is correctly detected in your Arch environment after this refactor?

singular0:
@rdmark nope, i get mixed results here, still on it, will come back when issues will be definitively localized

rdmark:
@singular0 I see. Would it be detrimental to your effort if I merged this to main, for starters? I can see an improvement in our CI builds at least. The Arch build is no longer using the namespaced stylesheets, and all the other builds look good too.

You can of course propose further tweaks to docbook detection in your own PR.

PR #1813, opened 2024-12-11
Add native metadata storage for macOS hosts.

NJRoadfan:
-Store all extended attributes without the `user.` prefix on macOS. This will orphan Netatalk metadata created on past versions, but is needed for native extended attributes to work properly on shares. Unlike Linux, macOS doesn't prefix user created extended attributes with `user.`.

-Store resource forks natively by using the `/..namedfork/rsrc` path instead of creating AppleDouble files. No need for separate files since the file system supports it natively.

-Synchronize native FinderInfo (stored in `com.apple.FinderInfo`) with Netatalk. Any changes made to FinderInfo by Netatalk clients or by the host OS are now visible to each other. Note that the contents of `com.apple.FinderInfo` take priority of that over what is stored in Netatalk's extended attributes.

### DIFF
--- a/doc/manual/upgrade.xml
+++ b/doc/manual/upgrade.xml
@@ -129,7 +129,14 @@
           <listitem>
             <para>stores Mac Metadata (eg FinderInfo, AFP Flags, Comment,
             CNID) in an Extended Attributed named
-            “<filename>org.netatalk.Metadata</filename>”</para>
+            “<filename>org.netatalk.Metadata</filename>”<itemizedlist>
+                <listitem>
+                  <para>Additionally, on macOS hosts running Netatalk 4.1.0 or 
+                  later, FinderInfo is natively stored in the file system and
+                  appears as an Extended Attribute named
+                  “<filename>com.apple.FinderInfo</filename>”</para>
+                </listitem>
+              </itemizedlist></para>
           </listitem>
 
           <listitem>
@@ -143,7 +150,12 @@
                 <listitem>
                   <para>an extra AppleDouble file named
                   “<filename>._file</filename>” for a file named
-                  “<filename>file</filename>”</para>
+                  “<filename>file</filename>” or</para>
+                </listitem>
+
+                <listitem>
+                  <para>natively stored in the resource fork
+                  on macOS hosts as of Netatalk 4.1.0.</para>
                 </listitem>
               </itemizedlist></para>
           </listitem>
--- a/meson.build
+++ b/meson.build
@@ -1192,6 +1192,9 @@ elif cc.has_function('getxattr', dependencies: attr)
             cdata.set(define, 1)
         endif
     endforeach
+    if host_os == 'darwin'
+        cdata.set('HAVE_EAFD', 1)
+    endif
 else
     have_ea = false
 endif
--- a/doc/ja/manual/upgrade.xml
+++ b/doc/ja/manual/upgrade.xml
@@ -127,7 +127,13 @@
           <listitem>
             <para>Mac のメタデータ（すなわち FinderInfo、AFP フラグ、コメント、CNID）は
             “<filename>org.netatalk.Metadata</filename>”
-            という名前の拡張属性に保存される。</para>
+            という名前の拡張属性に保存される。<itemizedlist>
+                <listitem>
+                  <para>さらに、Netatalk 4.1.0以降を実行しているmacOSホストでは、FinderInfo
+                  はファイルシステムにネイティブに保存され、"com.apple.FinderInfo"
+                  という名前の拡張属性として表示されます。</para>
+                </listitem>
+              </itemizedlist></para>
           </listitem>
 
           <listitem>
@@ -142,6 +148,9 @@
                   <para>ないしは、ファイル名が “<filename>file</filename>”
                   であるものに対して各々、“<filename>._file</filename>” という名の別の
                   AppleDouble ファイルに保存される。</para>
+                </listitem>
+                <listitem>
+                  <para>Netatalk 4.1.0 以降、macOS ホスト上のリソース フォークにネイティブに保存されています。</para>
                 </listitem>
               </itemizedlist></para>
           </listitem>
--- a/etc/afpd/file.c
+++ b/etc/afpd/file.c
@@ -105,6 +105,18 @@ void *get_finderinfo(const struct vol *vol, const char *upath, struct adouble *a
             memcpy((char *)data + FINDERINFO_FRFLAGOFF, &ashort, sizeof(ashort));
         }
     }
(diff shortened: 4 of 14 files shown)


NJRoadfan:
Marked draft because this will need some testing. To see the power of this integration, download some Stuff-It archives on your macOS host and then extract them using The Unarchiver to your Netatalk share. Open up the folder in your share, and you should see your extracted files complete with their resource forks and proper file types/creator.

rdmark:
Would it be feasible to automatically convert the orphaned meta data to the new format, akin to how AppleDouble v2 is converted to EA when detected?

NJRoadfan:
Netatalk re-creates a new `org.netatalk.Metadata` EA from the CNID database. Any Finder Info would auto populate from `com.apple.FinderInfo` at that point.

rdmark:
Aha, so the orphaning process just means that there will be junk metadata files floating around on the host file system, correct?

If so, we should provide some cleanup facility. If not built into afpd, then as a remark in the release notes, and perhaps a simple shell script.

NJRoadfan:
The command `xattr -d -r user.org.netatalk.Metadata *` run at the root of a share should work.

NJRoadfan:
Test suite has been modified to reflect changes in Netatalk under macOS. I haven't tested it, but it should run without issue.

NJRoadfan:
Extended attributes should be working for macOS clients now. There is a subtle difference in how the getxattr() functions work under macOS. They will not return the size of an EA unless you point to a null buffer when calling. Otherwise the function fails if the buffer passed is too small. Linux doesn't care in these cases and will always return the EA size if it exists.

rdmark:
I added instructions to the wiki how to run the spectests locally.

https://github.com/Netatalk/netatalk/wiki/Testing#run-the-spectest-test-suite

These are written for Debian, but if you modify the user/group steps it should apply to macOS too. If you want to try to run the testsuite. :)

rdmark:
@NJRoadfan I got around to setting up and running the testsuite on my macOS Sequoia host. For the most part, it works well with these changes. However, exactly one test goes from passing to failing after this PR.

* FPOpenFork:test431: check AppleDouble conversion from v2 to ea - FAILED

As a side note, 8 other tests are failing on the macOS host, both before and after this PR.

* FPGetFileDirParms:test420: FPGetFileDirParms an open file is renamed with local fs - FAILED
* FPMoveAndRename:test302: file renamed by someone else, cnid not updated - FAILED
* FPResolveID:test331: Resolve ID file modified with local fs - FAILED
* FPResolveID:test360: Resolve ID file modified with local fs and a file is opened - FAILED
* FPResolveID:test412: Resolve ID file modified with local fs - FAILED
* FPResolveID:test413: Resolve ID file modified with local fs - FAILED
* Error:test102: access error but not cname - FAILED
* Error:test103: did access error - FAILED

NJRoadfan:
At least nothing broke with the change to the resource fork storage. I haven't tested adouble_v2 conversions yet, so not surprised.

There are still some issues that need to be dealt with in terms of using the native FinderInfo and files that were added to the share from outside of Netatalk (ie: you copy files from your downloads folder into a share). Right now, the calls only read the native FinderInfo if there is already a Netatalk AppleDouble EA for the file.

MacOS9 and OSX both only call FPGetFileDirParms when opening folders (OS8 also calls FPSetFileParams, creating the Netatalk EAs). Reading `com.apple.FinderInfo` will be needed during the calls, not just in the adouble library.

NJRoadfan:
Mostly there. Only major problem that remains is the conversion from AppleDouble v2 to AppleDouble EA isn't bringing over FinderInfo (leading to an afptest fail) for some reason. This may be due to other bugs running on macOS. Otherwise this is working 100% on classic MacOS.

The story for MacOS X clients is a bit more annoying. Unlike the old Finder, MacOS X doesn't update FinderInfo by just opening a folder as it uses .DS_Store instead. As a result, nothing immediately triggers creation of the `org.netatalk.Metadata` EA on files that were copied into the share from outside of Netatalk. Things generally work OK in this "halfway" state, but if you try to alter any FinderData (like add a color label), it will be immediately erased when Netatalk creates its EA and then re-read's the native macOS FinderInfo. Additionally, files lacking the `org.netatalk.Metadata` EA will not return the resource fork or FinderInfo when using `xattr` from a terminal.

Both of these likely can be fixed, although it would likely incur a performance hit with large directories.

NJRoadfan:
Converting from adouble_v2 to adouble_ea should work now. The one afptest fail should now be resolved.

rdmark:
Yep, it passes now.

```
FPOpenFork:test431: check AppleDouble conversion from v2 to ea - PASSED
```

Would you mind rebasing this PR branch with the latest main? It would be nice to see if the entire suite passes cleanly now.

NJRoadfan:
Rebased and updated documention to reflect the changes for macOS hosts.

doc/manual/upgrade.xml
```diff
@@ -129,7 +129,14 @@
...
             CNID) in an Extended Attributed named
-            “<filename>org.netatalk.Metadata</filename>”</para>
+            “<filename>org.netatalk.Metadata</filename>”<itemizedlist>
+                <listitem>
+                  <para>Additionally, on macOS hosts, FinderInfo is 
+                  natively stored in the file system and appears as an
+                  Extended Attribute named
+                  “<filename>com.apple.FinderInfo</filename>”</para>
+                </listitem>
+              </itemizedlist></para>
```
rdmark, line 139: I suggest we qualify this statement with something like "As of Netatalk 4.1.0..."

doc/manual/upgrade.xml
```diff
@@ -143,7 +150,12 @@
                 <listitem>
                   <para>an extra AppleDouble file named
                   “<filename>._file</filename>” for a file named
-                  “<filename>file</filename>”</para>
+                  “<filename>file</filename>” or</para>
+                </listitem>
+
+                <listitem>
+                  <para>natively stored in the resource fork
+                  on macOS hosts.</para>
```
rdmark, line 158: Same as the above.

meson.build
```diff
@@ -1196,6 +1196,10 @@ else
     have_ea = false
 endif
 
+if host_os == 'darwin'
+    cdata.set('HAVE_EAFD', 1)
+endif
```
rdmark, line 1201: There's a block starting on line 1120 with all of the platform specific conditions. Maybe you can merge these?

rdmark:
Test suite passes cleanly now.

```
===================
TEST RESULT SUMMARY
-------------------
  Passed:     285
  Failed:     0
  Skipped:    22
  Not tested: 0
```

rdmark (approved):

PR #1818, opened 2024-12-21
Update atalkd.conf.5.xml

thecloudexpanse:
Zone names should be enclosed in quotation marks, not parentheses.

### DIFF
--- a/doc/manpages/man5/atalkd.conf.5.xml
+++ b/doc/manpages/man5/atalkd.conf.5.xml
@@ -149,7 +149,7 @@
           <para>Specifies a specific zone that this interface should appear on
           (example: <option>-zone "Parking Lot"</option>). Please note that
           zones with spaces and other special characters should be enclosed in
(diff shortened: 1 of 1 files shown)


thecloudexpanse:
atalkd.conf man page incorrectly says to enclose zone names in parentheses. They should actually be enclosed in quotation marks. This is only a documentation issue, but it was confusing as I was setting up atalkd with zones for the first time.

rdmark:
Thanks for your contribution!

If you want to challenge yourself, the Japanese localization (doc/ja) needs the fix as well. It should only be a matter of replacing the word for parentheses - 括弧 - with the word ダブルクォーテーション. :)

rdmark (approved):

rdmark:
Japanese localization updated with https://github.com/Netatalk/netatalk/commit/682ac70d40254b25641e8dad2b26877093f33325

PR #1822, opened 2024-12-24
meson: Introduce with-kerberos-path option for Heimdal compatibility

rdmark:
Heimdal's krb5 compatibility headers, shared libraries and binaries lives in `/usr/local/heimdal` on (at least) OpenBSD. This introduces a `-Dwith-kerberos-path` that can be used to get Meson to look for dependencies in this location.

Additionally, introduced a break when finding first krb.h header which avoids including broken `krb5/krb5.h` header on NetBSD.

With these changes, the krbV UAM can now be built on NetBSD, OpenBSD, and openSUSE

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -296,6 +296,7 @@ jobs:
             flex \
             gawk \
             gcc \
+            krb5-devel \
             libacl-devel \
             libavahi-devel \
             libdb-4_8-devel \
@@ -584,7 +585,6 @@ jobs:
               -Dbuildtype=release \
               -Dwith-appletalk=true \
               -Dwith-dtrace=false \
-              -Dwith-krbV-uam=false \
               -Dwith-tests=true
             meson compile -C build
             cd build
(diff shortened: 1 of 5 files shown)


rdmark:
@hfath Heads-up: With this change, you can now build the krbV UAM on NetBSD via the heimdal libraries. On NetBSD the main problem seems to have been that the `krb5/krb5.h` header is either vestigial and should not be used, or has to be used in a particular context. I.e. this error:

```
  In file included from ../etc/afpd/status.c:16:
  /usr/pkg/include/krb5/krb5.h:42:10: fatal error: krb5-types.h: No such file or directory
     42 | #include <krb5-types.h>
        |          ^~~~~~~~~~~~~~
  compilation terminated.
```

With this PR, the `/usr/pkg/include/krb5.h` header will be detected first, then we break out of the loop so that `/usr/pkg/include/krb5/krb5.h` won't be detected and used instead.

Please let me know if there's a more correct way to do this in NetBSD.

PR #1836, opened 2025-01-02
macipgw: Default port value for zip/ddp service

APCCV:
macipgw is not starting in OpenWrt as getservbyname() is not returning a hit for zip/ddp service and that is a terminal error in macipgw.
While the reasons why getservbyname() are failing should probably be addressed, most of the other components of netatalk default port numbers to known values if getservbyname() fails.
This change copies code working OK from getzones.


APCCV:
Bummer.... Sorry for the overflow in commit title.
If that's an issue just let me know here and I'll rebase it.

APCCV:
As pointed out in issue discussion, musl (and it's getservbyname() implementation) doesn't support anything other than "udp" or "tcp".

rdmark (approved):
LGTM. Thanks!

PR #1841, opened 2025-01-03
Loosen AppleDouble checks for DragonflyBSD and OpenBSD

rdmark:
Applying the fix for macOS in https://github.com/Netatalk/netatalk/pull/1829


NJRoadfan:
I just realized the changes made to fix the problem on macOS might cause problems with servers running both Samba and Netatalk. The vfs_fruit module use the [exact same code](https://github.com/samba-team/samba/blob/master/source3/lib/adouble.c#L280) to verify AppleDouble file headers. It would obviously fail a check in `ad_entry_check_size()` if the length of a PRIVDEV field is 4 vs. 8. As far as I know, OpenBSD and DragonflyBSD are supported Samba platforms. DragonflyBSD likely can't use vfs_fruit due to the lack of extended attribute support (is this still the case? Netatalk disables support) so likely wouldn't be a problem. Pinging @slowfranklin

rdmark:
Discarding this for now until we can confirm the interoperability with Samba.

PR #1842, opened 2025-01-04
docker: Bump base image to Alpine 3.21

rdmark:
Upgrade from Alpine 3.20 to 3.21

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,6 @@
 ARG RUN_DEPS="\
     acl \
     avahi \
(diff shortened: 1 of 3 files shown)


rdmark:
The transition from tracker to localsearch seems to have introduced this error:

```
(localsearch-3:34): Tracker-WARNING **: 06:33:59.614: Could not launch metadata extractor: Failed to execute child process “/usr/libexec/localsearch-extractor-3” (Permission denied)
```

PR #1849, opened 2025-01-05
Remove the ability to run AFP commands with root privileges

rdmark:
This modifies the afprun() function to execute shell commands only as the current user, not as root.

Removing the following runtime options:

- root preexec
- root postexec
- root preexec close


rdmark:
While testing, I noticed that in the macOS (Sequoia) client, both the preexec and postexec commands are executed the moment you authenticate and get the shared volume listings. Then the preexec gets executed again when you open the specific volume, and postexec executed when you disconnect the same volume.

PR #1850, opened 2025-01-05
Define all lockfiles through the Meson build system

rdmark:
Repurpose the `with-lockfile-path` meson option as the root path to the dir that houses all three lockfiles (netatalk, papd, atalkd).

Move a handful of papd path definitions from C code to the build system.

Retire the paths.h shared header.


rdmark:
@NJRoadfan I noticed that the spooldir paths for CUPS and lp and defined differently. Should they be consolidated into a single definition?

NJRoadfan:
They should remain separate. CUPS generally defaults to `/var/spool/cups` when its API is used to process a print job, while lpd has a standard definition (FHS) of `/var/spool/lpd`. If CUPS support isn't present in papd, spool files are sent to `/var/spool/lpd`.

rdmark:
In my testing on NetBSD, Alpine and Ubuntu, this works correctly now, creating the correct lockfiles in the correct places.

I have not tested papd printing end to end though, since I don't have my retro gear handy.

PR #1852, opened 2025-01-06
macipgw: Introduce a configuration file

rdmark:
Makes the macipgw daemon look for a macipgw.conf file in the default netatalk pkgconf path.  Without the config file, macipgw should behave as before.

Point to conf file in custom location with the `-f` terminal option.

Sample macipgw.conf file:

```
[Global]
network = 192.168.151.0
netmask = 255.255.255.0
nameserver = 8.8.8.8
zone = My Zone
unprivileged user = dmark
```

This changeset also fixes a few type safety issues.

### DIFF
--- a/config/macipgw.conf
+++ b/config/macipgw.conf
@@ -1,0 +1,5 @@
+# MacIP Gateway daemon configuration (netatalk 4.x)
+[Global]
+network = 192.168.151.0
+netmask = 255.255.255.0
+nameserver = 8.8.8.8
--- a/config/meson.build
+++ b/config/meson.build
@@ -44,7 +44,7 @@ endif
 static_conf_files = ['extmap.conf']
 
 if have_appletalk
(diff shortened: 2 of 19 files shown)


NJRoadfan:
One thing to keep in mind, Netatalk ships with a full INI file parser in `libatalk/iniparser`.

rdmark:
Yes, I tried using iniparser a few months ago. The learning curve was steep for such a simple usecase. I found the process of rolling my own solution more interesting. :)

rdmark:
Strike that; Using iniparser was a lot easier than I thought at first glance. It's been done.

The main drawback that I see right now is that our bundled fork of iniparser is modified to expect a [Global] section. It's arguably awkward for macipgw.conf to have a [Global] section because there are only global options. I don't really want to modify the bundled iniparser code even more, either. Makes it even harder to upgrade the library with upstream releases in the future.

PR #1855, opened 2025-01-06
Protect Netatalk metadata EA

NJRoadfan:
Currently, a client can create new files on the server with the same name as the Netatalk metadata extended attribute (`org.netatalk.Metadata`). If this EA has valid AppleDouble data (ex: from another Netatalk host), the server will treat the EA as read-only and not be able to update the data.

Additionally clients will not be able to update metadata on the file such as comments or FinderInfo. The only way to fix the problem is deleting the extended attribute on the server side, or deleting the file.

### DIFF
--- a/include/atalk/ea.h
+++ b/include/atalk/ea.h
@@ -88,6 +88,7 @@ enum {
 
 /* Names for our Extended Attributes adouble data */
 #define AD_EA_META "org.netatalk.Metadata"
+#define AD_EA_META_LEN (sizeof (AD_EA_META) - 1)
 #ifdef __APPLE__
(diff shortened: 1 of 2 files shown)


rdmark (approved):

rdmark:
If I understand this fix correctly, now we validate that metadata that calls itself "org.netatalk.Metadata" actually contains the EA `org.netatalk.Metadata`? So if you, say, create a file called "org.netatalk.Metadata" and then add the `org.netatalk.Metadata` EA to it, we still get unwanted behavior?

NJRoadfan:
The PR prevents an end user from tampering with or reading the metadata EA by using getxattr()/setxattr() on a client machine.

PR #1856, opened 2025-01-07
docs: Fixes for spelling and grammar

EricFromCanada:
Found while reading the docs and running `hunspell` against the source files.

### DIFF
--- a/doc/manual/configuration.xml
+++ b/doc/manual/configuration.xml
@@ -313,7 +313,7 @@ basedir regex = /usr/home</programlisting></para>
       <emphasis>multibyte</emphasis> character sets.</para>
 
       <para>One standardized multibyte charset encoding scheme is known as
-      <ulink url="http://www.unicode.org/">unicode</ulink>. A big advantage of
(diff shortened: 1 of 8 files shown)


doc/manual/configuration.xml
```diff
@@ -1165,7 +1165,7 @@ uams_dhx2.so -&gt; uams_dhx2_pam.so</programlisting> then you're using PAM,
 aclmode = passthrough</screen>
 
           <para>For an explanation of what this knob does and how to apply it,
-          check your hosts ZFS documentation (eg man zfs).</para>
+          check your hosts ZFS documentation (e.g man zfs).</para>
```
rdmark, line 1168: Nitpick: This should also have the trailing period. "e.g" -> "e.g."

rdmark (approved):
Thanks for contributing. Some of the typos may have been in place since the late 90s. ;)

doc/manual/configuration.xml
EricFromCanada: Thanks, fixed.

PR #1857, opened 2025-01-08
iconv fixes for OpenWrt

APCCV:
Addresses a build issue with cross-compiling with iconv as meson cannot run code to determine if INTERNAL iconv support is available.
Addresses linking issues when iconv is supported by libiconv (detected or with path provided) instead of the more common libc6 implementation in Linux (maybe MacOS and others too).

### DIFF
--- a/bin/aecho/meson.build
+++ b/bin/aecho/meson.build
@@ -1,8 +1,15 @@
+aecho_deps = []
+
+if have_iconv
+    aecho_deps += iconv
+endif
+
 executable(
     'aecho',
     'aecho.c',
     include_directories: root_includes,
     link_with: libatalk,
+    dependencies: aecho_deps,
     install: true,
     build_rpath: rpath_libdir,
     install_rpath: rpath_libdir,
--- a/bin/getzones/meson.build
+++ b/bin/getzones/meson.build
@@ -1,8 +1,15 @@
+getzones_deps = []
+
+if have_iconv
+    getzones_deps += iconv
+endif
+
 executable(
     'getzones',
     'getzones.c',
     include_directories: root_includes,
     link_with: libatalk,
+    dependencies: getzones_deps,
     install: true,
     build_rpath: rpath_libdir,
     install_rpath: rpath_libdir,
--- a/bin/misc/meson.build
+++ b/bin/misc/meson.build
@@ -6,12 +6,18 @@ fce_sources = ['fce.c']
 
 afpldaptest_sources = ['uuidtest.c']
 
+misc_deps = []
+
+if have_iconv
+    misc_deps += iconv
+endif
+
 executable(
     'netacnv',
     netacnv_sources,
     include_directories: root_includes,
     link_with: libatalk,
-    dependencies: mysqlclient,
+    dependencies: [mysqlclient, misc_deps],
     install: false,
 )
 
@@ -20,7 +26,7 @@ executable(
     loggertest_sources,
     include_directories: root_includes,
     link_with: libatalk,
-    dependencies: mysqlclient,
+    dependencies: [mysqlclient, misc_deps],
     install: false,
 )
 
@@ -29,7 +35,7 @@ executable(
     fce_sources,
     include_directories: root_includes,
     link_with: libatalk,
-    dependencies: mysqlclient,
+    dependencies: [mysqlclient, misc_deps],
     install: false,
 )
 
@@ -45,7 +51,7 @@ if have_ldap
         afpldaptest_sources,
         include_directories: root_includes,
         link_with: libatalk,
-        dependencies: [afpldaptest_deps, ldap],
+        dependencies: [afpldaptest_deps, ldap, misc_deps],
         c_args: confdir,
         install: true,
         build_rpath: rpath_libdir,
--- a/bin/nbp/meson.build
+++ b/bin/nbp/meson.build
@@ -1,8 +1,15 @@
+npb_deps = []
+
+if have_iconv
+    npb_deps += iconv
+endif
+
 executable(
     'nbplkup',
     'nbplkup.c',
     include_directories: root_includes,
     link_with: libatalk,
+    dependencies: npb_deps,
     install: true,
     build_rpath: rpath_libdir,
     install_rpath: rpath_libdir,
@@ -12,6 +19,7 @@ executable(
     'nbprgstr.c',
     include_directories: root_includes,
     link_with: libatalk,
+    dependencies: npb_deps,
     install: true,
     build_rpath: rpath_libdir,
     install_rpath: rpath_libdir,
@@ -21,6 +29,7 @@ executable(
     'nbpunrgstr.c',
     include_directories: root_includes,
     link_with: libatalk,
+    dependencies: npb_deps,
     install: true,
     build_rpath: rpath_libdir,
     install_rpath: rpath_libdir,
--- a/bin/pap/meson.build
+++ b/bin/pap/meson.build
@@ -1,8 +1,15 @@
+pap_deps = []
+
+if have_iconv
+    pap_deps += iconv
+endif
+
 executable(
     'pap',
     'pap.c',
     include_directories: root_includes,
     link_with: libatalk,
+    dependencies: pap_deps,
     install: true,
     build_rpath: rpath_libdir,
     install_rpath: rpath_libdir,
@@ -12,6 +19,7 @@ executable(
     'papstatus.c',
     include_directories: root_includes,
     link_with: libatalk,
+    dependencies: pap_deps,
     install: true,
     build_rpath: rpath_libdir,
     install_rpath: rpath_libdir,
--- a/contrib/a2boot/meson.build
+++ b/contrib/a2boot/meson.build
@@ -1,3 +1,9 @@
+a2boot_deps = []
+
+if have_iconv
+    a2boot_deps += iconv
+endif
+
 a2boot_c_args = [
     '-D_PATH_A_GS_BLOCKS="'
     + pkgconfdir
@@ -14,6 +20,7 @@ executable(
     'a2boot.c',
     include_directories: root_includes,
     link_with: libatalk,
+    dependencies: a2boot_deps,
     c_args: a2boot_c_args,
     install: true,
     install_dir: sbindir,
--- a/contrib/macipgw/meson.build
+++ b/contrib/macipgw/meson.build
@@ -1,3 +1,9 @@
+macipgw_deps = []
+
+if have_iconv
+    macipgw_deps += iconv
+endif
+
 macipgw_sources = [
 	'atp_input.c',
 	'macip.c',
@@ -17,6 +23,7 @@ executable(
     macipgw_sources,
     include_directories: [root_includes, macipgw_includes],
     link_with: libatalk,
+    dependencies: macipgw_deps,
     c_args: macipgw_c_args,
     export_dynamic: true,
     install: true,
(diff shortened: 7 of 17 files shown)


APCCV:
Tested working for basic file sharing and atalkd functionality.

Binaries are linked to libiconv as expected:
```
root@TestWrt-63:~# ldd /usr/sbin/afpd
	/lib/ld-musl-armhf.so.1 (0xb6f0a000)
	libatalk.so.19 => /usr/lib/libatalk.so.19 (0xb6ea8000)
	libgcc_s.so.1 => /lib/libgcc_s.so.1 (0xb6e9b000)
	libc.so => /lib/ld-musl-armhf.so.1 (0xb6f0a000)
	libiconv.so.2 => /usr/lib/libiconv.so.2 (0xb6dbd000)
```

APCCV:
Tested build in Ubuntu. Succeeds.
Tested few binaries from Ubuntu build. Work OK.
While Ubuntu build was configured with -Dwith-libiconv=yes and configuration reports it as a user option, unsure how to test it. Same meson config behaviour in Linux than in OpenWrt.
```
Dependency iconv found: YES unknown (cached)
Library iconv found: NO
```
```
  User defined options
    with-appletalk             : true
    with-libiconv              : true
```
It seems to have built w/ iconv support in Linux... links vs libc6 which has internal iconv support:
```
me@here:/home/work/netatalk/netatalk$ ldd ./build//etc/atalkd/atalkd
	linux-vdso.so.1 (0x00007ffd50b82000)
	libatalk.so.19 => /home/work/netatalk/netatalk/./build//etc/atalkd/../../libatalk/libatalk.so.19 (0x000072e2b83f6000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x000072e2b8000000)
	/lib64/ld-linux-x86-64.so.2 (0x000072e2b849b000)
```

meson_log shows all compiles and executes happy:
```
Using cached compile:
Cached command line:  cc /home/work/netatalk/netatalk/build/meson-private/tmpdik8lial/testfile.c -o /home/work/netatalk/netatalk/build/meson-private/tmpdik8lial/output.obj -c -D_FILE_OFFSET_BITS=64 -O0 -std=c11

Code:

    #include <iconv.h>
    int main(void) {
        iconv_t cd = iconv_open("ASCII", "UTF-8");
        if (cd == 0 || cd == (iconv_t)-1) return -1;
    return 0;
    }
    
Cached compiler stdout:
 
Cached compiler stderr:

Using cached compile:
Cached command line:  cc /home/work/netatalk/netatalk/build/meson-private/tmpmb7lpszg/testfile.c -o /home/work/netatalk/netatalk/build/meson-private/tmpmb7lpszg/output.obj -c -D_FILE_OFFSET_BITS=64 -O0 -std=c11

Code:

        #include <stdlib.h>
        #include <iconv.h>
        extern
        #ifdef __cplusplus
            "C"
        #endif
        #if defined(__STDC__) || defined(__cplusplus)
            size_t iconv (iconv_t cd, char * *inbuf, size_t *inbytesleft, char * *outbuf, size_t *outbytesleft);
        #else
            size_t iconv();
        #endif

Cached compiler stdout:

Cached compiler stderr:

Running compile:
Working directory:  /home/work/netatalk/netatalk/build/meson-private/tmpxegbv9p4
Code:

                #include <iconv.h>
                int main(void) {
                    iconv_t cd = iconv_open("ASCII", "UCS-2-INTERNAL");
                    if (cd == 0 || cd == (iconv_t)-1) return -1;
                    return 0;
                }
```

bin/aecho/meson.build
```diff
@@ -1,8 +1,15 @@
+iconv_deps = []
```
rdmark, line 1: This variable should be named `aecho_deps` to align with our Meson naming conventions.

bin/getzones/meson.build
```diff
@@ -1,8 +1,15 @@
+iconv_deps = []
```
rdmark, line 1: Rename this `getzones_deps`

bin/misc/meson.build
```diff
@@ -6,12 +6,18 @@ fce_sources = ['fce.c']
 
 afpldaptest_sources = ['uuidtest.c']
 
+iconv_deps = []
```
rdmark, line 9: Rename `misc_deps`

bin/nbp/meson.build
```diff
@@ -1,8 +1,15 @@
+iconv_deps = []
```
rdmark, line 1: `npb_deps`

bin/pap/meson.build
```diff
@@ -1,8 +1,15 @@
+iconv_deps = []
```
rdmark, line 1: `pap_deps`

contrib/a2boot/meson.build
```diff
@@ -1,3 +1,9 @@
+iconv_deps = []
```
rdmark, line 1: `a2boot_deps`

contrib/macipgw/meson.build
```diff
@@ -1,3 +1,9 @@
+iconv_deps = []
```
rdmark, line 1: `macipgw_deps`

contrib/timelord/meson.build
```diff
@@ -1,8 +1,15 @@
+iconv_deps = []
```
rdmark, line 1: `timelord_deps`

etc/atalkd/meson.build
```diff
@@ -11,11 +11,18 @@ atalkd_sources = [
 
 atalkd_c_args = ['-D_PATH_ATALKDCONF="' + pkgconfdir + '/atalkd.conf"', dversion]
 
+iconv_deps = []
```
rdmark, line 14: `atalkd_deps`

etc/cnid_dbd/meson.build
```diff
@@ -23,6 +23,12 @@ if use_dbd_backend
         cnid_dbd_deps += mysqlclient
     endif
 
+    iconv_deps = []
```
rdmark, line 26: `cnid_dbd_deps` is already defined, so please reuse that!

etc/papd/meson.build
```diff
@@ -32,12 +32,18 @@ papd_c_args = [
     dversion,
 ]
 
+iconv_deps = []
```
rdmark, line 35: `papd_deps`

etc/spotlight/meson.build
```diff
@@ -14,6 +14,12 @@ run_command(
 
 srp_sources = ['sparql_map.c', 'sparql_parser.c', 'spotlight_rawquery_lexer.c']
 
+iconv_deps = []
```
rdmark, line 17: `spotlight_deps`

rdmark (changes requested):

bin/aecho/meson.build
APCCV: Done

bin/getzones/meson.build
APCCV: Done

bin/misc/meson.build
APCCV: Done

bin/nbp/meson.build
APCCV: done

bin/pap/meson.build
APCCV: Done

contrib/timelord/meson.build
APCCV: Done

etc/atalkd/meson.build
APCCV: Done

etc/cnid_dbd/meson.build
APCCV: Done

etc/papd/meson.build
APCCV: Done

APCCV: Done

etc/spotlight/meson.build
APCCV: Done

contrib/a2boot/meson.build
APCCV: Done

contrib/macipgw/meson.build
APCCV: Done

APCCV:
This seems to be the only side effect. In MacOS build (from GitHub CI logs):
```
[94/274] Linking target libatalk/libatalk.19.dylib
ld: warning: ignoring duplicate libraries: '-liconv'
```

APCCV:
Alright. Got rid of most warnings but for MacOS libatalk linking.
I took out the change to meson.build for libatalk altogether (guess should've thought of it earlier, but a dynamic library doesn't need to link to another dynamic library) so that warning must have been there before I touched this.

I can't claim I fully understand meson as somehow it picks up the path from one dependency and applies to another. Whatever. Change is clean and simple now, and code now have *_deps variable for many more executables.

rdmark:
I would prefer to avoid introducing a linker warning like that. A warning can turn into an error in the future. :)

...and as I was typing up this comment I see that you resolved the warning. Great job!

rdmark (approved):

APCCV:
Yay!

Thank you @NJRoadfan for the solution to check if code is being cross-compiled.

meson.build
```diff
@@ -1515,17 +1515,19 @@ else
...
-        #include <iconv.h>
-        int main(void) {
-            iconv_t cd = iconv_open("ASCII", "UCS-2-INTERNAL");
-            if (cd == 0 || cd == (iconv_t)-1) return -1;
-            return 0;
-        }
-        ''',
-        ).returncode() == 0
-            cdata.set('HAVE_UCS2INTERNAL', 1)
+        if not meson.is_cross_build()
```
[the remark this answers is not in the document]
rdmark: Thank you for the suggestion. Meson and cross compiling is new territory for me. Let me read up on it. 

If you know off the top of your head the modifications required, we do very much appreciate patches / PRs!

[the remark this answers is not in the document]
APCCV: I'm no expert either... heh... I have no clue even on a good day, but it seems like a 1:1 replacement and less restrictive than is_cross_build(). 
I'll give it a try and PR if it works for Ubuntu (non-cross compile) and OpenWrt (cross compile w/no chance ever of host running the test). The middle ground someone else will have to test.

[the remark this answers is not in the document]
eli-schwartz: Yes, it should be a 1:1 replacement.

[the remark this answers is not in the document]
rdmark: Filed https://github.com/Netatalk/netatalk/issues/1921 so that it doesn't get forgotten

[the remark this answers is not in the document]
APCCV: Sorry... I've been nursing the PRs to get 4.1.1 into Snapshot and the future release of OpenWrt. Wanted to add a bunch of admin stuff to them. Good news, that's finally done!

I just need to finalize getting completely frustrated with another OpenWrt issue... and then will poke at this.

I can test it in Ubuntu, and I can test in OpenWrt... but the qemu scenario I'll have to leave to someone else. 
@eli-schwartz is that something you can test?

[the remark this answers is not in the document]
eli-schwartz: Unfortunately, I'm not personally a user of cross builds. I just know that this is a common approach and that "people do it" and that the overall theory being employed here is relied on by others.

Void Linux makes use of our qemu integration, but they don't package netatalk specifically.
https://github.com/void-linux/void-packages/blob/73a9b3dfc9132dd5e6ad1d3a63ba932bea7442f0/common/build-helper/meson.sh#L47-L59

PR #1858, opened 2025-01-08
Refactor retreival of native FinderInfo EA on macOS hosts.

NJRoadfan:
Check size of `com.apple.FinderInfo` EA to verify it exists before returning.

### DIFF
--- a/libatalk/adouble/ad_open.c
+++ b/libatalk/adouble/ad_open.c
@@ -992,22 +992,24 @@ static int ad2openflags(const struct adouble *ad, int adfile, int adflags)
 #ifdef __APPLE__
 int ad_open_native_finderinfo(const char *path, char *ret)
 {
+    ssize_t size;
     LOG(log_debug, logtype_ad,
         "ad_open_native_finderinfo(\"%s\"): BEGIN",
         fullpathname(path));
-    if (sys_getxattr(path,EA_FINFO,ret,ADEDLEN_FINDERI))
-    {
+    size = sys_getxattr(path,EA_FINFO,0,0);
+    if (size < 0) {
         LOG(log_debug, logtype_ad,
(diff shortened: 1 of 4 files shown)


rdmark:
@NJRoadfan Does this address the Mac OS 9.0.x bug that you were observing? That FinderInfo isn't updated if TYPE is not set.

NJRoadfan:
The only way around that bug is to have the server force a file type on any file that does not have that information. This isn't ideal for many reasons. Netatalk used to do it, but was removed when MacOS X and AFP 3.0 became mainstream, likely because it was creating FinderInfo for files that didn't need it. Modern macOS still looks at FinderInfo file type info to associate files if the extension is missing or not associated with an application in its database.

rdmark:
Would activating the file type mappings in extmap.conf work around this, for file types that has a mapping that is?

libatalk/adouble/ad_open.c
```diff
@@ -992,22 +992,24 @@ static int ad2openflags(const struct adouble *ad, int adfile, int adflags)
...
         LOG(log_debug, logtype_ad,
-            "ad_open_native_finderinfo(\"%s\"): INFO FOUND",
+            "ad_open_native_finderinfo(\"%s\"): Native macOS FinderInfo not found",
             fullpathname(path));
-        return 1;
+        return 0;
     }
     else
     {
+        sys_getxattr(path,EA_FINFO,ret,32);
```
rdmark, line 1008: Could we define a macro for "32" to avoid having a magic number in the code?

NJRoadfan:
This PR actually allows `extmap.conf` to work again. On MacOS 8-9, some applications do support automatically associating themselves to file extensions. SimpleText does it for `.txt`, while Quicktime does it for various image and video formats.

rdmark (approved):
Thanks for resolving the magic numbers elsewhere too!

PR #1859, opened 2025-01-08
launchd item: Start netatalk in non-forking mode

EricFromCanada:
An issue with the current implementation of the launchd item is that unloading or stopping the it (e.g by running `sudo brew services stop netatalk`) doesn't actually stop the `netatalk` process. This is because the launchd is running the equivalent of `sudo netatalkd start` (which starts `netatalk` (which itself starts `afpd` and `cnid_metad`)) and immediately exits. 

Since Apple's [docs for processes managed by launchd](https://developer.apple.com/library/archive/documentation/MacOSX/Conceptual/BPSystemStartup/Chapters/CreatingLaunchdJobs.html) specify that they be non-forking, a workaround is to invoke the `netatalk` process directly with its `-d` flag. When the launchd item is stopped, `netatalk` is ended along with `afpd` and `cnid_metad`.

### DIFF
--- a/distrib/initscripts/macos.netatalk.plist.in
+++ b/distrib/initscripts/macos.netatalk.plist.in
@@ -11,8 +11,8 @@
 	<string>io.netatalk.daemon</string>
 	<key>ProgramArguments</key>
 	<array>
-		<string>@bindir@/netatalkd</string>
-		<string>start</string>
+		<string>@sbindir@/netatalk</string>
+		<string>-d</string>
 	</array>
 	<key>RunAtLoad</key>
 	<true/>


rdmark (approved):
Thanks for your contribution!

EricFromCanada:
This came up during the development of the [`netatalk` Homebrew formula](https://github.com/Homebrew/homebrew-core/pull/2027130). Essentially, formulae can define services that can be started and stopped with the `brew services` command, which behind the scenes is just (on macOS) running `launchctl` commands like the ones above.

This change adjusts how the `io.netatalk.daemon` launchd item starts netatalk by ensuring it's done without forking, so that when the launchd item is stopped, then the netatalk process and its subprocess stop as well; this wasn't happening originally. 

The `netatalkd` script still serves as a handy way to start and stop netatalk on-demand in the shell, but it's actually written to be a [StartupItem](https://developer.apple.com/library/archive/documentation/MacOSX/Conceptual/BPSystemStartup/Chapters/StartupItems.html), which is how user-level background processes were managed in early versions of OS X until being replaced by launchd in 10.4 Tiger and eventually phased out.

rdmark:
As an historical view: The original purpose of the `netatalkd` script was to activate the `OBJC_DISABLE_INITIALIZE_FORK_SAFETY` flag and stop macOS frameworks from instantly killing netatalk upon launch. We figured out recently how to avoid linking with Objective-C libraries, so this workaround is no longer needed.

I agree that losing lockfile housekeeping is a drawback. In systemd, if you point to the lockfile with the `PIDFile=` option, systemd does the housekeeping for you. I wonder if launchd has a similar facility?

PR #1885, opened 2025-01-14
Add a security patching end-of-life policy

rdmark:
Create an official policy of 12 month patch cycle after superseding release.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -6,16 +6,18 @@ The reporter of an accepted and patched vulnerability will be given credit in th
 
 ## Supported Versions
 
-This table indicates the Netatalk release series that are current guaranteed to get security patches.
+This table indicates the Netatalk release series that are currently guaranteed to get security patches.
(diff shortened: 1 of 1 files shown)


rdmark:
@NJRoadfan What do you think about this policy? I want to have something that signals clearly EOL for legacy release series, notably 3.1.x.

The notable distros that still ship 3.1.x are Alpine and Gentoo. I want to start nudging / supporting them to upgrade.

NJRoadfan:
Seems reasonable. Maybe there is a reason they still ship 3.1.x?

rdmark:
The [Gentoo package](https://packages.gentoo.org/packages/net-fs/netatalk) is flagged as abandoned so that would be a reason. The [Alpine package](https://pkgs.alpinelinux.org/package/edge/community/x86/netatalk) was updated to 3.1.19 a month ago. I can reach out to the package maintainer and offer support.

Issue #1887, opened 2025-01-15
macipgw segmentation fault when config file is missing


APCCV:
**Describe the bug**
Attempting to start macipgw results in a segmentation fault

**To Reproduce**
Execute macipgw in any way or form. By itself. With flags. With bogus arguments. Doesn't matter.

**Expected behavior**
It should run.

**Environment**
 - Server OS: OpenWrt snapshot
 - Client OS: N/A
 - Netatalk Version: 4.1.0
 
**Logs**
Nothing shows on logs.

**Additional context**
Reverting commit ec796ad (macipgw: Introduce a configuration file) restores normal behaviour.
OpenWrt scripts are not aware of the macipgw.conf file so therefore it doesn't exist my OpenWrt install. _Maybe_ macipgw will not fault if the file exists. Didn't get to test that, but given that it faults immediately on startup the problem is something else.

APCCV:
Yep... if /etc/macipgw.conf is missing, it segfaults.
Wish I had tried that earlier instead of doing all the nonsense I did.

Would something like this be OK with you?

```
From 54c4a09b54e8676f9e190940ceb239b11b8eb52b Mon Sep 17 00:00:00 2001
From: Antonio Pastor <antonio.pastor@gmail.com>
Date: Tue, 14 Jan 2025 23:41:26 -0500
Subject: [PATCH] macipgw: make use of config file optional

---
 contrib/macipgw/main.c | 56 ++++++++++++++++++++++--------------------
 1 file changed, 30 insertions(+), 26 deletions(-)

diff --git a/contrib/macipgw/main.c b/contrib/macipgw/main.c
index 91d9a344..a4c3a122 100644
--- a/contrib/macipgw/main.c
+++ b/contrib/macipgw/main.c
@@ -173,6 +173,8 @@ macip_options * read_options(const char *conf)
 
     config = atalk_iniparser_load(conf);
 
+    if (config == NULL)
+        return NULL;
     options->network = atalk_iniparser_getstrdup(config, INISEC_GLOBAL, "network", "");
     options->netmask = atalk_iniparser_getstrdup(config, INISEC_GLOBAL, "netmask", "");
     options->nameserver = atalk_iniparser_getstrdup(config, INISEC_GLOBAL, "nameserver", "");
@@ -247,32 +249,34 @@ int main(int argc, char *argv[])
 
 	mio = read_options(conffile);
 
-	if ((mio->network[0] != '\0') && (net == 0)) {
[...]

PR #1891, opened 2025-01-16
macipgw: Handle scenario when config file is missing

APCCV:
config file is optional with this change.

Addresses https://github.com/Netatalk/netatalk/issues/1887


rdmark (approved):
Thanks for the fix!

PR #1894, opened 2025-01-17
Prevent illegal null pointer calls

rdmark:
Addressing a number of code quality issues raised by Sonarqube


rdmark:
@NJRoadfan This prevents some hypothetical fail states in atalkd and papd. They look safe to me, but let me know if I broke something.

PR #1905, opened 2025-01-19
Convert documentation from XML to Markdown

rdmark:
Leveraged pandoc's docbook filter to convert to CommonMark, then cmark-gfm (or cmark for man pages only) to transcode to roff or html.

This gives us a much, much leaner toolchain for generating documentation through the build system. No need to provide the heavy XML and Docbook suites.

Localization is done with po4a, which uses gettext in the backend. We can finally do away with having the Japanese localization as an entire separate copy of the docs.

Introduces with-docs, with-docs-l10n, and with-website Meson options.

Removes with-manual, with-manual-l10n, and with-readme Meson options.

The resulting roff (man page) and html page layouts are slightly different, but the overall structure is identical. Notably, man pages get a bit more spacing around blockquotes, and has less bold font text. The note/warning box element could not be converted and is now represented as a blockquote.

Out of the OSes that we support, only Solaris 11 seems to lack a `cmark` package. Although it may just be named something else. Solaris users may need to provide their own cmark to build the man pages.

A known issue is that cmark outputs fragmentary html without styles, so they're less pleasing to read locally.

Next up will be to integrate this with the web publishing pipeline, which I will handle in the next PR.


rdmark:
Note to self: meson can't fine po4a by default on Arch. But it works on Debian and macOS.

PR #1907, opened 2025-01-19
meson: remove hardcoded file command path.

erichelgeson:
NixOS stores find in a different location


rdmark (approved):
Thanks for your contribution!

PR #1911, opened 2025-01-24
meson: Look for shared Berkeley DB library in versioned subdir too

rdmark:
Some environments keep the Berkeley DB shared libraries in a versioned subdir, like what is common with the Berkeley DB include dir. This change makes it so that checking the subdir is a fallback when looking for libraries.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -560,6 +560,10 @@ foreach dir : bdb_dirs
 
             foreach name : bdb_libnames
                 libdb = cc.find_library(name, dirs: bdb_libdir, required: false)
+                if not libdb.found()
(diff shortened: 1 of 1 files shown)


rdmark:
@trodemaster Would you be able to test if this works in your MacPorts environment?

At first, please try without any bdb parameters.

trodemaster (approved):
I have tested this change in on my dev system and it's working! Removing both of the build flags for bdp simplifies the config as well. Thanks!

rdmark:
That's great, thanks for testing! Please let us know if you run into other roadblocks with MacPorts.

PR #1926, opened 2025-01-29
Avoid compiler complaints re: getpwnam() argument being NULL

sqrmax:
I didn't immediately find a smaller way to prevent compiler warnings with gcc 11.4.0.  Originally, gcc complained thus.

[131/205] Compiling C object etc/afpd/libafpd.a.p/uam.c.o
../etc/afpd/uam.c: In function ‘uam_getname’:
../etc/afpd/uam.c:214:25: warning: argument 1 null where non-null expected [-Wnonnull]
  214 |                 pwent = getpwnam(bdata(princ));
      |                         ^~~~~~~~
In file included from ../include/atalk/uam.h:8,
                 from ../include/atalk/globals.h:23,
                 from ../include/atalk/dsi.h:17,
                 from ../etc/afpd/uam.c:30:
/usr/include/pwd.h:116:23: note: in a call to function ‘getpwnam’ declared ‘nonnull’
  116 | extern struct passwd *getpwnam (const char *__name) __nonnull ((1));
      |                       ^~~~~~~~

### DIFF
--- a/etc/afpd/uam.c
+++ b/etc/afpd/uam.c
@@ -210,7 +210,11 @@ struct passwd *uam_getname(void *private, char *name, const int len)
             princ = bformat("%s@%s", name, obj->options.addomain);
         else
             princ = bformat("%s%s%s", obj->options.ntdomain, obj->options.ntseparator, name);
-        pwent = getpwnam(bdata(princ));
+        if (bdata(princ) != NULL) {
+            /* squelch compiler complaints re: getpwnam() receiving a NULL argument */
+            char *bdatum = bdata(princ);
+            pwent = getpwnam(bdatum);
+        }
         bdestroy(princ);
 
         if (pwent) {


etc/afpd/uam.c
```diff
@@ -210,7 +210,11 @@ struct passwd *uam_getname(void *private, char *name, const int len)
             princ = bformat("%s@%s", name, obj->options.addomain);
         else
             princ = bformat("%s%s%s", obj->options.ntdomain, obj->options.ntseparator, name);
-        pwent = getpwnam(bdata(princ));
+        if (bdata(princ) != NULL) {
+            /* squelch compiler complaints re: getpwnam() receiving a NULL argument */
```
rdmark, line 214: In the future, there's no need to leave code comments like this. The code should speak for itself.

Of course, I leave it up to your judgment. Code comment are sometimes warranted in a convoluted codebase like this.

etc/afpd/uam.c
```diff
@@ -210,7 +210,11 @@ struct passwd *uam_getname(void *private, char *name, const int len)
             princ = bformat("%s@%s", name, obj->options.addomain);
         else
             princ = bformat("%s%s%s", obj->options.ntdomain, obj->options.ntseparator, name);
-        pwent = getpwnam(bdata(princ));
+        if (bdata(princ) != NULL) {
```
rdmark, line 213: Side note: There might be a case for adding some warning or debug logging somewhere around here. In this branch, the user has clearly tried to register with a Microsoft domain, but the input data is bad.

Anyways, out of scope for this PR.

rdmark (approved):
Thanks for your contribution!

etc/afpd/uam.c
sqrmax: Agreed re: code speaking for itself.  In this case I'd rather leave the comments in.  The approach is a kludge directed at the compiler.  One would have wanted to just test bdata(princ) for NULL and then call getpwnam() with bdata(princ).  But if you do that without the intervening assignment, gcc still complains.  The comment is meant to reflect awareness that this was seen as weird when written.

PR #1935, opened 2025-02-01
uams: Check account validity after calling pam_authenticate()

rdmark:
Flagged by Sonarqube:

> You should call "pam_acct_mgmt" after calling "pam_authenticate" to check account validity

Also addresses the surrounding code smells: error handling for seteuid(), and global pointer pointing to local variable address.


rdmark:
I will ignore the failed code duplication quality gate here. There is a high level of duplication between the UAMs which would require a rewrite to abstract away, which is out of scope for this PR.

rdmark:
As a side note, I considered making a failed seteuid() a fatal error, but decided against it in the end. The added logging should tell us if the user change failed, and will provide context for any subsequent failed operation.

PR #1936, opened 2025-02-01
Avoid TOCTOU conditions

rdmark:
Addresses a number of file access race conditions, as flagged by Sonarqube.

### DIFF
--- a/bin/ad/ad_mv.c
+++ b/bin/ad/ad_mv.c
@@ -282,13 +282,20 @@ static int do_move(const char *from, const char *to)
             SLOG("Couldn't resolve CNID for %s", from);
             return -1;
         }
-
-        if (stat(from, &sb) != 0) {
-            SLOG("Can't stat %s: %s", to, strerror(errno));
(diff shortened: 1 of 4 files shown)


rdmark:
@NJRoadfan FYI, this takes care of everything that static analysis points out, except for one in `ea_open()` in libatalk/vfs/ea_ad.c

This one is particularly tricky because there a stat() on the metadata file in the local function, then it goes to create_ea_header() to do some file system operations with file handlers, then comes back and does an open() on the metadata file. So I'm honestly not sure exactly how safe access of these files ought to look like...

rdmark:
After sleeping on it, I decided to reduce the scope of this PR to the simpler / lower risk fixes to begin with. Namely: ad, dbd scanvol, papd lp, and speedtest.

What remains from Sonar flagged security issues all lie in libatalk and are more involved: vfs/vfs.c, vfs/ea_ad.c and cnid/cnid.c

The latter will be addressed separately.

PR #1942, opened 2025-02-10
[4.1] webmin: Fix `-router` switch in Webmin atalkd module

rdmark:
The `-router` option was incorrectly called `-route` in the webmin module


rdmark:
This was a mistake. Should have gone into the 4.1 branch.

PR #1951, opened 2025-02-14
Fix pointer declaration so it is a const

sqrmax:


rdmark (approved):
Thanks for the fix! It will make SonarQube happy. ;)

PR #1957, opened 2025-02-14
libatalk: Use file descriptors and exclusive access to avoid TOCTOU conditions

rdmark:
Refactoring write_ea operation in the libatalk vfs module to be more concurrency safe.


rdmark:
This breaks ea_open() for adouble v2. Back to the drawing board.

PR #1968, opened 2025-02-20
Config file handling compatible with upstream iniparser

rdmark:
This addresses two issues following the move to linking with a regular shared iniparser library.

- Introduce `name =` option to volumes which sets a custom volume name. This works around the fact that section names are forced lowercase in iniparser. And arguably, this is a better pattern, since ini file section names are really only meant as internal identifiers as the original iniparser design.
- Treat section names as all lowercase internally, so that the Homes dir isn't ignored.
- Rewrite the introduction and Homes section in the afp.conf man page

### DIFF
--- a/doc/manpages/man5/afp.conf.5.md
+++ b/doc/manpages/man5/afp.conf.5.md
@@ -16,12 +16,12 @@ The file consists of sections and parameters. A section begins with the
 name of the section in square brackets and continues until the next
 section begins. Sections contain parameters of the form:
 
-        name = value
+        option = value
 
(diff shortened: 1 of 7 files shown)


cheesestraws (approved):
looks sensible to me

Not that my approval means a huge amount, but I said I'd drop by and review some code when I had the mental health for it :-).  So here we are.

rdmark:
@cheesestraws Cheers, I appreciate when you lean it! Having more eyes on the code is very valuable.

I simplified the second commit. Only the change to the header was actually required. Also pushed documentation updates now, including the Japanese localization.

PR #1970, opened 2025-02-21
docs: Correct description of hostname option in afp.conf

rdmark:
The afp.conf man page was still using the description of the old `-hostname` option in netatalk 2.x verbatim, despite the behavior having changed completely as of 3.x.

Thanks to thecloud on 68kmla for reporting this.


rdmark:
Note: reverted in https://github.com/Netatalk/netatalk/pull/1974

PR #1985, opened 2025-02-28
libatalk: Remove bundled ftw libary and use system shared library instead

rdmark:


rdmark:
too much effort, I will attempt to upgrade to a newer version of GNU ftw instead

BSD ftw (e.g. macOS) lack callback that we use to traverse up in the dir structure (`upfunc` callback), as well as the return status `FTW_SKIP_SUBTREE` that we use to filter out files to skip. Both would have to be reimplemented locally, at which point we might as well keep the GNU ftw library.

PR #1998, opened 2025-03-07
Update asip-status.in license grant and copyright year

rdmark:
With approval from the original author, this adds the standard GPL v2 license grant, and updates the copyright year as per the (now defunct) home page for this script.

Also removing outdated information from the header.


jwa (approved):
Original author here. LGTM!

PR #2036, opened 2025-03-22
meson: Cross-platform crypt library detection; always use shadow when available

rdmark:
These changes make non-PAM auth work on FreeBSD, while we correctly alert that only PAM auth is available on macOS.

- Split up the crypt and crypt.h checks. On *BSD the crypt library is in unistd.h
- Remove the with-shadow option, and always build with shadow support if possible
- Only throw warning with neither crypt nor shadow is available

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -234,6 +234,7 @@ endif
 check_headers = [
     'acl/libacl.h',
     'attr/xattr.h',
+    'crypt.h',
     'dlfcn.h',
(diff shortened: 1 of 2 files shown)


rdmark:
This got non-PAM auth working on FreeBSD.

However, it's still busted on OpenBSD. We may have to use `crypt_checkpass()` instead on this platform. for https://man.openbsd.org/crypt_checkpass.3

Issue #2050, opened 2025-03-28
'getzones' with a parameter does not work


cheesestraws:
**Describe the bug**

The man page for 'getzones' says that it accepts an address, and that it will query the AppleTalk router at that address.  This does not appear to work, instead printing an 'atp_open: Can't assign requested address' error.

This error is printed regardless of the address passed, except when it is the address of the local machine.  For example, if the address of the host 'prometheus' is 4.115,

```
prometheus$ getzones 4.115
Ethernet
LocalTalk
AirTalk
```

But if I try to query any other address, it fails:

```
prometheus$ getzones 8.31
atp_open: Can't assign requested address
```

It fails this way regardless of what address is provided, or whether anything is listening for ZIP ATP requests on that address, or even whether that address corresponds to anything else at all.

I haven't had time to dig any further here: I thought that maybe it was trying to send traffic out the wrong interface, but removing the AppleTalk address on lo0 didn't seem to achieve anything.

**To Reproduce**
Run getzones with a parameter.  

**Expected behavior**
I expected to see a GetZoneList ATP request being sent to the address given in the parameter, and the zone list from the reply to that packet being printed.

**Environment**
Reproduced with both NetBSD with netatalk 2.3.1 and Debian trixie 4.1.2~ds-4 (so this is either some quirk of my environment or nobody else uses this parameter, or they'd have notice it didn't work).  atalkd.conf is default.

cheesestraws:
Merrily replying to my own issue here:

If we look here:

https://github.com/Netatalk/netatalk/blob/main/bin/getzones/getzones.c#L88

'saddr' is being set to the address provided.  This is then passed to atp_open on line 84.  atp_open then copies this parameter into its local 'addr' and passes that as the first parameter to netddp_open (line 64).  netddp_open passes this address to 'bind': https://github.com/Netatalk/netatalk/blob/main/libatalk/netddp/netddp_open.c#L40

So this is actually trying to bind to the address specified on the command line, not just send to it, I think?  And the last parameter of atp_open should be whatever the AppleTalk equivalent of INADDR_ANY is?

cheesestraws:
Never mind, PR #2051

PR #2051, opened 2025-03-28
getzones: do not attempt to bind to the address we're also going to send packets to

cheesestraws:
Having said I didn't have time to investigate issue #2050, I decided to investigate it rather than sleep.  This PR appears to fix it.

From reading the definition of atp_open(), it looks like passing NULL to this is the intended idiomatic "bind to any address" syntax.

### DIFF
--- a/bin/getzones/getzones.c
+++ b/bin/getzones/getzones.c
@@ -91,7 +91,7 @@ int main( int argc, char *argv[])
 	}
     }
 
-    if (( ah = atp_open( ATADDR_ANYPORT, &saddr.sat_addr )) == NULL ) {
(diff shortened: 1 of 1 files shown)


rdmark:
Thanks for the patch!

Out of curiosity, I looked at when the bug was introduced. In v1.4b2, as well as the ASUN fork, atp_open() had a simpler interface that probably worked as intended. But as of the 1.4.99 pre-release code (the very beginning of Subversion revision control) we see the offending code. So the folks who were refactoring the codebase in preparation for publishing on Sourceforge must have introduced it.

rdmark (approved):

cheesestraws:
Grief, I'm the first person who has used this functionality in a long time, then!  Thanks for the merge :-)

PR #2056, opened 2025-03-29
getiface.c: fix headers order for Apple

barracuda156:
Fix headers’ order to avoid build error on some macOS versions.

### DIFF
--- a/libatalk/util/getiface.c
+++ b/libatalk/util/getiface.c
@@ -9,6 +9,7 @@
 #endif
 
 #include <errno.h>
+#include <sys/socket.h>
 #include <net/if.h>
 #include <netinet/in.h>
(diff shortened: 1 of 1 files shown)


rdmark (approved):
Thanks for the PR!

Out of curiosity, what are examples of affected macOS versions?

barracuda156:
@rdmark On 10.6 it fails with this:
```
/usr/include/net/if.h:265:34: error: field 'ifru_addr' has incomplete type
```
I don’t remember exact version threshold when these headers sequence became irrelevant.

rdmark:
Does netatalk v4 work well otherwise on 10.6? I don't have any older OSX machines handy myself, so I'm curious what it looks like.

PR #2103, opened 2025-04-12
Fall back to no EA rather than AD when the EA support check fails

rdmark:
This brings back the behavior of netatalk 4.1 and earlier

Updating documentation to reflect the changed behavior

### DIFF
--- a/doc/manpages/man5/afp.conf.5.md
+++ b/doc/manpages/man5/afp.conf.5.md
@@ -1105,11 +1105,10 @@ ea = <sys|samba|ad|none\> (default: auto detect) **(V)**
 > Specify how Extended Attributes and Classic Mac OS resource forks are
 stored.
 
-> By default, we attempt to enable **sys** with a fallback to **ad**.
+> By default, we attempt to enable **sys** with a fallback to **none**.
 For the auto detection to work, the volume needs to be writable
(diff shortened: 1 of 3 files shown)


NJRoadfan:
With this PR, if native EA support isn't found, AppleDouble reverts to v2, but the user has no way of enabling the old-style ea=v2 extended attribute support that uses the prefixed "EA::" files in the .AppleDouble directory. Either a new option needs to be added to explicitly enable this support, or the old pre-4.2 configuration options need to come back.

rdmark:
Right, the reason for removing the `appledouble` option wholesale was because I couldn't come up with a single valid reason for wanting the combination of appledouble dot folders with EA style metadata files. We had a brief conversation about this at the time, if I don't misremember. :)

If you see practical value with this combination, I'm happy to revive the `appledouble` option.

NJRoadfan:
Looking back, the old documentation was actually wrong for years. If ea = auto and the check fails, it went to "none" prior to 4.2, not "ad"! Looks good otherwise.

rdmark:
Exactly, I noticed the discrepancy between the fallback behavior described in the afp.conf man page with the actual logic in the code while refactoring the code to remove the `appledouble` option. My first instinct was to correct the docs, but then I thought an AD fallback would be slightly more useful. Little did I know...

PR #2128, opened 2025-04-20
docker: Introduce DISABLE_SPOTLIGHT flag for turning off indexing

rdmark:
Setting the new DISABLE_SPOTLIGHT env variable to non-zero will set spotlight = no in afp.conf

Additionally, we explicitly set appletalk = no when no atalkd interface is defined


rdmark:
@flori pardon the fork spying, but I noticed that you wanted a way to disable Spotlight in a container.

Here's a runtime solution for your problem. You would set f.e. `- DISABLE_SPOTLIGHT=1` in a docker compose file for this to take effect.

Would this work for you?

Issue #1921, opened 2025-01-26
You should be using meson.can_run_host_binaries() instead


rdmark:
You should be using [meson.can_run_host_binaries()](https://mesonbuild.com/Reference-manual_builtin_meson.html#mesoncan_run_host_binaries) instead, which allows people who:
- are doing cross builds
- have qemu installed
- configure meson with qemu-user as an "exe_wrapper"

to do accurate checks by running the `cc.run()` binary using qemu and seeing if it works.

_Originally posted by @eli-schwartz in https://github.com/Netatalk/netatalk/pull/1857#discussion_r1923952609_

rdmark:
Fixed in https://github.com/Netatalk/netatalk/pull/2140

PR #2140, opened 2025-04-27
netatalk: fix iconv dependency build issue

APCCV:
OpenWrt build for 4.2.2 (and 4.2.1) fails if iconv is enabled as the dependency is not present for some of the bin/misc artefacts.

Also picked up change suggested here: https://github.com/Netatalk/netatalk/issues/1921

### DIFF
--- a/bin/misc/meson.build
+++ b/bin/misc/meson.build
@@ -1,10 +1,13 @@
+misc_deps = []
+
 if have_iconv
+    misc_deps += iconv
     executable(
         'netacnv',
         ['netacnv.c'],
         include_directories: root_includes,
         link_with: libatalk,
-        dependencies: [iconv],
+        dependencies: [misc_deps],
         install: false,
     )
 endif
@@ -14,6 +17,7 @@ executable(
     ['logger_test.c'],
     include_directories: root_includes,
     link_with: libatalk,
+    dependencies: [misc_deps],
     install: false,
 )
 
@@ -22,7 +26,7 @@ executable(
     ['fce.c'],
     include_directories: root_includes,
     link_with: libatalk,
-    dependencies: [iniparser],
+    dependencies: [iniparser, misc_deps],
     install: true,
 )
 
@@ -38,7 +42,7 @@ if have_ldap
         ['uuidtest.c'],
         include_directories: root_includes,
         link_with: libatalk,
-        dependencies: [afpldaptest_deps],
+        dependencies: [afpldaptest_deps, misc_deps],
         c_args: confdir,
         install: true,
         build_rpath: rpath_libdir,
--- a/meson.build
+++ b/meson.build
@@ -1417,7 +1417,7 @@ else
             cdata.set('ICONV_CONST', 'const')
         endif
 
-        if not meson.is_cross_build()
+        if meson.can_run_host_binaries()
                 if cc.run(
                     '''
                 #include <iconv.h>


bin/misc/meson.build
```diff
@@ -22,7 +26,7 @@ executable(
     ['fce.c'],
     include_directories: root_includes,
     link_with: libatalk,
-    dependencies: [iniparser],
+    dependencies: [iniparser,iconv_deps],
```
rdmark, line 29: I don't think the iconv dependency is needed for this executable target.

bin/misc/meson.build
```diff
@@ -38,7 +42,7 @@ if have_ldap
         ['uuidtest.c'],
         include_directories: root_includes,
         link_with: libatalk,
-        dependencies: [afpldaptest_deps],
+        dependencies: [afpldaptest_deps,iconv_deps],
```
rdmark, line 45: I don't think the iconv dependency is needed for this executable target.

bin/misc/meson.build
```diff
@@ -1,10 +1,13 @@
+iconv_deps = []
```
rdmark, line 1: A pattern that is more consistent with the rest of the codebase would be to call this variable `misc_deps`, named after the current "module".

rdmark:
> Also picked up change suggested here: https://github.com/Netatalk/netatalk/issues/1921

When doing a commit that resolves a GitHub issue ticket, it's great if you can put that ticket number in the commit message. This makes it easier to compile the changelog later, as well as lets GitHub automatically close the issue ticket when the PR is merged.

This is described in https://github.com/Netatalk/netatalk/wiki/Developer-Notes#commit-messages

bin/misc/meson.build
APCCV: I'll take a look

APCCV:
> I just don't understand why right now.

What I'll never understand is why iconv is so special that it needs a sub-directory in /lib (or /usr/lib, or similar). I think that's the root of all evil. In most distributions that probably gets added to LD_LIBRARY_PATH and no one is the wiser. 
On OpenWrt where everything is skimpy by design, if the library is not under /lib it's not found.
The explicit dependency is needed for the -L<path> directive for the linker. If iconv is included in build, then libatalk has references to it, and any linkage without an explicit iconv dependency fails.
At least I know to look out for this. 
This PR was more of a knee-jerk reaction with the patch I added to OpenWrt to build. I can reformat it. In short it does two things:
- have a means to extend the logical scope of if has_iconv by creating the variable (which I can rename). The original file just covers one. Many ways to make this prettier; I just went for the 2nd option that I though of (the first was even uglier).
- include the variable as a dependency on all three (four?) components being linked.

I'll revisit when I get another break. Cheers!

rdmark:
This theory sounds like a plausible reason for the behavior that you observed, that linking with libatalk implicitly adds the iconv dependency. I'm still not entirely sure why this problem is only observed on OpenWrt, though.

I wonder if explicitly linking iconv (and libiconv) with the libatalk shared library will make a difference? Right now the only spot in libatalk where we link with iconv is the libcharsets static library (in the unicode module)

bin/misc/meson.build
rdmark, line 29: nitpick: the common coding style in this project is that a comma is always followed by a space

rdmark:
@APCCV have you had a chance to look at this again? I think this solution is totally fine, if you can please just update the variable naming convention.

APCCV:
Sorry... was busy with PR to get 4.2.2 into OpenWrt.
OK with your comments. 
The afpldaptest could also be refactored into misc_deps... post a comment if OK and I'll fold it into misc_deps (by dropping afpldaptest). Makes sense too to leave it as is.

APCCV:
> When doing a commit that resolves a GitHub issue ticket, it's great if you can put that ticket number in the commit message. This makes it easier to compile the changelog later, as well as lets GitHub automatically close the issue ticket when the PR is merged.

Done

rdmark (approved):
cheers!

PR #2143, opened 2025-05-02
libatalk: MySQL query error log level is dropped to debug

rdmark:
Previously, the MySQL query errors were all logged at info level, which led to very noisy logs because this module uses duplicate entry error handling as a common pattern

Also, additional debug level logging for volume entry duplication

### DIFF
--- a/libatalk/cnid/mysql/cnid_mysql.c
+++ b/libatalk/cnid/mysql/cnid_mysql.c
@@ -236,7 +236,7 @@ static int cnid_mysql_execute(MYSQL *con, char *fmt, ...)
     rv = mysql_query(con, sql);
 
     if (rv) {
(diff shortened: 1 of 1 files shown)


rdmark:
Example log message that appears in large numbers during normal use.

```
afpd[130] {cnid_mysql.c:239} (info:CNID): MySQL query "INSERT INTO volumes (VolUUID,Volpath,Stamp,Depleted) VALUES('AC6DE124CDFFA5DBAE9CA5174CCBC5C5','/mnt/afpshare','_9h\0\0\0\0',0)", error: Duplicate entry 'AC6DE124CDFFA5DBAE9CA5174CCBC5C5' for key 'PRIMARY'
```

PR #2149, opened 2025-05-03
Read the appletalk option only when netatalk built with DDP

rdmark:
This puts the appletalk afp.conf read behind the NO_DDP flag and updates documentation

### DIFF
--- a/doc/manpages/man5/afp.conf.5.md
+++ b/doc/manpages/man5/afp.conf.5.md
@@ -937,8 +937,9 @@ clients. See also the Charset options elsewhere in this manual.
 
 appletalk = *BOOLEAN* (default: *no*) **(G)**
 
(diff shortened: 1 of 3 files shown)


rdmark:
@NJRoadfan are you comfortable with this change? Was there a particular reason for keeping the appletalk option outside of the NO_DDP preprocessor macro?

NJRoadfan:
It was an omission at the time. I'd lean towards it being disabled by default as most people are likely to not have kernel support or use AppleTalk. While harmless, enabling the option will throw errors on systems where DDP support is compiled in, but atalkd isn't running.

rdmark:
Fair enough, I can see the argument, although the counter argument could be that the error would point a user towards what to do to enable appletalk support.

Perhaps this is worth revisiting when/if a userland DDP network stack becomes widely available.

PR #2152, opened 2025-05-05
Beautify C and meson code

rdmark:
This is a wholesale reformatting of the entire C and meson codebases, using astyle for C and muson for meson, automated via shell script and the CI workflow

The new shell script depend on astyle, muon and git, and will return 0 if no code was modified, and 1 if the code was modified meaning that it wasn't compilant with the astyle rules or muon's opinionated coding style

Depends on muon 0.4.0 or later and astyle 3.6.9 or later

Note that we have been using muon for this purpose at an ad-hoc basis for a long time, and this just formalized/automates it. See https://github.com/Netatalk/netatalk/wiki/Developer-Notes#formatting

Also, special thanks to Lars Möllendorf of the iniparser project for the inspiration and starting point for the astyle configuration


rdmark:
I encountered a significant drawback with using astyle and muon to enforce coding style. Namely, that different versions of these style checkers produce significantly different results. For instance, the latest Debian Trixie pre-release ships astyle 3.1, while Homebrew on macOS provides 3.6.9. With the same rc file, the two produce different formatting.

Right now, the only solution I can think of would be to mandate a specific canonical version of each and use that for styling exclusively. We don't want to have a bunch of noise in the commit history caused by going back and forth between style checker versions.

This also means that it will be cumbersome to enforce style checks in CI, since there's no guarantee that each and every contributor will be able to run the specific style checker version that we mandate, and therefore not able to address their own stylistic infractions effectively.

We may want to continue researching more stable C formatters, in particular.

rdmark:
Curiously, even when running the exact same astyle version, 3.6.9 on macOS (homebrew) locally vs. in the GitHub workflow, we are getting different formatting results. There must be something environmental that affects how astyle interprets the rules.

rdmark:
Repurposed this PR to only cover meson code reformatting with muon.

Note that the GitHub CI job for Formatting is currently using macOS because 1) meson in Ubuntu 24.04 is too old (0.3.0) and 2) we cannot use container images as runners because GitHub will not clone the netatalk git repo in the checkout workflow but rather fetch a zip archive (we rely on git to track changes after running muon). The proper solution would be for us to build our own runner container for this purpose that comes with git and muon 0.4.0 pre-installed. A project for another day.

rdmark:
After additional experimentation, I'm back to using astyle 3.6.9 as the C code formatter. The discrepancies with v3.1 are not that many, and when scrutinizing the diff, all clear improvements (bugfixes in astyle).

As to what appeared to be environment discrepancies, turned out to be a matter of a handful of changes that took effect only upon a second run of the formatter. These were changes to particular lines that had gotten line breaks in the first round, and therefore subsequent changes to the latter fragment of the same line fell off.

To deal with the fact that v3.6.9 isn't available by default on many platforms yet, I have added verbose diff output in the CI logs, so that contributors can see exactly which changes are needed in case they have to make the edits by hand.

cheesestraws (approved):

cheesestraws:
This PR seems to have accidentally reverted my nbplkup patches

rdmark:
Pardon, my effort to rebase with main clearly failed.

I pushed a commit directly to main that should bring this back: https://github.com/Netatalk/netatalk/commit/8aaf34f4f7fc37b89286c7eb74fd162f95c33948

Please let me know if anything else is missing.

cheesestraws:
Looks like 2164 is missing also

rdmark:
It looks like it was only the sort order of header includes in nbplkup.c that got reverted. Fixed with direct commit to main:

https://github.com/Netatalk/netatalk/commit/a771359e8542329496c74b6ee43df2ab66ec1442

cheesestraws:
Cheers! I thought the PAP change had gone walkies too but that was me misreading - apologies

PR #2156, opened 2025-05-09
Use native doubly linked list when available

rdmark:
Netatalk's list.h is borrowed from the Linux kernel, so use the native header when available


rdmark:
For future reference, the Linux headers aren't available to other programs by default, but I attempted locally on Debian to install the Linux header package and force Meson to include those headers. This introduced a massive amount of symbol conflict so I don't think the Linux headers are meant for other programs to include.

PR #2157, opened 2025-05-10
nbplkup: provide for script-friendly output (and some code cleanup)

cheesestraws:
This PR modifies nbplkup to provide output in a format that's fixed enough for a script to easily parse.  On its way to doing this, it normalises whitespace and factors out some of the side-effect-y if() conditions - the latter of which is admittedly a purely personal preference.

Tested locally but only on NetBSD; not that anything in here should be OS-specific.

Side note: does netatalk target any specific revision of the C standard?  I've introduced a reference to stdbool.h here, in the interests of striking out for sanity.

### DIFF
--- a/bin/nbp/nbplkup.c
+++ b/bin/nbp/nbplkup.c
@@ -45,9 +45,9 @@ static char *Zone = "*";
 
 static void Usage(char *av0)
 {
-    char *p;
+    char *p = strrchr(av0, '/');
 
-    if ((p = strrchr(av0, '/')) == NULL) {
+    if (p == NULL) {
         p = av0;
     } else {
         p++;
@@ -87,7 +87,8 @@ int main(int ac, char **av)
             nresp = atoi(optarg);
             break;
         case 'm':
-            if ((charset_t)-1 == (chMac = add_charset(optarg))) {
+            chMac = add_charset(optarg);
+            if ((charset_t)-1 == chMac) {
                 fprintf(stderr, "Invalid Mac charset.\n");
                 exit(1);
             }
@@ -99,7 +100,8 @@ int main(int ac, char **av)
         }
     }
 
-    if ((nn = (struct nbpnve *)malloc(nresp * sizeof(struct nbpnve))) == NULL) {
+    nn = (struct nbpnve *)malloc(nresp * sizeof(struct nbpnve));
+    if (nn == NULL) {
         perror("malloc");
         exit(1);
     }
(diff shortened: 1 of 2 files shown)


bin/nbp/nbplkup.c
```diff
@@ -45,69 +45,68 @@ static char *Zone = "*";
...
-    extern int		optind;
+    struct nbpnve *nn;
+    char *name;
+    int i, c, nresp = 1000;
+    struct at_addr addr;
+    char *obj = NULL;
+    size_t obj_len;
+    charset_t chMac = CH_MAC;
+    char * convname;
+
```
rdmark, line 73: nitpick: Some trailing whitespace made it into the changeset

bin/nbp/nbplkup.c
```diff
@@ -64,7 +64,8 @@ int main(int ac, char **av)
     int i, c, nresp = 1000;
     struct at_addr addr;
     char *obj = NULL;
-    size_t obj_len;
+    char *type = NULL;
+    size_t obj_len, type_len;
```
rdmark, line 69: I have a mild preference for having a single declaration per line. Makes for a cleaner commit history, while arguably making the code more readable and less mistake prone.

doc/manpages/man1/nbp.1.md
```diff
@@ -50,6 +53,15 @@ Find all devices of type *LaserWriter* in the local zone.
...
     example%
+    
+Find all devices of type *netatalk* in the local zone, providing script-friendly
+output.
+
+	example% nbplkup -s :netatalk
+	5.42:4 netatalk-build:netatalk
+	4.162:4 prometheus:netatalk
+	8.31:4 Tiryns:netatalk
+	example%
```
rdmark, line 64: The commonmark spec expects four space indentation for code blocks, rather than tabs.

https://spec.commonmark.org/0.31.2/#indented-code-blocks

GitHub seems to still parse this tab indentation correctly, but there's no guarantee that all cmark parsers will.

bin/nbp/nbplkup.c
```diff
@@ -76,7 +78,7 @@ int main(int ac, char **av)
     set_charset_name(CH_MAC, MACCHARSET);
 
     memset(&addr, 0, sizeof(addr));
-    while ((c = getopt(ac, av, "r:A:m:")) != EOF) {
+    while ((c = getopt(ac, av, "r:A:m:s")) != EOF) {
```
rdmark, line 81: The convention for oparg strings in this codebase is: alphabetic order, with bool options first and parameterized options second. Makes for more readable, less error-prone code. (I dealt with some obnoxious bugs in the past because of this.)

In this case, the ideal optarg string would be: `"sA:m:r:"`

Since the number of options is still small, there's no rush to organize this one. But perhaps next time an option gets added. :)

doc/manpages/man1/nbp.1.md
```diff
@@ -20,7 +20,10 @@ be advertised.
 registered on the AppleTalk internet. *nbpname* is parsed by
 **nbp_name**(3). An \`*=*' for the *object* or *type* matches anything,
 and an \`*\**' for *zone* means the local zone. The default values are
-taken from the NBPLKUP environment variable, parsed as an *nbpname*.
+taken from the NBPLKUP environment variable, parsed as an *nbpname*.  If
+-s is specified, output is printed in a script-friendly format: for each
+response, first the address is printed, followed by a single space, followed 
+by the name and type, followed by a linefeed.
```
rdmark, line 26: Suggestion: It would be great to have this in a separate paragraph. I think it reads better that way, plus it makes localization with gettext (to Japanese) easier.

rdmark (changes requested):

bin/nbp/nbplkup.c
cheesestraws: Might as well do it while I'm here!

cheesestraws:
I think that should address all your comments :-)

rdmark (approved):
Perfect, thanks!

PR #2161, opened 2025-05-10
nbplkup: replace dubious malloc/strlen dance with strndup...

cheesestraws:
... and appease static analysis things which (rightly) think that unbounded string calls look like a liability.

### DIFF
--- a/bin/nbp/nbplkup.c
+++ b/bin/nbp/nbplkup.c
@@ -132,29 +132,23 @@ int main(int ac, char **av)
             exit(1);
         }
 
-        name = (char *)malloc(strlen(Obj) + 1);
-        if (name == NULL) {
-            perror("malloc");
+        Obj = strndup(Obj, 32);
+        if (Obj == NULL) {
+            perror("strndup(Obj)");
             exit(1);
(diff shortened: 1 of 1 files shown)


bin/nbp/nbplkup.c
```diff
@@ -132,29 +132,20 @@ int main(int ac, char **av)
             exit(1);
         }
 
-        name = (char *)malloc(strlen(Obj) + 1);
-        if (name == NULL) {
-            perror("malloc");
-            exit(1);
+        Obj = strndup(Obj, 32);
+        if (Obj == NULL) {
+            perror("strndup(Obj)");
```
rdmark, line 137: You don't think we need to exit the program anymore when this check fails? I'm genuinely curious if you perhaps saw that error handling is robust enough elsewhere. :)

cheesestraws: No, that was a copy and paste fail on my part, not any indication of confidence... this is why code review is a good thing.  One moment please, and I will fix.

rdmark (approved):
looks reasonable to me now

PR #2162, opened 2025-05-11
nbplkup: allow specification of NBP operation and of destination address

cheesestraws:
The intent of this PR is to allow nbplkup to probe NBP speakers that are not the local host.

To do this, it adds two features:

- A -D option, which specifies the address of the remote NBP speaker to direct queries to
- -f and -l options, which override the default BrRq request type and enable either LkUp or FwdReqs to be generated.

While I was passing I also ran astyle on libatalk/nbp/nbp_lkup.c - hope this doesn't interfere with your coding style work.

Tested only on NetBSD, but again, I can't see any reason this wouldn't be portable.

### DIFF
--- a/doc/manpages/man1/nbp.1.md
+++ b/doc/manpages/man1/nbp.1.md
@@ -8,7 +8,7 @@ nbplkup, nbprgstr, nbpunrgstr — tools for accessing the NBP database
 
 **nbprgstr** [-A *address*] [-m *Mac charset*] [-p *port*] *obj:type@zone*
 
-**nbpunrgstr** [-A *address*] [-m *Mac charset*] [-s] *obj:type@zone*
+**nbpunrgstr** [-A *address*] [-D *address*] [-m *Mac charset*] [-s] [-f | -l] *obj:type@zone*
 
 # Description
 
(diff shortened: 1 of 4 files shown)


doc/manpages/man1/nbp.1.md
```diff
@@ -22,6 +22,11 @@ registered on the AppleTalk internet. *nbpname* is parsed by
 and an \`*\**' for *zone* means the local zone. The default values are
 taken from the NBPLKUP environment variable, parsed as an *nbpname*.
 
+If -A is specified, this address is used as the local address to be used for the lookup.
+If -D is specified, this address is used as the destination address to send the lookup to.
+The -f and -l options allow FwdReq and LkUp operations to be used instead of the
+default BrRq.
```
rdmark, line 28: Might be worth adding a sentence on what distinguishes these three operations. I have to be honest that I don't know myself off the top of my head. (Unless I take the time to read up on it.)

rdmark:
The new logic looks sound in my reading. Cheers!

My only suggestion is regarding how to present this functionality to users in the man page. Perhaps adding another example for illustration would be helpful as well?

cheesestraws:
Yes, you're right it could be better explained.  LkUp vs BrRq is quite easy to explain, FwdReq less so - and probably less useful unless you're either troubleshooting a really nasty caching issue in a network or testing an NBP implementation.  Let me have a think about how better to present this.

cheesestraws:
How does that read?  Any better?  I'm trying to avoid protocol details and talk about actual uses.

rdmark (approved):
Yes, this version of the nbp man page explains the high level purpose of these options much better.

If a user wants more technical descriptions, they should read Inside AppleTalk or the netatalk code.

PR #2164, opened 2025-05-14
Small code quality fixups

cheesestraws:
Just a couple of small fixups not really connected to anything else but which were trivial to fix while I was passing

1. A slight tidying of includes and formatting.
2. While `#define FUCKED` gave me a real surge of empathy for whoever wrote that code, it is in fact redundant, since it looks like it's been permanently defined for the last 25 years; I think we can get rid of the ifdefs now.

### DIFF
--- a/bin/getzones/getzones.c
+++ b/bin/getzones/getzones.c
@@ -31,15 +31,15 @@ static void usage(char *s)
 
 int main(int argc, char *argv[])
 {
-    struct atp_handle	*ah;
-    struct atp_block	atpb;
-    struct sockaddr_at	saddr;
-    struct servent	*se;
-    char		reqdata[4], buf[ATP_MAXDATA];
-    struct iovec	iov;
-    short		temp, index = 0;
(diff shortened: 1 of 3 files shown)


cheesestraws:
Weird.  I ran astyle locally and it didn't do that.  OK, will reformat again...

cheesestraws:
I personally do prefer it - I thought 1tbs was meant to have the brace on the opening line, like all the other multi-line blocks.

rdmark:
The astyle interpretation of 1tbs is as follows, quoting the relevant passage:

> Opening braces are broken from namespaces, classes, and function definitions. The braces are attached to everything else, including arrays, structs, enums, and statements within a function. 

https://astyle.sourceforge.net/astyle.html#_style=1tbs

cheesestraws:
Well, I obviously misremembered what brace styles were called then, which isn't that surprising... sigh.

Anyway, here it is formatted according to the *current* rules, since we may as well be consistent

rdmark (approved):

PR #2165, opened 2025-05-15
Change from 1tbs to google base ruleset for astyle

rdmark:
This gives us back the non-broken opening braces which was to a great degree the standard in this codebase before applying astyle in the first place

### DIFF
--- a/.astylerc
+++ b/.astylerc
@@ -1,7 +1,7 @@
-# astyle base ruleset: One True Brace Style
---style=1tbs
+# astyle base ruleset
+--style=google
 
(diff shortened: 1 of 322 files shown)


rdmark:
Discarding. Turns out the google base ruleset expects two space indentation, which makes it slightly buggy with four space indentation, notable for switch statements.

For the record, I tried the java ruleset as well which (somewhat surprisingly) aligns perfectly with what we want, but this one was also somewhat buggy with a few of our switch statements, and I couldn't immediately see what caused it.

PR #2169, opened 2025-05-19
getzones: allow all ZIP query types to be generated (and sundry other enhancements)

cheesestraws:
The intent of this PR is to get getzones to the point it can meaningfully be used to probe ZIP with more query types than just the ATP ones.  I only ran astyle at the end, because my brain was addled, so apologies for any stylistic infelicities in earlier commits.

Most of the commit messages hopefully speak for themselves, but perhaps the strdup fix could do with a little discussion even though the code is tiny and trivial.  Essentially, the code as it is cannot work, because strdup will run off the end of the pascal string and just chunter off into memory for who knows how long.  Hooray.  However, this only ever is triggered when string conversion fails, and I'm not sure it realistically does fail except in circumstances where the fallback would also fail.  So in this PR I haven't bothered adding it to the further-down string converstions, but if you think it's worthwhile I'll do that (or if you think the whole thing is silly, I'll remove it from the place where it was).

Examples in the man page are for my own network; the AirTalk isn't there for product placement :p it's just that that's what the zone is called on my own network.  If you'd prefer to have a different zone name, though, I don't object to changing it, was just too lazy.

### DIFF
--- a/doc/manpages/man1/getzones.1.md
+++ b/doc/manpages/man1/getzones.1.md
@@ -4,7 +4,7 @@ getzones — list AppleTalk zone names
 
 # Synopsis
 
-**getzones** [-m | -l] [*address*]
+**getzones** [-m | -l] [-c *Mac charset*] [*address*]
 
 # Description
 
@@ -24,6 +24,11 @@ ZIP GetMyZone request.
 
 > List the local zones; this is accomplished by sending a GetLocalZones
 request.
+
+**-c**
+
+> Sets the Macintosh character set to use when interpreting zone names.  If not specified,
+defaults to MacRoman.
 
 *address*
 
--- a/doc/manpages/man1/getzones.1.md
+++ b/doc/manpages/man1/getzones.1.md
@@ -4,7 +4,7 @@ getzones — list AppleTalk zone names
 
 # Synopsis
 
-**getzones** [-g | -m | -l] [-c *Mac charset*] [*address*]
+**getzones** [-g | -m | -l | -z *zone*] [-c *Mac charset*] [*address*]
 
 # Description
 
@@ -30,6 +30,13 @@ ZIP GetMyZone request.
 
 > List the local zones; this is accomplished by sending a GetLocalZones
 request.
+
+**-z**
+
+> Verifies whether *zone* is a valid zone for this network by sending a GetNetInfo.
+If *zone* is valid, exit with 0; if *zone* is not valid, exit with 2.  Also prints the
+network configuration including, if the zone is invalid, the default zone to use instead
+of the one requested.
 
 **-c**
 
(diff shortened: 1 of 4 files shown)


rdmark:
I'll do a proper code review tomorrow!

In the meantime, just a stylistic remark that this codebase uses classic C style `/* */` code comments throughout.  (Codified in the style guide on the wiki.) It would look nicer (IMHO) if the newly added comments followed suite.

Very cool changeset, BTW!

cheesestraws:
Oops.  I'll fix them when I get home.  You'll probably have to tell me this again in future contributions I'm afraid, comment style is one of those things that seems to fall out of my brain...

cheesestraws:
Comments changed - don't think I missed any - and rebased onto master.  Weirdly, astyle here (3.6.8) appears not to be applying the rules that astyle in CI is applying (??) - it certainly doesn't seem to be respecting --align-pointer=name and it looks like it's complaining about whitespace too...

rdmark:
Can you please double check that astyle is reading the .astylerc rules? Maybe you're running it from a subdir and forgot to change the relative path in `astyle --options=.astylerc` ?

cheesestraws:
Oh, worked out what's happening.  Even if .astylerc is in the same directory, it ignores it unless you explicitly give it --options=./.astylerc.  For heaven's sake.  I am beginning to find this tool quite annoying.  One moment please, revised astyle commit coming now...

rdmark:
Reading the astyle documentation, I figured out now that you have the option to pass `--project` to astyle which will then read `.astylerc` in the current working dir, saving you the effort of having to type out the whole file name. Which isn't super intuitive, arguably.

There's a recent feature request on this topic. I upvoted it. You can too if you feel motivated. :)

https://gitlab.com/saalen/astyle/-/issues/64

.astylerc
```diff
@@ -18,3 +18,4 @@
 --pad-oper
 --suffix=none
 --unpad-paren
+
```
rdmark, line 21: Do you think a trailing newline is appropriate here, or was it a mistake?

doc/manpages/man1/getzones.1.md
```diff
@@ -15,21 +15,67 @@ running **atalkd**(8).
 
 # Options
 
-**-m**
+**-g**
```
rdmark, line 18: Please don't forget to document the `-c` option as well. :)

rdmark:
Very neatly organized and well code-commented changeset! While I cannot claim to understand the correctness and efficiency of the added logic, I can get a sense for the intentionality.

In the future, it would love to see the astyle styling applied on each commit before submission. Will make for a much cleaner changeset for posterity. But no worries this time.

.astylerc
cheesestraws: Mistake - and I'm not sure how I managed it.  Let me remove...

doc/manpages/man1/getzones.1.md
cheesestraws: ooooops!  Fixed now.  Also noticed that the corresponding -m option wasn't documented in nbplkup's man page either, so added a note in there.

rdmark (approved):

PR #2171, opened 2025-05-21
Code formatting script runs both meson and C formatter by default

rdmark:
Rather than making it mandatory to select one or the other, both astyle and muon are now ran, and will be skipped with a warning if either is missing rather than bailing out

We now allow for running without git as well, although no diff will be produced afterwards

### DIFF
--- a/contrib/shell_utils/codefmt.sh
+++ b/contrib/shell_utils/codefmt.sh
@@ -14,13 +14,10 @@
 #  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
 #  GNU General Public License for more details.
 
-if ! git rev-parse --is-inside-work-tree >/dev/null 2>&1; then
-    echo "Error: Not inside a git repository"
-    exit 3
-fi
-
 FORMATTER_CMD=""
 SOURCE_TYPE=""
+IS_GIT=0
+FORMATTED=0
 VERBOSE=0
 
 usage() {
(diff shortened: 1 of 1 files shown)


rdmark:
@cheesestraws This should make the `./contrib/shell_utils/codefmt.sh` convenience script a little bit more convenient, so you can run it as-is to reformat the entire codebase before each commit for instance. No need to juggle with astyle parameters.

cheesestraws:
On my machine the script is continually reformatting test/testsuite/afparg.c, which I'm guessing it isn't in CI - any idea why?

rdmark:
Can you please share the diff on afparg.c that you get locally?

cheesestraws:
Certainly - should have done that in my original post.  Here:

```
diff --git a/test/testsuite/afparg.c b/test/testsuite/afparg.c
index a77286ed..3e95f51a 100644
--- a/test/testsuite/afparg.c
+++ b/test/testsuite/afparg.c
@@ -19,11 +19,11 @@ struct test_fn {
 #define FN_N(a,b) { # a, FN(a) , # a " " # b},
 
 static struct test_fn Test_list[] = {
-    FN_N(FPResolveID, <file CNID>)
-    FN_N(FPEnumerate, <dir>)
-    FN_N(FPCopyFile, <source> <dest>)
-    FN_N(FPLockrw, d | r <file>)
-    FN_N(FPLockw, d | r <file>)
+    FN_N(FPResolveID, < file CNID >)
+    FN_N(FPEnumerate, < dir >)
+    FN_N(FPCopyFile, < source > < dest >)
+    FN_N(FPLockrw, d | r < file >)
+    FN_N(FPLockw, d | r < file >)
 
     {NULL, NULL},
 };
```

It's adding spaces around the chevrons,

rdmark:
Very curious! To my best knowledge, the chevrons here aren't C syntax. Just part of arbitrary token strings used in the macros. I suggest we remove them: https://github.com/Netatalk/netatalk/pull/2173

PR #2172, opened 2025-05-21
getzones: Consistently use uint8_t buffer data type

rdmark:
This is to appease Apple clang 17.0.0 on macOS


rdmark:
@cheesestraws FYI, this was something clang on macOS Sequoia started complaining about after the recent changes to getzones.

Not that we can use getzones on macOS right now of course, but I like to keep the AppleTalk code compiling cleanly on macOS to keep the dream alive...

rdmark:
This is the clang warning in question, for posterity

```
warning: passing 'uint8_t *' (aka 'unsigned char *') to parameter of type 'char *' converts between pointers to integer types where one is of the unique plain 'char' type and the other is not [-Wpointer-sign]
```

PR #2173, opened 2025-05-22
Remove chevrons from placeholder tokens in macro

rdmark:
The chevrons aren't standard C syntax, and their inclusion outside of quotation marks confuse astyle in certain circumstances


cheesestraws (approved):
I am always amazed, given its restrictions, how bonkers are some of the domain specific language type bits people manage to come up with the C preprocessor.  It's bizarrely inspiring.

rdmark:
I took me a minute to understand what was going on. C macros are an art. :)

PR #2178, opened 2025-05-24
Add rtmpqry utility to allow querying of routing information from neighbour routers

cheesestraws:
This PR introduces rtmpqry, a utility to allow querying of what neighbouring routers are advertising using RTMP.  The man page and the commit messages should hopefully be enough to give you an idea of what's going on here; essentially, we can solicit an RTMP data dump using an RTMP Route Data Request, or we can ask for network metadata through an RTMP Query.  These are very confusingly named, and I've done my best to add clarity to the distinction where possible, but there's only so much can be done.

### DIFF
--- a/bin/rtmpqry/rtmpqry.c
+++ b/bin/rtmpqry/rtmpqry.c
@@ -34,10 +34,12 @@ int setup_ddp_socket(const struct at_addr *local_addr,
 void do_rtmp_rdr(int sockfd, const struct sockaddr_at *sa_remote, bool get_all,
                  int secs_timeout);
 void print_rtmp_data_packet(const uint8_t *buf, size_t len);
+void do_rtmp_request(int sockfd, struct sockaddr_at *sa_remote);
 
 static void usage(char *s)
 {
-    fprintf(stderr, "usage:\t%s [ -A local_address ] [remote_address]\n", s);
+    fprintf(stderr, "usage:\t%s [-a | -r] [ -A local_address ] [remote_address]\n",
+            s);
     exit(1);
 }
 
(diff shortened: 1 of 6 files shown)


doc/manpages/man1/rtmpqry.1.md
```diff
@@ -0,0 +1,58 @@
...
+	example$
+
+Get all routes from the router at 12.1:
+
+	example$ build/bin/rtmpqry/rtmpqry -a 12.1
+	3 - 10 distance 0 via 12.1
+	11 distance 0 via 12.1
+	12 distance 0 via 12.1
+
+	example$
```
rdmark, line 50: Please use four space indentation for code blocks in Markdown to be Commonmark compliant

cheesestraws:
apologies for noise, trying to fix const correctness and I somehow did something horrible to the commit history.  I blame git rebase.

doc/manpages/man1/rtmpqry.1.md
cheesestraws: Sorry, thought I'd fixed my text editor to do this properly after last time... obviously I hadn't.  Fixed now.

bin/rtmpqry/rtmpqry.c
```diff
@@ -0,0 +1,447 @@
...
+
+        /* Skip the DDP type which we checked above */
+        uint8_t *cursor = buf + 1;
+        uint16_t router_net = ntohs(*(uint16_t*)cursor);
+        cursor += 2;
+        uint8_t id_len = *cursor++;
+
+        if (id_len != 8) {
+            fprintf(stderr, "RTMP packet contained bad id length %"PRIu8" - are you "
+                            "living in an alternate timeline where DDP was extended?", id_len);
```
rdmark, line 200: Perhaps the user was a Timelord riding the Tardis!

doc/manpages/man1/rtmpqry.1.md
```diff
@@ -0,0 +1,58 @@
...
+
+    example$
+
+# See Also
+
+atalk_aton(3), atalkd(8), getzones(1)
+
+# Author
+
+[Contributors to the Netatalk Project](https://netatalk.io/contributors)
```
rdmark, line 58: Since you are the sole author in this case, you have the option to credit yourself here. (Compare macipgw). It's up to you!

rdmark (approved):
The one general comment here is that it's often "nice" to define magic numbers as global constants, e.g. in a relevant header file. Makes it more transparent to future readers of the code, and more resilient to change than code comments.

Nothing critical for this changeset. Just on top of mind since I've worked to reduce arbitrary magic numbers in this codebase.

cheesestraws:
Are you referring to 0x82 / 0x80?  I left those as magic numbers because they're treated as a naked numbers in Inside Appletalk.  0x82 is, in the text, referred to in a couple of places as "the RTMP version number", but not in the actual packet structure specifications, and as for 0x80, your guess is as good as mine, it's in the packet format specifications as an integer which is never explained or named, as far as I can see.  I've no idea what it's for or what it's doing...

So I could factor those out but it would make the code look further away from the packet diagrams as they're actually published.  Which might not be a bad thing since it's unlikely many other people will work on it, but swings, roundabouts, you know.  Happy to change them if you would like me to.

PR #2180, opened 2025-05-24
Introduce Markdown style guide and reformat all sources

rdmark:
First off, define a ruleset based on the markdownlint standard:

https://github.com/DavidAnson/markdownlint/blob/main/README.md#rules--aliases

Then touch up and reformat portions of Markdown sources to be compliant, and generally nicer to read and parse.

Finally, run the markdownlint-cli2 action in the Tests GitHub workflow.


rdmark:
@cheesestraws Heads-up that this is a first step towards standardizing Markdown format and syntax, as well

cheesestraws:
Is it intentional that the formatting rules are also being applied to the bug report template and the issue template?

rdmark:
I was on the fence about it, but yes it’s intentional. 

We also have the _Sidebar and _Navbar files which are not real pages but fragments to be embedded when the static website is built. 

In the end I had to make only very minor concessions in the linting rules to accommodate these pages so I thought it worth to keep them in scope of ljnting. 

What do you think?

cheesestraws:
For what it's worth, I agree; if there's going to be a consistent style for markdown might as well make it actually consistent across the whole repo.  But I could see arguments the other way as well, so just wanted to make sure it was intentional!

cheesestraws (approved):

PR #2184, opened 2025-05-25
GitHub CI: Use native meson to build on Solaris

rdmark:
The Oracle Solaris 11.4.81 CBE release now ships a recent enough meson package to be able to build netatalk

release notes:

https://blogs.oracle.com/solaris/post/whats-new-in-the-oracle-solaris-11481-cbe-release

This changeset also disables PAM for Solaris for the time being since it won’t build on this CBE. Will be followed up in an issue ticket.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -700,10 +700,9 @@ jobs:
               gcc \
               libevent \
               libgcrypt \
+              meson \
               ninja \
-              pkg-config \
-              python/pip
-            pip install meson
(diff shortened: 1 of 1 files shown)


rdmark:
Also pertinent to netatalk is that Oracle dropped mDNS in favor of Avahi in this upgrade, which our build system seems to handle gracefully. 

Unfortunately, there seems to have been a change to the Solaris PAM library which causes a compilation failure now.

PR #2186, opened 2025-05-25
Fix missing 'free' of remote_addr in rtmpqry

cheesestraws:
Should address your note on #2181.  Not sure it really *matters* here hugely, since it's a command line tool with a very finite runtime, but we should be neat and not put snares in future maintainers' paths.  (I think this same laxness is perpetrated in some of the other CLI tools; I'll check as I have the energy)


rdmark (approved):
The way I see it is that the program might be expanded in the future or used as a module in a larger program.

cheesestraws:
Yeah, exactly.  Let's not leave unexpected gotchas for future maintainers

PR #2187, opened 2025-05-26
Use markdown headings in the GitHub issue templates

rdmark:
Disable the emphasis as heading rule override for markdownlint, and convert the issue template markdown to use proper headings instead


rdmark:
Turns out this override was hiding a bunch of other non compliant usage of emphasis as headings. Will look closer.

rdmark:
This concerns the man pages, namely the parameters that precedes the block quote descriptions. I think we should keep this as is for now.

PR #2189, opened 2025-05-27
meson: Check for brew include dir before adding it to list

rdmark:
Protect against the case where the homebrew base dir exists but not the include dir underneath

Should prevent a build failure on MacPorts


rdmark:
@trodemaster would it be possible to apply this patch to the MacPorts port and see if it fixes the build failure?

PR #2192, opened 2025-05-29
meson: Build and link with Homebrew libraries is now opt-in

rdmark:
Before, the build system would always attempt to detect an appropriate brew_prefix and then add Homebrew include and lib search paths

This causes issues when using Homebrew on Linux and attempt to build Netatalk on the side, or when building MacPorts packages on macOS

Therefore, Homebrew builds are now opt-in with the new -Dwith-homebrew boolean meson option

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -395,6 +395,7 @@ jobs:
         run: |
           meson setup build \
             -Dbuildtype=release \
+            -Dwith-homebrew=true \
             -Dwith-tests=true \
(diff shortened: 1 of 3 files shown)


rdmark:
@EricFromCanada Heads-up that I plan to introduce this breaking change for Homebrew builds in the next netatalk release version.

In practice, it means you have to pass `-Dwith-homebrew=true` when building against Homebrew libraries.

Please let me know if you have any thoughts or concerns!

PR #2197, opened 2025-05-31
spotlight: Use idiomatic NULL check to avoid compiler warning on Solaris

rdmark:
Rather than casting NULL as a pointer, use the idiomatic implicit comparison in the control structure

Before, compiling on Solaris with gcc 14.2.0 led to:

```
# meson compile -C build
INFO: autodetecting backend as ninja
INFO: calculating backend command to run: /usr/bin/ninja -C /root/netatalk/build
ninja: Entering directory `/root/netatalk/build'
[93/366] Compiling C object etc/spotlight/liblibspotlight.a.p/sparql_parser.c.o
sparql_parser.y: In function ‘isodate2unix’:
sparql_parser.y:156:48: warning: comparison between pointer and integer
[134/366] Compiling C object etc/spotlight/srp.p/sparql_parser.c.o
sparql_parser.y: In function ‘isodate2unix’:
sparql_parser.y:156:48: warning: comparison between pointer and integer
[366/366] Linking target test/afpd/afpdtest
```

### DIFF
--- a/etc/spotlight/sparql_parser.y
+++ b/etc/spotlight/sparql_parser.y
@@ -153,7 +153,7 @@ static time_t isodate2unix(const char *s)
 {
     struct tm tm;
 
-    if (strptime(s, "%Y-%m-%dT%H:%M:%SZ", &tm) == NULL)
+    if (!strptime(s, "%Y-%m-%dT%H:%M:%SZ", &tm))
         return (time_t)-1;
     return mktime(&tm);
 }


cheesestraws:
Hmmm, if so, this is a bit weird and naughty of Solaris here, since POSIX mandates that NULL should be of pointer type.  And there are also comparisons like this all over the rest of the codebase that presumably are not generating the warnings (? I don't actually know this, I don't have access to a Solaris box to compile on).

I do note, though, that `etc/spotlight/sparql_parser.y` doesn't include stddef.h, where NULL is meant to be defined.  So it might be worth trying to include that to see whether that makes the warning go away.

In general, I prefer (though this is probably just personal taste) explicit NULL tests on pointers, although I don't think it affects readability a huge amount here; so if fixing the missing header file doesn't make the warning go away, so be it...

cheesestraws (approved):
That's doubly weird, then.  Never mind, this is readable.

I may well have a go at setting up Solaris - I think it's supported on the VM host I use, I just have never quite got around to it.  Maybe doing more netatalk stuff will, weirdly, give me the energy I need to do it...

(Please keep adding me into reviews btw, at present this is being both interesting and good for my mental health)

PR #2199, opened 2025-05-31
Reformat yacc code with astyle

rdmark:
Ran the project astyle rules against the yacc code, and then some manual touch-up of overcorrection due to astyle not understanding yacc properly

Therefore we will not run continuous styling of this code

### DIFF
--- a/etc/spotlight/sparql_parser.y
+++ b/etc/spotlight/sparql_parser.y
@@ -1,45 +1,45 @@
 %{
-  #include <stdbool.h>
-  #include <stdio.h>
-  #include <string.h>
-  #include <time.h>
+#include <stdbool.h>
+#include <stdio.h>
+#include <string.h>
+#include <time.h>
 
-  #include <gio/gio.h>
+#include <gio/gio.h>
 
-  #include <talloc.h>
+#include <talloc.h>
 
-  #include <atalk/errchk.h>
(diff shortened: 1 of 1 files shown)


etc/spotlight/sparql_parser.y
```diff
@@ -85,45 +90,61 @@ expr                           {
...
+    {
         $$ = talloc_asprintf(ssp_slq, "%s", $1);
+    }
 }
 | match                        {$$ = $1; if ($$ == NULL) YYABORT;}
 | function                     {$$ = $1;}
 | OBRACE expr CBRACE           {$$ = talloc_asprintf(ssp_slq, "%s", $2);}
 | expr AND expr                {
-    if (!ssp_slq->slq_allow_expr) {
+    if (!ssp_slq->slq_allow_expr)
```
cheesestraws, line 121: are we intentionally using a different brace style here than in the rest of the C? (e.g., if we're using 1tbs everywhere else, {s should be on the same line as ifs)

rdmark: good catch; not sure why astyle didn't apply the braces rule consistently

rdmark: manually fixed all instances of if and else

cheesestraws (approved):

PR #2200, opened 2025-05-31
Use the pl file name suffix for installable Perl scripts

rdmark:
For historical reasons, the Perl scripts that get processed for installation had the .in file name suffix

This uses the standard .pl file name suffix instead, which has several upsides, not least that they're more easily identified as Perl scripts

Naturally, since the perl shebang is a placeholder before meson has processed them, you have to explicitly pass them to a perl interpreter when invoking them from the source tree


rdmark:
@cheesestraws FYI my ulterior motive for this change is to apply Perl beautification using `perltidy`. For this purpose I need to be able to glob for perl files.

PR #2201, opened 2025-05-31
Reformat Perl source code with perltidy

rdmark:
Add support for formatting Perl code to the codefmt shell script, leveraging perltidy

Run the formatting on the entire Perl codebase, including the webmin module

See the .perltidyrc file for a list of custom rules that are applied; a full list of rules can be found in the documentation:

https://metacpan.org/dist/Perl-Tidy/view/bin/perltidy


contrib/shell_utils/asip-status.pl
```diff
@@ -28,617 +28,584 @@
...
-	}
-
-	return ( $host, $port );
-}
+sub parse_hostport {
+    my ($hostport, $port) = @_;
+    my ($host);
+
+    if (($port) && ($hostport =~ /\[\S+\]/)) {
+
```
cheesestraws: These superfluous newlines after ifs are a bit odd - I noticed them in addump.pl too.  Is this a perltidy rule doing this?

contrib/shell_utils/asip-status.pl
```diff
@@ -28,617 +28,584 @@
...
+            my ($utf8name) = extract(1, $utf8_name_offset + 1, @packet);
+            print "UTF8 Servername: $utf8name\n";
+        }
+        $offset += 2;
+    }
+
+    # Undocumented Bit12
+    if ($flags & (1 << 12)) {
+
+        # looks like a SPNEGNO negTokenInit blob with the extensions from MS-SPNG
```
cheesestraws, line 343: ... is it something to do with ifs that start with a comment?

cheesestraws (approved):
I can't see anything that looks semantics-altering in here, though admittedly it's a bit of an epic diff so my eyes were glazing over by the end - one note on some ifs looking a bit weird but nothing blocking I think

contrib/shell_utils/asip-status.pl
rdmark: you're right, perltidy seems to want to inject a newline before every comment

let me see if there's a rule to override this behavior

contrib/shell_utils/asip-status.pl
rdmark: I found a rule that overrides this behavior: `--noblanks-before-comments`

contrib/shell_utils/asip-status.pl
rdmark: found the rule --noblanks-before-comments which has been applied, and reformat commit refreshed

rdmark:
Ran sanity tests on all the scripts. Let's ship it!

PR #2204, opened 2025-06-01
Restore the correct ouput of the Unicode precomposition script

rdmark:
A previous commit added two whitespace indentations to the precomposition tables, so we need to also adjust regex and substring index in the subsequent operations


rdmark:
@cheesestraws Ran into this one corner case where changing the code style introduced a bug. This script is doing some pretty rigid regex matching and operations on substrings at a static index, which obviously breaks when introducing indentation.

rdmark:
I think the fundamental problem with this script is that it first formats the Unicode entity data into C code, then operates on the C code.

PR #2205, opened 2025-06-01
Separate scripts into bin_utils and scripts depending on purpose

rdmark:
This creates a clear separation between *bin_utils* which are end user scripts that get packaged and installed,
with *scripts* which are meant for project internal or advanced user purposes

The changeset also contains an overhaul of the codefmt script, which (re)introduces proper non-verbose and verbose modes


rdmark:
@cheesestraws I would like to hear if you have another suggestion for how to organize these scripts.

rdmark:
Then let's leave this open until tomorrow. Maybe a better idea will present itself.

rdmark:
How about "utility_scripts"?

rdmark:
Strike that; here's a good one I think :)

- "bin_utils" for the installable Perl scripts
- "scripts" for the rest

What do you think?

cheesestraws:
Yeah, let's go with that

PR #2206, opened 2025-06-01
Add support for macOS in the macusers script

rdmark:
Defines a macOS compatible ps format string so that the script can be run on this OS as well


cheesestraws:
Why not just use the existing BSD one directly above this one?  As far as I can see, the output of the ps command is exactly the same on macOS and *BSD. 

(The only difference between your darwin command and the BSD one is that the darwin one truncates the command column, and the BSD one does not because it has ww in its options)

rdmark:
Well spotted. That's even better. Update pushed.

cheesestraws (approved):

PR #2207, opened 2025-06-01
addump.pl: enable 'use strict'

cheesestraws:
This PR achieves three main goals:

- Enables the strict pragma for this script, which will make it considerably easier to work in, and fixes implicit global variable scopes that 'strict vars' didn't like (and which we shouldn't like, either).
- Fixes an old bug where the extended attributes file handle wouldn't actually be closed (I think).  Perl's close only takes one file handle; the second parameter would have been ignored, which is good, because it also I think would have got flattened into a string constant (thanks, the 1990s).
- Makes the trauma-ridden part of me that twitches every time it sees a perl script without 'use strict' at the top rest a little calmer ;-)

### DIFF
--- a/contrib/shell_utils/addump.pl
+++ b/contrib/shell_utils/addump.pl
@@ -436,10 +436,7 @@ exit 0;
 
 sub filedatesdump {
     my ($ofst, $len) = @_;
-    my ($datedata);
-    my ($i);
-    my ($datestr);
-    my $buf;
+    my ($datedata, $datestr, $buf);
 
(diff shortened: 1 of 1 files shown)


rdmark (approved):
Looks a lot safer now, cheers.

rdmark:
AFAICT it's only the libatalk/unicode/make-precompose.h.pl script that's not using the strict pragma yet.

And the entire webmin module codebase too of course, but that's a can of worms that I'm not ready to open yet.

PR #2210, opened 2025-06-02
meson: Use single array option to control CNID backends to build

rdmark:
Rather than atomic boolean meson options, use a single option that takes a comma separated array to enable the backends

Usage examples:

-Dwith-cnid-backends=dbd,last,mysql

-Dwith-cnid-backends=dbd

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1466,9 +1466,9 @@ cnid_backends = ''
 
 # Determine whether or not to use Database Daemon CNID backend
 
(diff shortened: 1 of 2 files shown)


rdmark:
This is partially in preparation for the sqlite CNID backend.

cheesestraws (approved):
Looks reasonable to me (but my meson-fu is not strong)

PR #2211, opened 2025-06-02
Use strict code validation in precompose.h Perl script

rdmark:
This consistently adds the `my` keyword to all variables to give them local scope, while fixing one typo that caused a line to be dropped from the resulting header file. Now we can enable both strict and warning pragmas for the Perl interpreter for peace of mind.


rdmark:
@cheesestraws This wasn't too bad! Please let me know if you spot other improvement areas.

cheesestraws (approved):

cheesestraws:
Woo!  Yes, I wasn't sure what the script was actually doing, which possibly made the maze of twisty little temporary files all alike look a bit scarier than it actually is...

IIRC, this is the last perl script in the repo that wasn't run in strict mode, which is a nice milestone.

rdmark:
Please don't be startled, but there is actually one final bastion of dodgy Perl code: the webmin module :)

This should take care of it https://github.com/Netatalk/netatalk/pull/2212

cheesestraws:
aghhh I forgot about the webmin module, possibly out of self-preservation.

I'll have a look in the morning :-)

PR #2212, opened 2025-06-02
Apply the strict and warning pragmas to webmin module Perl code

rdmark:
In order to be compliant with the strict pragma, a number of minor logic bugs and archaic syntax have been fixed

- Use properly scoped variables everywhere
- Explicitly declare global webmin variables in each file
- Don't use bareword file handlers
- Return the full %in hash in one place


rdmark:
The strict pragma revealed a handful of bugs, but overall the state of the code wasn't as bad as I had feared!

cheesestraws (approved):
Ooh yeah, that wasn't that bad, was it!  Nice one.

rdmark:
As with the other Perl scripts, I wish I'd done this work years ago, too!

Debugging webmin module code when running in webmin is an absolute nightmare.

PR #2217, opened 2025-06-08
Beautify shell scripts with shfmt

rdmark:
Create an .editorconfig code styling config file with shfmt compatible directives, while adding support for shell formatting in the codefmt.sh shell script

The codefmt.sh script has been refactored to be POSIX compatible, which enables us to enforce POSIX standards on all shell scripts in this repo

Finally, run shfmt with the new rules against all shell scripts in the repo to beautify them

These formatting rules are now enforced in the Tests CI workflow

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,10 @@
+[*.sh]
+indent_style       = space
+indent_size        = 4
+shell_variant      = posix
+binary_next_line   = true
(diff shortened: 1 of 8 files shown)


cheesestraws (approved):
Looks reasonable to me, but I haven't used shfmt before so I wouldn't necessarily spot any issues...

PR #2218, opened 2025-06-08
Rewrite ICN hex to C byte array script from Python to Perl

rdmark:
This removes the final traces of Python code from this codebase, which in turn allows us to disable static analysis of Python in the SonarQube Cloud configuration


rdmark:
Truly! When I joined this project we had a good amount of Python code, but it's all been deprecated or rewritten now. I've grown to appreciate Perl for its portability and headache-free dependency management.

PR #2221, opened 2025-06-13
docs: Flesh out netatalk and dbd man pages

rdmark:
In the netatalk man page, describe the daemon structure in more detail

In the dbd man page, explain a few practical usage scenarios

### DIFF
--- a/doc/manpages/man8/netatalk.8.md
+++ b/doc/manpages/man8/netatalk.8.md
@@ -10,9 +10,18 @@ netatalk — Netatalk AFP server service controller daemon
 
 # Description
 
-**netatalk** is the service controller daemon responsible for starting and
-restarting the AFP daemon **afpd** and the CNID daemon **cnid_metad**. It is
-normally started at boot time by an init system.
+**netatalk** is the daemon used to control the Netatalk AFP file server,
+which is made up of multiple forking daemons.
+For most deployments you would use **netatalk** for centralized control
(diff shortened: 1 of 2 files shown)


doc/manpages/man8/netatalk.8.md
```diff
@@ -32,11 +41,11 @@ normally started at boot time by an init system.
 
 SIGTERM
 
-> Stop Netatalk service, AFP and CNID daemons
+> Stop the Netatalk AFP and CNID daemons.
 
 SIGHUP
 
-> Sending a *SIGHUP* will cause the AFP daemon reload its configuration
+> Sending a *SIGHUP* will cause the AFP daemon to reload its configuration
```
cheesestraws: Given the multiple daemons in play here, this phrasing is still a bit confusing, I think, and it might be better to specify the daemon(s) by name - is it only afpd that gets its configuration reloaded, or is it all child daemons of netatalk?

rdmark: In practice it means all child daemons of netatalk. It's only netatalk and afpd that have explicit handlers for SIGHUP, but netatalk will kill all cnid_* daemons in its own callback `sighup_cb()` if I am reading the code correctly.

rdmark: Hopefully this new phrasing should make it clearer. :)

cheesestraws (approved):
nice!

PR #2224, opened 2025-06-15
docs: Improve AppleTalk manual and developer readmes

rdmark:
- Move info on AppleTalk protocols from the DEVELOPER readme to the html manual
- Opening paragraphs in the DEVELOPER readme get a clearer intention
- Sample C code in DEVELOPER readme should follow our own style guide
- Call out which C compilers are recommended in INSTALL readme

### DIFF
--- a/DEVELOPER.md
+++ b/DEVELOPER.md
@@ -1,15 +1,15 @@
 Information for Netatalk Developers
 ===================================
 
-For basic installation instructions, see the Installation chapter
-in the html manual published on the [Netatalk homepage](https://netatalk.io)
-and the [Installation Quick Start Guide](https://netatalk.io/install).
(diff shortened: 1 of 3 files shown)


rdmark:
Our AppleTalk chapter in the manual is now consistently within the top 10 search results for the AppleTalk keyword on google, so my intention here is to make that page is bit more useful.

cheesestraws:
I might have a bash at cleaning up that AppleTalk text, if that sounds useful?  I'm not sure that talking about routed is going to be particularly useful any more, and worse, might result in someone actually using routed/RIP...

rdmark:
Yes please, that document is overdue for an overhaul by a real expert.

PR #2247, opened 2025-06-27
papd: Make handler function pointers and definitions C11/C23-compliant

rdmark:
- Update all handler function pointers in struct papd_comment and struct genquery to use full prototypes.
- Replace K&R-style and implicit function declarations with modern C prototypes.
- Flag parameter in cq_end() with unused macro

### DIFF
--- a/etc/papd/comment.h
+++ b/etc/papd/comment.h
@@ -8,11 +8,14 @@
 
 #include <sys/types.h>
 
+struct papfile;
+struct sockaddr_at;
+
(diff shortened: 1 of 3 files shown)


rdmark:
@NJRoadfan I don't think this should have an averse effect on papd, but please keep an eye out for any side effects.

With this change, the whole codebase compiles cleanly with Apple clang (finally).

cheesestraws (approved):

Issue #2113, opened 2025-04-15
Selinux Policy Request


specializedtom:
**Is your feature request related to a problem? Please describe.**
By default, Fedora (and other flavours of Linux that use SeLinux) do not have an SeLinux policy . This presents a security issue, particularly for those people trying to get Mac's using Time Machine that are stuck using AFP. (eg. v12.7)

**Describe alternatives you've considered**
I've considered other non-SeLinux flavored Linux OS's, however considering other system applications I am also running, I would rather have SeLinux running on my server to isolate access from other running processes . 

If someone can write an SeLinux policy specific for Netatalk, that would help immensely. I am sure other Netatalk admin have also come across this issue.

rdmark:
This is a new area for me, so I had to do some reading. The most informative resource I could find is this [custom policy guide from Red Hat](https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/8/html/using_selinux/writing-a-custom-selinux-policy_using-selinux#creating-and-enforcing-an-selinux-policy-for-a-custom-application_writing-a-custom-selinux-policy).

IINM, in addition to a standard netatalk policy, you'd have to customize it to allow afpd (and perhaps other daemons too) access to all of the shared volume dirs that you define.

Anyways, no guarantees, but I'll try to tinker with these policies when I have time to spare.

rdmark:
An selinux policy has been added to contrib/selinux in the netatalk code tree in the main branch (bleeding edge development code). See the readme and run netatalk.sh to build and install the policy. Consider this alpha level software: should work in theory but hasn't been proven yet in a production environment. Please share your feedback if you get to try it out!

specializedtom:
Thanks for following through on this issue @rdmark . When I get a spare chance, I'll test the changes out and let you know the outcome.

PR #2248, opened 2025-06-28
Create SELinux policy for netatalk daemons

rdmark:
This is an attempt at creating an SELinux policy for the netatalk -> afpd -> cnid_metad -> cnid_dbd daemon set using the Red Hat guide.

https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/8/html/using_selinux/writing-a-custom-selinux-policy_using-selinux

### DIFF
--- a/distrib/selinux/README.md
+++ b/distrib/selinux/README.md
@@ -1,0 +1,43 @@
+# Netatalk SELinux Policy Module
+
+This directory contains a SELinux policy module for Netatalk that provides security contexts and rules for
+running Netatalk services on SELinux-enabled systems.
+
+## Quick Start
+
+To build and install the policy module:
+
+```shell
+./netatalk.sh
+```
+
+This will compile the policy, install it, and generate an RPM package for distribution.
+
+## Policy Overview
+
+The policy defines the following security contexts:
+
+- _netatalk_t_ - Domain for Netatalk processes
+- _netatalk_exec_t_ - Type for executable files
+- _netatalk_etc_t_ - Type for configuration files
+- _netatalk_var_lib_t_ - Type for variable data
+- _netatalk_lock_t_ - Type for lock files
+- _netatalk_log_t_ - Type for log files
+
+## Troubleshooting
+
+If you encounter AVC denials, you can update the policy by running:
+
+```shell
+./netatalk.sh --update
+```
+
+This will analyze the audit log for new required permissions and offer to update the policy accordingly.
+
+## Further Reading
+
+For detailed information about SELinux policies and security contexts:
+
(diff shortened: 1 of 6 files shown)


rdmark:
@knight-of-ni I understand SELinux is primarily an RHEL and Fedora concern. Do you have any recommendations regarding the best practices for bundling and distributing SELinux policies?

I was pondering whether it would make sense to automate the building and enabling of the policy via Meson, e.g. introduce something like -Dwith-selinux that opts you in

knight-of-ni:
Ultimately, this new netatalk policy would need to be added to the fedora selinux policy package:
https://src.fedoraproject.org/rpms/selinux-policy

I have reached out to fedora selinux group, to learn what that workflow looks like. I'll let you know as soon as I have more information.

rdmark:
This would be much preferable, thanks for following up!

knight-of-ni:
I submitted a PR to selinux-policy, to add the policy files in this PR. 

You can monitor the progress here:
https://github.com/fedora-selinux/selinux-policy/pull/2863

rdmark:
~~Thank you! Let me discard this PR for now. We can continue the discussion in the issue ticket https://github.com/Netatalk/netatalk/issues/2113~~

Edit: Strike that, I see now that a decentralized policy is not discouraged: https://fedoraproject.org/wiki/SELinux/IndependentPolicy#Decentralised_SELinux_Policy

rdmark:
As per the discussion in https://github.com/fedora-selinux/selinux-policy/issues/2761, it is considered beneficial to ship an SELinux policy with the software package that it's for, in addition to the centralized repository that Redhat manages.

After review, I would like to merge this. I chose not to add meson support for compiling and installing the SELinux policy, but rather leave it up to the user to manually run the included netatalk.sh script.

I would expect most users to pull the policy from the centralized repository (when made available) so the version we include is more for reference and perhaps special cases where the user wants to customize the policy for their deployment needs.

knight-of-ni:
FYI, I am not willing to include this in the netatalk package in Fedora. I simply don't have the time to learn selinux well enough and then spend time troubleshooting it when the eventual bug report is filed against it. I am at my limits with the amount of time I am spending with Fedora package management (not just with netatalk).

Sorry for the hard stance, but that is why I submitted the PR to selinux-policy. Let them handle the issues.

Alternatively, if you could find another Fedora packager, I would gladly add them as an admin to the package.

andylemin (approved):
Nice! LGTM

rdmark:
@knight-of-ni right, by centralized repository I meant *selinux-policy* rather than the Fedora rpm.

PR #2254, opened 2025-06-30
Remove obsolete uint type definitions from endian.h

rdmark:
The endian.h header contains ancient compatibility type definitions which should be safe to remove now, leaving only endianness code in endian.h


cheesestraws (approved):
This PR gives me joy.

Are you following any specific rule as to whether to include/not include stdint.h?  It seems a little inconsistent - not that I've got any particular opinions on the point, more just for my own information

rdmark:
@cheesestraws No rules in place right now, I'm mainly importing the header when the compiler complains.

What do you think is a good standard? To import headers as close to the code that uses the symbols as possible, or hidden away in another global header file? Or something in between?

cheesestraws:
No idea, I was hoping you had one :-D 

Any rule that isn't enforceable automatically isn't going to be consistently used, anyway - it's too easy to get wrong.  I'm not sure we need a standard anyway, was mostly curious.

PR #2256, opened 2025-07-01
testsuite: Test ProDOS Info Bit in FPGetFileDirParams

rdmark:


rdmark:
@NJRoadfan Anything else you think we should test here?

NJRoadfan:
In addition to the current FPGetFileParams tests:

-Set a file with a Macintosh filetype of `APPL` and creator of `TEST`. This file should return a ProDOS type of `$00` and auxtype of `$0000` on a FPGetFileDirParams call since it has no valid mapping.
-Perform a FPGetFileDirParams call to a directory, it should return ProDOS type of `$0F` and auxtype of `$0200`.

Add a few FPSetFileParams tests:
-Have client set a file's ProDOS type to `$50` and auxtype to `$5445` (this maps to a GS/OS Teach document). Verify that the Macintosh filetype is `pPTE` and creator is `pdos`.
-Have client set a file's ProDOS type to `$04` and auxtype to `$0000` (this maps to a generic text file). Verify that the Macintosh filetype is `TEXT` and creator is `pdos`. This tests a special mapping case in Netatalk.

Reference for ProDOS information block packet is in "Inside AppleTalk Second Edition" pages 13-18 to 13-20.

rdmark:
@NJRoadfan Made good progress on adding the tests that you proposed. The first two are working, but the latter two I'm stuck at setting the PDINFO bits and then getting the correct Mac creator/type back. Could have a look and see if you spot any incorrect assumptions?

NJRoadfan:
Maybe an endianness issue? Looking at an actual ProDOS info bit set or read (Type `$50`, Aux Type `$5445`) in WireShark shows the following:

`50 00 45 54 00 00`

The test file may also need some FinderInfo attached to it. Weird things happen with Netatalk when there is no AppleDouble information on a file. This might be one of them! This is rarely a problem in actual AFP sessions since clients are aggressive in updating things like modification dates and filenames to force creation of AppleDouble EAs.

NJRoadfan:
A few notes:

-Currently the PR has the high and low bytes flipped for the aux type in test 434.
-The function `afp_filedir_unpack()` might need another FPGetFileDirParams call with just the ProDOS info bit set in the file bitmap before running.

rdmark:
> -Currently the PR has the high and low bytes flipped for the aux type in test 434.

Right, it's good to read the manual. :)

`Two special formats are designed to encode Pro DOS Info. The first is denoted by an f dType
made up of the letter 'p ' followed by a 1-byte ProDOS File Type and a 2-byte ProDOS Aux Type
(high order byte first). `

Which must be why you saw the two bytes flipped in Wireshark, no?

> -The function afp_filedir_unpack() might need another FPGetFileDirParams call with just the ProDOS info bit set in the file bitmap before running.

Is that something you have to do in AFP, always running FPGetFileDirParams before a FPSetFileParams of the same bitmap?

BTW, this prompted me to take a closer look at what's happening in afp_filedir_unpack() (afpcli.c); turns out there are a bunch of hacks for UTF-8 in FILPBIT_PDINFO branch. That might be why we're getting unexpected results with the ProDOS info bits.

NJRoadfan:
I suspected that there are issues in `afp_filedir_unpack()`. Apple reused that bit to indicate a UTF8 filename in AFP3.0, so handling is different. I just spitballing with the extra FPGetFileDirParams call just in case the data returned wasn't suitable for `afp_filedir_unpack()`. You may be able to pack that struct directly without calling `afp_filedir_unpack()` again avoiding any UTF8 code issues.

Netatalk's code is clearly seeing the request to alter the ProDOS information, but is receiving junk data (0x00 0x00 0x09) instead of the correct info.

rdmark:
Tried unpacking manually in the tests but the outcome is the same. I'll have to sleep on this one!

rdmark:
@NJRoadfan have you recently tested that the ProDOS type to Mac type logic works in netatalk? 

The Mac type assignment is happening in `setfilparams()` (file.c) and it looks sound at a glance. But the buffer that gets fed to the function has bad data in it. I haven't been able to track the rest of the callstack yet.

NJRoadfan:
Whatever build I'm using is working fine (VM is saying 4.2.0), including the "special" case of setting the filetype to TEXT. I have a Desk Accessory for GS/OS that allows one to read and write ProDOS filetype/auxtypes directly on any file.

NJRoadfan:
Has it been ruled out that the testsuite itself is sending packets with correct data? Running a packet sniffer between a machine running the testsuite and the Netatalk server would reveal this. Remember, garbage in-garbage out.

rdmark:
@NJRoadfan attaching a zipped pcap from when running test434, captured with CocoaPacketAnalyzer. Do you have pointers how to interpret the data? I'm weak when it comes to network analysis.

[07272025-230827.zip](https://github.com/user-attachments/files/21456153/07272025-230827.zip)

NJRoadfan:
If this is the output from the test suite, something is definitely wrong with the `FPSetFileDirParam()` command used to set ProDOS information (packet 27 in your saved dump). After byte 105 (pad), there should only be 6 additional bytes containing the ProDOS information you want to set on the given file.

It looks like the test suite is "helping" by creating a packet for a Unicode name somewhere. You can see the correct format of a ProDOS info packet by issuing a `FPGetFileDirParam()` command, just save the output of `test440()`. Note that WireShark will throw an exception error as the packet dissector only expects Unicode names in the result. This is normal.

rdmark:
Since it seems more and more likely that the testsuite AFP client will require some work, I spun off the PDINFO type conversion discussion to https://github.com/Netatalk/netatalk/issues/2299 

I propose we merge the partial test coverage in this PR for starters.

cheesestraws:
The code looks sensible structurally but I don't really understand AFP enough to have a worthwhile opinion on correctness.  Happy to approve it on this basis, so long as you know that that's what the basis is. :-)

rdmark:
@cheesestraws I'm happy to receive your code review based on your surface level reading of syntax and logic. :)

PR #2264, opened 2025-07-11
webmin: Convert UI icon images from gif to png

rdmark:
imagemagick was used to convert from the proprietary gif format to the open png format

This also brings back the traditional daemon icon to represent netatalk, appropriate since the module icon is only displayed when in Legacy mode in Webmin

For the record, icon.gif has to be in gif format due to hard coding in Webmin itself


rdmark:
FTR, this is what Webmin 2.x looks like when run in Legacy mode. This is a setting that's hard to find, and I doubt very many users activate it.

And as mentioned above, Legacy mode is the only time when the module icon.gifs are displayed. I think the old netatalk daemon icon looks better in this context than the new logo.

Screenshot 2025-07-12 at 9 32 25

cheesestraws (approved):
I actually really like the old icon.  There's something intensely relatable at the look of surprise the demon has while staring at its own network...

rdmark:
Agreed, it's a quite clever and icon design! I appreciate how it ties back to the 4.4BSD roots of netatalk.

PR #2265, opened 2025-07-11
Build and run netatalk without Berkeley DB

rdmark:
This changeset modifies both the build system and the netatalk executable to remove the hard dependency on Berkeley DB.

Regarding the build system piece: in the absence of a Berkeley DB library, it is now possible to build and use another CNID backend as the default. When the dbd backend is not built, we don't install the cnid_metad / cnid_dbd daemons or man pages.

In the netatalk controller daemon, when built without dbd support it will no longer attempt to launch the cnid_metad.

It introduces new logic for picking the default CNID backend which is more flexible, as well as printing the default backend in both the meson summary and in the output of afpd -V.

There is also surrounding refactoring of the build system for libgcrypt and the CNID backend libraries.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -39,14 +39,28 @@ documentation for more details.
 
 ### Required
 
+These are the libraries that are hard requirements for netatalk.
+
 | Package      | Details |
 |--------------|---------|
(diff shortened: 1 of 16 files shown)


rdmark:
I'm happy if you can comment on the C code portions. :)

PR #2266, opened 2025-07-11
tests: Use the sqlite CNID backend in the afpd integration tests

rdmark:
The sqlite backend is zero-configuration and run embedded in afpd so it's ideal for running the afpd tests

The last CNID backend can now be made permanently read-only again


rdmark:
problem is that the sqlite library gets dynamically linked so the test module cannot run independently anymore

it can surely be solved but I'll put it on the back burner

PR #2267, opened 2025-07-12
testsuite: appl and icon operations fail with mysql backend

rdmark:


rdmark:
Not sure why the Sonar code duplication heuristic gets triggered here. I think it can be safely ignored.

libatalk/cnid/mysql/cnid_mysql.c
```diff
@@ -877,10 +877,33 @@ struct _cnid_db *cnid_mysql_open(struct cnid_open_args *args)
...
     MYSQL_ROW row;
     struct vol *vol = args->cnid_args_vol;
     MYSQL *temp_conn = NULL;
+    snprintf(dirpath, sizeof(dirpath), "%sCNID/%s", _PATH_STATEDIR,
+             vol->v_localname);
+    become_root();
+
+    /* We need the volume dir under CNID to store .AppleDesktop,
+       although this backend does not keep any database file here. */
+    if (mkdir(dirpath, 0755) != 0) {
```
cheesestraws, line 891: It feels a bit weird to me that these permissions are hardcoded but I can't work out why it feels weird.  I can't think of any situations where they would need to be different, but I'm just flagging this up as a "hm, this feels odd" in case that's useful to you.

test/testsuite/spectest.c
```diff
@@ -365,7 +365,9 @@ void usage(char *av0)
     fprintf(stdout, "\t-7\tAFP 3.4 version (default)\n");
     fprintf(stdout, "\t-v\tverbose\n");
     fprintf(stdout, "\t-V\tvery verbose\n");
-    fprintf(stdout, "\t-x\tdon't run tests that require special setup\n");
+#if 0
```
cheesestraws: Is there a reason you're just #if 0-ing this instead of just removing it completely, since you've unconditionally removed it from the usage string above?

libatalk/cnid/mysql/cnid_mysql.c
rdmark: It's a bit icky, I agree. This time I followed what the dbd CNID backend does (specifically, when cnid_metad creates the db dir.) I followed suite when making the sqlite backend as well.

When it comes to the mysql backend, I have only observed afpd creating the .AppleDesktop dir there, but the tricky part with netatalk is that we have this cluster of daemons running as various users with various permissions attempting to write to either the shared volume dir or CNID dir, so it's kind of a headache to keep track of all potential scenarios.

This is my reasoning anyways. But it's a bad reason for not tightening up permissions further. ;)

Do you have a suggestion for safer permissions?

test/testsuite/spectest.c
rdmark: It's a minor thing, but it's meant for making it easy to bring the option back in the future while remembering to use the same letter code.

I can remove it for cleanliness sake!

rdmark: Removed. For future needs to skip tests, I say we introduce more fine-grained methods tied to certain capabilities.

libatalk/cnid/mysql/cnid_mysql.c
rdmark: Another consideration is that a volume can be converted from one CNID scheme to another, so having consistent permissions on the volume CNID dirs is important to avoid troublesome fail states.

cheesestraws: Nah, I think you're probably right here; it was a heuristic worry rather than an actual concrete worry.  Let's consider this resolved unless or until either of us has a better idea

cheesestraws (approved):

PR #2268, opened 2025-07-13
Introduce bstring as subproject and remove vendored bstring code

rdmark:
We now check for a system bstring dependency and use it instead of the previously vendored bstring library. But since bstring is currently not widely packaged downstream, we introduce a subproject in Meson that pulls in the [upstream](https://github.com/msteinert/bstring) library code when a system dependency wasn't found.

Moved wrappers and utility macros from the bstradd module to other parts of the codebase where they are used (no need to keep them as global symbols since they were locally used in a handful of places.)

Shore up error handling around a handful of bstring API calls


rdmark:
~~The git wrapper for the subproject doesn't work consistently, so I'm working with the upstream bstring project to cut a new release tarball to use instead.~~

PR has been updated with the wrap-file module in the meson subproject wrapper file, pointing to an upstream release tarball!

cheesestraws (approved):

PR #2269, opened 2025-07-14
cnid: Make the mysql CNID backend a bit more resilient

rdmark:
Addresses a range of code quality and convention issues, in part aligning with improvements in the sqlite backend

- Use preformatted SQL strings in execute function rather than dynamic formatting
- The mysql CNID module uses functions declared in unix.h, notable become_root() so let's include it explicitly
- Use standard while loop instead of goto statement
- Explicit type casting, remove unused variable, merge if statements, etc.


cheesestraws (approved):
Like all of this.

PR #2270, opened 2025-07-19
Link with shared bstring library as subproject

rdmark:
Pull in the bstring library from git, build and link it as a shared library; fallback to existent bstring if found

Consequently, remove the vendored version of bstring from netatalk


rdmark:
Depends on https://github.com/Netatalk/netatalk/pull/2268 being merged first, so keeping as draft for the time being.

PR #2273, opened 2025-07-20
Handle return codes from function calls consistently

rdmark:
Consistently handle return values from non-void functions, and exit the function/program early with logging on failure

This appeases the -Wunused-result compiler flag

Also includes a small meson change to always link with glib when building POSIX ACLs, not just on Linux.

### DIFF
--- a/etc/netatalk/netatalk.c
+++ b/etc/netatalk/netatalk.c
@@ -92,6 +92,7 @@ static int set_sl_volumes(void)
     EC_INIT;
     const struct vol *volumes, *vol;
     struct bstrList *vollist = bstrListCreate();
+    int sysret;
     bstring sep = bfromcstr(", ");
     bstring volnamelist = NULL, cmd = NULL;
     EC_NULL_LOG(volumes = getvolumes());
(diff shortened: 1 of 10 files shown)


rdmark:
Code duplication is unavoidable here I think, given that we use similar macros and patterns for error handling.

etc/netatalk/netatalk.c
```diff
@@ -190,7 +202,12 @@ static void sigterm_cb(evutil_socket_t fd _U_, short what _U_, void *arg _U_)
     event_del(sighup_ev);
     event_del(timer_ev);
 #ifdef WITH_SPOTLIGHT
-    system(INDEXER_COMMAND " -t");
+
+    if (system(INDEXER_COMMAND " -t") < 0) {
```
cheesestraws: It feels like this and the following should also test for > 0 in case the command fails?

etc/netatalk/netatalk.c
```diff
@@ -200,7 +217,12 @@ static void sigquit_cb(evutil_socket_t fd _U_, short what _U_, void *arg _U_)
 {
     LOG(log_note, logtype_afpd, "Exiting on SIGQUIT");
 #ifdef WITH_SPOTLIGHT
-    system(INDEXER_COMMAND " -t");
+
+    if (system(INDEXER_COMMAND " -t") < 0) {
```
cheesestraws: see note on line 206

etc/netatalk/netatalk.c
```diff
@@ -536,7 +558,12 @@ int main(int argc, char **argv)
         sleep(1);
         set_sl_volumes();
         LOG(log_note, logtype_default, "Starting indexer: " INDEXER_COMMAND " -s");
-        system(INDEXER_COMMAND " -s");
+
+        if (system(INDEXER_COMMAND " -s") < 0) {
```
cheesestraws: see note on line 206

etc/netatalk/netatalk.c
rdmark: Good call. Elsewhere in this file they use the `if (system(...) != 0)` pattern so I applied this consistently.

etc/netatalk/netatalk.c
rdmark: PR updated

etc/netatalk/netatalk.c
rdmark: PR updated

rdmark:
There's a failing spectest right now, not caused by this PR. Will fix separately.

> FPGetACL:test398: check ACL support

(on Alpine + sqlite only)

rdmark:
Also note that with recent changes to the workflows, the Static Analysis and Container jobs are run detached from PRs. You can find the results on the Actions page.

The reason for this chance is to allow bots and non-member PRs to run those jobs.

rdmark:
~~I realized now that these latest changes have led to the SonarQube analysis report no longer getting pushed to this PR page either. Let me see if I can fix that.~~

~~In the meantime, you can get those reports under Actions or in sonarcloud.io~~

I reverted back to the previous workflows structure. There were too many drawbacks and glitches with the workflow triggers.

etc/netatalk/netatalk.c
cheesestraws: Does errno get set when the command returns a non-zero exit status?  The man page I've got for 'system' here is a bit ... underspecified about when errno is set or not.

rdmark: A brief investigation confirms that erro is indeed not set when the shell command itself fails.

Pushed a new solution that checks for both scenarios and tells you when something goes wrong with a command. Sample log:

```
set_sl_volumes: command 'gsettings -X set org.freedesktop.LocalSearch3 index-recursive-directories "['/mnt/afp', '/mnt/afpbackup', '/mnt/afpshare']"' exited with status 1
```

rdmark:
Thanks for the review! I updated the PR message to justify the second commit.

PR #2275, opened 2025-07-20
Properly support LocalSearch and remove Tracker < v3 support

rdmark:
Tighten up dependency checks for LocalSearch and TinySPARQL in meson

When detected, use TinySPARQL header and LocalSearch API, rather than the deprecated Tracker backward compatibility fallbacks

This fixes a bug where LocalSearch failed to register a D-Bus service using a too old D-Bus name (pre-Tracker3)

At the same time, remove compatibility checks and code for Tracker older than v3, which have been deprecated since 2020

The obsoleted meson options -Dwith-tracker-install-prefix and -Dwith-tracker-prefix have been removed as well

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -260,6 +260,7 @@ jobs:
             libdb-devel \
             libgcrypt-devel \
             libtalloc-devel \
+            localsearch \
             mariadb-connector-c-devel \
             meson \
             ninja-build \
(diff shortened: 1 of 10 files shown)


cheesestraws (approved):
This looks reasonable but to warn you, my knowledge of LocalSearch is somewhat south of zero, so all I can do here is a sanity check

rdmark:
That's fair. Apart from the D-Bus name and included header, the logic should not have changed at all here. My reading of the GNOME docs, they didn't noticeably change the internal API between tracker3 and localsearch. It's a naming change.

PR #2276, opened 2025-07-20
docker: Remove support for Spotlight compatible search

rdmark:
Since the LocalSearch/Tracker indexer is running on the D-Bus, we won't get the behavior we want when in a container context

When using the D-Bus inside the container, we run into issues with privileges (which may be masking some other constraints) while when using the host D-Bus we end up with other issues such as LocalSearch/Tracker version mismatch between the host and the container

Therefore, let's build the containerized netatalk without Spotlight support altogether, to be revisited if we introduce another mechanism for indexing (which is something that has been discussed)


rdmark:
As an additional boon, we will get lighter containers since the LocalSearch/Tracker stack has a lot of dependencies

cheesestraws (approved):

PR #2278, opened 2025-07-21
GitHub CI: Container test runs and static analysis in separate workflows

rdmark:
The container test runs and static analysis are now a separate respective workflows that are triggered by other workflows

This should allow an unprivileged service user such as dependabot to trigger the container test runs that rely on repo secrets

As per the GitHub security model, service users aren't allowed to access repo secrets


rdmark:
The workflows will only trigger if the workflow files exist on the main branch, so this has to be merged to be tested

https://docs.github.com/en/actions/reference/events-that-trigger-workflows#workflow_run

PR #2279, opened 2025-07-21
GitHub CI: Arrange for all privileged jobs to run as workflow triggers

rdmark:
A few tweaks to the workflows so that only non-privileged jobs are triggered directly by the PR, with the rest triggering indirectly to grant them access to secrets

Run the container test runs only on container build success: Only a successful container build workflow will result in usable containers

Containers workflow triggered off of the Tests workflow: Also loosens up the trigger criteria for other workflows

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -8,9 +8,7 @@ on:
       - "**.md"
       - "**/COPYING"
       - "**/README*"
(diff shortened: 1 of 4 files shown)


rdmark:
This cannot be tested without merging to main, so skipping code review

PR #2284, opened 2025-07-21
Move ftw code from libatalk into the ad binary

rdmark:
The file tree walker functions are only used narrowly in the ad binary, so let's move it out of the shared libatalk library to reduce bloat

Removes the ftw.h header file from a few source files where the symbols aren't used

At the same time, fix a handful of code smells and security hotspots reported by SonarQube

### DIFF
--- a/bin/ad/ftw.c
+++ b/bin/ad/ftw.c
@@ -113,7 +113,6 @@
 # define NFTW_NAME nftw
 # define NFTW_OLD_NAME __old_nftw
 # define NFTW_NEW_NAME __new_nftw
-# define INO_T ino_t
 # define STAT stat
 # define LXSTAT(V,f,sb) lstat (f,sb)
 # define XSTAT(V,f,sb) stat (f,sb)
@@ -137,7 +136,7 @@ struct dir_data {
 
 struct known_object {
     dev_t dev;
-    INO_T ino;
+    ino_t ino;
 };
 
 struct ftw_data {
@@ -163,7 +162,7 @@ struct ftw_data {
     const int *cvt_arr;
 
     /* Callback function.  We always use the `nftw' form.  */
-    NFTW_FUNC_T func;
+    nftw_func_t func;
 
     /* Device of starting point.  Needed for FTW_MOUNT.  */
     dev_t dev;
@@ -756,7 +755,7 @@ static int ftw_startup(const char *dir,
        every case the callback using the format of the `nftw' version
        and get the correct result since the stack layout for a function
        call in C allows this.  */
-    data.func = (NFTW_FUNC_T) func;
+    data.func = (nftw_func_t) func;
     /* Since we internally use the complete set of FTW_* values we need
        to reduce the value range before calling a `ftw' callback.  */
     data.cvt_arr = is_nftw ? nftw_arr : ftw_arr;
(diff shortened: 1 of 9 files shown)


rdmark:
@cheesestraws I did sanity testing of all of the `ad` commands. Nothing is malfunctioning AFAICT.

While doing this I noticed that the `ad find` command isn't documented, so filed https://github.com/Netatalk/netatalk/issues/2296 to follow up.

Anyways, this is ready for your review. :)

bin/ad/ftw.c
```diff
@@ -293,13 +293,14 @@ open_dir_stream(int *dfdp, struct ftw_data *data, struct dir_data *dirp)
...
-            const size_t MAX_BUFFER_SIZE = 1024 * 1024; /* 1MB buffer limit */
+            const size_t MAX_ENTRY_SIZE = 255;
+            const size_t MAX_BUFFER_SIZE = 1024 * 1024;
 
             while ((d = __readdir64(st)) != NULL) {
                 size_t this_len = NAMLEN(d);
 
-                if (this_len > UTF8FILELEN_EARLY) {
-                    this_len = UTF8FILELEN_EARLY;
+                if (this_len > MAX_ENTRY_SIZE) {
```
cheesestraws: what's this doing?  I assume it's doing something sensible, I just don't know what UTF8FILELEN_EARLY means, or why it's being replaced

cheesestraws (approved):
Nice tidying up!  I've left one question for my own curiosity, but it doesn't block merge I think.

bin/ad/ftw.c
rdmark: This is ultimately me not being entirely sure what a reasonable max file name length is in this context.

My first attempt used an arbitrary 4096 maximum: https://github.com/Netatalk/netatalk/commit/0773cec53258accae0cc45acefb6c3f543d92735

Then in the second attempt I thought UTF8FILELEN_EARLY (which resolves to 255) is reasonable as the first iteration of OSX file name max length: https://github.com/Netatalk/netatalk/commit/2c289b9bfa5db94a93c5ff21ce9c670e01c2074d

And now when decoupling the ftw.c code from libatalk we don't have access to UTF8FILELEN_EARLY anymore so I'm defining a local constant MAX_ENTRY_SIZE to the same 255 value, which means the logic doesn't change.

Do you have a suggestion how to do this more gracefully and accurately?

rdmark:
By moving code out of libatalk (iniparser, bstring, ftw) we have slimmed down the compiled shared library by about 100KB.

When libatalk is built on Debian Trixie without appletalk, in bytes:

4.3dev: 1088568
4.2: 1112632
4.1: 1188552

It's still a pretty phat library at 1.1MB

Edit: Building with appletalk adds about 145k to the above

bin/ad/ftw.c
cheesestraws: Nope, not at all - the maximum length of something is inevitably going to be a magic number.  If it's intended to be the max file name length under OS X, though, I might suggest renaming the constant to be something like MAXFILENAMELEN or something, since 'MAX_ENTRY_SIZE' is a bit opaque in terms of semantics (this feels a bit like bikeshedding; if so, sorry!)

rdmark: This is a good point, and if nothing else it makes the code more readable to future readers which is always a good thing. Changeset updated!

PR #2288, opened 2025-07-22
meson: Add glib runtime dependency for ACLs

rdmark:
Notably on Alpine Linux, for ACLs to function you need the GNOME glib runtime dependency

This hasn't triggered before because the glib dependency has been linked with implicitly through other features such as Spotlight, but when disabling Spotlight it came to light

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -52,6 +52,7 @@ jobs:
             dbus-dev \
             flex \
             gcc \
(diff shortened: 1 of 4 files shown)


rdmark:
@cheesestraws After merging this, the spectest failure that blocks the other PRs should get resolved.

cheesestraws (approved):

PR #2297, opened 2025-07-27
docs: Document ad find command in man page

rdmark:
The find subcommand for ad was the last one added back in 2010, but the man page was never updated at the time

This adds a section on ad find to the ad(1) man page

### DIFF
--- a/doc/manpages/man1/ad.1.md
+++ b/doc/manpages/man1/ad.1.md
@@ -4,7 +4,7 @@ ad - AppleDouble file utility suite
 
 # Synopsis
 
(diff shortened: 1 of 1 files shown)


NJRoadfan:
I'm starting to wonder if this tool is actually finished. Note that it only works with the BDB CNID backend. No code was ever written for searches on the other backends.

rdmark:
Right, it might be worth putting a big disclaimer in the man page about known limitations with this tool.

cheesestraws (approved):

PR #2298, opened 2025-07-27
docs: Split out Legal Notices into a separate chapter

rdmark:
The legal notices have outgrown the Introduction chapter, so let's keep them on its own page at the end of the manual

Add the legal notice for glibc (the ftw.c code)

Change to preformatted text for all legal notice paragraphs

Sort the legal notices alphabetically, except for the original netatalk notice which is listed first


rdmark:
Frustratingly, the test failure in AFP spec test (cnid:sqlite) AFP 3.4 - Alpine (Prod) is still recurring intermittently, in maybe about 20% of CI runs. Always the same spectest _FPGetACL:test398: check ACL support_ that's failing. Currently at a loss what's causing it. :-/

It started occurring when I stopped building the production container with Spotlight support.

Edit: Also worth noting that I have yet to succeed in reproducing the issue locally.

cheesestraws (approved):

PR #2309, opened 2025-07-28
docker: Build production container without ACL support

rdmark:
ACLs have proven flaky when in a container context, so since there arguably is little utility for ACLs within a docker deployment let's disable it for now


rdmark:
Merging without review to stop the pipelines from failing intermittently

PR #2315, opened 2025-07-30
Update Japanese localization for netatalk 4.3 release

rdmark:
This should take care of all translatable strings, while refreshing gettext references and touching up line endings

Also includes a refresh of the compilation readme


cheesestraws (approved):
Since I neither understand Japanese nor the localisation framework, this approval may be a matter of form largely.  But the PR looks sensible as far as I understand it...

PR #2321, opened 2025-08-03
testsuite: Prevent potential buffer overflow in FPGetFileDirParms test

rdmark:
Use snprintf to write to buffer with enforced max length, and bail out if the operation isn't successful,
which should make gcc 13.3.0 on Ubuntu 24.04 happy


rdmark:
@cheesestraws one final fix for the release. do you have time for code review? :)

cheesestraws (approved):

PR #2324, opened 2025-08-04
docker: Build production container with dbd and mysql CNID backends

rdmark:
Explicitly build the dbd and mysql CNID backends, and exclude last and sqlite

The last backend is only used for testing, and sqlite is experimental

We will bring the sqlite backend back in the container once it's flagged as production ready

Also includes a minor fix for the container CI workflow, to properly skip one last test run with dependabot


rdmark:
Merging to unblock pipeline for dependabot

PR #2331, opened 2025-08-04
docs: Create separate man pages for each afptest tool

rdmark:
Previously, all of the documentation for the test tools in the afptest testsuite were documented in a single afptest.1 man page

This changeset introduces a separate man page for each tool, each with more thorough coverage of options and usage patterns

The afp_lantest man page which was introduced earlier gets a minor overhaul to consolidate information and style


andylemin:
Consistency is nice 😉

PR #2351, opened 2025-08-11
Introduce yaml code style guide and automatic formatting

rdmark:
Apply yaml formatting rules based on the [yamlfmt style guide](https://github.com/google/yamlfmt/blob/main/docs/config-file.md) with the following configuration:

```
formatter:
  type: basic
  eof_newline: true
  include_document_start: false
  include_document_end: false
  indent: 2
  indentless_arrays: false
  line_ending: lf
  max_line_length: 120
  pad_line_comments: 1
  retain_line_breaks_single: true
```

- Introduce a configuration file for yamlfmt
- Add support for formatting yaml files to codefmt.sh
- GitHub CI: Check for yaml coding style compliance in test workflow
- GitHub CI: Remove redundant steps when building release webmin container
- GitHub CI: Consistently leave records of version tags for each action

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -33,7 +33,7 @@ jobs:
     container:
       image: alpine:latest
     steps:
-      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
       - name: Install dependencies
         run: |
           apk add \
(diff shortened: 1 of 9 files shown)


rdmark:
@cheesestraws Do you have any feedback regarding the yaml style guide I have landed on here?

The number of lines of yaml code has increased rapidly, so I think it's time to put it under automatic stylistic control now. :)

cheesestraws (approved):
This looks pretty decent and a reasonable set of defaults to me.

Man, YAML is a mess, isn't it.

rdmark:
Thanks for the review!

It's easy to grow a distaste for any language where indentation is part of the syntax, when you start dealing with a larger codebase.

cheesestraws:
Yes - but it's not just that indentation can be a part of the syntax, although that's ew, it's that indentation can mean multiple different things depending on punctuation that might be many lines away, which just... is the worst of both worlds.

PR #2354, opened 2025-08-13
testsuite: afp_lantest, Added Linux IO monitoring to afp_lantest and refactored results display

andylemin:
Key Changes:
- Added afp_lantest IO monitoring module to track disk operations for afpd/cnid_dbd processes when run on the same host as Netatalk
- Tracks read/write operations via /proc/[pid]/io during test execution
- Refactored 400+ line displayresults() into 4 functions
- Added TOTAL_AFP_OPS constant (80,686) and AFP operation counts per test
- Added aggregates summary with total/average times and throughput metrics
- Fixed is_there() calls to include volume parameter
- Added comprehensive documentation
- Docker: Added IO_MONITORING env variable and auto-mount of secondary /proc_io mount for systemd distributions
- Build: Added lantest_io_monitor.c to meson.build
- Updated afp_lantest docs and translation

This enables developers to quantify IO impact of AFP operations and validate performance optimizations.
IO monitoring is optional and backward compatible.

Statistics: 7 files changed, +1,392 lines (1,672 insertions, 280 deletions)

```
% docker run -it --rm --privileged --network host --cap-add=NET_ADMIN --volume "/Volumes/case-sensitive/test:/mnt/afpshare" --volume "/Volumes/case-sensitive/test:/mnt/afpbackup" --env IO_MONITORING=1 --env TESTSUITE=lan --env TEST_FLAGS="-n 2" --env AFP_USER=test --env AFP_PASS=test --env AFP_GROUP=test --env SHARE_NAME='File Sharing' --env INSECURE_AUTH=true --name netatalk-test netatalk-test:latest
*** Setting up environment
*** Setting up users and groups
*** Configuring shared volume
*** Fixing permissions
*** Removing residual lock files
*** Configuring Netatalk
NOTE: Set the `ATALKD_INTERFACE' environment variable to start DDP services.
Successfully created /proc_io with hidepid=0,gid=0 for I/O monitoring (gid=0 as TESTSUITE tests run as root)
*** Starting AFP server
*** Running testsuite: lan

+ afp_lantest -n 2 -7 -h 127.0.0.1 -p 548 -u test -w test -s 'File Sharing'
Connecting to host 127.0.0.1:548
IO monitoring: /proc_io is available
Looking for cnid_dbd processes with -u test in command line
Found cnid_dbd process: PID 40
Looking for afpd processes owned by user 'test' (UID: 1000)
Found privilege-dropped afpd process: PID 36
IO monitoring enabled (afpd: 36, cnid_dbd: 40)

Run 1 => Open, stat and read 512 bytes from 1000 files [8,000 AFP ops]        1923 ms
         IO Operations; afpd: 6000 READs, 7002 WRITEs | cnid_dbd: 0 READs, 2 WRITEs
Run 1 => Writing one large file [103 AFP ops]                                  136 ms for 100 MB (avg. 771 MB/s)
         IO Operations; afpd: 0 READs, 299 WRITEs | cnid_dbd: 0 READs, 0 WRITEs
Run 1 => Reading one large file [102 AFP ops]                                   39 ms for 100 MB (avg. 2688 MB/s)
         IO Operations; afpd: 100 READs, 100 WRITEs | cnid_dbd: 0 READs, 0 WRITEs
Run 1 => Locking/Unlocking 10000 times each [20,000 AFP ops]                   799 ms
         IO Operations; afpd: 0 READs, 20000 WRITEs | cnid_dbd: 0 READs, 0 WRITEs
Run 1 => Creating dir with 2000 files [4,000 AFP ops]                         4061 ms
         IO Operations; afpd: 2000 READs, 10005 WRITEs | cnid_dbd: 4 READs, 6150 WRITEs
Run 1 => Enumerate dir with 2000 files [~51 AFP ops]                           637 ms
         IO Operations; afpd: 1960 READs, 49 WRITEs | cnid_dbd: 0 READs, 0 WRITEs
Run 1 => Deleting dir with 2000 files [2,000 AFP ops]                         3176 ms
         IO Operations; afpd: 4000 READs, 4004 WRITEs | cnid_dbd: 2 READs, 6104 WRITEs
Run 1 => Create directory tree with 1000 dirs [1,110 AFP ops]                 1885 ms
         IO Operations; afpd: 0 READs, 4445 WRITEs | cnid_dbd: 4 READs, 2351 WRITEs
Run 1 => Directory cache hits (100 dirs + 1000 files) [11,000 AFP ops]        3625 ms
         IO Operations; afpd: 10000 READs, 11100 WRITEs | cnid_dbd: 0 READs, 100 WRITEs
Run 1 => Mixed cache operations (create/stat/enum/delete) [820 AFP ops]       1134 ms
         IO Operations; afpd: 820 READs, 1621 WRITEs | cnid_dbd: 0 READs, 1201 WRITEs
Run 1 => Deep path traversal (nested directory navigation) [3,500 AFP ops]     965 ms
         IO Operations; afpd: 2500 READs, 3550 WRITEs | cnid_dbd: 0 READs, 50 WRITEs
Run 1 => Cache validation efficiency (metadata changes) [30,000 AFP ops]      8529 ms
         IO Operations; afpd: 30000 READs, 30100 WRITEs | cnid_dbd: 0 READs, 100 WRITEs
Run 2 => Open, stat and read 512 bytes from 1000 files [8,000 AFP ops]        2453 ms
         IO Operations; afpd: 6000 READs, 7002 WRITEs | cnid_dbd: 0 READs, 2 WRITEs
Run 2 => Writing one large file [103 AFP ops]                                   87 ms for 100 MB (avg. 1205 MB/s)
         IO Operations; afpd: 0 READs, 299 WRITEs | cnid_dbd: 0 READs, 0 WRITEs
Run 2 => Reading one large file [102 AFP ops]                                   36 ms for 100 MB (avg. 2912 MB/s)
         IO Operations; afpd: 100 READs, 100 WRITEs | cnid_dbd: 0 READs, 0 WRITEs
Run 2 => Locking/Unlocking 10000 times each [20,000 AFP ops]                   769 ms
         IO Operations; afpd: 0 READs, 20000 WRITEs | cnid_dbd: 0 READs, 0 WRITEs
Run 2 => Creating dir with 2000 files [4,000 AFP ops]                         3442 ms
         IO Operations; afpd: 2000 READs, 10005 WRITEs | cnid_dbd: 7 READs, 6140 WRITEs
Run 2 => Enumerate dir with 2000 files [~51 AFP ops]                           805 ms
         IO Operations; afpd: 1960 READs, 49 WRITEs | cnid_dbd: 0 READs, 0 WRITEs
Run 2 => Deleting dir with 2000 files [2,000 AFP ops]                         2475 ms
         IO Operations; afpd: 4000 READs, 4003 WRITEs | cnid_dbd: 4 READs, 6180 WRITEs
Run 2 => Create directory tree with 1000 dirs [1,110 AFP ops]                 1701 ms
         IO Operations; afpd: 0 READs, 4442 WRITEs | cnid_dbd: 2 READs, 2267 WRITEs
Run 2 => Directory cache hits (100 dirs + 1000 files) [11,000 AFP ops]        2962 ms
         IO Operations; afpd: 10000 READs, 11100 WRITEs | cnid_dbd: 0 READs, 100 WRITEs
Run 2 => Mixed cache operations (create/stat/enum/delete) [820 AFP ops]        598 ms
         IO Operations; afpd: 820 READs, 1621 WRITEs | cnid_dbd: 2 READs, 1242 WRITEs
Run 2 => Deep path traversal (nested directory navigation) [3,500 AFP ops]     796 ms
         IO Operations; afpd: 2500 READs, 3550 WRITEs | cnid_dbd: 0 READs, 50 WRITEs
Run 2 => Cache validation efficiency (metadata changes) [30,000 AFP ops]      8431 ms
         IO Operations; afpd: 30000 READs, 30100 WRITEs | cnid_dbd: 0 READs, 100 WRITEs

Netatalk Lantest Results (Averages and standard deviations (±) for all tests, across 2 iterations (default))
============================================================================================================

Test                                                                Time_ms  Time± AFPD_R AFPD_R± AFPD_W AFPD_W± CNID_R CNID_R± CNID_W CNID_W±   MB/s
------------------------------------------------------------------ -------- ------ ------ ------- ------ ------- ------ ------- ------ ------- ------
Open, stat and read 512 bytes from 1000 files [8,000 AFP ops]          2188  374.8   6000     0.0   7002     0.0      0     0.0      2     0.0      0
Writing one large file [103 AFP ops]                                    111   34.7      0     0.0    299     0.0      0     0.0      0     0.0    900
Reading one large file [102 AFP ops]                                     37    2.2    100     0.0    100     0.0      0     0.0      0     0.0   2702
Locking/Unlocking 10000 times each [20,000 AFP ops]                     784   21.2      0     0.0  20000     0.0      0     0.0      0     0.0      0
Creating dir with 2000 files [4,000 AFP ops]                           3751  437.7   2000     0.0  10005     0.0      5     2.2   6145     7.1      0
Enumerate dir with 2000 files [~51 AFP ops]                             721  118.8   1960     0.0     49     0.0      0     0.0      0     0.0      0
Deleting dir with 2000 files [2,000 AFP ops]                           2825  495.7   4000     0.0   4003     1.0      3     1.4   6142    53.7      0
Create directory tree with 1000 dirs [1,110 AFP ops]                   1793  130.1      0     0.0   4443     2.2      3     1.4   2309    59.4      0
Directory cache hits (100 dirs + 1000 files) [11,000 AFP ops]          3293  468.8  10000     0.0  11100     0.0      0     0.0    100     0.0      0
Mixed cache operations (create/stat/enum/delete) [820 AFP ops]          866  379.0    820     0.0   1621     0.0      2     0.0   1221    29.0      0
Deep path traversal (nested directory navigation) [3,500 AFP ops]       880  119.5   2500     0.0   3550     0.0      0     0.0     50     0.0      0
Cache validation efficiency (metadata changes) [30,000 AFP ops]        8480   69.3  30000     0.0  30100     0.0      0     0.0    100     0.0      0
------------------------------------------------------------------ -------- ------ ------ ------- ------ ------- ------ ------- ------ ------- ------
Sum of all AFP OPs = 80686                                            25729         57380          92272             13          16069               

Aggregates Summary:
-------------------
Average Time per AFP OP: 0.319 ms
Average AFPD Reads per AFP OP: 0.711
Average AFPD Writes per AFP OP: 1.144
```

Backwards compatibility;
```
% docker run -it --rm --network host --cap-add=NET_ADMIN --volume "/Volumes/case-sensitive/test:/mnt/afpshare" --volume "/Volumes/case-sensitive/test:/mnt/afpbackup" --env TESTSUITE=lan --env TEST_FLAGS="-n 2" --env AFP_USER=test --env AFP_PASS=test --env AFP_GROUP=test --env SHARE_NAME='File Sharing' --env INSECURE_AUTH=true --name netatalk-test netatalk-test:latest 
*** Setting up environment
*** Setting up users and groups
*** Configuring shared volume
*** Fixing permissions
*** Removing residual lock files
*** Configuring Netatalk
NOTE: Set the `ATALKD_INTERFACE' environment variable to start DDP services.
NOTE: TESTSUITE set to 'lan' for afp_lantest test, however IO_MONITORING=1 envvar not set (mounts /proc_io filesystem). IO monitoring disabled.
*** Starting AFP server
*** Running testsuite: lan

+ afp_lantest -n 2 -7 -h 127.0.0.1 -p 548 -u test -w test -s 'File Sharing'
Connecting to host 127.0.0.1:548
WARNING: /proc_io directory not found. IO Monitoring disabled.
To enable IO monitoring, mount proc filesystem at /proc_io. Ensure /proc_io mounted with 'gid' matching group of user running afp_lantest.
Eg, as root:
  mkdir -p /proc_io && mount -t proc -o hidepid=0,gid=0 proc /proc_io

Run 1 => Open, stat and read 512 bytes from 1000 files [8,000 AFP ops]        2116 ms
Run 1 => Writing one large file [103 AFP ops]                                  246 ms for 100 MB (avg. 426 MB/s)
Run 1 => Reading one large file [102 AFP ops]                                   42 ms for 100 MB (avg. 2496 MB/s)
Run 1 => Locking/Unlocking 10000 times each [20,000 AFP ops]                   788 ms
Run 1 => Creating dir with 2000 files [4,000 AFP ops]                         3743 ms
Run 1 => Enumerate dir with 2000 files [~51 AFP ops]                          1097 ms
Run 1 => Deleting dir with 2000 files [2,000 AFP ops]                         2539 ms
Run 1 => Create directory tree with 1000 dirs [1,110 AFP ops]                 1703 ms
Run 1 => Directory cache hits (100 dirs + 1000 files) [11,000 AFP ops]        3600 ms
Run 1 => Mixed cache operations (create/stat/enum/delete) [820 AFP ops]        885 ms
Run 1 => Deep path traversal (nested directory navigation) [3,500 AFP ops]    1007 ms
Run 1 => Cache validation efficiency (metadata changes) [30,000 AFP ops]      8556 ms
Run 2 => Open, stat and read 512 bytes from 1000 files [8,000 AFP ops]        2298 ms
Run 2 => Writing one large file [103 AFP ops]                                  101 ms for 100 MB (avg. 1038 MB/s)
Run 2 => Reading one large file [102 AFP ops]                                   41 ms for 100 MB (avg. 2557 MB/s)
Run 2 => Locking/Unlocking 10000 times each [20,000 AFP ops]                   793 ms
Run 2 => Creating dir with 2000 files [4,000 AFP ops]                         3420 ms
Run 2 => Enumerate dir with 2000 files [~51 AFP ops]                           686 ms
Run 2 => Deleting dir with 2000 files [2,000 AFP ops]                         3076 ms
Run 2 => Create directory tree with 1000 dirs [1,110 AFP ops]                 1896 ms
Run 2 => Directory cache hits (100 dirs + 1000 files) [11,000 AFP ops]        2992 ms
Run 2 => Mixed cache operations (create/stat/enum/delete) [820 AFP ops]        636 ms
Run 2 => Deep path traversal (nested directory navigation) [3,500 AFP ops]    1048 ms
Run 2 => Cache validation efficiency (metadata changes) [30,000 AFP ops]      8448 ms

Netatalk Lantest Results (Averages and standard deviations (±) for all tests, across 2 iterations (default))
============================================================================================================

Test                                                               Time_ms Time±   MB/s
------------------------------------------------------------------ ------- ------ ------
Open, stat and read 512 bytes from 1000 files [8,000 AFP ops]         2207  128.7      0
Writing one large file [103 AFP ops]                                   173  102.5    578
Reading one large file [102 AFP ops]                                    41    1.0   2439
Locking/Unlocking 10000 times each [20,000 AFP ops]                    790    3.6      0
Creating dir with 2000 files [4,000 AFP ops]                          3581  228.4      0
Enumerate dir with 2000 files [~51 AFP ops]                            891  290.6      0
Deleting dir with 2000 files [2,000 AFP ops]                          2807  379.7      0
Create directory tree with 1000 dirs [1,110 AFP ops]                  1799  136.5      0
Directory cache hits (100 dirs + 1000 files) [11,000 AFP ops]         3296  429.9      0
Mixed cache operations (create/stat/enum/delete) [820 AFP ops]         760  176.1      0
Deep path traversal (nested directory navigation) [3,500 AFP ops]     1027   29.0      0
Cache validation efficiency (metadata changes) [30,000 AFP ops]       8502   76.4      0
------------------------------------------------------------------ ------- ------ ------
Sum of all AFP OPs = 80686                                           25874              

Aggregates Summary:
-------------------
Average Time per AFP OP: 0.321 ms
```

### DIFF
--- a/doc/manpages/man1/afp_lantest.1.md
+++ b/doc/manpages/man1/afp_lantest.1.md
@@ -4,14 +4,13 @@ afp_lantest — AFP LAN performance and directory cache testing tool
 
 # Synopsis
 
-**afp_lantest** [-34567GgVvbC] [-h *host*] [-p *port*] [-s *volume*] [-u *user*] [-w *password*]
+**afp_lantest** [-34567bcGgKVv] [-h *host*] [-p *port*] [-s *volume*] [-u *user*] [-w *password*]
 [-n *iterations*] [-f *tests*] [-F *bigfile*]
 
 # Description
 
 **afp_lantest** is a comprehensive AFP (Apple Filing Protocol) performance testing tool designed to
-benchmark various aspects of AFP servers, with special focus on directory cache performance
-improvements in netatalk.
+benchmark various aspects of AFP servers, including directory cache performance in netatalk.
 
 The tool runs a series of tests that measure file operations, directory traversal, and caching
 efficiency. It includes both traditional file system benchmarks and specialized cache-focused tests
@@ -38,26 +37,29 @@ features.
 **-b**
 : Debug mode
 
-**-C**
-: Run cache-focused tests only (tests 9-12)
-
-**-f** *tests*
-: Specific tests to run, specified as digits (e.g., "134" runs tests 1, 3, and 4)
+**-c**
+: Output results in CSV format (default: tabular)
 
 **-F** *bigfile*
 : Use existing file in volume root for read test (file size must match -g/-G options)
 
-**-g**
-: Optimize for gigabit networks (increases file test size to 1 GB)
+**-f** *tests*
+: Specific tests to run, specified as digits (e.g., "134" runs tests 1, 3, and 4)
 
 **-G**
 : Optimize for 10-gigabit networks (increases file test size to 10 GB)
 
+**-g**
+: Optimize for gigabit networks (increases file test size to 1 GB)
+
 **-h** *host*
 : Server hostname or IP address (default: localhost)
 
+**-K**
+: Run cache-focused tests only (tests 9-12)
+
 **-n** *iterations*
-: Number of test iterations to run (default: 1)
+: Number of test iterations to run (default: 2). For iterations > 5 outliers will be removed
 
 **-p** *port*
 : Server port number (default: 548)
@@ -68,11 +70,11 @@ features.
 **-u** *user*
 : Username for authentication (default: current uid)
 
-**-v**
-: Verbose output
-
 **-V**
 : Very verbose output
+
+**-v**
+: Verbose output
 
 **-w** *password*
 : Password for authentication
@@ -129,7 +131,7 @@ The following tests are available:
 
 Tests 9-12 are specifically designed to highlight directory cache performance improvements in
 netatalk. These tests benefit significantly from optimized cache validation and probabilistic
-validation features. Use the **-C** option to run only these cache-focused tests.
+validation features. Use the **-K** option to run only these cache-focused tests.
 
 # Examples
 
@@ -139,7 +141,7 @@ Run all tests with default settings:
 
 Run only cache-focused tests:
 
-    afp_lantest -C -h server.example.com -u testuser -w password -s TestVolume
+    afp_lantest -K -h server.example.com -u testuser -w password -s TestVolume
 
 Run specific tests (file operations):
 
@@ -153,74 +155,109 @@ Run multiple iterations for statistical analysis:
 
     afp_lantest -n 5 -h server.example.com -u testuser -w password -s TestVolume
 
-Run the afp_lantest benchmark using AFP 3.4.
+# Example with IO Monitoring
 
-    % afp_lantest -7 -h localhost -p 548 -u test -w test -s "File Sharing" -n 2
-    Run 0 => Opening, stating and reading 512 bytes from 1000 files   3237 ms
-    Run 0 => Writing one large file                                    146 ms for 100 MB (avg. 718 MB/s)
-    Run 0 => Reading one large file                                     36 ms for 100 MB (avg. 2912 MB/s)
-    Run 0 => Locking/Unlocking 10000 times each                        772 ms
-    Run 0 => Creating dir with 2000 files                             3615 ms
-    Run 0 => Enumerate dir with 2000 files                             755 ms
-    Run 0 => Deleting dir with 2000 files                             1245 ms
-    Run 0 => Create directory tree with 10^3 dirs                     1724 ms
-    Run 0 => Directory cache hits (1000 dir + 10000 file lookups)     3056 ms
-    Run 0 => Mixed cache operations (create/stat/enum/delete)          484 ms
-    Run 0 => Deep path traversal (nested directory navigation)         377 ms
-    Run 0 => Cache validation efficiency (metadata changes)           8822 ms
-    Run 1 => Opening, stating and reading 512 bytes from 1000 files   3448 ms
-    Run 1 => Writing one large file                                     79 ms for 100 MB (avg. 1327 MB/s)
-    Run 1 => Reading one large file                                     37 ms for 100 MB (avg. 2833 MB/s)
-    Run 1 => Locking/Unlocking 10000 times each                        779 ms
-    Run 1 => Creating dir with 2000 files                             3731 ms
-    Run 1 => Enumerate dir with 2000 files                             587 ms
-    Run 1 => Deleting dir with 2000 files                             1156 ms
-    Run 1 => Create directory tree with 10^3 dirs                     1802 ms
-    Run 1 => Directory cache hits (1000 dir + 10000 file lookups)     3006 ms
-    Run 1 => Mixed cache operations (create/stat/enum/delete)          463 ms
-    Run 1 => Deep path traversal (nested directory navigation)         247 ms
-    Run 1 => Cache validation efficiency (metadata changes)           8565 ms
+IO Monitoring (Linux only) requires proc filesystem mounted at /proc_io with hidepid=0.
+Set gid to the group ID of the user running afp_lantest.
+For example, to run as root;
+`mkdir -p /proc_io && mount -t proc -o hidepid=0,gid=0 proc /proc_io`
 
-    Netatalk Lantest Results (average times across 2 iterations)
-    ===================================
+    afp_lantest -n 2 -7 -h 127.0.0.1 -p 548 -u test -w test -s 'File Sharing'
+    Connecting to host 127.0.0.1:548
+    IO monitoring: /proc_io is available
+    Looking for cnid_dbd processes with -u test in command line
+    Found cnid_dbd process: PID 40
+    Looking for afpd processes owned by user 'test' (UID: 1000)
+    Found privilege-dropped afpd process: PID 36
+    IO monitoring enabled (afpd: 36, cnid_dbd: 40)
 
-    Test 1: Opening, stating and reading 512 bytes from 1000 files
-     Average:   3342 ms ± 149.2 ms (std dev)
+    Run 1 => Open, stat and read 512 bytes from 1000 files [8,000 AFP ops]        1923 ms
+            IO Operations; afpd: 6000 READs, 7002 WRITEs | cnid_dbd: 0 READs, 2 WRITEs
+    Run 1 => Writing one large file [103 AFP ops]                                  136 ms for 100 MB (avg. 771 MB/s)
+            IO Operations; afpd: 0 READs, 299 WRITEs | cnid_dbd: 0 READs, 0 WRITEs
+    Run 1 => Reading one large file [102 AFP ops]                                   39 ms for 100 MB (avg. 2688 MB/s)
+            IO Operations; afpd: 100 READs, 100 WRITEs | cnid_dbd: 0 READs, 0 WRITEs
+    Run 1 => Locking/Unlocking 10000 times each [20,000 AFP ops]                   799 ms
+            IO Operations; afpd: 0 READs, 20000 WRITEs | cnid_dbd: 0 READs, 0 WRITEs
+    Run 1 => Creating dir with 2000 files [4,000 AFP ops]                         4061 ms
+            IO Operations; afpd: 2000 READs, 10005 WRITEs | cnid_dbd: 4 READs, 6150 WRITEs
+    Run 1 => Enumerate dir with 2000 files [~51 AFP ops]                           637 ms
+            IO Operations; afpd: 1960 READs, 49 WRITEs | cnid_dbd: 0 READs, 0 WRITEs
+    Run 1 => Deleting dir with 2000 files [2,000 AFP ops]                         3176 ms
+            IO Operations; afpd: 4000 READs, 4004 WRITEs | cnid_dbd: 2 READs, 6104 WRITEs
+    Run 1 => Create directory tree with 1000 dirs [1,110 AFP ops]                 1885 ms
+            IO Operations; afpd: 0 READs, 4445 WRITEs | cnid_dbd: 4 READs, 2351 WRITEs
+    Run 1 => Directory cache hits (100 dirs + 1000 files) [11,000 AFP ops]        3625 ms
+            IO Operations; afpd: 10000 READs, 11100 WRITEs | cnid_dbd: 0 READs, 100 WRITEs
+    Run 1 => Mixed cache operations (create/stat/enum/delete) [820 AFP ops]       1134 ms
+            IO Operations; afpd: 820 READs, 1621 WRITEs | cnid_dbd: 0 READs, 1201 WRITEs
+    Run 1 => Deep path traversal (nested directory navigation) [3,500 AFP ops]     965 ms
+            IO Operations; afpd: 2500 READs, 3550 WRITEs | cnid_dbd: 0 READs, 50 WRITEs
+    Run 1 => Cache validation efficiency (metadata changes) [30,000 AFP ops]      8529 ms
+            IO Operations; afpd: 30000 READs, 30100 WRITEs | cnid_dbd: 0 READs, 100 WRITEs
+    Run 2 => Open, stat and read 512 bytes from 1000 files [8,000 AFP ops]        2453 ms
+            IO Operations; afpd: 6000 READs, 7002 WRITEs | cnid_dbd: 0 READs, 2 WRITEs
+    Run 2 => Writing one large file [103 AFP ops]                                   87 ms for 100 MB (avg. 1205 MB/s)
+            IO Operations; afpd: 0 READs, 299 WRITEs | cnid_dbd: 0 READs, 0 WRITEs
+    Run 2 => Reading one large file [102 AFP ops]                                   36 ms for 100 MB (avg. 2912 MB/s)
+            IO Operations; afpd: 100 READs, 100 WRITEs | cnid_dbd: 0 READs, 0 WRITEs
+    Run 2 => Locking/Unlocking 10000 times each [20,000 AFP ops]                   769 ms
+            IO Operations; afpd: 0 READs, 20000 WRITEs | cnid_dbd: 0 READs, 0 WRITEs
+    Run 2 => Creating dir with 2000 files [4,000 AFP ops]                         3442 ms
+            IO Operations; afpd: 2000 READs, 10005 WRITEs | cnid_dbd: 7 READs, 6140 WRITEs
+    Run 2 => Enumerate dir with 2000 files [~51 AFP ops]                           805 ms
+            IO Operations; afpd: 1960 READs, 49 WRITEs | cnid_dbd: 0 READs, 0 WRITEs
+    Run 2 => Deleting dir with 2000 files [2,000 AFP ops]                         2475 ms
+            IO Operations; afpd: 4000 READs, 4003 WRITEs | cnid_dbd: 4 READs, 6180 WRITEs
+    Run 2 => Create directory tree with 1000 dirs [1,110 AFP ops]                 1701 ms
+            IO Operations; afpd: 0 READs, 4442 WRITEs | cnid_dbd: 2 READs, 2267 WRITEs
+    Run 2 => Directory cache hits (100 dirs + 1000 files) [11,000 AFP ops]        2962 ms
+            IO Operations; afpd: 10000 READs, 11100 WRITEs | cnid_dbd: 0 READs, 100 WRITEs
+    Run 2 => Mixed cache operations (create/stat/enum/delete) [820 AFP ops]        598 ms
+            IO Operations; afpd: 820 READs, 1621 WRITEs | cnid_dbd: 2 READs, 1242 WRITEs
+    Run 2 => Deep path traversal (nested directory navigation) [3,500 AFP ops]     796 ms
+            IO Operations; afpd: 2500 READs, 3550 WRITEs | cnid_dbd: 0 READs, 50 WRITEs
+    Run 2 => Cache validation efficiency (metadata changes) [30,000 AFP ops]      8431 ms
+            IO Operations; afpd: 30000 READs, 30100 WRITEs | cnid_dbd: 0 READs, 100 WRITEs
 
-    Test 2: Writing one large file
-     Average:    112 ms ± 47.4 ms (std dev)
-     Throughput: 936 MB/s (Write, 100 MB file)
+    Netatalk Lantest Results (Averages and standard deviations (±) for all tests, across 2 iterations (default))
+    ============================================================================================================
 
-    Test 3: Reading one large file
-     Average:     36 ms ± 1.0 ms (std dev)
-     Throughput: 2912 MB/s (Read, 100 MB file)
+    Test                                                                Time_ms  Time± AFPD_R AFPD_R± AFPD_W AFPD_W± CNID_R CNID_R± CNID_W CNID_W±   MB/s
+    ------------------------------------------------------------------ -------- ------ ------ ------- ------ ------- ------ ------- ------ ------- ------
+    Open, stat and read 512 bytes from 1000 files [8,000 AFP ops]          2188  374.8   6000     0.0   7002     0.0      0     0.0      2     0.0      0
+    Writing one large file [103 AFP ops]                                    111   34.7      0     0.0    299     0.0      0     0.0      0     0.0    900
+    Reading one large file [102 AFP ops]                                     37    2.2    100     0.0    100     0.0      0     0.0      0     0.0   2702
+    Locking/Unlocking 10000 times each [20,000 AFP ops]                     784   21.2      0     0.0  20000     0.0      0     0.0      0     0.0      0
+    Creating dir with 2000 files [4,000 AFP ops]                           3751  437.7   2000     0.0  10005     0.0      5     2.2   6145     7.1      0
+    Enumerate dir with 2000 files [~51 AFP ops]                             721  118.8   1960     0.0     49     0.0      0     0.0      0     0.0      0
+    Deleting dir with 2000 files [2,000 AFP ops]                           2825  495.7   4000     0.0   4003     1.0      3     1.4   6142    53.7      0
+    Create directory tree with 1000 dirs [1,110 AFP ops]                   1793  130.1      0     0.0   4443     2.2      3     1.4   2309    59.4      0
+    Directory cache hits (100 dirs + 1000 files) [11,000 AFP ops]          3293  468.8  10000     0.0  11100     0.0      0     0.0    100     0.0      0
+    Mixed cache operations (create/stat/enum/delete) [820 AFP ops]          866  379.0    820     0.0   1621     0.0      2     0.0   1221    29.0      0
+    Deep path traversal (nested directory navigation) [3,500 AFP ops]       880  119.5   2500     0.0   3550     0.0      0     0.0     50     0.0      0
+    Cache validation efficiency (metadata changes) [30,000 AFP ops]        8480   69.3  30000     0.0  30100     0.0      0     0.0    100     0.0      0
+    ------------------------------------------------------------------ -------- ------ ------ ------- ------ ------- ------ ------- ------ ------- ------
+    Sum of all AFP OPs = 80686                                            25729         57380          92272             13          16069               
 
-    Test 4: Locking/Unlocking 10000 times each
-     Average:    775 ms ± 5.0 ms (std dev)
+    Aggregates Summary:
+    -------------------
+    Average Time per AFP OP: 0.319 ms
+    Average AFPD Reads per AFP OP: 0.711
+    Average AFPD Writes per AFP OP: 1.144
 
-    Test 5: Creating dir with 2000 files
-     Average:   3673 ms ± 82.0 ms (std dev)
+## Result Columns
 
-    Test 6: Enumerate dir with 2000 files
-     Average:    671 ms ± 118.8 ms (std dev)
+    Time(ms) = Test runtime in milliseconds
+    Time±    = Test runtime standard deviation
+    AFPD_R   = afpd process IO Read operations
+    AFPD_R±  = afpd process IO Read operation standard deviation
+    AFPD_W   = afpd process IO Write operations
+    AFPD_W±  = afpd process IO Write operation standard deviation
 
-    Test 7: Deleting dir with 2000 files
-     Average:   1200 ms ± 62.9 ms (std dev)
+    CNID_*   = IO measurements for the cnid_dbd process (optional)
 
-    Test 8: Create directory tree with 1000 dirs
-     Average:   1763 ms ± 55.2 ms (std dev)
-
-    Test 9: Directory cache hits (1000 dir + 10000 file lookups)
-     Average:   3031 ms ± 35.4 ms (std dev)
-
-    Test 10: Mixed cache operations (create/stat/enum/delete)
-     Average:    473 ms ± 14.9 ms (std dev)
-
-    Test 11: Deep path traversal (nested directory navigation)
-     Average:    312 ms ± 91.9 ms (std dev)
-
-    Test 12: Cache validation efficiency (metadata changes)
-     Average:   8693 ms ± 181.7 ms (std dev)
+For more information see [Testing Wiki](https://netatalk.io/docs/Testing)
 
 # Return Codes
 
--- a/test/testsuite/lantest.c
+++ b/test/testsuite/lantest.c
@@ -1,3 +1,18 @@
+/*
+ * Copyright (c) 2025, Andy Lemin (andylemin)
+ * Credits; Based on work by Rafal Lewczuk, Didier Gautheron, Frank Lahm, and Netatalk contributors
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ */
+
 #include "specs.h"
 
 #include <getopt.h>
@@ -7,18 +22,33 @@
 #include <stdlib.h>
 #include <stdarg.h>
 #include <sys/types.h>
+#include <sys/stat.h>
 #include <pwd.h>
 #include <pthread.h>
 #include <sys/socket.h>
 #include <math.h>
 #include <stdint.h>
+#include <limits.h>
+
+#ifdef __linux__
+#include "lantest_io_monitor.h"
+#endif
 
 #include "afpclient.h"
 #include "test.h"
 
+/* IO Monitoring measurement types */
+#define MEASURE_TIME_MS 0
+#define MEASURE_AFPD_READ_IO 1
+#define MEASURE_AFPD_WRITE_IO 2
+#define MEASURE_CNID_READ_IO 3
+#define MEASURE_CNID_WRITE_IO 4
+#define NUM_MEASUREMENTS 5
+
 #define FPWRITE_RPLY_SIZE 24
 #define FPWRITE_RQST_SIZE 36
 
+/* Test IDs */
 #define TEST_OPENSTATREAD 0
 #define TEST_WRITE100MB 1
 #define TEST_READ100MB 2
@@ -33,18 +63,37 @@
 #define TEST_CACHE_VALIDATION 11
 #define LASTTEST TEST_CACHE_VALIDATION
 #define NUMTESTS (LASTTEST+1)
+#define TOTAL_AFP_OPS 80686
 
 CONN *Conn;
 int ExitCode = 0;
+int Version = 34;
+int Mac = 0;
 char Data[300000] = "";
 char    *Vol = "";
 char    *User = "";
 char    *Path;
-int     Version = 34;
-int     Mac = 0;
 char    *Test = NULL;
+static uint16_t vol;
+static DSI *dsi;
+static char    *Server = "localhost";
+static int32_t Port = DSI_AFPOVERTCP_PORT;
+static char    *Password = "";
+static uint8_t Iterations = 2;
+static uint8_t Iterations_save;
+static uint8_t iteration_counter = 0;
+static struct timeval tv_start;
+static struct timeval tv_end;
+static struct timeval tv_dif;
+static pthread_t tid;
+static uint8_t active_thread = 0;  /* Track if we have an active thread */
 
-/* Unused */
+/* Global Debug flag */
+uint8_t Debug = 0;
+/* Global Quiet flag override for lantest (stdout output enabled) */
+uint8_t lantest_Quiet = 0;
+
+/* Remote linker names */
 CONN *Conn2;
 uint16_t VolID;
 int PassCount = 0;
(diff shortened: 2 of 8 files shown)


doc/manpages/man1/afp_lantest.1.md
```diff
@@ -131,6 +133,212 @@ Tests 9-12 are specifically designed to highlight directory cache performance im
...
+each test generates.
+
+## DSI Quantum Size Impact
+
+The DSI (Data Stream Interface) protocol default quantum size is **1MB (1048576 bytes)**, which affects how many
+read/write operations are performed for large data transfers.
+
+## Test AFP Operation Counts
+
+### 1. TEST_OPENSTATREAD - Open, stat and read 512 bytes from 1000 files
```
rdmark: I'd like to discuss alternatives to this way of documenting the tests.

While it's a plus to keep all docs for the lantest in one place, there are a few drawbacks.

- The maintenance cost of keeping the man page up to date as you change the test code (and risk of the two getting out of sync)
- The overhead of translating all of this text to Japanese
- It can be cumbersome to navigate a lot of text in the man page reader

Off the top of my head, alternatives:

- Put it in a README
- Put it in the wiki
- Implement a feature in afp_lantest for querying available tests and get descriptions printed to stdout
- Doxygen (or similar) inline docs in the code

Thoughts? :)

doc/manpages/man1/afp_lantest.1.md
```diff
@@ -153,74 +361,92 @@ Run multiple iterations for statistical analysis:
...
-
-    Test 10: Mixed cache operations (create/stat/enum/delete)
-     Average:    473 ms ± 14.9 ms (std dev)
-
-    Test 11: Deep path traversal (nested directory navigation)
-     Average:    312 ms ± 91.9 ms (std dev)
-
-    Test 12: Cache validation efficiency (metadata changes)
-     Average:   8693 ms ± 181.7 ms (std dev)
+Run the afp_lantest benchmark using AFP 3.4 (IO Monitoring automatically enabled).
```
rdmark: It might be worth abbreviating the test run log. Readers probably are most interesting in seeing the summary report.

doc/po/afp_lantest.1.md.ja.po
```diff
@@ -6,7 +6,7 @@
 msgid ""
 msgstr ""
 "Project-Id-Version: Netatalk 4.3.0\n"
-"POT-Creation-Date: 2025-08-05 13:58+0200\n"
+"POT-Creation-Date: 2025-08-13 22:14+1000\n"
```
rdmark: I think I prefer not having the po file changes included in the PRs. It results in large and noisy diffs, and risks getting us stuck in merge conflict hell again.

test/testsuite/lantest.c
```diff
@@ -43,8 +59,26 @@ char    *Path;
...
+static pthread_t tid;
+static int active_thread = 0;  /* Track if we have an active thread */
+
+/* Global Debug flag */
+int Debug = 0;
+/* Global Quiet flag override for lantest (stdout output enabled) */
+int lantest_Quiet = 0;
 
-/* Unused */
+/* Remote linker names */
```
rdmark, line 96: I don't understand why this comment was changed. :)

test/testsuite/lantest.c
```diff
@@ -55,54 +89,48 @@ char FailedTests[1024][256] = {{0}};
...
+static int validation_iterations = 100;  /* validation iterations */
+
+/* Forward declarations */
+void fatal_failed(void);
 static unsigned long long numrw;
-static char teststorun[NUMTESTS];
-static unsigned long (*results)[][NUMTESTS];
+static char teststorun[NUMTESTS];  /* Bool array of which tests to run */
+static unsigned long (
+    *results)[][NUMTESTS][NUM_MEASUREMENTS];  /* [iteration][test][measurement_type] */
```
rdmark: Let's put the code comment on a separate line, since it's causing an ugly line break after the opening parenthesis.

test/testsuite/lantest.c
```diff
@@ -1060,7 +1519,7 @@ void run_test(const int dir)
         uint32_t *deep_dirs = calloc(deep_dir_levels, sizeof(uint32_t));
 
         if (!deep_dirs) {
-            fprintf(stderr, "Memory allocation failed for deep directories\n");
+            fprintf(stderr, "Memory allocation failed for 'deep_dirs'\n");
```
rdmark: Doesn't the original message read better?

test/testsuite/lantest.c
```diff
@@ -1220,21 +1682,22 @@ void usage(char *av0)
...
     fprintf(stdout, "\t-7\tAFP 3.4 version (default)\n");
-    fprintf(stdout, "\t-n\thow many iterations to run (default: 1)\n");
-    fprintf(stdout, "\t-v\tverbose\n");
-    fprintf(stdout, "\t-V\tvery verbose\n");
     fprintf(stdout, "\t-b\tdebug mode\n");
-    fprintf(stdout, "\t-g\tfast network (Gbit, file testsize 1 GB)\n");
-    fprintf(stdout,
-            "\t-G\tridiculously fast network (10 Gbit, file testsize 10 GB)\n");
+    fprintf(stdout, "\t-c\toutput results in CSV format (default: tabular)\n");
+    fprintf(stdout, "\t-C\trun cache-focused tests only (tests 8-11)\n");
```
rdmark: In all the other test tools, **-C** means "turn off ANSI colors in terminal output". Wouldn't the lantest also benefit from some ANSI colors to enhance readability?

rdmark: Although I remember now that -C for cache tests is not new in this PR, so counter productive to change it now.

test/testsuite/lantest.c
```diff
@@ -43,8 +59,26 @@ char    *Path;
...
+static struct timeval tv_start;
+static struct timeval tv_end;
+static struct timeval tv_dif;
+static pthread_t tid;
+static int active_thread = 0;  /* Track if we have an active thread */
+
+/* Global Debug flag */
+int Debug = 0;
+/* Global Quiet flag override for lantest (stdout output enabled) */
+int lantest_Quiet = 0;
```
rdmark: Why do we need an override for the global Quiet flag? All the other tools use it and I don't think there's any conflict.

rdmark:
I've done a superficial review now. Will look closer at the logic tomorrow. :)

doc/manpages/man1/afp_lantest.1.md
andylemin: Agree, manpage is not really the best place as man is for guidance how to run a tool, not deep dive how a tool works and what it does. Not worried about "getting out of sync", as trying to get the performance testing tool to a place of stability, as a stable reference point for changes over time. Translation is overhead for sure. So either README or wiki seems the best.. As this is trying to improve testing mechanisms, maybe this should be in a wiki so it can accompany other information discussing structured change testing? I'm even tempted to integrate this into a git pre-push hook? Need your guidance on how to create/update wiki pages

doc/manpages/man1/afp_lantest.1.md
andylemin: I have reordered it for now, everything below `# AFP Operation Counts Analysis` until `# Return Codes` should be moved to wiki as you mention

doc/po/afp_lantest.1.md.ja.po
andylemin: Agree. I included only the .po for the one manpage being updated, to coordinate the change. Are you requesting we should not even cherry pick like his? Ie, we let .po's drift, and just periodically update all .po files in dedicated PRs? I will drop this .po from the CR.

test/testsuite/lantest.c
andylemin: I found it confusing, so I initially deleted them assuming they were unused. They are not "Unused", they are required to enable remote linking? 😉

test/testsuite/lantest.c
andylemin: Agree. Some of the others read poorly, so I normalised all to reference the vars.
Have reverted this, and also changed all Memory allocation errors to use natural language.
I can be a bit OCD for consistency 😉

test/testsuite/lantest_io_monitor.h
```diff
@@ -0,0 +1,65 @@
+/*
+ * Copyright (c) 2025 Netatalk Project
```
rdmark: Would you feel comfortable putting yourself as the copyright owner here?

Even if the code would be considered derivative of existing netatalk code, I think it's preferable to have an individual's name to indicate authorship.

What do you think?

test/testsuite/lantest_io_monitor.c
```diff
@@ -0,0 +1,660 @@
+#ifdef HAVE_CONFIG_H
```
rdmark, line 26: May I ask you to put a copyright blurb in this file, like the one you put in the header file?

In fact, by some standards (e.g. Oracle vs Google) you cannot claim copyright of code headers, so the blurb is arguably optional for headers, but mandatory for proper source code.

test/testsuite/lantest.c
andylemin: Interesting point! The new `-c` for CSV output is intended as the main use case, in a developer context to store and compare performance data over a series of commits. But ANSI colours are also nice when testing changes before marking a commit.
As `-C` was only introduced a week ago in 4.3.0, and is not incorporated into any tool chains yet, we could safely change this now? I would prefer it to be consistent.. I will wait for your guidance before changing?

test/testsuite/lantest_io_monitor.h
andylemin: Lol, sure 😉

andylemin: What about lantest.c? I have updated more than 50% of the code now?

test/testsuite/lantest.c
andylemin: So this was the source of my previous pain with stdout suppression..
Simply, I needed Quiet to be accessible by the new `lantest_io_monitor` module. However if I exposed `Quiet` from lantest.c, it conflicted with the already exposed Quiet in afpcli.c (https://github.com/Netatalk/netatalk/blob/main/test/testsuite/afpcli.c#L10) - which is currently defaulted to 1 and the root cause of the stdout suppression.

This is why non-error fprintf lines have been using stderr everywhere, which is not nice.
It was either update afpcli.c to default to 0 (ideal), or override it with a sidecar name to keep it local to afp_lantest for now. Both Quiet and Debug should be 0 by default IMO.

test/testsuite/lantest_io_monitor.h
rdmark: IANAL, I think it's appropriate that you claim copyright of that file now.

The original authors of the testsuite code are Rafal and Didier, as captured in the [COPYING file](https://github.com/Netatalk/netatalk/blob/main/test/testsuite/COPYING). But they had the habit of not claiming explicit copyright in each source file.

At a related note: last year I was able to get in touch with Rafal over LinkedIn actually (he authored much of the CNID interface code as well). He said he only contributed the initial structure and stubs for the testsuite, and Didier did most of the implementation.

test/testsuite/lantest_io_monitor.c
```diff
@@ -0,0 +1,674 @@
+/*
+ * Copyright (c) 2025, Andy Lemin (@andylemin), Netatalk Project
```
rdmark: Great to see your name here now. :)

Is `@andylemin` referring to your GitHub handle? And does the "Netatalk Project" indicate that you're a member of the project, or that Netatalk Project is a co-author so to speak?

Sorry for being a stickler for the details, but I'm preparing for future Debian packaging requirements. They have tools for parsing copyright blurbs and compiling a so-called *debian/copyright* file with consolidated copyright and license information.

Here's the one for the netatalk package: https://salsa.debian.org/netatalk-team/netatalk/-/blob/debian/latest/debian/copyright

Debian usually expects an email address as the unique identifier for the author, so if you want to use your GitHub handle, perhaps something like `@andylemin@github` is better. Just your name without an identifier is an option, too.

test/testsuite/lantest.c
rdmark: This rings a bell. I'm ok with introducing the local Quiet flag in this PR. There are some broader structural problems, for instance that you have to mirror all global variables from afpcli in each tool even if you don't use them.

One reason for Quiet=1 being the default is because of the spectest suite, where verbose output makes the test log overwhelmingly huge. You only want to turn it on when debugging a failure, really.

test/testsuite/lantest.c
rdmark: I'm fine with doing a quick switcheroo between 4.3.0 and 4.3.1 for the lantest -C option. Do you have another letter in mind?

test/testsuite/lantest.c
rdmark: I can empathize with your predicament. ;)

test/testsuite/lantest.c
rdmark: Aha, this refers to the static linking with libafptest?

At some point I'd like to try to refactor away these global variables from afpcli. I don't see much value to this apart from having a single place to define defaults?

doc/po/afp_lantest.1.md.ja.po
rdmark: Right, I propose that we let po files drift during phases of active development. Then right before taggin a stable release, merge the po files and update the Japanese translation. I can formalize this in the developer documentation.

doc/manpages/man1/afp_lantest.1.md
rdmark: Wiki sounds good to me. I have a skeleton of a page at https://github.com/Netatalk/netatalk/wiki/Testing that you can either flesh out or branch off of if you prefer.

I like the idea of a git pre-push hook. It's something I've had in the back of my mind for a while but not prioritized yet. What checks would you propose we run in such a hook?

andylemin: Nice, building out your existing page sounds easy 🙂
I'll move the new tests deep-dive to clear this PR, and then I'll build it out.
Re; pre-push hook, the minimum should probably be the formatter

test/testsuite/lantest.c
andylemin: Agreed 😉 Defaults are nice as long as they make sense for the whole testsuite scope

test/testsuite/lantest.c
andylemin: So you want to leave it as -C for now and swap later rather than right now?
Best I can think of is -K for cache?

test/testsuite/lantest.c
andylemin: Ahhh ok... in that case I'll update spectest logs to use `if (Debug)` (another PR) so we can make the global defaults sane?

test/testsuite/lantest_io_monitor.h
andylemin: lol, same 😉 Ok I'll add one and include them

test/testsuite/lantest_io_monitor.c
andylemin: Good question. Not keen on sharing my email as I don't want the spam. I'll use andylemin@github then 👍
I added 'Netatalk Project' to imply the rights are shared with the rest of the project authors.
If this is not valid let me know what you prefer? Including the team is my preference

test/testsuite/lantest.c
rdmark: The thing is in the spectest, we have two distinct layers of verbose output. The first level (verbose), Quiet = 0, logs AFP method calls in addition to some other informative messages. The second level (very verbose), Debug = 1, logs DSI responses, in addition to everything from the first level.

I'm open to rethinking how to structure this. The most important requirement is to keep the DSI logging and AFP logging separate, in my opinion, since they're both so chatty but yet important for troubleshooting test failures.

Details to worked out in a future PR. :)

test/testsuite/lantest.c
rdmark: Sorry for being unclear. In my opinion, I think it's better to change it now rather than later.

test/testsuite/lantest_io_monitor.c
rdmark: From my experience working with Debian copyright policy, copyright is something that a person claims, not something that you assign to someone else. 

One option would be to add a credits section after the copyright blurb, for instance calling out that the code is based on work by Rafal Lewczuk, Didier Gautheron, and Frank Lahm. What do you think?

test/testsuite/lantest.c
andylemin: Done.

test/testsuite/lantest.c
andylemin: Yea, We already have -v and -V for this.
So
```
-b = Show everything, max verbose
-V = Very verbose
-v = verbose
(Nothing) = Normal
-q = Very little
```

This is the standard? its debatable whether -b comes before or after -Vv in this hierarchy, but I don't see an issue to make this better

test/testsuite/lantest_io_monitor.h
andylemin: Done

test/testsuite/lantest_io_monitor.c
andylemin: Yes this is much nicer to call credits 🙏 Done

test/testsuite/lantest_io_monitor.c
```diff
@@ -0,0 +1,675 @@
+/*
+ * Copyright (c) 2025, Andy Lemin (andylemin@github.com)
```
rdmark: Dunno if it's just me, but doesn't `andylemin@github.com` make it look like you're a GitHub employee?

My initial thought was to use `@github` as a suffix without the TLD. Or maybe alternatively `@noreply.github.com` to signal that it's not a real email address? What do you think?

test/testsuite/lantest.c
rdmark: Looks good to me. But perhaps we don't need something above -V very verbose, if we introduce the -q Extra quiet mode?

test/testsuite/lantest.c
```diff
@@ -127,162 +172,561 @@ static unsigned long timediff(void)
...
+{
+    int i, test, maxindex, minindex;
+    unsigned long long sum;
+    unsigned long max, min;
+    int measure_type;
 
-        if (Iterations_save > 1) {
-            fprintf(stderr, "\n\t  Average: %6llu ms", avg);
+    /* Eliminate runaways for all measurement types */
+    if (Iterations_save > 5) {
```
rdmark, line 589: Random thought: Do we want to document in the man page this logic for eliminating outliers? And perhaps indicate in the test results that outliers have been eliminated from the results?

doc/manpages/man1/afp_lantest.1.md
```diff
@@ -4,14 +4,13 @@ afp_lantest — AFP LAN performance and directory cache testing tool
 
 # Synopsis
 
-**afp_lantest** [-34567GgVvbC] [-h *host*] [-p *port*] [-s *volume*] [-u *user*] [-w *password*]
+**afp_lantest** [-34567GgVvbcK] [-h *host*] [-p *port*] [-s *volume*] [-u *user*] [-w *password*]
```
rdmark: Nitpick: Let's keep the letter options sorted in alphabetic order. I.e. `[-34567bcGgKVv]`

When I started maintaining the testsuite code, I found several cases of letter options being duplicated in both documentation and code, and spent considerable time getting it cleaned up and sorted. ;)

doc/manpages/man1/afp_lantest.1.md
```diff
@@ -38,7 +37,10 @@ features.
 **-b**
 : Debug mode
 
-**-C**
+**-c**
+: Output results in CSV format (default: tabular)
+
+**-K**
```
rdmark: As above, let's get the option paragraphs resorted to maintain alphabetic order!

test/testsuite/lantest.c
```diff
@@ -1208,7 +1687,7 @@ void usage(char *av0)
 {
     int i = 0;
     fprintf(stdout,
-            "usage:\t%s [-34567GgVv] [-h host] [-p port] [-s vol] [-u user] [-w password] "
+            "usage:\t%s [-34567CcGgVv] [-h host] [-p port] [-s vol] [-u user] [-w password] "
```
rdmark: Don't forget to update the `K` letter option here!

Also I just noticed that the `b` debug letter option isn't documented.

test/testsuite/lantest.c
```diff
@@ -1254,15 +1734,14 @@ void usage(char *av0)
...
-    int Debug = 0;
     static char *vers = "AFP3.4";
     static char *uam = "Cleartxt Passwrd";
 
     if (ac == 1) {
         usage(av[0]);
     }
 
-    while ((cc = getopt(ac, av, "1234567bCGgVvF:f:h:n:p:s:u:w:")) != EOF) {
+    while ((cc = getopt(ac, av, "1234567bKcGgVvF:f:h:n:p:s:u:w:")) != EOF) {
```
rdmark: As above, let's restore alphabetic order. :)

test/testsuite/lantest.c
```diff
@@ -1526,9 +2012,23 @@ int main(int ac, char **av)
...
+        fprintf(stderr, "Error: Failed to open volume '%s'\n", Vol);
+        fatal_failed();
+    }
+
+#ifdef __linux__
+    /* IO monitoring setup - only after AFP connection is established */
+    init_io_monitoring(User);
+
+    /* Warn if not running on localhost */
+    if (strcmp(Server, "localhost") != 0 && strcmp(Server, "127.0.0.1") != 0) {
```
rdmark: Do we want to support IPv6 `::` localhost address?

test/testsuite/lantest_io_monitor.c
andylemin: Good point. When you mentioned about the Debian license expectations I figured that means it wants a valid email address. A quick google search seems to say "Copyright (C) 2025 YourGitHubUsername" is acceptable?

test/testsuite/lantest.c
andylemin: YES! Great point.. That is not mentioned anywhere but in the code. I'll add this 👍

doc/manpages/man1/afp_lantest.1.md
andylemin: lol, I have been sorting them as well. Missed this one

test/testsuite/lantest_io_monitor.c
```diff
@@ -0,0 +1,675 @@
...
+#include <limits.h>  /* For PATH_MAX */
+
+/* Global variables for IO monitoring */
+int io_monitoring_enabled = 0;
+pid_t afpd_pid = 0;
+pid_t cnid_dbd_pid = 0;
+unsigned long long afpd_start_reads = 0, afpd_start_writes = 0;
+unsigned long long cnid_start_reads = 0, cnid_start_writes = 0;
+unsigned long long afpd_end_reads = 0, afpd_end_writes = 0;
+unsigned long long cnid_end_reads = 0, cnid_end_writes = 0;
```
rdmark: I have a preference for explicit integer types, like `uint64_t` over `unsigned long long`. The netatalk codebase has been gradually converted over to the explicit types over the years, but the testsuite code is still overall archaic since it was merged with netatalk proper fairly recently.

doc/manpages/man1/afp_lantest.1.md
andylemin: Agreed.. I'll reorder them

test/testsuite/lantest.c
andylemin: Well spotted. Few misses here. Thanks!

test/testsuite/lantest_io_monitor.c
```diff
@@ -0,0 +1,675 @@
...
+/* Safe PID string parsing helper */
+static pid_t safe_parse_pid(const char *pid_str)
+{
+    char *endptr;
+    long val;
+    /* Define PID_MAX if not available from system headers */
+#ifndef PID_MAX
+    /* Conservative default maximum PID value */
+#define PID_MAX 99999
+#endif
```
rdmark: I'm generally favoring having global constants defined in the headers rather than inline in the functions where they're used.

Over the years I've found the same or similar constants defined multiple times across the codebase, and I've made an effort of consolidating them in appropriate headers.

But this is not a strict rule; if you think this one makes more sense as a one-off local constant here that is reasonable too.

testsuite_alp.Dockerfile
```diff
@@ -1,18 +1,22 @@
-#
-# To test using this Dockerfile
-# Build Container Image: docker build -t netatalk-test:latest .
 
-# Init Container and Netatalk: docker run -d --network host --cap-add=NET_ADMIN --volume "<path to local folder>:/mnt/afpshare" --volume "<path to local folder>:/mnt/afpbackup" --env AFP_USER=test --env AFP_PASS=test --env SHARE_NAME='File Sharing' --env AFP_VERSION=7 --env AFP_GROUP=afpusers --env TZ=Europe/London --env INSECURE_AUTH=true --name netatalk-test netatalk-test:latest
+# Build Container Image: docker build --no-cache -f testsuite_alp.Dockerfile -t netatalk-test:latest .
```
rdmark: Just as a note: these code comments would be a great fit for your future lantest wiki page! And then we may not need them here in the dockerfile anymore?

Also, I wonder if it would make sense to have a paragraph about the testing docker images in the [DOCKER readme](https://github.com/Netatalk/netatalk/blob/main/DOCKER.md)? To make them more discoverable. Case in point, you didn't know about them before I pointed it out. ;)

rdmark (changes requested):

test/testsuite/lantest_io_monitor.c
rdmark: Right, just your nickname/online handle is a fine substitute for the email address, I think.

So "Copyright (c) 2025, Andy Lemin (andylemin)" then?

doc/manpages/man1/afp_lantest.1.md
rdmark: So did we decide to remove this section from the man page for now, then?

andylemin: Just did 🙂 https://github.com/Netatalk/netatalk/wiki/Testing#afp_lantest

doc/manpages/man1/afp_lantest.1.md
andylemin: Done

doc/manpages/man1/afp_lantest.1.md
andylemin: Done

test/testsuite/lantest.c
andylemin: Done

test/testsuite/lantest.c
andylemin: Done

test/testsuite/lantest.c
andylemin: Done

test/testsuite/lantest_io_monitor.c
andylemin: I really like this suggestion for consistency and performance reasons 🙂 I have replaced every single non-explicit data type 🚀

test/testsuite/lantest_io_monitor.c
andylemin: PID_MAX is general in nature, so I think it should be common. Moved it to the header file.

testsuite_alp.Dockerfile
andylemin: Good call. My concern is that users may not find the information, so do you mind if I add a reference for the Testing wiki into the testing Dockerfiles?

andylemin: Done

doc/manpages/man1/afp_lantest.1.md
andylemin: Overhauled the whole page! 😉

rdmark: Looks good!

test/testsuite/lantest.c
```diff
@@ -1251,18 +1733,17 @@ void usage(char *av0)
...
+    int32_t cc, i, t;
     static char *vers = "AFP3.4";
     static char *uam = "Cleartxt Passwrd";
 
     if (ac == 1) {
         usage(av[0]);
     }
 
-    while ((cc = getopt(ac, av, "1234567bCGgVvF:f:h:n:p:s:u:w:")) != EOF) {
+    while ((cc = getopt(ac, av, "1234567bcGgKVvF:f:h:n:p:s:u:w:")) != EOF) {
```
rdmark: I spotted a preexisting bug here: The optarg string checks for `1` and `2` but the following switch statement doesn't cover those cases.

The lantest cannot handle AFP 2.1 or 2.2 the last time I tested, but crashes after the first test. I haven't tested it after your changes though.

doc/manpages/man1/afp_lantest.1.md
```diff
@@ -4,14 +4,13 @@ afp_lantest — AFP LAN performance and directory cache testing tool
 
 # Synopsis
 
-**afp_lantest** [-34567GgVvbC] [-h *host*] [-p *port*] [-s *volume*] [-u *user*] [-w *password*]
+**afp_lantest** [-1234567bcGgKVv] [-h *host*] [-p *port*] [-s *volume*] [-u *user*] [-w *password*]
```
rdmark: The lantest doesn't support AFP 2.1/2.2 so we shouldn't add -1 and -2 to the optarg parameters here.

rdmark (changes requested):

test/testsuite/lantest.c
```diff
@@ -33,76 +63,91 @@
...
 CONN *Conn;
-int ExitCode = 0;
+int32_t ExitCode = 0;
 char Data[300000] = "";
 char    *Vol = "";
 char    *User = "";
 char    *Path;
-int     Version = 34;
-int     Mac = 0;
+int32_t Version = 34;
```
rdmark: Don't we have to change the data type for this variable as declared in test.h as well? Right now it is: `extern int Version;`

The same goes for another handful of variables below that are declared in test.h – and changing these data types will of course cascade to all the other test tools. So perhaps something for a future PR?

andylemin: Oh good catch, Thank you! Got over zealous.. Reverted;
`Version` (test.h), `Mac` (specs.h), `ExitCode` (specs.h), `PassCount` (test.h), `FailCount` (test.h), `SkipCount` (test.h), and `NotTestedCount` (test.h) to `int` for now.

For some reason `Debug` is not in test.h. Although this poses a question I wanted to ask - I don't like using `int` for a `bool`, what is the minimum C standard for Netatalk? Can we use <stdbool.h> header?

For now I have changed `Debug`, `lantest_Quiet`, `active_thread`, and `io_monitoring_enabled` (all bools) to `uint8_t` for bool.
And `Iterations` should not be higher than 8 bit, so I also changed `Iterations`, `Iterations_save` and `iteration_counter` to `uint8_t` and added 256 limit check.

doc/manpages/man1/afp_lantest.1.md
andylemin: Done

test/testsuite/lantest.c
rdmark: We target the C11 standard, as documented in https://github.com/Netatalk/netatalk/wiki/Developer-Notes#coding-standards

I've been gradually introducing stdbool.h when I see the opportunity, so please by all means continue that process!

andylemin: Ok great. I wont start messing with bool just yet. This PR is already extremely large.. I can revisit later.

test/testsuite/lantest.c
andylemin: Done.

test/testsuite/lantest.c
```diff
@@ -150,6 +151,26 @@ static const char *test_names[NUMTESTS] = {
...
+static const ErrorDescription error_descriptions[] = {
+    {0, "Normal Exit", "Clean shutdown"},
+    {2, "Memory Allocation", "Buffer/cache allocation failures"},
+    {3, "File/Directory Ops", "Create/delete/access failures"},
+    {4, "Fork Operations", "Fork open/close failures"},
+    {5, "Network/Protocol", "AFP protocol/read/write errors"},
+    {6, "Thread Operations", "pthread_create failures"},
+    {7, "Validation/Test", "Test validation errors"},
+    {8, "Configuration", "Invalid parameters/out of range"},
+    {9, "Volume Operations", "Volume open/access failures"},
```
rdmark: How about you define macro constants for the error codes, f.e.:

```
#define AFPTEST_NORMAL_EXIT 0
...
#define AFPTEST_CONFIGURATION_ERROR 8
```

...then you don't have to lean on code comments to explain the magic numbers:

```
        clean_exit(AFPTEST_CONFIGURATION_ERROR);
```

What do you think?

rdmark:
@andylemin Looks great! Only one optional refactoring suggestion above.

Can you please squash the commits into a single commit? It was handy to have the incremental commits for code review purposes, but let's clean it up now. ;)

test/testsuite/lantest.c
andylemin: Done

rdmark (approved):
This is amazing. Let's ship it. :)

PR #2356, opened 2025-08-14
Add support for beautifying markdown to codefmt.sh

rdmark:
It is now possible to run `markdownlint-cli2 --fix` through the codefmt.sh script

With this feature, check markdown formatting in Code Formatting job instead of using the Markdownlint workflow


rdmark:
@andylemin If I'm not mistaken, this is what you were asking for! You can now reformat Markdown on the fly with codefmt.sh

Mind you, in my experience markdownlint isn't always the most capable reformatter. It gets hung up easily on non-trivial or ambiguous issues. But for basic typos and rule violations, this will do the the trick.

andylemin:
Great, thanks @rdmark Yea, it does not matter if it is pedantic. As long as it matches what is in the build pipeline so we can avoid some extra pushes to fix build markdown lint failures.
PS; Can we add an option to codefmt.sh which formats for all filetypes? Maybe `-s all`, this could then be used as a pre-push git hook? Current it requires a series of commands to format for each type

rdmark:
@andylemin this is actually already possible, if you run the codefmt.sh script without any parameters!

I've added a line to the usage helptext now that helps users understand this feature.

PR #2364, opened 2025-08-15
afp.conf.5: update logtypes list

EricFromCanada:
Retrieved from here: https://github.com/Netatalk/netatalk/blob/d24b7b560df1b47a7a1aa9f25b5ca7d27e4a4a03/include/atalk/logger.h#L32-L46


doc/manpages/man5/afp.conf.5.md
```diff
@@ -678,7 +678,7 @@ should be logged.
 > By default afpd logs to syslog with a default logging setup equivalent
 to **default:note**
 >
-> logtypes: default, afpdaemon, logger, uamsdaemon
+> logtypes: default, logger, cnid, afpd, dsi, atalkd, papd, uams, fce, ad, sl
```
rdmark: Let's not add atalkd and papd logtypes to this list because they're not used by the daemons that are controlled by afp.conf

Regardless, great catch! These were never updated for the 3.x release series it seems.

doc/po/afp.conf.5.md.ja.po
```diff
@@ -1866,7 +1866,7 @@ msgid ""
 "> By default afpd logs to syslog with a default logging setup equivalent\n"
 "to **default:note**\n"
 ">\n"
-"> logtypes: default, afpdaemon, logger, uamsdaemon\n"
+"> logtypes: default, logger, cnid, afpd, dsi, atalkd, papd, uams, fce, ad, sl\n"
```
rdmark: Same as the above

rdmark (changes requested):

PR #2368, opened 2025-08-18
testsuite: afp_lantest, Fix portability and thread safety issues

andylemin:
Sonarcube;
- Use portable PRIu64 format specifiers for uint64_t
- Replace magic numbers with named constants
- Use thread-safe getpwnam_r() instead of getpwnam()
- Update afp_lantest documentation

### DIFF
--- a/doc/manpages/man1/afp_lantest.1.md
+++ b/doc/manpages/man1/afp_lantest.1.md
@@ -257,6 +257,25 @@ For example, to run as root;
 
     CNID_*   = IO measurements for the cnid_dbd process (optional)
 
+## Aggregates Summary
+
+The aggregate values are purely Intrinsic Metrics, as AFP operations are a mixture of reads, writes,
+and connection related operations.
+
+Therefore the measurements are meaningful only within their own context.
+They provide a way to assess changes relative to itself,
+but lack any external coherence or reference point for comparison with other systems.
+
+Generally, values less than 1 indicate efficient operation (for example due to batching and caching etc),
+and values greater than 1 indicate sub-optimal operation (for example amplification of a single operation,
+causing additional downstream operations).
+
+Note; As total AFP Ops are a mix of reads, writes, and other ops the effective 1:1 (AFP:IO) thresholds are not equal to 1.
+Accurate thresholds can be calculated on a per test basis using the information shown in the Testing Wiki,
+and comparing relevant AFP read/write ops with the relevant AFPD_R/AFPD_W values.
+
+Ideally code changes should observe aggregate values as reducing.
+
 For more information see [Testing Wiki](https://netatalk.io/docs/Testing)
 
 # Return Codes
--- a/test/testsuite/lantest_io_monitor.c
+++ b/test/testsuite/lantest_io_monitor.c
@@ -29,22 +29,34 @@
 
 #ifdef __linux__
 
+/* Standard C library includes */
+#include <ctype.h>
+#include <dirent.h>
+#include <errno.h>
+#include <fcntl.h>
+#include <inttypes.h>
+#include <limits.h>
+#include <pwd.h>
+#include <stdint.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
-#include <unistd.h>
-#include <dirent.h>
-#include <pwd.h>
-#include <sys/types.h>
 #include <sys/stat.h>
-#include <fcntl.h>
-#include <errno.h>
-#include <ctype.h>
-#include <stdint.h>
+#include <sys/types.h>
+#include <unistd.h>
 
+/* Netatalk library includes */
 #include "lantest_io_monitor.h"
 
-#include <limits.h>  /* For PATH_MAX */
+/* Constants for magic numbers */
+#define MAX_PROCESSES_TO_TRACK 10
+#define PROC_SCAN_THRESHOLD 3
+#define PROC_IO_WAIT_SECONDS 1
+#define MAX_RETRY_ATTEMPTS 3
+#define PATH_BUFFER_SIZE 256
+#define LINE_BUFFER_SIZE 256
+#define COMM_BUFFER_SIZE 256
+#define CMDLINE_BUFFER_SIZE 1024
 
 /* Global variables for IO monitoring */
 uint8_t io_monitoring_enabled = 0;
@@ -122,8 +134,8 @@ int32_t check_proc_io_availability(void)
     closedir(proc_dir);
 
     /* If /proc_io appears to be empty or nearly empty, it's likely not properly mounted */
-    if (entry_count < 3) {
-        char message[256];
+    if (entry_count < PROC_SCAN_THRESHOLD) {
+        char message[PATH_BUFFER_SIZE];
         snprintf(message, sizeof(message),
                  "/proc_io directory appears to be empty (found %d entries)",
                  entry_count);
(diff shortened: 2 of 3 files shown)


doc/manpages/man1/afp_lantest.1.md
```diff
@@ -257,6 +257,19 @@ For example, to run as root;
 
     CNID_*   = IO measurements for the cnid_dbd process (optional)
 
+## Aggregates Summary
+
+The aggregate values are purely Intrinsic Metrics, as AFP operations are a mixture of reads, writes, and connection related operations.
+
+Therefore the measurements are meaningful only within their own context. They provide a way to assess changes relative to itself, but lack any external coherence or reference point for comparison with other systems.
```
rdmark: Nitpick: The line length exceeds 120 chars.

I'm fixing the code formatting script now, because it wasn't being run recursively and missed this case!

andylemin: Ahhh. No worries. I'll run the formatter again once you've merged it 🫡

andylemin: I saw a comment about adding the markdown to the code format options. Should `-s c` be sufficient now? Or do we need `-s markdown` as well?
Ideal would be `-s all`

andylemin: Hi @rdmark 
Just saw your comment on https://github.com/Netatalk/netatalk/pull/2356 and looked at codefmt.sh
I was missing `brew install markdownlint-cli2`..

However running `./contrib/scripts/codefmt.sh` while in the project base folder is not working.
`markdownlint-cli2 --fix .` does not seem to find doc/manpages/..

Only way to make it work is to run `markdownlint-cli2 --fix doc/manpages/man1`.
This uses `markdownlint.json` and found the issues in my afp_lantest.1.md correctly. However `--fix` is not working when using json config.

Seems to be a known issue https://github.com/DavidAnson/markdownlint-cli2/issues/110
Looks like the format of markdownlint.json might need tweaking to add `"fix": true` into the json. Looks like json overrides command line options.

andylemin: I fixed it manually for now 👍

rdmark: Does the fix in https://github.com/Netatalk/netatalk/pull/2370 resolve it for you as well?

`markdownlint-cli2 --fix "**/*.md"` worked for me without modifying markdownlint.json

doc/manpages/man1/afp_lantest.1.md
```diff
@@ -257,6 +257,25 @@ For example, to run as root;
...
+The aggregate values are purely Intrinsic Metrics, as AFP operations are a mixture of reads, writes,
+and connection related operations.
+
+Therefore the measurements are meaningful only within their own context.
+They provide a way to assess changes relative to itself,
+but lack any external coherence or reference point for comparison with other systems.
+
+Generally, values less than 1 indicate efficient operation (for example due to batching and caching etc),
+and values greater than 1 indicate sub-optimal operation (for example amplification of a single operation,
+causing additional downstream operations).\
```
rdmark: The trailing backslash is a typo, I assume?

doc/manpages/man1/afp_lantest.1.md
```diff
@@ -257,6 +257,25 @@ For example, to run as root;
...
+NB; as the total AFP Ops are a mix of reads, writes, and other ops the effective 1:1 (AFP:IO) thresholds are not equal to 1.
```
rdmark: I had to look up what *NB* stands for. Not all readers might be familiar with the Latin term. :)

doc/manpages/man1/afp_lantest.1.md
```diff
@@ -257,6 +257,25 @@ For example, to run as root;
...
+
+Accurate thresholds can be calculated on a per test basis using the information shown in the
+[Testing Wiki](https://netatalk.io/docs/Testing), and comparing relevant AFP ops with the relevant AFPD_R and AFPD_W values.
```
rdmark: Since you're adding a hyperlink to https://netatalk.io/docs/Testing in this paragraph, would it make sense to remove the final paragraph that refers to the same URL?

test/testsuite/lantest_io_monitor.c
```diff
@@ -41,11 +41,22 @@
 #include <errno.h>
 #include <ctype.h>
 #include <stdint.h>
+#include <inttypes.h>  /* For PRIu64 */
 
 #include "lantest_io_monitor.h"
 
 #include <limits.h>  /* For PATH_MAX */
```
rdmark: Late comment, but unless there's a logic/syntax reason I prefer having C library header includes in the same block, listed in alphabetical order.

We actually a coding guideline for how to group and order headers: https://github.com/Netatalk/netatalk/wiki/Developer-Notes#rules

When I started working on this codebase, we found dozens upon dozens of duplicated header includes all over the place. My solution was to introduce sorting rules. But perhaps you know of a more sophisticated way to reach the same outcome?

test/testsuite/lantest_io_monitor.c
```diff
@@ -41,11 +41,22 @@
 #include <errno.h>
 #include <ctype.h>
 #include <stdint.h>
+#include <inttypes.h>  /* For PRIu64 */
```
rdmark: Minor, but I don't think you need to justify why you include a header.

Besides, the comment may become outdated once a subsequent changeset starts using other symbols from that header. Just sayin'.

doc/manpages/man1/afp_lantest.1.md
andylemin: Actually no. Wanted a line return without a double space. So it's intended.

doc/manpages/man1/afp_lantest.1.md
andylemin: Interesting the things you take for granted 😉 I'll remove

doc/manpages/man1/afp_lantest.1.md
andylemin: I thought about this too. The final line to trying to say there is lots of additional info, whilst this line is saying info for this specific item can be found here. So their coverage/scope is technically not the same?

I think I'll keep the last line, and remove this inline link

test/testsuite/lantest_io_monitor.c
andylemin: Sounds good to me. I'll move limits.h into the same group as well and sort them. Limits.h is also a C standard system library

test/testsuite/lantest_io_monitor.c
andylemin: Yea sorry. Thats a left over trying to keep track of things whilst trying things out. PRIu64 was new for me.

doc/manpages/man1/afp_lantest.1.md
andylemin: No, `markdownlint-cli2 --fix "**/*.md"` does not work.
`markdownlint-cli2 --fix .` sees the .markdownlint.json config, but does not find the files.
And `markdownlint-cli2 --fix "**/*.md"` finds the files but not the json!
Sadly it seems very broken on MacOS 15.5.
Is `--fix` really working for you? It shows me errors when I run it in the man1 folder, and set `--config` explicitly, but it still does not actually fix anything?

doc/manpages/man1/afp_lantest.1.md
andylemin: Done

doc/manpages/man1/afp_lantest.1.md
andylemin: Done

test/testsuite/lantest_io_monitor.c
andylemin: Done

test/testsuite/lantest_io_monitor.c
andylemin: Done, and minor tidy up

doc/manpages/man1/afp_lantest.1.md
andylemin: Done

doc/manpages/man1/afp_lantest.1.md
rdmark: Odd, more testing is needed then.

Let's continue this discussion in the other PR.

rdmark (approved):

PR #2370, opened 2025-08-18
Recursively reformat markdown files in codefmt script

rdmark:
The markdown formatting was being done only for the working dir, so doing a proper posix shell compatible recursive routine for the entire dir structure, while also introducing a quiet output mode akin to the other code formatting routines

Also splits up the GitHub job for code formatting so that each language has its own step, and brings back the dedicated markdownlint-cli2-action workflow so that all linting errors, not just the ones that can be automatically fixed, get flagged as CI failures

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -49,3 +49,14 @@ jobs:
         run: ./contrib/scripts/codefmt.sh -v -s shell
       - name: Check yaml formatting
         run: ./contrib/scripts/codefmt.sh -v -s yaml
+
+  markdown-linting:
+    name: Markdown Linting
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


contrib/scripts/codefmt.sh
```diff
@@ -70,7 +70,7 @@ fi
 
 if [ "$SOURCE_TYPE" = "markdown" ] || [ "$SOURCE_TYPE" = "" ]; then
     if command -v markdownlint-cli2 > /dev/null 2>&1; then
-        markdownlint-cli2 --fix .
+        markdownlint-cli2 --fix "**/*.md"
```
cheesestraws: ** is a bash-ism but the shebang line says to invoke /bin/sh

rdmark: good call; fixed

rdmark:
@andylemin When you have the opportunity, please test if this approach works for you!

andylemin:
Nice! Much better 🙂 Thank you

PR #2379, opened 2025-08-23
Fix code quality issues in rtmpqry and disambiguate comments

rdmark:
rtmqry: touch up a few variables to avoid implicit data type conversion

overall: rephrase and clean up code comments that SonarQube interpreted as commented-out C code

### DIFF
--- a/bin/rtmpqry/rtmpqry.c
+++ b/bin/rtmpqry/rtmpqry.c
@@ -125,7 +125,7 @@ int main(int argc, char** argv)
     } else {
         do_rtmp_rdr(sockfd, &sa_remote, all_routes, 2);
     }
-    
+
     free(remote_addr);
 }
 
@@ -165,8 +165,8 @@ void do_rtmp_request(int sockfd, struct sockaddr_at *sa_remote)
     uint8_t buf[600];
     buf[0] = DDPTYPE_RTMPR;
     buf[1] = RTMPROP_REQUEST;
-    int ret = netddp_sendto(sockfd, buf, 2, 0, (struct sockaddr*)sa_remote,
-                            (int)sizeof(struct sockaddr_at));
+    ssize_t ret = netddp_sendto(sockfd, buf, 2, 0, (struct sockaddr*)sa_remote,
+                                (int)sizeof(struct sockaddr_at));
 
     if (ret < 0) {
         perror("netddp_sendto");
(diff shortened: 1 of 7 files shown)


bin/rtmpqry/rtmpqry.c
```diff
@@ -165,8 +165,8 @@ void do_rtmp_request(int sockfd, struct sockaddr_at *sa_remote)
     uint8_t buf[600];
     buf[0] = DDPTYPE_RTMPR;
     buf[1] = RTMPROP_REQUEST;
-    int ret = netddp_sendto(sockfd, buf, 2, 0, (struct sockaddr*)sa_remote,
-                            (int)sizeof(struct sockaddr_at));
+    int ret = (int) netddp_sendto(sockfd, buf, 2, 0, (struct sockaddr*)sa_remote,
```
cheesestraws: Urgh, this is a mess.

Since netddp_sendto is a macro that just expands to sendto, I feel like it's probably better to give 'ret' the type ssize_t here, rather than casting the return of sendto to int.  

But its macroness also means that netddp_sendto has no actual defined prototype, which is a bit sad.  Maybe it should be a trivial inline function rather than a macro, so at least it has a consistent prototype?

rdmark: Right, I went back and forth on this one too. Don't we have the same situation with the netddp_sendto call? I pushed a rebased commit now that refactors those return code variables. Please have a look.

cheesestraws: Yeah, that looks nicer.

rdmark:
@andylemin It would be great if you could eye through these tweaks to the lantest code. This changeset should bring the SonarQube "new code" quality gate back to green. :)

andylemin:
> @andylemin It would be great if you could eye through these tweaks to the lantest code. This changeset should bring the SonarQube "new code" quality gate back to green. :)

Hi @rdmark don't worry, I'm half way through. Should have a patch to fix up every single recommendation (as well as the main one) in the next couple days.

rdmark:
Removed the lantest fixes since it's handled in a different PR now.

@cheesestraws Is there anything about theses changes you want to revisit in the light of #2383 ?

cheesestraws:
Nope, I'm still happy with this: the actual prototype of the two functions is exactly the same as it was, so it won't have any effect here.

PR #2380, opened 2025-08-23
test: Consistently initialize buffers before use in afpd component tests

rdmark:
Addressing uninitialized memory usage flagged by valgrind

With this fix, you can run the tests successfully in a valgrind wrapper, e.g. through meson:

```
meson setup build -Dwith-tests=true
meson compile -C build
meson test --wrapper='valgrind --leak-check=full --track-origins=yes' -C build
```

when run on Linux with valgrind installed this should give you:

```
ninja: Entering directory `/home/dmark/netatalk/build'
ninja: no work to do.
1/2 afpd integration tests - setup        OK              0.32s
2/2 afpd integration tests - run          OK              6.66s

Ok:                 2   
Expected Fail:      0   
Fail:               0   
Unexpected Pass:    0   
Skipped:            0   
Timeout:            0   

Full log written to /home/dmark/netatalk/build/meson-logs/testlog-valgrind.txt
```

And now when this is working, let's run valgrind in select Linux CI jobs!

### DIFF
--- a/test/afpd/afpfunc_helpers.c
+++ b/test/afpd/afpfunc_helpers.c
@@ -91,7 +91,7 @@ static int push_path(char **bufp, const char *name)
 
 char **cnamewrap(const char *name)
 {
-    static char buf[256];
+    static char buf[256] = { 0 };
     static char *p = buf;
     int len = 0;
     PUSHVAL(p, uint8_t, 3, len); /* path type */
@@ -104,8 +104,7 @@ char **cnamewrap(const char *name)
 
 int getfiledirparms(AFPObj *obj, uint16_t vid, cnid_t did, const char *name)
 {
-    const int bufsize = 256;
-    char buf[bufsize];
+    static char buf[256] = { 0 };
     char *p = buf;
     int len = 0;
     ADD(p, len, 2);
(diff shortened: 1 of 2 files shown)


test/afpd/afpfunc_helpers.c
```diff
@@ -94,6 +94,7 @@ char **cnamewrap(const char *name)
     static char buf[256];
     static char *p = buf;
     int len = 0;
+    memset(buf, 0, sizeof(buf));
```
cheesestraws: Why not just do:

`static char buf[256] = { 0 };`

on line 94?  IMO this is neater

rdmark: I force pushed a commit now that changes to this initialization pattern in all functions. What do you think?

cheesestraws: Much nicer.  I really think the ' = { 0 }' idiom for zero-initialising any array or struct is one of the big quality of life improvements in modern C, just makes life so much clearer than memsetting...

rdmark: Without a doubt. We target C11 so there's no reason to keep the archaic style.

PR #2383, opened 2025-08-24
netddp: remove macros, define interface for netddp_{close, recvfrom, sendto}

cheesestraws:
At present, netddp_close, netddp_recvfrom and netddp_sendto are just macros which fob off responsibility for their prototypes to the prototypes of their underlying functions defined in unistd.h.  In the interests of better typechecking and removing a barrier to using netddp as an actually useful abstraction boundary again, this PR replaces the macros with stub inline functions (which boils down to the same thing).  The prototypes chosen are those mandated by POSIX for the underlying socket calls, which seems the sensible option, except that I've removed 'restrict' at the insistance of sonarqube.

### DIFF
--- a/include/atalk/netddp.h
+++ b/include/atalk/netddp.h
@@ -15,16 +15,29 @@
 
 #include <sys/types.h>
 #include <sys/socket.h>
+#include <unistd.h>
+
 #include <netatalk/at.h>
 
 extern int netddp_open(struct sockaddr_at *, struct sockaddr_at *);
 
-#include <unistd.h>
-#include <sys/types.h>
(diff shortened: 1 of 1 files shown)


rdmark (approved):
LGTM – a more sane abstraction.

At a glance I think it'd be safe to merge this to 4.3 stable branch to be included with 4.3.1, but do you see a good reason to hold off until we open up the main branch for breaking changes?

cheesestraws:
No, aiming for 4.3.1 sounds sensible to me.  It shouldn't actually affect the generated binaries at all, if I understand the optimisations in play here.

PR #2389, opened 2025-08-28
docs: Flesh out the afpstats.1 man page

rdmark:
Spell out all preconditions for using afpstats, plus a handful of other tweaks

Fix code comment copy paste mistake in the afpstats D-Bus config file


rdmark:
Merging without waiting for review since I want to prepare for the next release today. :)

Issue #2399, opened 2025-08-30
Prefer mDNSResponder over Avahi on macOS if both are available


trodemaster:
**Describe the bug**

In working on a 4.3.1 port for macports I'm seeing Avahi preferred over mDNSResponder if Avahi is installed already. 

**To Reproduce**

Building using pre-release portfile https://github.com/trodemaster/blakeports/blob/main/net/netatalk4/Portfile Avahi is used for Zeroconf support. 

**Expected behavior**

macOS should prefer mDNSResponder for spotlight support at build time

**Environment**

- Server OS: macOS 15.6.1
- Client OS: Mac OS 9.2.2
- Netatalk Version: 4.3.1

**Configuration**

Build issue so conf not important

**Logs**

Build issue so conf not important

**Additional context**

I generated a patch that seems to work. 
https://github.com/trodemaster/blakeports/blob/main/net/netatalk4/files/prefer-mdns-on-darwin.diff

rdmark:
@trodemaster I get a 404 for the patch. Is it a private repo?

If you don't mind, please file a PR and we can take it from there. :)

PR #2400, opened 2025-08-31
meson.build: prefer mDNS over Avahi on Darwin systems

trodemaster:
## Description

This PR modifies the zeroconf detection logic in meson.build to prioritize Apple's built-in mDNS (Bonjour) support over Avahi on Darwin/macOS systems.

## Changes Made

* Swaps the priority order in zeroconf detection logic
* Checks  first (Apple's mDNSResponder)
* Falls back to  for non-Darwin systems
* Maintains backward compatibility

## Benefits

* macOS has native mDNS support via dns_sd framework
* Maintains Avahi as fallback for non-Darwin systems
* Better integration with macOS networking stack

## Testing

This change has been tested in the MacPorts build environment and resolves the issue where Avahi was being preferred over mDNSResponder on macOS systems.

## Related Issue

Fixes: https://github.com/Netatalk/netatalk/issues/2399


rdmark:
@trodemaster Unfortunately it's not as easy as always preferring mDNS over Avahi for all platforms. We've had headaches with Alpine in particular. It has native mDNS headers but they're incomplete. In the build jobs here I see that Arch and OmniOS have similar situations.

I can see two options. Either craft a more sophisticated check for certain symbols or other capabilities of the mDNS library, so that we don't trigger on an incomplete library. Or do a crude platform override for macOS so that we always use native mDNS there and keep the preference for Avahi elsewhere.

What do you think?

trodemaster:
I think the platform specific check is the way to go. Mostly from a simplicity perspective. Checking mDNS capabilities seems like it could change over time.

rdmark:
Agreed, let’s go with the platform check. There are other checks for Darwin elsewhere that you can use as example.

meson.build
```diff
@@ -944,18 +944,36 @@ have_dns = (
...
-        cdata.set('HAVE_MDNS', 1)
-        cdata.set('freebsd_zeroconf', 'mdnsd')
-        zeroconf_provider += 'mDNS'
+    if host_os == 'darwin'
+        # On Darwin/macOS, prefer mDNS (Bonjour) over Avahi
+        if have_dns
+            have_zeroconf = true
+            cdata.set('USE_ZEROCONF', 1)
+            cdata.set('HAVE_MDNS', 1)
+            cdata.set('freebsd_zeroconf', 'mdnsd')
```
rdmark, line 953: Minor remark: We don't need to set the freebsd string in the Darwin branch

meson.build
```diff
@@ -944,18 +944,36 @@ have_dns = (
...
+            zeroconf_provider += 'mDNS'
+        elif avahi.found()
+            have_zeroconf = true
+            cdata.set('USE_ZEROCONF', 1)
+            cdata.set('HAVE_AVAHI', 1)
+            cdata.set('freebsd_zeroconf', 'avahi_daemon')
```
rdmark, line 959: Same as above

rdmark:
@trodemaster Please squash the commits and force push the branch!

meson.build
trodemaster: If I remove or change the line cdata.set('freebsd_zeroconf', 'mdnsd') the build always ends up using Avahi.

rdmark: Using Avahi on macOS you mean? That's surprising. Are you sure you didn't change something else at the same time?

trodemaster: Yes. I edited just those specific lines. Testing on 2 x macOS virtual machines. Tried removiing the line and just removing the frebsd part.
 cdata.set('mdnsd')

rdmark: Alright then, let me accept your solution then. While it does introduce some code duplication, a less risky fix is better for the stable branch.

PR #2403, opened 2025-09-03
afpd: Revert to non-reentrant getpwnam() in the uam module

rdmark:
Since afpd isn't a threading application, there is no pressing need to use the reentrant-safe way to fetch the passwd entry in uam_getname()

The reverted solution had flaws that led to a critical failure when attempting to authenticate in a complex ActiveDirectory environment

### DIFF
--- a/etc/afpd/uam.c
+++ b/etc/afpd/uam.c
@@ -218,7 +218,6 @@ struct passwd *uam_getname(void *private, char *name, const int len)
 {
     AFPObj *obj = private;
     struct passwd *pwent = NULL;
-    struct passwd pwent_buf;
     static char username[256];
     static char user[256];
     static char pwname[256];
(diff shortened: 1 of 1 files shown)


andylemin (approved):
LGTM

NJRoadfan:
One suggestion might be to add a comment on why the non-reentrant function has to be used here. For anyone in the future that might read the code.

rdmark:
@NJRoadfan Added a comment. Is this in line with what you had in mind?

cheesestraws:
I'd prefer a slightly more detailed comment personally - "don't change this back to the reentrant version, it won't work" is fine, but a note about what the actual problem was and why the reentrant version was the problem would be even better

rdmark:
Maybe it's better to not leave a comment after all? I don't necessarily want to discourage future contributors to attempt a refactor of that function.

I think it should be possible to use the reentrant call, just that the handling of the buffer that contains the return value becomes complicated when accounting for multiple calls to uam_getname(). And I prefer not to deal with further risky changes in the stable branches, especially since I don't have an ActiveDirectory setup where I can reproduce the bug.

Just my 2 cents. :)

PR #2404, opened 2025-09-03
[4.2] afpd: Revert to non-reentrant getpwnam() in the uam module

rdmark:
Since afpd isn't a threading application, there is no pressing need to use the reentrant-safe way to fetch the passwd entry in uam_getname()

The reverted solution had flaws that led to a critical failure when attempting to authenticate in a complex ActiveDirectory environment


andylemin (approved):
LGTM

PR #2405, opened 2025-09-04
[4.2] GitHub CI: Bring build jobs up to date and secure them

rdmark:
Debian Trixie needs the gcc package, and bump OmniOS bootstrap

Use specific release hashes for docker images and vmactions actions

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -34,7 +34,7 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 12
     container:
-      image: alpine:latest
+      image: alpine:3.22.1
     steps:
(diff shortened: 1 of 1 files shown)


andylemin:
Hard coding versions instead of using `latest` is going to cause some extra admin?

andylemin (approved):

rdmark:
@andylemin you’re not wrong! We trade convenience for a more granular control of which dependencies we take in. The latter is more secure and also more predictable. No more scrambling when a Linux distribution pushes out a breaking change to the latest tag. :)

FWIW we already made this change in the main branch and this is only a backport. In the main branch we manage dependency updates with Dependaboy.

PR #2415, opened 2025-09-12
meson: Solaris compatible check for iconv parameters

rdmark:
These fixes should make the iconv capability checks more consistently compatible with Solaris and friends (Solaris 11, OpenIndiana, OmniOS)

Make sure the the _XPG6 flag is turned on in the iconv capability check so that the code is compiled with POSIX compatible prototypes

Explicitly use defined include_dirs for the compilation check since OmniOS may have multiple iconv.h headers in different locations

Flip the validation logic so that we check for the const type rather than the non-const

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1518,25 +1518,31 @@ else
     )
     if have_iconv
         cdata.set('HAVE_USABLE_ICONV', 1)
+        iconv_code = '''
+            #ifndef _XPG6
+            #define _XPG6 1
+            #endif
+            #include <stdlib.h>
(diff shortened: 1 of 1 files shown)


rdmark:
There's still something missing to build on OpenIndiana. The linker tries to link with the 32 bit libraries in /usr/lib rather than the 64 bit ones in /usr/lib/amd64 and I have yet to figure out what would get the linker do the right thing.

andylemin (approved):
Nice. LGTM

PR #2421, opened 2025-09-15
getzones: defensively release memory in optarg function

rdmark:
Should prevent a theoretical memory leak condition flagged by SonarQube


cheesestraws:
Does it tell you where it thinks this could have been set?  I'm a bit confused by this...

edit: ohhh, it'll be if someone puts -z on the command line twice, at a guess.  Fair play.

rdmark:
Right, this is my read, too. You might potentially hit the same case more than once with a malformed optarg buffer.

PR #2424, opened 2025-09-21
testsuite: Bugfixes for lantest.c. Docker: Updated Alpine linux to 3.22.1

andylemin:
testsuite: Bugfixes for lantest.c; correctly delete files and directories after tests and added a new recursive cleanup.
Use the highest found PID for afpd (most recently launched daemon in stead of the lowest) - required when multiple afpd processes with the same user.

Docker: Updated Alpine linux to 3.22.1


testsuite_alp.Dockerfile
```diff
@@ -46,7 +46,7 @@ ARG BUILD_DEPS="\
     tinysparql-dev \
     "
 
-FROM alpine:3.22@sha256:4bcff63911fcb4448bd4fdacec207030997caf25e9bea4045fa6c8c44de311d1 AS build
+FROM alpine:3.22.1@sha256:4bcff63911fcb4448bd4fdacec207030997caf25e9bea4045fa6c8c44de311d1 AS build
```
rdmark: The preferred way to bump container dependencies is through Dependabot!

If you go to https://github.com/Netatalk/netatalk/network/updates you can trigger an update job for Dockerfiles manually. They're scheduled to run automatically at a weekly cadence.

Besides, I think we already use 3.22.1 since "3.22" matches the latest patch version.

All that said, it is not a bad idea to use the patch version tag since this gives better transparency of which version we're actually using. I'm pretty sure Dependabot will keep it updated for us. Can we please update the main [Dockerfile](https://github.com/Netatalk/netatalk/blob/main/Dockerfile) too?

test/testsuite/lantest.c
```diff
@@ -796,6 +799,153 @@ void clean_exit(int exit_code)
     exit(exit_code);
 }
 
+/*!
+ * @brief depth-first recursively delete a directory tree
```
rdmark: Do I read it correctly that this is a general purpose function for deleting any dir?

I would have use for it in the spectest too! Many tests have the precondition of an empty dir and we rely on manually cleaning up.

Entirely optional for this PR, but worth discussing if we can move this into a shared module within the testsuite. Maybe afphelper.c?

testsuite_alp.Dockerfile
andylemin: Hi. Ok cool, I'll remove this change from the PR and trigger the update that way. 
The reason I updated was because some of the packages we use are no longer available in the repo for 3.22 anymore

rdmark: Which packages did we lose in 3.22?

test/testsuite/lantest.c
andylemin: 🙂 Yes, it's a general purpose recursive delete function (much better than trying to clean manually).
It was tricky figuring out the file and directory bitmaps to extract the directory and file IDs from the enumeration response.
Happy to move it for future use, pleased it's useful!

andylemin: What happened in #2249? Looks like you got somewhere but dropped it?

andylemin: Oh I think I see.. yea, you have to delete in reverse from leaf to root. Doing that with a loop would require tracking a whole tree, which is why I used nested recursion deleting files on the way, and directories in reverse as nested calls return.
I'll try and make it a bit nicer for a utility func

rdmark: Right, I saw the nested recursion in your solution; it makes sense now why I couldn't dig myself out of that attempt at solving the problem.

At a related note, afphelper.c seems to contain a mix of actual "AFP helper" functions, with pure test runner functions. Might be worth splitting the latter out to a separate module while adding more AFP helper functions. For instance testrunner.c?

testsuite_alp.Dockerfile
andylemin: Hmm, curious.. 3.22 is now succeeding! 3 packages were failing (can't rememeber exactly)
I wonder if it was transient whilst they pulled some CVE'ed ones and was just waiting for mirror replication..
Anyway, I want to try https://github.com/Netatalk/netatalk/network/updates

rdmark: A transient issue with their package repo mirrors sound plausible. In my experience one of the top reasons for intermittent failures in the CI pipeline.

rdmark (approved):
I think this is good to merge now. Did you have any further improvements in mind for this PR?

andylemin:
Hi @rdmark So I moved the new recursive function as you requested, and in the process did some more testing, only to find it was not working as expected. My afp_lantests were cleaning up properly (following a different fix) leading me to mistake it as the recursive function doing the cleaning (I thought there was still one test that needed clean up work).
After a lot of digging, I found something strange;

All current uses of FPEnumerate, FPEnumerateFull and FPEnumerateExt2Full are broken, including test300 in FPEnumerate.c.
In short FPEnumerate and FPEnumerate* use my_dsi_data_receive, while all other commands use my_dsi_cmd_receive - However my_dsi_data_receive stores data length in 'cmdlen', not 'datalen'?

my_dsi_cmd_receive (line 321) - receives data into DSI_CMDSIZ buffer (x->commands)
my_dsi_data_receive (line 327) - receives data into DSI_DATASIZ buffer (x->data)
So my_dsi_data_receive data is received into dsi->data buffer, not dsi->commands

The names and uses with my_dsi_data_receive and my_dsi_cmd_receive are all flipped around. Reading the naming, my_dsi_data_receive kinda makes more sense, so think someone was trying to flip them all but did not complete it creating confuson and silently breaking all uses of FPEnumerate in the testsuite.

andylemin:
@rdmark Anyway, you will now see a more robust version of delete_directory_tree() in afphelper.c which can handle very large trees and is much easier to understand.
We will need to update the uses of FPEnumerate* to match this working example. Ready for review and merge

test/testsuite/lantest.c
```diff
@@ -38,6 +38,7 @@
 /* Netatalk library includes */
 #include "afpclient.h"
 #include "test.h"
+#include "afphelper.h"
```
rdmark, line 41: nitpick: alphabetic sort order of header includes >:)

test/testsuite/lantest.c
```diff
@@ -140,6 +141,11 @@ static int32_t validation_iterations = 100;  /* validation iterations */
 
 /* Forward declarations */
 void clean_exit(int exit_code);
+static int cleanup_test_directory(CONN *conn, uint16_t vol,
+                                  char *dirname);
+int32_t is_there(CONN *conn, uint16_t volume, int32_t did, char *name);
+extern int delete_directory_tree(CONN *conn, uint16_t volume,
+                                uint32_t parent_did, char *dirname);
```
rdmark: the formatter wants one more space indent here

test/testsuite/afphelper.c
```diff
@@ -1267,3 +1267,200 @@ void assert_false(int real, char *file, int line, void (*fn)(), int log_level)
...
+                    dirname, ret);
+            break;
+        }
+
+        /* Process batch */
+        /* WARN: FPEnumerate and FPEnumerate* use my_dsi_data_receive, other commands use my_dsi_cmd_receive.
+         my_dsi_data_receive stores data length in 'cmdlen', not 'datalen'?
+         my_dsi_cmd_receive (line 321) - receives data into DSI_CMDSIZ buffer (x->commands)
+         my_dsi_data_receive (line 327) - receives data into DSI_DATASIZ buffer (x->data)
+         So my_dsi_data_receive data is received into dsi->data buffer, not dsi->commands. */
```
rdmark: this was a great find! I had a gut feel that something was up with the enumeration routines but couldn't put my finger on it.

do we even want to add a FIXME at the affected code so that we don't forget about it?

test/testsuite/lantest.c
andylemin: Interesting codefmt.sh did not fix this..

test/testsuite/afphelper.c
andylemin: I added a FIXME in front of the first use of FPEnumerate* in every function that uses them.
In most cases the tests do not interogate the actual entries, so the issue is benign. But there are several I can see which are processing the entiries and so silently failing (when reading the wrong fields, it just looks like there are no entires)..
The FIXME's can be removed once the test is properly reviewed and either fixed or confirmed no change required..

PR #2425, opened 2025-09-21
testsuite: refactor libafptest with a testhelper module

rdmark:
This moves functions that are strictly for managing tests from afphelper to a new testhelper module

Some dead code has been cleaned up as well


rdmark:
@andylemin This came out of our discussion in your latest lantest PR. There may be other fragments of test helper functions here and there, but for starters this makes the purpose of afphelper.c a bit tighter IMHO. Please let me know what you think.

PR #2426, opened 2025-09-21
atalkd: free rtmp entry after updating head node

rdmark:
Rather than freeing the rtmp entry in the rtmp_replace() function, do it after updating the head node in rtmp_packet() under the same conditions

This avoids a potential access of freed memory when the timeout is reached indicated by the RTMPTAB_BAD status


rdmark:
@cheesestraws do you see any potential problem with this change?

At first I attempted a more complicated solution, keeping track of when we encounter the RTMPTAB_BAD status (which I think means a route lookup timeout) in rtmp_replace() and then have some special handling in rtmp_packet() for this case so that we don't attempt to update the next node (which is when the freed memory is accessed.) But the more I looked at it, the more I got convinced that rtmp_free() handles the cleanup and that it makes the most sense to just skip the logic in question.

As a side note, it's puzzling to me why this bug has never been reported in the wild. I wonder if the RTMPTAB_BAD timeout condition is just extremely rare?

cheesestraws:
Urgh, this rtmp implementation gives me a headache.

RTMPTAB_BAD is when the route is bad, and this isn't when a route lookup times out, it's when the validity of the router times out, and actually corresponds to the route "aging out" of the routing table.  So I think this would only come up in a fairly small time window in an already fairly rare situation.

Is the code path we're worried about here the one where rtmp_replace is called on line 801, which then calls rtmp_free?

NJRoadfan:
One thing I will state about atalkd. The code is very much "here be dragons", and I would be hesitant to change something with fear of breaking in some obscure way. What SonarQube thinks is a problem might not actually be.

rdmark:
@cheesestraws Let me explain the code path that I'm concerned about:

1. rtmp_packet(): `if (rtmp_replace(rtmp) < 0)` on line 801
2. rtmp_replace(): `if (replace->rt_state == RTMPTAB_BAD)` on line 418 which calls rtmp_free()
3. rtmp_free(): will do some cleanup of the rtmp data structure then finally `free(rtmp)`
4. back in rtmp_packet(): we exit the nested control structure and hit `if (rtmp->rt_next != NULL)` on line 824
5. since rtmp has been freed, accessing the rt_next member will point to freed memory

Do you see the same, or is there a flaw in this interpretation?

cheesestraws:
@rdmark your interpretation is good but I don't think this code is going to do anything about it.

At line 801, the 'rtmp' pointer is being passed by _value_, don't forget.

So rtmp_replace has no way to set the rtmp pointer it's passed to be NULL.  So if it calls rtmp_free, then the rtmp variable in the caller will not be NULL, it will just point into unallocated memory, which a NULL test won't catch.

It looks like the intent here is for rtmp_replace to be where the caller basically gives up ownership of the rtmp pointer, and it shouldn't touch that pointer after it is called.  But then it does.

My preferred way to "fix" this would be to refactor rtmp_replace to take a pointer-to-pointer, and then if it calls rtmp_free it can set the pointer it is passed to NULL, which you can then null-check for.  This is my normal pattern for dealing with this kind of thing, anyway... and that won't change the semantics of what rtmp_replace is doing.

That whole microoptimisation about shuffling around the tuples is weird, though, and feels like someone trying to be overly clever; I'd be fascinated to learn whether it actually had a meaningful performance impact.  I'm almost inclined to suggest removing the "shuffle the tuples" dance at the end, but that's a more invasive change.

@NJRoadfan 
> One thing I will state about atalkd. The code is very much "here be dragons", and I would be hesitant to change something with fear of breaking in some obscure way. What SonarQube thinks is a problem might not actually be.

Aye, I really don't like this code.  I'm fairly sure this use-after-free _is_ a real thing though, I think @rdmark's interpretation is sound, even if it happens pretty rarely.  If this were less fragile code I'd be more gung-ho about removing the weird optimisation, but as it is I think we should go for the least semantically intrusive option, which is the pointer-to-pointer and the NULL check.

rdmark:
@cheesestraws This is the least intrusive solution I could think of, using your approach. What do you think?

I would like to attempt a deeper rewrite of these routines when we branch out the stable 4.3 branch and open up the main branch for breaking changes (which could be any day now. :) )

cheesestraws:
Yeah, this is what I meant - I can't think of anything better, certainly.  This code is such a hairball...

cheesestraws (approved):

PR #2432, opened 2025-09-24
Consider the sqlite CNID backend stable

rdmark:
Remove the warning when loading the sqlite backend library

Remove the notes in documentation about it being experimental

Refactor the description of the CNID backends


andylemin (approved):
Congratulations 🙂

PR #2437, opened 2025-09-26
testsuite: Clean test volumes between each spectest test execution

rdmark:
Enumerate and delete all files and directories in Vol and Vol2 through a new utility function called clear_volume() in afphelper

A new `-E` option to afp_spectest enables this mode (default: off) while the other test suites don't use it

Lingering files between test executions may lead to unexpected test results, so using this option can give you more consistent results, or give you a shortcut for cleaning up a test volume before running tests

Refactor the testsuite to use is_there() from afphelper globally


andylemin (approved):
:)

PR #2439, opened 2025-09-26
afpd: introduce runtime option for valid shellcheck

rdmark:
Remove the compile time disable valid shellcheck option in favor for a runtime option configurable in afp.conf

A known issue is that papd doesn't have the capability to configure valid shellcheck or even propagates the global papd settings to the UAM library;
for now we always keep valid shellcheck enabled for papd and leave a FIXME in the code

Also, fixes a bug where valid shellcheck was never done in the ClearTxt PAM UAM for afpd (but was doing it for papd auth)


andylemin (approved):
LGTM

PR #2441, opened 2025-09-28
getzones: protect against potential memory leak when exiting in optarg

rdmark:
Depending on the optarg input string, the loop may exit after the requested_zone variable has been dynamically allocated, so let's free the memory every time


andylemin (approved):
LGTM

PR #2443, opened 2025-09-28
testsuite: refactor headers and declarations for internal consistency

rdmark:
Symbol declarations and macros have been organized into headers which matches a module in the libafptest library

afpclient.h - AFP client implementation
afpcmd.h - AFP command wrapper functions
afphelper.h - test supporting utility functions using AFP commands
testhelper.h - pure test runner control functions, macros, and data structures

The meta header "specs.h" has been removed

Changeset also includes some refactoring of global headers to enable the testsuite improvements and make globals.h less monolithic:

- afp_util.c has been moved from afpd to libatalk, and the declarations have been moved from atalk/global.h to new header atalk/afp_util.h (because I want to use AfpNum2name() in the spectest)
- atalk/iniparser_util.h has been added and wrapper macros moved from globals.h

And finally since the afpclient.c code was touched, SonarQube flagged all code quality bugs, which are addressed here to keep the quality gate green.


rdmark:
@andylemin turns out the problem we had with compiling on Solaris was solved by simply including "config.h" in lantest.c!

This is now the only header that always has to be included first. ;)

andylemin:
Ok, so config.h replaces specs.h as only one needing to be first. This is nicer as config.h is smaller and looks like all its contents have a reason to be early.

andylemin:
Any thoughts on where the new safe atoi should live now?

Also I think you should merge this first, and I will rebase and update mine against this?

andylemin (approved):
Big changes, will help a bunch of things 👍

rdmark:
> Any thoughts on where the new safe atoi should live now?

How about declaring the function(s) in include/atalk/util.h and then add a new file under libatalk/util/ to house it?

> Also I think you should merge this first, and I will rebase and update mine against this?

Let's do that, since this PR is ready to be merged now. Pardon the extra busywork for your PR.

Issue #2358, opened 2025-08-15
crash on macOS when client connects to server with afpstats enabled


EricFromCanada:
**Describe the bug**

While attempting to get `afpstats` working on macOS, I found that `afpd` often crashes when a client connects to a server that has `afpstats = yes` set, whether dbus is running or not.

**To Reproduce**

1. Install `netatalk` from Homebrew
2. set `afpstats = yes` in [Global] of afp.conf
3. ensure dbus is not running
4. `sudo netatalkd start`
5. (optional) run `afpstats` to get error: `org.freedesktop.DBus.Error.NoServer: Failed to connect to socket /opt/homebrew/var/run/dbus/system_bus_socket: Connection refused`
6. from another Mac: connect to the server

**Expected/Actual behavior**

- expected: The client Mac would connect.
- actual: The client Mac gives "There was a problem connecting to the server" right after authenticating, and afpd sometimes crashes on the server

Simply removing `afpstats = yes` allows netatalk to work properly.

**Environment**

- Server OS: macOS Sonoma arm64
- Client OS: macOS Monterey x86_64
- Netatalk Version: 4.3.0

**Configuration**

```
[Global]
log file = /var/log/netatalk.log
log level = default:maxdebug
afpstats = yes

[Homes]
basedir regex = /Users
```

**Logs**

see attachments

[afpd crash.zip](https://github.com/user-attachments/files/21800383/afpd.crash.zip)

rdmark:
Good catch! I suspect we're not failing gracefully when the dbus connection fails in afpstats.c which corrupts the stack and leads to the crash later. The callstack is almost certainly a red herring.

rdmark:
I spent the weekend hacking on this, and can unfortunately only conclude that afpstats cannot easily work on macOS with the current architecture. After adding thorough error handling during thread initialization, attempting various ways to check the status of D-Bus on the system as early as possible, afpd child processes still consistently hit a SIGABRT after authentication in this scenario. At this stage, I am suspecting that the non-fork-safe nature of macOS is a factor here, with bad things happening when a forked afpd attempts to interact with GThread.

Like with Darwin ACLs, I propose adding a build system check that disables afpstats on macOS with a warning. Thoughts?

PR #2448, opened 2025-10-02
afpd: afpstats init error handling; don't build on macOS

rdmark:
A set of minor changes to afpstats for better logging, error checking during startup, while not building on macOS with warning message

- thorough error handling while initializing afpstats
- don't build afpstats on macOS (addresses #2358)
- touch up format of afpstats logging messages


andylemin (approved):
LGTM

PR #2450, opened 2025-10-03
meson: error out if no writable CNID backend selected for compilation

rdmark:
Since the last backend is read-only, we want to error out rather than automatically selecting it as the default CNID backend in the absence of any other backend

You can still force the last backend as default with -Dwith-cnid-default-backend if you absolutely want to (but I cannot see when this would be useful)

This effectively introduces a hard build dependency on at least one of the three database libraries: bdb, mysql, or sqlite


andylemin (approved):
LGTM

PR #2459, opened 2025-10-07
Update Configuration.md

andylemin:
Added Directory Cache Tuning section, and moved the SSH tunneling section to the bottom of the page to keep more relevant items at the top.

### DIFF
--- a/doc/manual/Configuration.md
+++ b/doc/manual/Configuration.md
@@ -490,92 +490,6 @@ A small overview of the officially supported UAMs.
 | Server support   | uams_guest.so | uams_cleartxt.so | uams_randnum.so  | uams_dhx.so  | uams_dhx2.so  | uams_gss.so      |
 | Password storage | None          | Either system auth or PAM | Passwords stored in clear text in a separate text file | Either system auth or PAM | Either system auth or PAM | At the Kerberos Key Distribution Center |
 
-### SSH tunneling
-
-Tunneling and VPNs usually have nothing to do with AFP authentication
-and UAMs. But since Apple introduced an option called "Allow Secure
-Connections Using SSH" and many people tend to confuse the two, we'll
-cover this here.
-
-#### Manually tunneling an AFP session
-
-This works since the first AFP servers that spoke "AFP over TCP"
-appeared in networks. One simply tunnels the remote server's AFP port to
-a local port different than 548 and connects locally to this port
-afterwards. On macOS this can be done by
-
-    ssh -l $USER $SERVER -L 10548:127.0.0.1:548 sleep 3000
-
-After establishing the tunnel one will use `"afp://127.0.0.1:10548"` in
-the "Connect to server" dialog. All AFP traffic including the initial
-connection attempts will be sent encrypted over the wire since the local
-AFP client will connect to the Mac's local port 10548 which will be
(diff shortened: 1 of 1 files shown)


doc/manual/Configuration.md
```diff
@@ -785,6 +699,80 @@ ACL_MASK entry is present too, afpd recalculates its value so that the
 new group rights become effective and existing entries of type ACL_USER
 or ACL_GROUP stay intact.
 
+## Directory Cache Tuning
```
rdmark, line 702: I would love to see a preamble paragraph here that explains at a high level the mechanism of netatalk's dircache and why a user should invest time into configuring it. For instance, that we use conservative defaults but if netatalk is the only process accessing the shared volume you can expect significant performance gains. What do you think?

doc/manual/Configuration.md
```diff
@@ -785,6 +699,80 @@ ACL_MASK entry is present too, afpd recalculates its value so that the
...
+
+If Netatalk is the only process accessing the volume you can safely
+set a value of 1800.
+
+**Example** (Netatalk only access to volume):
+dircache validation freq = 100
+dircache metadata window = 3600
+dircache metadata threshold = 1800
+
+**Note**: Monitor cache effectiveness with `afpstats` or check log files
```
rdmark: How do you use afpstats for this purpose? I don't recall you adding any new functionality to it, but I could be overlooking something. :)

rdmark (approved):
Good call moving the SSH tunneling section further down, and fixing the heading hierarchy. We might even consider moving that section out of the manual and to a Tech Notes wiki page for instance, since the use case is arguably very narrow (and maybe not even support in macOS anymore?)

rdmark:
Oh, and please squash the two commits!

NJRoadfan:
I would remove the SSH tunneling section completely. I think Apple dropped support for it in OS X 10.6 and the OSes that do support it are too old to support modern SSH ciphers.

rdmark:
I created a wiki page with the ssh tunneling sections and listed it under "Retro AFP" in the sidebar: https://github.com/Netatalk/netatalk/wiki/SSH-Tunneling

Ready to be removed from the manual now!

doc/manual/Configuration.md
andylemin: Thanks for spotting this one. afpstats for these was not implemented in the end. The PR got too large with all the bug fixes that this work surfaced.

doc/manual/Configuration.md
andylemin: Done :)

andylemin:
All changes done 👍

doc/manual/Configuration.md
```diff
@@ -843,3 +852,4 @@ must use the global option **dbus daemon** to point to the path, e.g. for
 Solaris with Tracker from OpenCSW:
 
     dbus daemon = /opt/csw/bin/dbus-daemon
+
```
rdmark: The linter doesn’t like this added newline ;)

rdmark:
It would be good to have a line in the commit message about why we removed the SSH section. Handy for future perusers of the commit log.

doc/manual/Configuration.md
andylemin: Done

rdmark (approved):

PR #2461, opened 2025-10-07
nad: remove limits to operate on shared volumes using any CNID backend

rdmark:
This removes validations for CNID backend, EA metadata, and adouble format.

Also, tweaks the error logging in the sqlite CNID backend to avoid overly verbose logging when using the module in nad.


rdmark:
@NJRoadfan was there any other modifications to (n)ad that you did in the past?

NJRoadfan:
I have tested this with the check turned off in the past. It seemed to work with other CNID backends at the time. The "find" utility doesn't work because only dbd has support for it.

bin/nad/nad_util.c
```diff
@@ -110,28 +108,15 @@ int openvol(AFPObj *obj, const char *path, afpvol_t *vol)
...
-    /* Sanity checks to ensure we can touch this volume */
-    if (vol->vol->v_adouble != AD_VERSION2
-            && vol->vol->v_adouble != AD_VERSION_EA) {
-        ERROR("Unsupported adouble versions: %u", vol->vol->v_adouble);
-    }
-
-    if (vol->vol->v_vfs_ea != AFPVOL_EA_AD && vol->vol->v_vfs_ea != AFPVOL_EA_SYS) {
-        ERROR("Unsupported Extended Attributes option: %u", vol->vol->v_vfs_ea);
-    }
-
```
NJRoadfan, line 126: These sanity checks should remain in the code. This can be an extra guard against running these tools in a directory that is not a Netatalk volume.

rdmark: Fair enough, but let me add support for AFPVOL_EA_NONE

PR #2462, opened 2025-10-08
webmin: support newly added dircache tuning options

rdmark:
Adding support for three new directory cache tuning options, including the defaults and valid ranges


andylemin (approved):
Nice! Thanks for doing this @rdmark 🙂

PR #2464, opened 2025-10-08
cnid: implement cnid_find() in the sqlite backend

rdmark:
The find API function in the CNID library is used specifically by the "nad find" command, which can be handy in a pinch

### DIFF
--- a/libatalk/cnid/sqlite/cnid_sqlite.c
+++ b/libatalk/cnid/sqlite/cnid_sqlite.c
@@ -182,6 +182,29 @@ EC_CLEANUP:
     EC_EXIT;
 }
 
+static int init_prepared_stmt_find(CNID_sqlite_private *db)
+{
+    EC_INIT;
+    char *sql = NULL;
+
+    if (db->cnid_find_stmt) {
+        sqlite3_finalize(db->cnid_find_stmt);
+    }
+
+    EC_NEG1(asprintf(&sql, "SELECT Id FROM \"%s\" WHERE Name LIKE ? ORDER BY Id",
+                     db->cnid_sqlite_voluuid_str));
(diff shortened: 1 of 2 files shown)


rdmark:
@NJRoadfan it wasn't too bad to make this work with sqlite

libatalk/cnid/sqlite/cnid_sqlite.c
```diff
@@ -931,6 +957,7 @@ cnid_t cnid_sqlite_get(struct _cnid_db *cdb, cnid_t did, const char *name,
     return id;
 }
 
+
```
andylemin: Was this extra new line intended?

libatalk/cnid/sqlite/cnid_sqlite.c
```diff
@@ -1031,14 +1058,67 @@ int cnid_sqlite_getstamp(struct _cnid_db *cdb, void *buffer,
 }
 
 
-int cnid_sqlite_find(struct _cnid_db *cdb _U_, const char *name,
-                     size_t namelen _U_,
-                     void *buffer _U_, size_t buflen _U_)
+int cnid_sqlite_find(struct _cnid_db *cdb, const char *name, size_t namelen,
```
andylemin, line 1059: Interesting to see you removing the undefined macro. I was wondering about removing this myself in cache.c

libatalk/cnid/sqlite/cnid_sqlite.c
rdmark: This was a copy paste mistake!

libatalk/cnid/sqlite/cnid_sqlite.c
rdmark: Right, if we start using parameters, let's remove the `_U_` macros!

PR #2467, opened 2025-10-09
docs: generate dynamic developer documentation with Doxygen

rdmark:
This introduces a new meson option -Dwith-docs=developer which will generate documentation on the netatalk C codebase using Doxygen

We repurpose the DEVELOPER.md as the landing page for the generated developer docs

The generated code will be installed into /usr/share/doc/netatalk/developer or equivalent for your platform


rdmark:
Note that with this structure, we can include any number of markdown pages as supplementary developer docs. It's just a matter of adding them to the `FILE_PATTERNS` filter. Anything apart from the hard coded index page will be kept under a _Related Pages_ heading.

rdmark:
Published pre-release doxygen docs to the website (not linked to from anywhere yet)

https://netatalk.io/developer/

PR #2471, opened 2025-10-11
meson: refactor the docs installation dir logic

rdmark:
This renames the -Dwith-manual-install-dir option to -Dwith-docs-install-dir

The new option specifies the prefix for all docs, rather than the html manual subdir

In this way, we can use the same prefix for all kinds of generated docs: html manual, converted readmes, and the doxygen docs

Notably, when used in tandem with the -Dwith-website=true option, this helps this project's maintainers with automating the website maintenance

### DIFF
--- a/doc/manpages/man1/meson.build
+++ b/doc/manpages/man1/meson.build
@@ -57,6 +57,6 @@ endforeach
 
 if get_option('with-website')
     foreach page : manfiles
-        install_data(page + '.1.md', install_dir: manual_install_path + '/en')
+        install_data(page + '.1.md', install_dir: docs_install_path + '/manual/en')
     endforeach
 endif
--- a/doc/manpages/man3/meson.build
+++ b/doc/manpages/man3/meson.build
@@ -25,6 +25,6 @@ endforeach
 
 if get_option('with-website')
     foreach page : manfiles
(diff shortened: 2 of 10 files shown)


rdmark:
FTR, with this change, a project maintainer would configure and install netatalk like this when updating the website for a new release, where /path/to/dev/netatalk.io is where the [static website code](https://github.com/Netatalk/netatalk.io) is cloned.

```
meson setup build -Dwith-appletalk=true -Dwith-testsuite=true -Dwith-website=true -Dwith-docs-l10n=true -Dwith-docs-install-path=/path/to/dev/netatalk.io
meson compile -C build
meson install -C build
```

Afterwards, you commit the changes to the netatalk.io code and push to remote. The Cloudflare webhook will then trigger the build and deployment of the static website.

rdmark:
@andylemin No, this is not something that we're documenting in detail on the website currently. I think we mention in a few places in the docs that you can run `meson configure` in the root of the source tree to get a list of available build options. Meson is self documenting in this sense.

Beyond this, I make sure to list any new or changed build options in the release notes.

Open to suggests how to make this even better for end users! Maybe we can pipe the output of `meson configure` to a markdown page and publish that somewhere on the website. :)

PR #2473, opened 2025-10-11
docs: create a chapter in the developer docs for DDP / AppleTalk

rdmark:
The new chapter is made up of content from the AppleTalk section in the html manual, the APPLETALK.md readme, and the AppleTalk Kernel Module wiki article

At the same time, some contents have moved from the readme into the wiki article. The APPLETALK.md readme file is removed altogether.

The new structure is as follows:
- Internal configuration concerning netatalk and atalkd is in the manual
- Architectural design and code references are in the developer docs
- OS specific capability and configuration details are in the wiki


rdmark:
The new developer docs have been published to the website for demonstration: https://netatalk.io/developer/md_developer_2ddp

The neat thing about this new structure is that you can create any arbitrary number of markdown files under _doc/developer_ and doxygen will pick it up and add it to the index.

PR #2482, opened 2025-10-13
docker: restructure containerization support files into distrib/docker

rdmark:
Consolidating all files that have to do with Docker into a single subdir, for discoverability and maintainability

The drawback is that that you have do specify the file that docker build and docker compose should read, ex:

```
% docker build -t netatalk:latest -f distrib/docker/netatalk.Dockerfile .
% docker compose -f distrib/docker/docker-compose.yml up
```

Additionally, refactor meson and dockerfiles to copy only required build files over to workdir

By explicitly copying only what's absolutely necessary for the configuration, we save on build runtime as well as protect against future unexpected file inclusions

### DIFF
--- a/distrib/docker/netatalk.Dockerfile
+++ b/distrib/docker/netatalk.Dockerfile
@@ -48,7 +48,22 @@ RUN apk update \
     $BUILD_DEPS
 
 WORKDIR /netatalk-code
-COPY . .
+COPY bin/ ./bin/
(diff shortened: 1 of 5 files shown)


rdmark:
If you think this is a misguided change then I would love to know! Are there particular files that you should should stay where they are?

rdmark:
Testing wiki updated!

PR #2484, opened 2025-10-14
docs: custom header and footer in doxygen docs

rdmark:
The base header and footer html code is generated by doxygen, and then modified for netatalk

Modifications are:

- titlearea rewritten to use CSS block style instead of a table
- import mermaid.min.js script for rendering mermaid markdown
- netatalk logo is now a hyperlink back to the website root

Kudos to Pieter P and Tapani Hyvamaki for helpful examples in the [doxygen-mermaid](https://github.com/tttapa/doxygen-mermaid) project

Now we pull in two wiki pages as developer docs: dircache and libatalk

### DIFF
--- a/doc/developer/libatalk.md
+++ b/doc/developer/libatalk.md
@@ -1,0 +1,153 @@
+# Using libatalk from third party apps
+
+Third party developers wishing to add Netatalk support to their applications have two choices
+
+1. Directly integrate libatalk from the Netatalk source tree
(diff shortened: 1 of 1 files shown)


rdmark:
The new dircache doxygen page has been deployed to the website as a demonstration! It looks really good if I may say so myself.

https://netatalk.io/developer/md_developer_2dircache

andylemin (approved):
Nice! The renderings look great..

PR #2485, opened 2025-10-15
improve code documentation and remove dead code

rdmark:
Touch up all code documentation syntax warnings thrown by doxygen, standardizing on Qt style with consistent application of command tags, while correcting typos and grammar

Remove parameters and entire functions that are unused, rather than fixing the code documentation for dead code

Now since the code documentation is clean and compliant, let's treat warnings as errors in doxygen

Introducing new meson option `-Dwith-doxygen-strict=true` which enables error-on-warning behavior in doxygen; the reason for making this opt-in, is because doxygen will treat any unknown Doxyfile option as a warning=>error, so if you use anything but the very latest doxygen 1.14 you get a ton of errors (because doxygen apparently keeps adding new options all the time)

Finally, build doxygen docs in a new job in the Tests workflow so that we can get an early warning when invalid code documentation gets introduced

### DIFF
--- a/doc/Doxyfile.in
+++ b/doc/Doxyfile.in
@@ -1062,7 +1062,7 @@ EXCLUDE_SYMLINKS       = NO
 # Note that the wildcards are matched against the file with absolute path, so to
 # exclude all test directories for example use the pattern */test/*
 
-EXCLUDE_PATTERNS       = */build/*
+EXCLUDE_PATTERNS       = */build/* */dummy.c */meson_config.h
 
 # The EXCLUDE_SYMBOLS tag can be used to specify one or more symbol names
 # (namespaces, classes, functions, etc.) that should be excluded from the


rdmark:
This was a deeper rabbit hole than I anticipated. When I started digging, there where all sorts of code documentation in unusual formats, and corner cases that required additional doxygen configuration or refactoring of surrounding code.

I'm pretty happy with how it has turned out now. The latest branch revision has been published to the website again: https://netatalk.io/developer/

The one ugly workaround right now is some hard coded preprocessor mapping of the function parameter macros that the libatalk/vfs module uses. See `PREDEFINED` in the Doxyfile. I could not get doxygen to expand those macros organically despite much effort. It might be an argument for refactoring in the vfs module in the long run. It confuses everybody.

PR #2488, opened 2025-10-19
Reduce the usage of preprocessor macros to enable platform code

rdmark:
**macipgw: MacIP tunnel platform code chosen by build system**

Rather than depending on preprocessor macros to check for Linux or *BSD, use the mson build scripts to the same effect

Among other things, this enables doxygen code documentation of both tunnel modules to be generated at once

**POSIX and Solaris remove ACL functions separation**

Before this change, the remove_acl_vfs() function was named the same between the POSIX and Solaris implementations, separated by preprocessor macros

This makes the naming unique so that both can be enabled without a namespace conflict, e.g. for doxygen documentation purposes

**control building of the last CNID backend through the build system**

No need to put the last backend code behind a macro here

### DIFF
--- a/doc/Doxyfile.in
+++ b/doc/Doxyfile.in
@@ -2450,6 +2450,7 @@ PREDEFINED             = \
     HAVE_LDAP=1 \
     HAVE_LIBQUOTA=1 \
     HAVE_MDNS=1 \
+    HAVE_NFSV4_ACLS=1 \
     HAVE_POSIX_ACLS=1 \
     HAVE_SPLICE=1 \
     TCPWRAP=1 \


rdmark:
There is definitely other places in the code where we can potentially cut down on feature flag macros. If you see the Doxyfile PREDEFINED list, we currently do this (which in itself is far from a complete list, just what I spotted at a glance):

```
    HAVE_ACLS=1 \
    HAVE_AVAHI=1 \
    HAVE_CUPS=1 \
    HAVE_DBUS_GLIB=1 \
    HAVE_LDAP=1 \
    HAVE_LIBQUOTA=1 \
    HAVE_MDNS=1 \
    HAVE_NFSV4_ACLS=1 \
    HAVE_POSIX_ACLS=1 \
    HAVE_SPLICE=1 \
    TCPWRAP=1 \
    USE_CRACKLIB=1 \
    WITH_DTRACE=1 \
    WITH_RECVFILE=1 \
    WITH_SENDFILE=1 \
    WITH_SPOTLIGHT=1 \
```

Issue #2024, opened 2025-03-15
nad: segfault when modifying files on volume owned by other user


rdmark:
If you run `ad cp` or `ad mv` (etc.) on a file on a shared volume where the user running `ad` is also the owner of the dir/files on the shared volume, the operations succeed without issue.

However, if the dir/files are owned by another user, ad throws a segfault.

This is not a new bug. I could reproduce as far back as 3.1.12.

```
dmark@pretrix:/srv/afptest1$ ad cp 1 11
Mar 15 15:13:05.099331 ad[125273] {netatalk_conf.c:2024} (error:Default): getvolbypath("/srv/afptest1/11"): path does not match basedir regex "/home"
Mar 15 15:13:05.099435 ad[125273] {netatalk_conf.c:2024} (error:Default): getvolbypath("/srv/afptest1/1"): path does not match basedir regex "/home"
Segmentation fault
```

Sample backtrace

```
Program received signal SIGSEGV, Segmentation fault.
copy (path=0xaaaaaaaca850 "README", statp=0xffffffffc6d0, tflag=0, ftw=0xffffffffc778) at ../bin/ad/ad_cp.c:376
376	    if (!dvolume.vol->vfs->vfs_validupath(dvolume.vol, dir))
(gdb) bt
#0  copy (path=0xaaaaaaaca850 "README", statp=0xffffffffc6d0, tflag=0, ftw=0xffffffffc778) at ../bin/ad/ad_cp.c:376
#1  0x0000fffff7f4dea8 in ftw_startup (dir=0xfffffffff6ff "README", is_nftw=1, func=0xaaaaaaaa3694 <copy>, 
    up=0xaaaaaaaa2d28 <upfunc>, descriptors=20, flags=19) at ../libatalk/util/ftw.c:817
#2  0x0000fffff7f4dfc4 in nftw (path=0xfffffffff6ff "README", func=0xaaaaaaaa3694 <copy>, up=0xaaaaaaaa2d28 <upfunc>, 
    descriptors=20, flags=19) at ../libatalk/util/ftw.c:862
#3  0x0000aaaaaaaa35d4 in ad_cp (argc=1, argv=0xfffffffff498, obj=0xffffffffc990) at ../bin/ad/ad_cp.c:338
#4  0x0000aaaaaaaaa568 in main (argc=4, argv=0xfffffffff488) at ../bin/ad/ad.c:68
```

NJRoadfan:
[...]

PR #2491, opened 2025-10-20
libatalk: refactor vfs module with explicit function definitions

rdmark:
Replace VFS function macros with explicit function implementations and clean up the module initialization code

Each VFS operation is now defined as a plain C function with proper type definitions in the header file

This improves code readability and maintainability by removing macro magic

The VFS module chain limit of 3 modules is preserved, supporting the existing layers: primary (adouble), secondary (EA) and tertiary (ACL) modules.

In addition, all instances of the `restrict` C keyword has been removed (previously broadly used in the vfs module) because they offer marginal memory savings with the tradeoff of heightened risk of memory corruption

### DIFF
--- a/libatalk/vfs/ea_ad.c
+++ b/libatalk/vfs/ea_ad.c
@@ -969,7 +969,8 @@ exit:
  * @note Copies EA size into rbuf in network order. Increments *rbuflen +4.
  */
 int get_easize(const struct vol *vol, char *rbuf, size_t *rbuflen,
-               const char *uname, int oflag, const char *attruname, int fd)
+               const char *uname, int oflag _U_, const char *attruname,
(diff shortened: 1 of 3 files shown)


andylemin (approved):
Wow this is great! Definitely will need some soak testing. I'll deploy this into my test setup soon and share any issues.
I'm hoping this might solve some other edge issues I've seen with orphaned file locks on ZFS filesystems. Not sure if that's a race or corruption issue yet, this will help diagnose at the very least 🚀

rdmark:
I'm hoping it will have a positive effect on the memory corruption I observed in `nad` as well, which I could track down to execution paths deep in the vfs module code. https://github.com/Netatalk/netatalk/issues/2024

PR #2495, opened 2025-10-24
testsuite: use the global afp.h interface for AFP commands

rdmark:
Do away with the local custom fork of the AFP interface in favor of the global afp.h header

From the removed afp.conf, we only salvage BITERR_* error code macros; everything else was redundant


rdmark:
With this change, only test/testsuite/dsi.h remains as forked code from netatalk proper.

I spent a few hours tinkering with this last puzzle piece today, but the refactoring is complicated by 8e5e83d for starters, which changed the dsi->commands header from a fixed length static buffer to a dynamically allocated one. This requires pretty broad changes across the testsuite. Secondly, both the command and data buffers got a lot larger - previously fixed at 8192 bytes and fit in 16 bit integers, but now broad changes will be required to accommodate 32 bit buffers. As a result, we get cascading refactoring of network byte order handling to deal with as well.

A project for another day!

andylemin (approved):
Nice, always good to remove redundant duplication 🙂

PR #2506, opened 2025-11-03
Make the nad commands work more reliably on any OS

rdmark:
Changes to the nad binary:

allow non-owning users to do file operations in volumes

This is a collection of improvements that make nad commands work reliably on f.e. Linux, in particular when operating on volumes that aren't owned by the current user but by group membership

- explicitly set AFP uid to the calling user, otherwise we get treated as root
- collect a users's group belongings when initializing nad, so that group based allow lists work
- shore up error handling in cp and rm commands to bail out early and avoid segfaults
- set log level as per the afp.conf settings rather than hard coding to default:info
- improve debug logging across the module
- fleshes out the nad.1 man page with notes on user permissions and other improvements

Changes to the libatalk library:

- bypass volume access check for root users in util module
- improves the surrounding code documentation

Before, the root user was denied access to shared volumes when using the "valid users" afp.conf option, notably when operating on the shared volume on the host running "nad" as root (e.g. with sudo)

### DIFF
--- a/doc/manpages/man1/nad.1.md
+++ b/doc/manpages/man1/nad.1.md
@@ -16,12 +16,16 @@ directories or files in **.AppleDouble** directories) and the CNID
 databases are updated as appropriate when files in a shared Netatalk
 volume are modified.
 
-This is preferable over using the equivalent system file operation commands
-on files in a Netatalk shared volume to preserve the integrity of Mac OS
-metadata and accuracy of the CNID database.
+Using **nad** is preferable over the operating system's native file operation commands
+on files and directories in a Netatalk shared volume, because it preserves the integrity
+of Mac OS metadata and accuracy of the CNID database.
 
 It depends on Netatalk running on the host system and the AFP volume being
 shared by Netatalk.
+
+Only users with appropriate permissions to access the files and directories
+can use **nad** to manipulate them.
+It is sensitive to afp.conf settings such as *valid users* and *invalid users*.
 
 # Available Commands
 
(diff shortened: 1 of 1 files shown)


rdmark:
@NJRoadfan this seems to have done the trick with stabilizing nad (ad), at least for basic usage patterns. Tested on Debian, Alpine, macOS, FreeBSD and OpenIndiana.

The main limitation is that a user cannot perform actions on the shared volume dir that it doesn't have standard file system permissions for, regardless of afp.conf settings. The workaround is to run nad as root (e.g. sudo) which now works after a small change to libatalk.

andylemin (approved):
Nice! Great to see `nad` looking more complete.

NJRoadfan:
That Darwin specific macro duplication is driving me crazy. Any reason why a `#define` wasn't used here instead?

Something like:

```
#ifdef __APPLE__
#define gid_t int
#endif
```

Technically `gid_t` is an alias for an unsigned int in the GNU C library. I'm surprised that Darwin doesn't define this type. Docs say it should be there.

rdmark:
`git_t` is typedefed to unsigned int on macOS, but the awkward thing is that the getgrouplist() prototype on macOS is `int getgrouplist(const char *, int, int *, int *__ngroups)` while on f.e. linux `int getgrouplist(const char *, gid_t, gid_t *, int *ngroups)`

So at a glance there is this weird mismatch / bug in the macOS C library. Maybe there's a good reason for this.

Anyways, I see somewhat of a risk to redefining the git_t type here since it overrides the type globally after the macro (unless I'm mistaken.)

NJRoadfan:
If this is the only place `getgrouplist()` is used, define it as a third type and use that in the function instead:

```
#ifdef __APPLE__
#define my_gid_type int
#else
#define  my_gid_type gid_t
#endif
```

Along with plenty of code documentation on why this was done, it shouldn't be a problem for a future maintainer. Just trying to prevent duplication of code and possible problems down the road when one half of the macro definition receives a revision and the other half is missed.

rdmark:
That's a clever trick, I like it! And yes, getgrouplist() is only called in this function in netatalk.

rdmark:
I will go ahead and merge this now without re-review, to unblock another branch I'm working on.

PR #2507, opened 2025-11-04
libatalk: create utility function convert_utf8_to_mac()

rdmark:
The convert_utf8_to_mac() function is based on utompath() which is implemented in etc/afpd/desktop.c but of a simpler, earlier form – we remove two duplicated implementations in nad and dbd respectively

In comparison, utompath() can handle other input charsets than UTF-8, while also doing Mac filename mangling after charset conversion

As a side note, utompath() in etc/afpd/desktop.c cannot easily be moved to libatalk since the mangling logic also sits in the afpd code, and has deep coupling with other afpd logic such as dirlookup(). A much larger refactoring would be needed here.

### DIFF
--- a/bin/nad/nad_cp.c
+++ b/bin/nad/nad_cp.c
@@ -590,7 +590,7 @@ static int copy(const char *path,
             ad_setid(&ad, st.st_dev, st.st_ino, did, pdid, dvolume.db_stamp);
 
             if (dvolume.vol->v_adouble == AD_VERSION2) {
-                ad_setname(&ad, utompath(dvolume.vol, basename(to.p_path)));
+                ad_setname(&ad, convert_utf8_to_mac(dvolume.vol, basename(to.p_path)));
             }
 
             ad_setdate(&ad, AD_DATE_CREATE | AD_DATE_UNIX, (uint32_t) st.st_mtime);
@@ -667,7 +667,7 @@ static int copy(const char *path,
             ad_setid(&ad, st.st_dev, st.st_ino, cnid, did, dvolume.db_stamp);
 
             if (dvolume.vol->v_adouble == AD_VERSION2) {
-                ad_setname(&ad, utompath(dvolume.vol, basename(to.p_path)));
+                ad_setname(&ad, convert_utf8_to_mac(dvolume.vol, basename(to.p_path)));
             }
 
             ad_setdate(&ad, AD_DATE_CREATE | AD_DATE_UNIX, (uint32_t) st.st_mtime);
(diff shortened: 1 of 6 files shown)


andylemin:
I've been meaning to chat to you about this (UTF8 code page things) as I've noticed a few strange things in this area.

First setting the client format type (maccodepage) to UTF-MAC as documented results in errors in the afpd log saying it's unrecognised - intention to minimise conversion overhead from UTF-8 on storage to the client (by default it's MacRoman and want to force Decomposed UTF-8 (UTF8-MAC).

I also think there is something else funky, as you can create a file with '/' in the name (yes that's a bad thing to do) on the client. But it gets created fine, but you cannot then access it, edit or delete it thereafter as the '/' gets mangled to ':' on the fly.

rdmark:
To some extent the codepage shenanigans are there to enable compatibility with AFP2 and pre-OSX clients. But maybe this is beside the point here? You're saying that we can optimize afpd by changing the default internal representation of file names to UTF8-MAC consistently?

The forward slash file name thing seems like an outright bug to me – any reason not to disallow it? :)

In any case, I think the refactoring in this PR should make it easier to do future improvements.

andylemin:
Hi. Yes there is an analysis of the computational overhead for converting between the different charsets in the wiki dev docs I put together (in the performance tuning page). I'm not near my computer so can't share the link right now. But it discusses the overhead differences.

If I remember correctly, converting between UTF-8 and MacRoman (and others) requires a mapping lookup. While converting from composed (UTF8) to Decomposed (UTF8-MAC) does not require any lookups - so avoids a lot of memory ops.

This is why UTF8-MAC was part of AFP3 I believe, but I don't think Netatalk fully complied with this. There is also a whole thing about this in Apple's switch from AFP to APFS for the same performance reasons.

From our docs; "With OS X Apple introduced the AFP3 protocol. One of the big changes was, that AFP3 uses Unicode names encoded as Decomposed UTF-8 (UTF8-MAC)."

For AFP2, the default should stay MacRoman, but for AFP3 I think we should technically be defaulting to UTF8-MAC. But I failed to test due to the error when forcing UTF8-MAC. Also need to double check that Linux and Windows clients would gracefully interpret Decomposed UTF8 (they should).

It's been in the back on my mind for ages, and this PR just reminded me about it :) 

For the forward slash, I agree we should just block/error during initial file creation, as it's wrong for many reasons to have a forward slash in a filename? What do you think?

PS; agree no reason to delay this commit

andylemin (approved):

rdmark:
It will be particularly interesting to me to see how we can optimize the charset conversion logic without breaking AFP2 compatibility. How does afpd know that a connection is AFP2 for every time a charset conversion routine is called?  Maybe you already have done that analysis. :)

PR #2523, opened 2025-11-13
docker: expose dircache settings in production container

rdmark:
Adds four new env variables for the docker entrypoint script:

- AFP_DIRCACHESIZE
- AFP_DIRCACHE_VALIDATION_FREQ
- AFP_DIRCACHE_METADATA_WINDOW
- AFP_DIRCACHE_METADATA_THRESHOLD

The defaults are the same that are documented in the afp.conf.5 man page

The production container test jobs in GitHub CI are now using the optimized dircache settings, except for `dircache validation freq` which causes test failures

Also contains some tweaks to the Japanese localization for good measure


rdmark:
filed https://github.com/Netatalk/netatalk/issues/2524 for the test failures

rdmark:
the other optimization settings have no averse effect, so let's just turn `dircache validation freq` off for now until we investigate the failures

andylemin:
Which tests were failing when setting freq above 1?

I'll have a look at updating the tests to play nice with periodic validation in a later commit 👍

andylemin (approved):

rdmark:
This is the list of failures (when using sqlite):

```
  Failed tests:
    Dircache:test500: move and rename dir, enumerate new parent, stat renamed dir
    Dircache:test501: move and rename dir, then stat it
    Dircache:test502: move and rename dir, enumerate renamed dir
    Dircache:test503: move and rename dir, enumerate renamed dir
    Dircache:test504: rename topdir, stat file in subdir of renamed topdir
    Dircache:test505: rename dir, stat subdir in renamed dir
    Dircache:test506: stat subdir in poisoned path
    Error:test174: did error two users from parent folder did=<deleted> name=test174 name
```

I have created issue ticket #2524 and outlined my findings so far.

PR #2525, opened 2025-11-14
docs: split up developer docs per topic and make improvements

rdmark:
Rather than following the original readme structure, split into one page per topic

Tweaks some of the verbiage for clarity and reformat inline code for style compliance


andylemin (approved):
🙂

PR #2526, opened 2025-11-15
Revise README project description and verbiage

rdmark:
Tweaked the verbiage of the opening paragraphs to better convey the purpose of netatalk

Removed the project status badges, since they've proven flaky and/or of less value


rdmark:
@andylemin I should be good to go with the release after this!

PR #2527, opened 2025-11-22
note in README that the wiki is edited by project members

rdmark:


rdmark:
This PR is to start a conversation about the GitHub wiki. I would like to change the permissions so that only project members can edit. Right now, any registered GitHub user can edit. The reasoning is twofold:

First, over these last 3+ years, I have not seen a single good edit from non-project members. They have either been mistaken/misguided, or spam. By restricting write privileges, one type of maintenance overhead goes away (that is, monitoring wiki edits.)

Secondly, GitHub will flag a wiki for search engine indexing (controlled with the x-robots-tag header) only for projects with restrictive settings. We are now close to the 500 star mark where GitHub will [consider us a significant project](https://github.com/orgs/community/discussions/4992) and open up for indexing.

@andylemin @cheesestraws what do you think?

cheesestraws:
+1 from me here

(once again, apologies for my continued silence; life is being quite the thing.)

cheesestraws (approved):

PR #2534, opened 2025-11-24
Create git configuration file for bypassing git blame

rdmark:
A git configuration file to be used with:

git blame --ignore-revs-file .git-blame-ignore-revs

or

git config blame.ignoreRevsFile .git-blame-ignore-revs

This makes it easier to navigate large stylistic changests with git blame

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,0 +1,19 @@
+# remove trailing whitespace
+06e427ea02bbcea13edd9e5e0235a7d122b171b7
+1ec59c617be7b892760aa517d8cd7a34bdebea68
+c848b5f89e303a2c5c9421c456aac7edd589696e
+# reformat code in nbp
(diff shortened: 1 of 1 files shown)


rdmark:
@andylemin do you have an opinion on this? I have tested that it works as intended with `git blame` as well as the GitHub web UI for most files. A good test case is etc/spotlight/sparql_parser.y which got a big reformatting but not many other revisions recently, so now you can see the actual logic changes more clearly with git blame.

Where it *doesn't* work is for commits that restored files that were previously deleted – git blame still shows those commits even when I try to flag them for filtering. Case in point, all of the AppleTalk headers and modules in libatalk.

rdmark:
An added boon with this file is that VSCode's *Git Blame Status Bar Item* works more consistently when present.

I will merge now for my own selfish benefit. :)

PR #2563, opened 2025-12-23
nad: protect against potential memory leak in xgetcwd()

rdmark:
this adds handling of errors in malloc and realloc, which could potentially lead to NULL pointer being returned

### DIFF
--- a/bin/nad/ftw.c
+++ b/bin/nad/ftw.c
@@ -249,11 +249,26 @@ static char *xgetcwd(void)
     path_max = (unsigned) PATH_MAX;
     path_max += 2;        /* The getcwd docs say to do this. */
     cwd = malloc(path_max);
+
+    if (cwd == NULL) {
+        return NULL;
+    }
+
     errno = 0;
 
     while ((ret = getcwd(cwd, path_max)) == NULL && errno == ERANGE) {
+        char *new_cwd;
         path_max += 512;
-        cwd = realloc(cwd, path_max);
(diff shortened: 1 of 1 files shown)


bin/nad/ftw.c
```diff
@@ -249,11 +249,26 @@ static char *xgetcwd(void)
...
+
+    if (cwd == NULL) {
+        return NULL;
+    }
+
     errno = 0;
 
     while ((ret = getcwd(cwd, path_max)) == NULL && errno == ERANGE) {
+        char *new_cwd;
         path_max += 512;
```
andylemin, line 261: ExFAT and HFS+ on macOS support file names up to 255 UTF-16 characters, which means 512 bits, so agree this is a good limit. 
I assume this is consistent throughout the entire codebase?

Interestingly ZFS now allows upto 1023 bits (https://github.com/openzfs/zfs/issues/13043) for 511 characters which is becoming the more modern limit, so I am tempted to trial increasing this limit in Netatalk to see if it works with new MacOS versions.

There has also been talk in samba about increasing this, not sure of the current state of that however.

rdmark: The commonly used limit we have right now is defined in atalk/globals.h

```c
#define UTF8FILELEN_EARLY 255           /*!< Type3, early Mac OS X 10.0-10.4.? */
```

Elsewhere we use PATH_MAX from limits.h so yeah this area could use an overhaul.

PR #2570, opened 2025-12-25
docs: adapt libatalk DSI readme to the developer docs

rdmark:
touch up with markdown syntax and capitalization, update outdated implementation notes

touch up code comments in dsi.h concerning the functions that have been moved

### DIFF
--- a/include/atalk/dsi.h
+++ b/include/atalk/dsi.h
@@ -92,9 +92,7 @@ typedef struct DSI {
     int zeroconf_registered;
 #endif
 
-    /* protocol specific open/close, send/receive
-     * send/receive fill in the header and use dsi->commands.
-     * write/read just write/read data */
+    /* protocol specific open/close */
     pid_t (*proto_open)(struct DSI *);
     void (*proto_close)(struct DSI *);
 } DSI;
--- a/test/testsuite/dsi.h
+++ b/test/testsuite/dsi.h
@@ -19,7 +19,7 @@
  * @brief DSI (Data Stream Interface) protocol definitions
  * modified for the afptest test-suite
  *
- * The interface for the test-suite differts from the libatalk DSI interface
+ * The interface for the test-suite differs from the libatalk DSI interface
  * in that is retains a statically allocated buffer for commands
  * as well as smaller command and data buffer sizes that fit in 16 bit size fields.
  *
@@ -71,9 +71,7 @@ typedef struct DSI {
     const char *program;
     int socket, serversock;
 
-    /* protocol specific open/close, send/receive
-     * send/receive fill in the header and use dsi->commands.
-     * write/read just write/read data */
+    /* protocol specific open/close */
     pid_t (*proto_open)(struct DSI *);
     void (*proto_close)(struct DSI *);
 } DSI;


rdmark:
@andylemin do you have an idea what this *proto_send* thing they're talking about here is? I couldn't find traces of such a function neither in the historical netatalk codebase nor online.

doc/developer/dsi.md
```diff
@@ -0,0 +1,52 @@
...
+
+Performance Tweaking
+--------------------
+
+Sending complete packets or the header and a partial packet to the
+client should always be handled by proto_send. For dsi_tcp.c,
+proto_send will coalesce the header and data by using writev.
+In addition, AppleShare sessions often involve
+the sending and receiving of many small packets. As a consequence, I
+use TCP_NODELAY to speed up the turnaround time.
```
andylemin: Disabling Nagle's algorithm is generally good, as it is especially harmful on links above 1Gbps.
And it can be safely enabled for everything on the socket.

However with modern TCP congestion control algorithms (such as Cubic) this can have an unintended impact resulting in worse performance.

For example, Nagle coalesces many small packets into 1 to reduce congestion (but also increases buffer delay).
Disabling Nagle results in lower latency but more small packets, which means more chance of congestion.

Cubic uses time since last congestion for the congestion window. So paradoxically disabling Nagle with some TCP CC algorithms can have an opposite effect of reducing performance and increasing latency (window collapse) on networks which detect congestion often (Wi-Fi interference is detected as congestion).

1) Therefore as Cubic is now the default, we really should make the disabling of Nagle an AFPD config option (only set once at socket setup, so no lookup overhead impact).

Currently to overcome some issues of Cubic without Nagle over WiFi (large buffer WiFi such as wifi7), you have to switch the TCP CC back to Newreno on the Netatalk server.

2) Disabling Nagle is designed to reduce latency through packet coalescing which is great for signals and interactivity, however it can harm peak throughput during file transfers when you really do want to buffer and coalesce to send full MTU packets.

But as we have the DSI between AFP and the Network, I _think_ we are ok (the DSI should coalesce and send full datagrams instead, so Nagle or not, for transfers the packets should always be full during transfers - I should investigate this).

3) As these Signals are blocking and atomic, we should consider setting the TCP PUSH bit manually in the header when sending Signals - to stop the client/receiver buffering them.

andylemin:
> @andylemin do you have an idea what this *proto_send* thing they're talking about here is? I couldn't find traces of such a function neither in the historical netatalk codebase nor online.

This would have been part of the DSI layer. If proto_send is missing it was probably removed post implementation. DSI is adding a whole additional buffering and streaming layer on top of TCP CC - this is non-trivial, and Netatalk's DSI is actually quite advanced!

But the algorithm is also very complex, so I would not be surprised if the implementation was refactored after the initial design once tested with TCP's nuances.

doc/developer/dsi.md
rdmark: Pardon if I misunderstood your intention, but are you saying we should add context like this to the DSI readme? :)

rdmark:
You were right, I figured out what happened. The implementation change happened in 1998, before revision control was introduced. Specifically between netatalk-1.4b2+asun2.1.0 and netatalk-1.4b2+asun2.1.1

The changelog spells it out:

```
1998-11-17  a sun  <asun@purgatorius.zoology.washington.edu>
[...]
	* libatalk/dsi/dsi_stream.c: moved all of the read/write functions
	into here as they're pretty generic. now, the protocol specific
	stuff only handles open and close. 
```

I pushed a new commit that rewrites the relevant sections. Please take a look!

doc/developer/dsi.md
andylemin: Haha, 🙃 I was just thinking out loud as there is a decent case now to provide a config option to disable the addition of TCP_NODELAY.
The other comments are my musings about future improvements and testing.

Long story short, you cant have both low latency/responsive, and high throughput under the same TCP tuning.. (one requires no buffering, no latency and small packets, and the other requires steaming full size packets with pacing).
But with the DSI implementation we actually do have the best of both to some extent - but it needs modernising as TCP algos are quite different to 20 years ago.

Happy to have comments added to the DSI docs. The Newreno trick is confirmed (gives ~20% speed boost for file transfers over WiFi with Nagle disabled compared to Cubic with Nagle disabled).
Not sure how useful they will be to people until we have a tuneable?

rdmark: Curious, on this topic we actually have a USE_TCP_NODELAY preprocessor flag but it seems orphaned and only applied to the cnid_dbd module. Could definitely be a nice future project to expose something like this as a dynamic configuration in afp.conf instead!

PR #2576, opened 2025-12-27
GitHub CI: remove NetBSD packages with dbus dependencies

rdmark:
The dependency tree for dbus in NetBSD 10 is broken currently, which means that gnome-tracker and libcups cannot be installed

I've been told that dbus support in NetBSD is an afterthought; the downstream pkgsrc packaging doesn't these use dependencies either

https://cvsweb.netbsd.org/bsdweb.cgi/pkgsrc/net/netatalk4/

For simplicity's sake, let's stop building with Spotlight and CUPS support altogether on this platform


rdmark:
this is what should resolve the job failures in all of the other PRs

PR #2580, opened 2025-12-28
config: overhaul daemon configuration file comments

rdmark:
Tweak the config file comment headers of all daemons in the Netatalk suite to clarify context and basic usage


rdmark:
@andylemin FYI this came out of some feedback from a new user in a Discord chat. It wasn't clear to them the difference between afp.conf and atalkd.conf, and that the latter can be left empty for basic deployments. Hopefully these changes should give some more context to new users.

PR #2592, opened 2026-01-02
testsuite: cleanup of DSI struct pointers in test functions

rdmark:
Discard unused variables, use const keyword on immutable variables, assign on the same line as declaration, rearrange variable declaration, remove dead FIXME


rdmark:
This should get us back to a net decrease of static analysis bugs!

Somehow the recent testsuite overhaul triggered a large bunch of these "pointer to const" issues.

rdmark:
Merging trivial code changes without peer review for a quick turnaround on validating the test results

PR #2594, opened 2026-01-02
libatalk: regenerate utf16 case lookup tables with Unicode v17.0.0

rdmark:
Bump the supported Unicode character set from v16.0.0 to v17.0.0


andylemin (commented):
The issues we had last time with Unicode were related to changing the build mechanism, so this should be ok.
But there is still a risk of new user issues.
However I don't believe any new Unicode issues would be conflated with new dircache issues, so I'm not against adding this in the same release

rdmark:
Thanks for sharing your thoughts! After sleeping on it I suggest we postpone this after all, no need to introduce risk for a very marginal benefit. I expect it will take a good long while for most OSes to introduce support for the new glyphs in Unicode 17, so we have time to get this code into production.

rdmark:
@andylemin just checking that you're ok with merging this to main now?

please approve when ready :)

andylemin:
@rdmark approved and merged.

We have had no big issues reported since the dircache changes (other than endian issues with 32bit builds).

PR #2595, opened 2026-01-02
testsuite: flag immutable pointers as const in test functions

rdmark:
A handful of stragglers flagged by SonarQube after the last commit


rdmark:
Merging trivial code change without peer review to validate the SonarQube scan results

PR #2608, opened 2026-01-05
docs: revise CNID configuration details in afp.conf man page

rdmark:
Updated CNID backend options and configuration details in afp.conf man page for clarity

### DIFF
--- a/doc/manpages/man5/afp.conf.5.md
+++ b/doc/manpages/man5/afp.conf.5.md
@@ -403,12 +403,6 @@ afp read locks = *BOOLEAN* (default: *no*) **(G)**
 spec mandates this, but it's not really in line with UNIX semantics and
 is a performance hog.
 
-cnid listen = *ip address[:port] [ip address[:port] ...]* **(G)**
-
-> Specifies the IP address and port that the CNID server should listen on.
-This should match the address and port of the **cnid server** option
-for most deployments. The default is **localhost:4700**.
-
 disconnect time = *number* **(G)**
 
 > Keep disconnected AFP sessions for *number* hours before dropping them.
@@ -488,11 +482,17 @@ registration if Avahi or mDNSResponder were compiled in.
 
 ## CNID Database Backend Options
 
-Global configurations specific to the CNID database backend, which
+Global configurations specific to the CNID database backends, which
 controls how the database is stored and accessed.
 
-For granular control over the CNID database backend, see the Volume Parameters
-section below.
+The actual CNID backend scheme is configured per volume.
+See the **cnid scheme** option in the Volume Parameters section below.
+
+cnid listen = *ip address[:port] [ip address[:port] ...]* **(G)**
+
+> Specifies the IP address and port that the CNID server should listen on.
+This should match the address and port of the **cnid server** option
+for most deployments. The default is **localhost:4700**.
 
 cnid mysql host = *MySQL server address* **(G)**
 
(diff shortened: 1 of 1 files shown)


doc/manpages/man5/afp.conf.5.md
```diff
@@ -1120,25 +1120,23 @@ hosts deny = *IP host address/IP netmask bits* [ ... ] **(V)**
...
+Not all backends may be available with your netatalk installation.
+Run **afpd -v** to see a list of available backends, as well as which one is the default.
 >
-> *dbd* uses Berkeley DB, with database reads and writes managed through the **cnid_dbd** daemon.
+> *dbd*: uses Berkeley DB, with database reads and writes managed through the **cnid_dbd** daemon.
 It is recommended for most deployments.
 >
-> The *sqlite* backend uses the SQLite embedded database library.
-It is performant and lean, requiring no external database or daemon.
+> *last*: uses a read-only, in-memory Trivial Database.
```
andylemin: It's not clear how a read-only database can work?
Does it provide a NOOP backend to force all to disk? (Ie, What makes its special for testing?)

andylemin (approved):

doc/manpages/man5/afp.conf.5.md
rdmark: Very good questions! There's a bit of a history to this. The database itself isn't read-only per se, but there's logic in place that sets the read-only flag for the volume when using this backend. So we will write CNIDs to the database once, but never modify them, basically.

It's used for the "afpd" tests that you may have seen:

https://github.com/Netatalk/netatalk/tree/main/test/afpd

Using the last backend allows this tiny test suite to build with minimal dependencies and run very quickly in a contained environment; pseudo unit tests, you might say.

Now, if you look closer at the code you may see that when you build netatalk with `-Dwith-tests` the `last` backend actually becomes read/write. This was a hack on top of a hack that I did to be able to add more tests to the "afpd" suite. All this before I revived the testsuite BTW, so back then we had extremely few tests and I was desperate. ;)

This is where the nasty business is:

https://github.com/Netatalk/netatalk/blob/d6a7830b7eae435973123e0dec2cc9f4a3aa67a2/libatalk/util/netatalk_conf.c#L1327

I think the reason the previous maintainers made `last` read-only in the first place is because it's horribly broken for anything beyond very basic operations. Here's the commit where they made it read-only. Very terse commit message though:

https://github.com/Netatalk/netatalk/commit/caaaa67dd9ce1e564918a719393837377149a797

Some time ago I toyed with the idea of refactoring the "afpd" tests to use the sqlite backend instead, and get rid of `last` once and for all. But I ran into some roadblocks that I don't recall what it was. Something with linking perhaps.

Anyways, pardon the rant. Happy to rephrase the man page verbiage to make more sense!

andylemin: Thanks for the perfect explanation.

Definitely too much detail for the manpage, but a little more detail would be nice.

Maybe just "write-once-read-many / write-once-read-only, used exclusively for unit tests" would clarify.

One could even argue should it be in the man page at all?

rdmark: How about this verbiage, keeping it vague rather than trying to explain the source of inconsistency:

> It is not reliable enough for everyday use. We keep it around strictly for testing purposes.

I still think we should document it a little bit, since it appears here and there, like when you run afpd -v:

```
afpd 4.4.0 - Apple Filing Protocol (AFP) daemon of Netatalk

This program is free software; you can redistribute it and/or modify it under
the terms of the GNU General Public License as published by the Free Software
Foundation; either version 2 of the License, or (at your option) any later
version. Please see the file COPYING for further information and details.

afpd has been compiled with support for these features:

          AFP versions:	2.2 3.0 3.1 3.2 3.3 3.4 
      Transport layers:	TCP/IP
         CNID backends:	dbd last mysql sqlite 
  Default CNID backend:	dbd

              afp.conf:	/usr/local/etc/afp.conf
           extmap.conf:	/usr/local/etc/extmap.conf
       state directory:	/var/local/netatalk/
    afp_signature.conf:	/var/local/netatalk/afp_signature.conf
      afp_voluuid.conf:	/var/local/netatalk/afp_voluuid.conf
       UAM search path:	/usr/local/lib/netatalk/
  Server messages path:	/var/local/netatalk/msg/

```

andylemin: Yes makes sense

andylemin (approved):

PR #2610, opened 2026-01-06
tests: refactor the afpd tests to use the sqlite CNID backend

rdmark:
using the sqlite backend instead of the last backend in the afpd tests allows us to continuously test actual production code rather than the half-finished last backend

unlike dbd or mysql, the sqlite backend can function without root privileges which makes it ideal for automated tests

a handful of minor bugfixes and improvements to the tests:

- fixed a bug in the test helper where the network byte order was flipped
- ignore SIGPIPE to keep the test runner alive when writing to IPC pipe
- proper cleanup of data structures when exiting the tests (Valgrind flagged)
- create unique shared volume and CNID dirs every time
- set server signature, vol UUID, and vol dbpath to make test setup atomic
- capture afpd logs at debug level in the test log

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -478,6 +478,8 @@ jobs:
         uses: vmactions/dragonflybsd-vm@84277fec84c57e6f1b4eaa6a6ef3d36618d0c05b # v1.2.1
         with:
           copyback: true
+          sync-time: true
+          sync: nfs
           usesh: true
           prepare: |
             set -e
@@ -508,6 +510,7 @@ jobs:
               -Dwith-tests=true \
               -Dwith-testsuite=true
             meson compile -C build
+            meson test -C build
             meson install -C build
             netatalk -V
             afpd -V
@@ -680,6 +683,7 @@ jobs:
               -Dwith-tests=true \
               -Dwith-testsuite=true
             meson compile -C build
+            meson test -C build
             meson install -C build
             netatalk -V
             afpd -V


rdmark:
this PR got quite sprawling so I have started breaking out atomic changes into separate PRs

rdmark:
@andylemin if you're curious about why this is taking weeks, I have run into a real headscratcher of a sqlite CNID bug. The afpd tests create a "dir1" dir (located in volume root), that when it hits the dircache dir1 has did=1 which triggers an assert there. This happens *only* on FreeBSD, on other OSes did is 17 at that stage. I've tried to refactor/strengthen the DB initialization to eliminate potential race conditions but no luck so far.

"Luckily" the bug reproduces in my local FreeBSD VM as well, so that makes things easier. As the resident FreeBSD expert, do you have any ideas what to look closer at?

test step that fails in test/afpd/test.c

```
    TEST_expr(reti = createdir(&obj, vid, DIRDID_ROOT, "dir1"),
              reti == 0 || reti == AFPERR_EXIST);
```

afpd log snippet from FreeBSD

```
Jan 17 23:18:30.025070 [13740] {cnid_sqlite.c:526} (debug:CNID): cnid_sqlite_lookup: name: 'dir1', did: 2 is not in the CNID database
Jan 17 23:18:30.025131 [13740] {cnid_sqlite.c:725} (debug:CNID): cnid_sqlite_add: lookup returned id=0, errno=1
Jan 17 23:18:30.025201 [13740] {cnid_sqlite.c:741} (debug:CNID): cnid_sqlite_add: binding name='dir1', did=2, dev=7036080426116545447, ino=51
Jan 17 23:18:30.025338 [13740] {cnid_sqlite.c:763} (debug:CNID): cnid_sqlite_add: not using CNID hint, CNID set is depleted or hint not set
Jan 17 23:18:30.025503 [13740] {dircache.c:611} (error:AFPDaemon): dircache_add(): did:1 is less than the allowed 17. Try rebuilding the CNID database for: "dir1"
```

snippet from a successful Alpine test

```
Jan 17 23:16:17.806674 [2430] {cnid_sqlite.c:524} (debug:CNID): cnid_sqlite_lookup: name: 'dir1', did: 2 is not in the CNID database
Jan 17 23:16:17.806679 [2430] {cnid_sqlite.c:724} (debug:CNID): cnid_sqlite_add: lookup returned id=0, errno=1
Jan 17 23:16:17.806684 [2430] {cnid_sqlite.c:738} (debug:CNID): cnid_sqlite_add: binding name='dir1', did=2, dev=48, ino=3201649
Jan 17 23:16:17.806691 [2430] {cnid_sqlite.c:762} (debug:CNID): cnid_sqlite_add: not using CNID hint, CNID set is depleted or hint not set
Jan 17 23:16:17.806741 [2430] {dircache.c:673} (debug:AFPDaemon): dircache(did:17,'dir1'): {added}
```

andylemin:
Hmmm, it feels like something is missing here.

Thinking out loud;

CNID records accessed by id do not store the whole path, only the parent did record id and the target object name in the parent folder which the record represents.

To build the whole path, a chain of CNID records is required.

Once the parent did in a CNID record equals 2, we have reached the lowest possible CNID record (an object _in_ the vol root), and the whole path has been built. Having a CNID record whose id is 0, 1, or 2 is not possible and should never exist.

Once the whole path has been built we can call dircache_add.

The error `dircache_add(): did:1 is less than the allowed 17` implies we are calling dircache_add with a corrupt did.

I think you need to add two log lines;
1) A temporary log line to get back and print the contents of the CNID record after `binding name='dir1', did=2, dev=7036080426116545447, ino=51` 
2) A temp log line printing `did` and `curdir` just before calling dircache_add


As for thoughts about why FreeBSD?

FreeBSD has a more strict POSIX implementation which sometimes causes issues compared to Linux.

Other than that nothing significant should be different from the other distros.


Let me know how you get on with the extra logs.

andylemin:
This looks strange to me `createdir(&obj, vid, DIRDID_ROOT, "dir1")`

DIRDID_ROOT is 1 and does not exist. It is the ID for an abstract base struct for a CNID record.

Maybe the test itself is invalid

rdmark:
DIRDID_ROOT is defined as 2 actually, in directory.h:

```c
#define DIRDID_ROOT_PARENT    htonl(1)  /*!< parent directory of root */
#define DIRDID_ROOT           htonl(2)  /*!< root directory */
```

I will play around with logging today!

rdmark:
the data flow I'm seeing is:

- the test code (afpfunc_helpers.c) has a createdir() helper that calls afp_createdir() in directory.c
- afp_createdir() calls dir_add() in the same file which in turn calls dircache_add()
- when checking right before we call dircache_add() in dir_add(), we do have the bad did=1 at that point

the added logging just before calling dircache_add says

```
Jan 19 01:02:18.913339 [3138] {directory.c:1067} (debug:AFPDaemon): dir_add: before dircache_add did:1 for /tmp/AFPtestvolume-rVnUGr/dir1
```

the sqlite database as dumped from a failed run on FreeBSD has the did=1 as well

```
INSERT INTO E2A05364F48311F0BE4B0800274119A5 VALUES(1,'dir1',2,-2619697646098754333,58);
```

a good database row (from macOS here) looks like

```
INSERT INTO "269AD340515847A087EFD0404CBF5C1D" VALUES(17,'dir1',2,16777239,110283930);
```

next up, to trace where the bad row is inserted, and how come our CNID safeguards don't catch it

andylemin:
Nice. Ok so the parent did can be 2, so it's not that.

Good progress! Pleased we can see the corruption is happening earlier. 

I'm curious about why this one is only affecting FreeBSD.

Yes dir_add should have valid did checks. did >= 2 ?

andylemin:
Did these last fixes resolve it? 🙂

rdmark:
I have a lead that the afpd test helpers need to be overhauled with proper network byte order of the data that is sent over the wire. I got stuck battling some cascading errors that came out of those changes, and moved on to other tasks for the time being.

rdmark:
@andylemin the real bug turned out to be something very silly indeed: the use of double quotes instead of single quotes for database table names: in SQLite, double quotes are for identifiers and single quotes for string literals.

https://www.sqlite.org/lang_keywords.html

however, SQLite has a quirks mode where it falls back to interpreting a double quoted string as a string literal when it fails to read it as an identifier, as described in "8. Double-quoted String Literals Are Accepted" at

https://sqlite.org/quirks.html

my hypothesis that the version of the SQLite library that is shipped with FreeBSD is running in some sort of strict mode, but I haven't found evidence to prove it

I added additional debug logging and validation for the CNID table initialization logic, which gave us this very revealing output

```
  Feb 01 03:15:01.755656 [2388] {cnid_sqlite.c:273} (info:CNID): sqlite     
  query "UPDATE sqlite_sequence SET seq = 16 WHERE name =                   
  "C24E5F90FED011F0A83C0800274119A5";", error: no such column:              
  "C24E5F90FED011F0A83C0800274119A5" - should this be a string literal in   
  single-quotes?  
```

the correct fix is obviously to use the SQLite standards compliant single quotes when looking up database tables

I'll create a new PR with this and a few other minor fixes for the sqlite backend!

andylemin:
Nice! Very quirky 😉

FreeBSD is known for being strict, most notably its POSIX implementation - it's why I run critical production workloads on FreeBSD.

rdmark:
just to be clear, this depends on https://github.com/Netatalk/netatalk/pull/2672 before the jobs can pass in this PR!

rdmark:
@andylemin rebased, all green, and ready for your review!

andylemin (approved):

PR #2611, opened 2026-01-06
GitHub CI: save meson logs after each build job

rdmark:
The meson build and test logs are invaluable for troubleshooting failures, so we need to store them as build artifacts

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -104,6 +104,12 @@ jobs:
           timelord -V
       - name: Uninstall
         run: ninja -C build uninstall
+      - name: Upload meson logs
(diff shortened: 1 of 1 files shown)


rdmark:
keeping as draft because there's a bug: an error anywhere in the vmactions step leads to the copyback not taking effect and us not getting any logs to upload in the subsequent step

it works well with the regular container based jobs though!

rdmark:
~~This is close to working now; only DragonflyBSD and OmniOS are having trouble syncing artifacts to the next step in the job.~~ See https://github.com/vmactions/freebsd-vm/issues/130

Got them working by not specifying a particular sync method

PR #2619, opened 2026-01-07
libatalk: add volume option 'volume uuid' to define static UUID

rdmark:
this bypasses the reading and writing of UUID to/from the afp_voluuid.conf file

when all volumes define a UUID with this option, we never touch afp_voluuid.conf

this removes the requirement to run afpd as root for basic use cases (such as the afpd integration tests)


rdmark:
Keeping this as a draft for now, since the main branch has been unfrozen yet after the release of 4.4.0

PR #2625, opened 2026-01-08
cnid: refactor sqlite lookup to a safer two phase logic

rdmark:
Replaced the ambiguous single OR-clause SQL query in cnid_sqlite_lookup() with two separate queries for (DevNo, InodeNo) and (Did, Name) lookups, matching the dbd daemon's two-index architecture.

This enables clear decision logic with five distinct cases mirroring dbd_lookup.c and eliminates complex duplicate row handling.

Fixed database corruption bug where did=0 could be written during hint-based updates by adding explicit DID validation before calling cnid_sqlite_update(), preventing invalid AFP directory IDs from corrupting the database.

Breakdown of the decision cases:

Case | Condition | Action
-- | -- | --
1 | Neither found | Return NOT FOUND
2 | Both found, IDs match | Return ID ✅
3 | Both found, IDs differ | Delete both, return NOT FOUND
4 | devino only | Check hint → Update if match, else delete
5 | didname only | Delete stale entry, return NOT FOUND

### DIFF
--- a/include/atalk/cnid_sqlite_private.h
+++ b/include/atalk/cnid_sqlite_private.h
@@ -12,7 +12,8 @@ typedef struct CNID_sqlite_private {
     sqlite3 *cnid_sqlite_con;
     char *cnid_sqlite_voluuid_str;
     cnid_t cnid_sqlite_hint;
-    sqlite3_stmt *cnid_lookup_stmt;
+    sqlite3_stmt *cnid_lookup_devino_stmt;
+    sqlite3_stmt *cnid_lookup_didname_stmt;
     sqlite3_stmt *cnid_add_stmt;
     sqlite3_stmt *cnid_put_stmt;
     sqlite3_stmt *cnid_get_stmt;
--- a/libatalk/cnid/sqlite/cnid_sqlite.c
+++ b/libatalk/cnid/sqlite/cnid_sqlite.c
@@ -44,23 +44,49 @@
 #include <atalk/util.h>
 #include <atalk/volume.h>
 
-static int init_prepared_stmt_lookup(CNID_sqlite_private * db)
+static int init_prepared_stmt_lookup_devino(CNID_sqlite_private *db)
 {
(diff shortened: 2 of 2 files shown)


andylemin:
Nice, I wondered about the OR statement as well. This looks nicer

andylemin:
I'm not sure about 3, 4 and 5.

None of the other backends do this? - they rely on the main business logic to clean CNID records.

Is there something special about the SQLite backend that makes corruption more likely, making you want to have these cleanups in a low level interface?

Only protecting against CNID_INVALID is incomplete, we need to check DID >= CNID_START (16).

rdmark:
@andylemin you know, I also thought the CNID backends were light on business logic at first, but then I dug into the cnid_dbd code; see for instance dbd_lookup() which clearly has the logic that both the mysql and sqlite CNID backends are emulating.

https://github.com/Netatalk/netatalk/blob/main/etc/cnid_dbd/dbd_lookup.c

the doxygen code comment even spells it out

```
/*!
 * @brief This returns the CNID corresponding to a particular file.
 * @note It will also fix up the database if there's a problem.
 */
```

not saying that this is necessarily a good design, but it seems the CNID backends are expected to do database housekeeping right now.

did you see logic on the afpd side that is doing similar database repair?

andylemin:
Interesting. I'm not at computer so can't easily check right now.

Ok, that means the DBD backend must be doing >= CNID_START then and returning false based on previous testing.

If they are all the same then I'm happy 🙂

rdmark:
@andylemin can you please remind me: did you refactor away the testsuite logic that triggered the did==0 scenario that led to the test failures that we observed? if so, let's bring it back so that we can get test coverage for this corner case. :)

rdmark:
discarding for now; rather than making the CNID backends more complex, we want to see if the business logic can be moved out of the backends in https://github.com/Netatalk/netatalk/issues/2640

PR #2628, opened 2026-01-09
GitHub CI: use the bare minimum of dependencies in OpenIndiana

rdmark:


rdmark:
unfortunately, the bare minimum adds up to about 239MB, which is a small reduction from the 267MB that include the optional dependencies... it still takes a small eternity to install from pkg.openindiana.org and times out most of the time

PR #2632, opened 2026-01-09
cnid: disallow looking up did<2 in sqlite backend

rdmark:

### DIFF
--- a/libatalk/cnid/sqlite/cnid_sqlite.c
+++ b/libatalk/cnid/sqlite/cnid_sqlite.c
@@ -498,6 +498,13 @@ cnid_t cnid_sqlite_lookup(struct _cnid_db *cdb, const struct stat *st,
         EC_FAIL;
     }
 
+    if (ntohl(did) < 2) {
+        LOG(log_warning, logtype_cnid,
+            "cnid_sqlite_lookup: not looking up illegal did: %" PRIu32, ntohl(did));
+        errno = CNID_INVALID;
+        EC_FAIL;
+    }
+
     strlcpy(stmt_param_name, name, sizeof(stmt_param_name));
     stmt_param_name_len = len;
     stmt_param_did = ntohl(did);


rdmark:
@andylemin what do you think about this as a minimal workaround for the stable branch? I would like to keep the regression scope as little as possible.

andylemin:
CNID_START is the first lowest numbered usable DID number

rdmark:
did = 1 an abstraction for DIRDID_ROOT_PARENT and did = 2 for DIRDID_ROOT

shouldn't we allow repairs for did = 2, since that indicates a file (referenced by `retid`) located in the volume root? or maybe I'm misunderstanding the logic?

NJRoadfan:
Files themselves shouldn't have a DID = 1 or 2. If I recall, Netatalk should have the logic to handle this with volumes at the root level (files in the root should have a "Parent ID" of 2). Inside AppleTalk is a bit light on the details of how CNIDs work. The place to look is the HFS and HFS+ file system documentation, since that is where the whole concept is inherited from. CNIDs are basically the HFS equivalent to inode numbers and the first 16 are reserved for various system functions. 

https://developer.apple.com/library/archive/technotes/tn/tn1150.html#CatalogFile

This is also why Apple's server doesn't work on APFS volumes, since that file system dropped CNIDs for 64-bit inodes only. They deprecated the server at that point, so why bother adapting it. AFP itself is basically a network version of HFS.

rdmark:
@andylemin @NJRoadfan pushed the updated commit

NJRoadfan:
Re-reading the code, it looks like the PR may not attempt to repair files with a Parent CNID (referred to as "did" in the database schema) of "2" which is valid for files/folders in the root of the volume. Other than that, it should suffice.

On a related note, developer docs may want to cover the database schema used by the CNID backends.

andylemin:
Yes I agree with @NJRoadfan, its part of the AFP spec

There should never be a CNID record with DID=0, 1 or 2.
They may have a parent did = 2.

All the CNID backends should not accept any DID < CNID_START.

andylemin:
@rdmark @NJRoadfan if the `did` _is_ the parent did (as received by the backend), then did < CNID_START may be too aggressive after all?

Maybe it should be did < 2?

In the main code, did, usually means the target, and pdid the parent did. But they are used inconsistently. 

CNID_START is the lowest number DID for a target record. pdid = 2 is valid?

NJRoadfan:
@andylemin You appear to be correct. pdid = 2 is correct for files and folders in the root of a volume. I do agree the variables referencing IDs in Netatalk's code base can be pretty confusing as "did" implies "directory id" and not "file id". In practice it can mean both because the code doesn't differentiate between files and folders in this context and the value is going to be unique.

I did some reading on this. Technically there is no prohibition in the AFP specification against using IDs numbered 3 thru 16 for files or directories on a share. IDs < 17 is a HFS restriction and the AFP spec specifically defines only directory IDs 0 thru 2 as reserved and having a special purpose (0 = invalid, 1 = dummy parent ID for volume route, 2 = volume route).

The term "Catalog Node ID" itself comes not from the AFP spec, but the HFS spec! You'll never see a directory or file ID between 3 and 16 on a AFP server sharing files from a HFS volume, but I bet you will from Microsoft's Services for Macintosh or PCMacLAN.

The AFP 2.0 specification defined by "Inside AppleTalk" doesn't give too many details on the importance of file IDs, other than its a field and should be unique. Heck, they likely can overlap with directory IDs and even include ID 1 or 2! AFP 2.1 added commands to create and delete file IDs but in practice they were never used and later deprecated in AFP 3.0.

Overall the programmers of Netatalk's CNID system took the conservative approach and followed the HFS rules for ID assignment. No IDs under 17 and all objects (files and folders) receive a unique ID with no overlap.

rdmark:
In this context `did` (e.g. `lookup_result_did`) definitely means the CNID of the *directory* that houses the file that is being queried, as opposed to `id` (e.g. `lookup_result_id`).

example from the wild, one of my test volumes for a file that resides in the volume root, where the sqlite database row that says

```
INSERT INTO B7D57A12823F935CED6D6A5BB4272BC3 VALUES(54,'eng5_writing_Good_Life copy 2.pdf',2,65026,12464);
```

results in this log message generated in cnid_sqlite_lookup()

```
Jan 12 22:08:22.054188 afpd[1086250] {cnid_sqlite.c:538} (maxdebug:CNID): cnid_sqlite_lookup: id=54, did=2, name=eng5_writing_Good_Life copy 2.pdf, dev=65026, ino=12464
```

NJRoadfan:
That clears things up. I would be curious what would pass did = 0 to this function to begin with? If it is files that were added to the volume externally (ex: copied into the volume by the host system or via Samba), but not in the CNID database, they should be added to the database. If there is a database entry returned with did = 0 for a file, that entry should be deleted/repaired.

I guess what I'm getting at, if this function is called with did = 0 (or even 1), why even bother querying the database since the initial request is technically invalid to begin with?

rdmark:
This particular corner case - did=0 when querying in cnid_sqlite_lookup() - was observed by @andylemin when running test500 in the AFP spec test. Here's an example pipeline job where it triggered the bug:

https://github.com/Netatalk/netatalk/actions/runs/20556288957/job/59040968289

copy pasting in case the job gets purged in the future

```
############## entering test500 ##############
[FPOpenVolFull] Open Vol test1 bitmap 21
[FPCreateDir] Create Directory Vol 256 did : 0x2 <t500 dir>
directory ID 0x11
[FPCreateDir] Create Directory Vol 256 did : 0x11 <t500 subdir1>
directory ID 0x12
[FPCreateDir] Create Directory Vol 256 did : 0x11 <t500 subdir2>
directory ID 0x13
[FPMoveAndRename] Move and rename Vol: 256 did: 0x11 <t500 subdir1> ==> 0x13 <t500 renamedsubdir1>
dsi_stream_read(0): EOF
header.dsi_code       -1	
[../test/testsuite/T2_Dircache_attack.c:165] FPMoveAndRename(Conn2, vol2, dir_id, subdir2_id, subdir1, renamedsubdir1)
[FPEnumerate] Enumerate Vol 256 did : 0x13 <>
header.dsi_code       -5018	AFPERR_NOOBJ   
[../test/testsuite/T2_Dircache_attack.c:167] FPEnumerate(Conn, vol1, subdir2_id, "", (1 << FILPBIT_FNUM), (1 << DIRPBIT_PDID))
[FPGetFileDirParams] GetFileDirParams Vol 256 did : 0x13 <t500 renamedsubdir1>
header.dsi_code       -5018	AFPERR_NOOBJ   
[../test/testsuite/T2_Dircache_attack.c:170] FPGetFileDirParams(Conn, vol1, subdir2_id, renamedsubdir1, 0, bitmap)
	FAILED ffff0000 should be 12000000
	FAILED  should be t500 renamedsubdir1
[FPCloseVol] Close Vol 256
dsi_stream_read(0): EOF
header.dsi_code       -1	
[../test/testsuite/T2_Dircache_attack.c:191] FPCloseVol(Conn2, vol2)
[FPDelete] FPDelete conn 0x7f5eb4642020 Vol 256 did : 0x12 <>
[FPDelete] FPDelete conn 0x7f5eb4642020 Vol 256 did : 0x13 <>
[FPDelete] FPDelete conn 0x7f5eb4642020 Vol 256 did : 0x11 <>
Dircache:test500: move and rename dir, enumerate new parent, stat renamed dir - FAILED
```

What the test does is to create a dir with user 1 and then immediately try to move the dir with user 2 followed by enumeration by user 1

btw it would be "great" if we can get the test back to a failing state to prove the fix is good, but I'm not sure exactly what caused it to stop failing

andylemin:
> That clears things up. I would be curious what would pass did = 0 to this function to begin with? If it is files that were added to the volume externally (ex: copied into the volume by the host system or via Samba), but not in the CNID database, they should be added to the database. If there is a database entry returned with did = 0 for a file, that entry should be deleted/repaired.
> 
> I guess what I'm getting at, if this function is called with did = 0 (or even 1), why even bother querying the database since the initial request is technically invalid to begin with?

Hi @NJRoadfan 
The code uses `curdir` extensively as a global, and any changes to the current working directory (Eg movecwd()) requires `curdir` to be updated to the same. Therefore when did changes you often see `curdir = did`, and the passing of `curdir` directly to CNID backend to get a file from CWD.

However, dir_remove (and some other places) set `curdur = NULL` for safety (safer than a dangling pointer). This previously resulted in `did = 0` occasionally being passed to the CNID backend.

The other backends return an error for 0, which tells the main business logic the DID was invalid and to rebuild it. With SQLite it was not returning the error so the CNID and dircache was not being rebuilt - instead the DB was corrupted.

The original bug has been fixed by adding a missing `curdir = did`, after `did` was updated (in cname() if I remember correctly), which was before `curdir` was passed to CNID lookup. It only triggered when curdir was set to NULL, otherwise `curdir` still pointed to a valid `did`.

This PR is the low level fix to make SQLite behave the same as DBD and MYSQL backends.

NJRoadfan:
The FPMoveAndRename command in test500() is failing, which I'm guessing is the condition that was later fixed. When FPEnumerate was called in this case, it likely receives a NULL value for DID since the directory its expecting isn't there.

@andylemin Thanks for the clarification. There is ALOT of moving parts to trace thru here. Looking at cnid_dbd_lookup(), it does nothing (returns CNID_INVALID) if it doesn't receive back any database entries from the query, which should ALWAYS be the case when did=0 is passed to it.

Is it possible to just set these lookup functions to immediately return CNID_INVALID if did=0 is passed to them, or will that break something elsewhere? Just seems like a potential optimization and a way to avoid a possible DDoS by someone sending malformed requests that needlessly hit the database backend.

andylemin:
> The FPMoveAndRename command in test500() is failing, which I'm guessing is the condition that was later fixed. When FPEnumerate was called in this case, it likely receives a NULL value for DID since the directory its expecting isn't there.
> 
> @andylemin Thanks for the clarification. There is ALOT of moving parts to trace thru here. Looking at cnid_dbd_lookup(), it does nothing (returns CNID_INVALID) if it doesn't receive back any database entries from the query, which should ALWAYS be the case when did=0 is passed to it.
> 
> Is it possible to just set these lookup functions to immediately return CNID_INVALID if did=0 is passed to them, or will that break something elsewhere? Just seems like a potential optimization and a way to avoid a possible DDoS by someone sending malformed requests that needlessly hit the database backend.

Ahh, here is the interesting bit @NJRoadfan. This is what I said to @rdmark too.. If the DBD record does not exist it returns CNID_INVALID. However SQLite sees the record does not exist (as DID = 0), and tries to update the "other" record it found with the OR statement, with did = 0. This seems wrong to me as this feels like a primitive low level function trying to be too smart and making decisions which affect the main business logic.

andylemin:
> FPMoveAndRename

PS; FPMoveAndRename triggered the issue as after renaming a directory the did becomes invalid (as it has been renamed), so curdir gets set to NULL, until it moves to the new path

NJRoadfan:
If that is the case, I'll go with the KISS principle in handling this and check if the request is valid before doing anything. The AFP spec says "did = 0" isn't valid so I would have had the function immediately check the value of did on entry and if its "0" or NULL. If so immediately return CNID_INVALID and don't bother with any database queries.

I know I may be over thinking this, but its also possible that a bad actor could add a database entry with a did of 0 and purposely trigger it with an "invalid" query.... what happens then? A crash? A buffer overflow? Sanitize your inputs.

NJRoadfan:
...and to add to @andylemin's comment above, we shouldn't be attempting database repairs during lookup queries. It leads to unpredictable behavior, and in some cases, behavior that can't be reproduced.

NJRoadfan:
OK, looking at the code, it should protect against did = 0 already, see function `valide()` in `cnid.c`. Oddly this appears to only check the value after the relevant function in run in the CNID backend, but at least this prevents database corruption from causing further issues.

rdmark:
> ...and to add to @andylemin's comment above, we shouldn't be attempting database repairs during lookup queries. It leads to unpredictable behavior, and in some cases, behavior that can't be reproduced.

I agree with this stance in principle. It would be the better architecture. By all means, it is pretty unexpected that database repair is happening in CNID lookup.

If you look at `etc/cnid_dbd/dbd_lookup.c` the previous author describes in the uppermost code comment is some detail the fail state scenarios they're trying to recover from which provides some good insights. And then down in `dbd_lookup()` is where the database recovery logic is implemented, the logic that the MySQL and by extension SQLite backends are emulating.

Shall we draw up a new design for all three backends? :)

andylemin:
I think we all agree having this repair logic in the CNID layer is not desirable for many reasons, but a redesign will require some work to decide where it should be moved up too.

I would say let's strike a compromise for now.
The SQLite repair logic does replicate/duplicate DBD and MySQL backends, but the input check should be at the top of the function.

If DID < 2 just return CNID_INVALiD immediately, don't bother with any lookups. 
Currently the check is too late.

rdmark:
@andylemin like this?

andylemin:
@rdmark Thanks for working through this together.
I think we have the right solution for now and you can resolve the issue we raised.

rdmark:
thank you! it was a good exercise, and an opportunity to look closer at the cnid_dbd code which has been a blackbox to me before.

I filed https://github.com/Netatalk/netatalk/issues/2640 for looking into more fundamental changes to the CNID separation of concerns

PR #2641, opened 2026-01-17
[4.4] cnid: return CNID_INVALID when attempting to look up did<2

rdmark:


rdmark:
this is backporting https://github.com/Netatalk/netatalk/pull/2632 to the stable 4.4 branch (mistakes were made)

rdmark:
it seems like GitHub had some kind of system malfunction here: the history above is pointing to a commit that's not associated with this PR - bd1ffff - and the commit I thought we merged is just gone, nowhere to be found in the branch-netatalk-4-4 commit history... this is very disconcerting.

rdmark:
it was successful this time: https://github.com/Netatalk/netatalk/pull/2653 -- let to commit afc9f26

PR #2645, opened 2026-01-19
refactor: convert adf_lock macros to functions

### LINKED ISSUE #2600: refactor macro in ad_lock.h to function
the adf_lock_free macro in the libatalk/adouble/ad_lock.h header is quite complex and should better be refactored to a proper function (not least to allow static analysis to parse the code better)

andylemin:
Refactors adf_lock_init and adf_lock_free from macros to proper functions to improve static analysis capability and code maintainability.

Adds defensive NULL checks with error logging.

Removes ad_lock.h header file as it is no longer needed after the refactoring - all declarations are now in include/atalk/adouble.h.

Fixes #2600


rdmark (approved):
much cleaner!

one could argue that ad_lock.h has served its purpose and could just as well be refactored away now, but what do you think?

andylemin:
Agreed, I'll add it here.

And yes, the static analyser is working now - it raised some other minor things

andylemin:
@rdmark This is ready for review now

PR #2649, opened 2026-01-19
MySQL CNID Backend: improve charset performance, TCP performance, and test container config

andylemin:
MySQL CNID Backend: improve charset performance, TCP performance, and test container config

- Enable TCP_NODELAY for MySQL/MariaDB TCP connections to reduce query latency
- Reduce connection/read/write timeouts from 600s to 30s
- Set UTF-8 charset (utf8mb4) at connection time for UTF-8 volumes (avoid repeated dynamic detection overhead)
- Add improved logging for connection info and table creation
- Enhanced Docker entrypoint to support MySQL/MariaDB testing via TCP and Socket
- Add mariadb package to Alpine testsuite Dockerfile
- Update afp.conf documentation for MySQL backend options

### DIFF
--- a/distrib/docker/entrypoint_netatalk.sh
+++ b/distrib/docker/entrypoint_netatalk.sh
@@ -184,15 +184,59 @@ else
     AFP_VALIDUSERS2="$AFP_USER $AFP_USER2"
 fi
 
+# Validate MySQL/MariaDB CNID backend configuration
 if [ "$AFP_CNID_BACKEND" = "mysql" ]; then
     if [ -z $AFP_CNID_SQL_HOST ]; then
         AFP_CNID_SQL_HOST="localhost"
     fi
     if [ -z $AFP_CNID_SQL_USER ]; then
-        AFP_CNID_SQL_USER="root"
+        AFP_CNID_SQL_USER="netatalk"
     fi
     if [ -z $AFP_CNID_SQL_DB ]; then
         AFP_CNID_SQL_DB="cnid"
+    fi
+
+    echo "*** MySQL CNID backend configured: host='$AFP_CNID_SQL_HOST', user='$AFP_CNID_SQL_USER', db='$AFP_CNID_SQL_DB'"
+    # Start MariaDB server if using localhost or 127.0.0.1 - for container-local database
+    if [ "$AFP_CNID_SQL_HOST" = "localhost" ] || [ "$AFP_CNID_SQL_HOST" = "127.0.0.1" ]; then
+        echo "*** Starting MariaDB server for MySQL CNID backend"
+        mkdir -p /run/mysqld
+        chown mysql:mysql /run/mysqld
+        if [ ! -d /var/lib/mysql/mysql ]; then
+            echo "*** Initializing MariaDB data directory"
+            mysql_install_db --user=mysql --datadir=/var/lib/mysql > /dev/null 2>&1
+        fi
+        # Configure MariaDB to listen on TCP 127.0.0.1 in addition to Unix socket
+        cat > /etc/my.cnf.d/netatalk.cnf << 'MYCNF'
+[mysqld]
+bind-address = 127.0.0.1
+skip-networking = 0
+MYCNF
+        # Start MariaDB with TCP enabled
+        mariadbd-safe --user=mysql &
+        # Wait for MariaDB to be ready
+        echo "*** Waiting for MariaDB to start..."
+        for i in 1 2 3 4 5 6 7 8 9 10; do
+            if mysqladmin ping > /dev/null 2>&1; then
+                echo "*** MariaDB is ready"
+                break
+            fi
+            sleep 1
+        done
+        # Configure MySQL user with password authentication
+        # (MariaDB uses unix_socket auth by default, which only works when running as root)
+        # Create user for both 'localhost' (Unix socket) and '127.0.0.1' (TCP)
+        echo "*** Configuring MySQL user with password authentication"
+        if [ -z "$AFP_CNID_SQL_PASS" ]; then
+            AFP_CNID_SQL_PASS="netatalk"
+        fi
+        mariadb -u root << EOSQL
+CREATE USER IF NOT EXISTS '$AFP_CNID_SQL_USER'@'localhost' IDENTIFIED BY '$AFP_CNID_SQL_PASS';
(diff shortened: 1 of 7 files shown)


distrib/docker/entrypoint_netatalk.sh
```diff
@@ -184,16 +184,60 @@ else
...
 fi
 
+# Validate MySQL/MariaDB CNID backend configuration
 if [ "$AFP_CNID_BACKEND" = "mysql" ]; then
     if [ -z $AFP_CNID_SQL_HOST ]; then
         AFP_CNID_SQL_HOST="localhost"
     fi
     if [ -z $AFP_CNID_SQL_USER ]; then
-        AFP_CNID_SQL_USER="root"
+        AFP_CNID_SQL_USER="netatalk"
```
rdmark, line 193: you argue that 'netatalk' is a better username default than 'root' in this case?

distrib/docker/entrypoint_netatalk.sh
```diff
@@ -184,16 +184,60 @@ else
...
     fi
     if [ -z $AFP_CNID_SQL_DB ]; then
         AFP_CNID_SQL_DB="cnid"
     fi
+
+    echo "*** MySQL CNID backend configured: host='$AFP_CNID_SQL_HOST', user='$AFP_CNID_SQL_USER', db='$AFP_CNID_SQL_DB'"
+    # Start MariaDB server if using localhost or 127.0.0.1 - for container-local database
```
rdmark, line 200: do I understand it correctly that this is a local mariadb configuration strictly for testing Unix sockets in a container, or is it something that someone could reasonably choose for a production container deployment too?

depending on the answer, I may have follow-up suggestions. :)

distrib/docker/entrypoint_netatalk.sh
```diff
@@ -225,6 +269,9 @@ login message = $AFP_LOGIN_MESSAGE
 mimic model = $AFP_MIMIC_MODEL
 server name = ${SERVER_NAME:-Netatalk File Server}
 uam list = $UAMS
+mac charset = ${AFP_MAC_CHARSET:-MAC_ROMAN}
+unix charset = ${AFP_UNIX_CHARSET:-UTF8}
+vol charset = ${AFP_VOL_CHARSET:-UTF8}
```
rdmark, line 274: I think these could be useful for the general public, so please document them in distrib/docker/README.md !

doc/manpages/man5/afp.conf.5.md
```diff
@@ -496,7 +496,15 @@ for most deployments. The default is **localhost:4700**.
...
-> name or address of a MySQL server for use with the mysql CNID backend.
+> Name or address of a MySQL server for use with the mysql CNID backend.
+>
+> On Unix systems, setting this to **localhost** (or leaving it unset) causes
+> the MySQL client library to prefer Unix socket connections, which offers
+> better performance for a local database. Setting it to **127.0.0.1**
+> forces a TCP connection to the loopback interface.
+>
+> On Windows systems, TCP is always used regardless of this setting since
+> Unix sockets are not supported.
```
rdmark: since netatalk doesn't run on Windows, I wonder if it's redundant to describe this caveat?

libatalk/cnid/mysql/cnid_mysql.c
```diff
@@ -1019,19 +1019,51 @@ struct _cnid_db *cnid_mysql_open(struct cnid_open_args *args)
...
+     */
+    if (vol->v_volcharset == CH_UTF8) {
+        EC_ZERO(mysql_options(db->cnid_mysql_con, MYSQL_SET_CHARSET_NAME, "utf8mb4"));
+    }
+
     const AFPObj *obj = vol->v_obj;
+    LOG(log_info, logtype_cnid,
+        "MySQL CNID backend: host='%s', user='%s', database='%s'",
+        obj->options.cnid_mysql_host ? obj->options.cnid_mysql_host : "(default)",
+        obj->options.cnid_mysql_user ? obj->options.cnid_mysql_user : "(none)",
```
rdmark: I wonder if it can be seen as security sensitive to log usernames?

distrib/docker/entrypoint_netatalk.sh
andylemin: Yes.
Using root is generally a bad idea, but in this case it was causing issues with the local in-container database, as a root user always exists, and there is a root user record by default.

To make it work required clobbering the root user password, and presented complications around using CREATE or ALTER queries.
This will now work for any given username but root - which is desirable.

And while root access is possible without a password when connecting via unix socket, Netatalk drops to a different user so that method doesn't work anyway, and a valid password is still needed.

This way there is no user collision, and a new user record and password can be created and everything works

distrib/docker/entrypoint_netatalk.sh
andylemin: Haha, both!

It was setup to support setting the MySQL host to "localhost" (test Unix socket), or "127.0.0.1" (test TCP without Nagle).

But yes, you could add this to the production container with "localhost" and then it would be production ready

distrib/docker/entrypoint_netatalk.sh
andylemin: Agreed.
Added as I needed to test the charset used with MySQL.

Also want to debug why everything but MacRoman reports as an invalid charset for the Mac charset.

libatalk/cnid/mysql/cnid_mysql.c
andylemin: Good point. Logging is done safely and with the correct permissions, but yes I agree, I'll remove

distrib/docker/entrypoint_netatalk.sh
rdmark: makes total sense now, thanks for the explanation

distrib/docker/entrypoint_netatalk.sh
rdmark: then my suggestion would be that we test this configuration in the github workflow, in an additional job!

andylemin: Nice, yes agreed. I'll finish the clean up on this tonight

andylemin:
@rdmark This is ready for final review

PR #2652, opened 2026-01-19
meson: introduce option to force the building of subprojects

rdmark:
by enabling'-Dwith-subprojects' you make meson fetch the tarballs and build subprojects, even if the dependency was found on the system

this has the side effect that you can build netatalk distribution tarballs with bundled subprojects on any system

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -667,7 +667,7 @@ has_bgets = cc.compiles(bgets_test_code, name: 'bgets defined in standard C libr
 cdata.set('HAVE_BGETS', has_bgets)
 bstring = dependency('bstring', required: false)
 
-if not bstring.found()
+if not bstring.found() or get_option('with-subprojects')
     bstring_proj = subproject('bstring')
     bstring = bstring_proj.get_variable('bstring_dep')
 endif
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -178,6 +178,12 @@ option(
     description: 'Create the Netatalk state directories if they do not exist',
 )
 option(
+    'with-subprojects',
+    type: 'boolean',
+    value: false,
+    description: 'Always build subprojects',
+)
+option(
     'with-tcp-wrappers',
     type: 'boolean',
     value: true,


rdmark:
agreed that a primary goal should be to keep build system options consistent between feature releases

in this particular case, I'm literally the only user of this option at the moment, so changing it won't cause any pain at all I promise ;)

the release process / automation is described here if you're curious: https://github.com/Netatalk/netatalk/wiki/Release-Process

now I agree that it's arguably a bit icky to group the flags like this, the main argument is to make things easier for me (and future release managers) to just turn on this flag as a shorthand for everything that is needed to make a release

that said, maybe someone else would want to force the bstring subproject even if they have bstring already installed? but the counter argument is... that if you have bstring installed as a shared library already, building and installing the bstring tarball will only clutter your system with another bstring shared library.

but, I'm open to suggestions!

andylemin:
I've slept on this and I don't like it! 😉

with-website is clear for people who want the website, and with-dist is clear for us who need build things.

We should separate these. Ie, keep with-website so we don't upset downstream folks (even if we're the only ones using it, some users like local resources), and add with-dist.

Just renaming and not splitting it out is always going to be sub-optimal for someone. It's ok if with-dist automatically enables with-website..

rdmark:
hehe, alright let me rethink this.

at a related note, I have been playing with the idea of folding the website code (Netatalk/netatalk.io) into this one, effectively creating a monorepo for netatalk. this has a few upsides, as well as a few downsides.

pros

- no more duplication of markdown files
- no need to keep the generated doxygen html pages in revision control, just generate them on the fly when publishing the website
- the website can read the latest version number from the build system

cons

- questionable if we want to keep a bunch of binary website assets (images) in the main repo
- build system complexity (maybe we need a website publishing mode that bypasses building the C code?)

rdmark:
now the option is much more narrowly defined. what do you think?

andylemin (approved):

PR #2653, opened 2026-01-20
[4.4] cnid: handle error when attempting to look up did<2 in sqlite backend

rdmark:


rdmark:
this is a second attempt at merging this branch, the previous PR https://github.com/Netatalk/netatalk/pull/2641 malfunctioned and no commit ended up in the stable branch

PR #2658, opened 2026-01-21
meson: package release tarball with subprojects

rdmark:
distributing subprojects code, currently only bstring, enables you to build the tarball in environments that lack a network connection


rdmark:
forward-port of the hotfix I applied to the 4.4 branch to unblock the release

PR #2668, opened 2026-01-29
Add ARC (Adaptive Replacement Cache) option for directory cache

### LINKED ISSUE #2642: introduce Adaptive Replacement Cache (ARC)
ZFS style ARC cache has several benefits over LRU cache, as discussed in https://github.com/Netatalk/netatalk/discussions/2422

this ticket is about introducing ARC cache as either a full replacement or a user configurable option

andylemin:
Implements ARC algorithm alongside existing LRU (based on Megiddo & Modha, IBM 2004 paper).
https://lrita.github.io/images/posts/datastructure/ARC.pdf

LRU remains default for backward compatibility. ARC offers 10-50% better hit ratios, 2x improvement on sequential scans.

Implementation:
- 4 lists: T1 (recent), T2 (frequent), B1/B2 (ghosts for learning)
- REPLACE subroutine with curdir protection
- Self-tuning parameter p (adapts recency vs frequency)
- Ghost mechanism: Keep ALL fields (simplest, safest approach)
- Comprehensive statistics and ARC-aware dump
- Zero-allocation queue management: MRU moves and all queue transitions
  use node recycling (queue_move_to_tail, queue_move_to_tail_of)

Critical Fixes:
- **LRU crash fix: Evict cache BEFORE hash_alloc_insert to prevent overflow
  crashes when LRU full (pre-existing issue triggered by cache thrashing)**
- Dangling pointer fix: qidx_node = NULL after all dequeue operations (13 sites)
- Statistics accuracy: dir_remove() report_invalid parameter for correct counts

Performance optimisations:
- Inline hash_vid_did(), hash_comp_vid_did(), hash_didname(), hash_comp_didname()
- Inline hash_lookup() to eliminate function call overhead
- Increase hash table INIT_SIZE from 64 to 1024 (match MIN_DIRCACHE_SIZE)

Other Changes:
- Configuration: Added 'dircache size' option (consistent naming, backward compatible)
- Documentation: Comprehensive Configuration.md with examples, tuning guide
- Test infrastructure: AFP_DIRCACHESIZE=1024 in all 17 integration test jobs

Configuration (afp.conf):
```
  dircache mode = arc|lru           # Choose algorithm (default: lru)
  dircache size = <entries>         # Also accepts 'dircachesize' (legacy)
  dircache validation freq = 100    # Tuning for Netatalk-only access
```

Docker support:
```
AFP_DIRCACHE_MODE=arc
AFP_DIRCACHESIZE=1024
```

Files modified: 16 (~1850 lines)

Closes #2642

# Tests

## test509
test509 is a new cache stress test which creates a working set larger than cache size and runs multiple lookups mixed with scans to approximate a more realistic work pattern

### Validation Frequency = 1 (Strict)
| Metric | LRU | ARC | Improvement |
|--------|-----|-----|-------------|
| **Lookups** | 28,100 | 25,600 | -8.9% (fewer lookups needed) |
| **Cache Hits** | 3,724 (13.3%) | 6,162 (24.1%) | **+81% hit rate** |
| **Ghost Hits** | N/A | 761 (3.0%) | Predictive capability |
| **Total Hit Rate** | 13.3% | **27.0%** | **+103% improvement** |
| **Evictions** | 11,776 | 7,962 | **-32.4% fewer evictions** |
| **Added Entries** | 13,876 | 10,677 | -23.0% |

### Validation Frequency = 100 (Relaxed)
| Metric | LRU | ARC | Improvement |
|--------|-----|-----|-------------|
| **Lookups** | 28,100 | 25,600 | -8.9% |
| **Cache Hits** | 3,724 (13.3%) | 6,162 (24.1%) | **+81% hit rate** |
| **Ghost Hits** | N/A | 761 (3.0%) | Predictive capability |
| **Total Hit Rate** | 13.3% | **27.0%** | **+103% improvement** |
| **Evictions** | 11,776 | 7,962 | **-32.4% fewer evictions** |
| **Validations** | ~37 | ~69 | Different validation patterns |

### Validation Frequency = 1 (Strict)
| Metric | LRU | ARC | Improvement |
|--------|-----|-----|-------------|
| **Lookups** | 37,657 | 35,139 | -6.7% |
| **Cache Hits** | 7,134 (18.9%) | 8,694 (24.7%) | **+30.7% more hits** |
| **Ghost Hits** | N/A | 2,183 (6.2%) | Predictive capability |
| **Total Hit Rate** | 18.9% | **31.0%** | **+64% improvement** |
| **Evictions** | 13,824 | 12,264 | **-11.3% fewer evictions** |
| **Added Entries** | 18,452 | 14,714 | -20.3% |
| **Expunged** | 222 | 222 | Same |
| **Invalid on Use** | 203 | 203 | Same |

### Validation Frequency = 100 (Relaxed)
| Metric | LRU | ARC | Improvement |
|--------|-----|-----|-------------|
| **Lookups** | 37,660 | 35,143 | -6.7% |
| **Cache Hits** | 7,306 (19.4%) | 8,877 (25.3%) | **+30.5% more hits** |
| **Ghost Hits** | N/A | 2,179 (6.2%) | Predictive capability |
| **Total Hit Rate** | 19.4% | **31.5%** | **+62% improvement** |
| **Evictions** | 13,824 | 12,253 | **-11.4% fewer evictions** |
| **Validations** | ~73 | ~110 | More strategic validation |
| **Expunged** | 31 | 31 | Same |
| **Invalid on Use** | 22 | 22 | Same |


.github/workflows/containers.yml
```diff
@@ -367,6 +367,7 @@ jobs:
               -e AFP_VERSION=7 \
               -e AFP_REMOTE=1 \
               -e AFP_EXTMAP=1 \
+              -e AFP_DIRCACHESIZE=1024 \
```
rdmark, line 370: do we want to enable ARC cache in a subset of these test jobs too? it feels like we ought to test it continuously!

doc/manpages/man5/afp.conf.5.md
```diff
@@ -584,6 +584,17 @@ cache evictions if the working set is larger than the dircache size.
 **dircachesize** and file-intensive workflows. For volumes where Netatalk
 is the only accessor, enabling this with a large cache may improve performance.
 
+dircache mode = *lru* | *arc* (default: *lru*) **(G)**
+
+> Cache replacement algorithm. **lru** = Least Recently Used (legacy).
+**arc** = Adaptive Replacement Cache (self-tuning, 10-50% better hit ratios,
+2× better on sequential scans). ARC uses approximately **2× memory** (~100% overhead)
```
rdmark: thinking about how to convey these options to users, perhaps we can label LRU as "stable" and ARC as "experimental"? or alternatively "efficient" and "high performance"?

just thinking that the "legacy" label might suggest that LRU is going away in the future, which I don't think is the case given the difference in memory usage!

doc/manual/Configuration.md
```diff
@@ -110,7 +110,7 @@ There is a command line utility called **dbd** available which can be used
 to verify, repair and rebuild the CNID database.
 
 ***NOTE:*** There are some CNID related things you should keep in mind when
-working with netatalk:
+working with Netatalk:
```
rdmark: I guess it's old-school to write Unix commands / project names in all lower case ;)

doc/manual/Configuration.md
```diff
@@ -703,97 +703,100 @@ or ACL_GROUP stay intact.
...
+validation operations are largely redundant and can be safely reduced (increase
+'dircache validation freq'), significantly lowering IO on storage/page-cache/ZFS ARC
+layers.
+
+NB; If a cache entry is found to be invalid during use, it is automatically and
+immeditely rebuilt. Critical operations such a delete, rename, xattr or rights
+changes are always validated. These settings only affect the delay before detecting
+and showing external filesystem changes in Netalk.
+
+> **💡 Netatalk-Only Access Optimization:**
```
rdmark: It feels wrong somehow to have Unicode emojis in the manual. Probably because I remember when this text was ancient docbook XML, not long ago. But never mind this comment, I'm only musing. :D

you may have seen that I recently caved in and added emojis to the netatalk.io landing page, so who am I to complain!

doc/manpages/man5/afp.conf.5.md
```diff
@@ -541,9 +541,9 @@ each share.
 
 dircachesize = *number* **(G)**
```
rdmark, line 542: shouldn't this be updated to `dircache size` (with a space) as well?

doc/manual/Configuration.md
```diff
@@ -811,22 +814,124 @@ if the working set is larger than the configured dircachesize.
...
+    dircache size = 524288
     dircache files = yes
     dircache validation freq = 100
+    dircache mode = arc
 
 **Note**: Monitor dircache effectiveness by checking Netatalk log files
 for "dircache statistics:" lines when afpd shuts down gracefully (user
 disconnects).
 
+### Directory Cache Statistics Reference
```
rdmark: this whole statistics reference section doesn't seem at home in the Configuration manual chapter, in my opinion

the chapter is already very long and dense with technical information, while the below information is extremely niche even for power users

what do you say about creating a wiki page for this instead? other ideas?

etc/afpd/dircache.c
```diff
@@ -128,15 +131,96 @@ static volatile uint64_t validation_counter = 0;
...
     unsigned long long expunged;
     unsigned long long evicted;
-    unsigned long long invalid_on_use;  /*!< entries that failed when used */
+    /*!< entries that failed when used */
+    unsigned long long invalid_on_use;
 } dircache_stat;
 
+/********************************************************
+ * ARC (Adaptive Replacement Cache) Implementation
+ * Based on: Megiddo & Modha, IBM 2004
```
rdmark, line 151: it's this paper?

https://theory.stanford.edu/~megiddo/pdf/IEEE_COMPUTER_0404.pdf

.github/workflows/containers.yml
andylemin: Yes absolutely.

doc/manpages/man5/afp.conf.5.md
andylemin: Good points. Yes LRU will not go away, it is specifically intended for small memory systems, while ARC offers significant benefits, it uses twice the memory.. NB; ARC benefits are still significantly higher than just a double size LRU 😉

doc/manual/Configuration.md
andylemin: Grammar police! 😉😂

doc/manual/Configuration.md
andylemin: Haha, this is a bad habit I have to make text easier to search and identify land marks. The landing page is appropriate, however it feels out of context here

doc/manpages/man5/afp.conf.5.md
andylemin: Well spotted. Both are supported, but yes, manuals should be consistent.

doc/manual/Configuration.md
andylemin: agreed, this belongs in /Volumes/case-sensitive/github/netatalk/doc/developer/dircache.md with a reference to this in Configuration.md

etc/afpd/dircache.c
andylemin: Yes, exactly this one. Should we add the URL?

etc/afpd/directory.c
```diff
@@ -454,7 +454,7 @@ static struct path *path_from_dir(struct vol *vol, struct dir *dir,
         }
 
         ret->d_dir = NULL;      /* 4 */
-        dir_remove(vol, dir);   /* 5 */
+        dir_remove(vol, dir, 0);   /* 5 - Proactive cleanup, not invalid on use */
```
rdmark, line 460: not related tot his line, but you've made such massive changes to this source file that you may want to consider adding your name and year to the copyright header

etc/afpd/hash.c
```diff
@@ -27,7 +27,7 @@
 #ifdef KAZLIB_RCSID
 #endif
 
-#define INIT_BITS   6
+#define INIT_BITS   10
```
rdmark, line 31: what's the context for this change?

test/testsuite/T2_Dircache_attack.c
```diff
@@ -636,6 +636,445 @@ STATIC void test506()
     exit_test("Dircache:test506: stat subdir in poisoned path");
 }
 
+/*!
+ * @brief test508: ARC stress test with evictions and ghost hits
```
rdmark: would you say this test has value when run without the recommended settings (e.g. LRU cache)? we have the option to put it in a separate bucket and run only when passed a certain parameter, for instance.

test/testsuite/T2_Dircache_attack.c
```diff
@@ -636,6 +636,445 @@ STATIC void test506()
...
+        printf("  Phase 2 completed: 5 reaccess passes\n");
+    }
+
+    /* ============================================================
+     * PHASE 3: Enumerate to verify stability under load
+     * COMMENTED OUT: This phase is verification/cleanup, not part of
+     * the actual ARC stress test. Including it distorts statistics with
+     * ~2100 sequential lookups that inflate miss counts.
+     * ============================================================ */
+    /* PHASE 3 DISABLED - Uncomment to re-enable enumerate verification
```
rdmark: when would you ever enable this code, if it doesn't have value to the overall test goal?

btw, minor detail, but the convention in this codebase is to use `#if 0` blocks for disabled code

libatalk/util/queue.c
```diff
@@ -70,6 +70,78 @@ qnode_t *enqueue(q_t *q, void *data)
...
+ */
+qnode_t *queue_move_to_tail(q_t *q, qnode_t *node)
+{
+    if (!q || !node || node == q) {
+        return NULL;  // Invalid parameters or trying to move sentinel
+    }
+
+    /* Fast path: already at tail (MRU)? */
+    if (node->next == q) {
+        return node;  // Nothing to do
```
rdmark: nitpick: it would be great if we could use `/* */` style code comments for consistency :)

etc/afpd/hash.c
andylemin: Yes the comment detailing this one in the summary is quite subtle. "Increase hash table INIT_SIZE from 64 to 1024 (match MIN_DIRCACHE_SIZE)".

We start with a queue size of just 64 entries.

This changes it so we start with an initial queue size of 1024 entries, as 1024 is still small enough even for the smallest of systems and avoids repeated memory reallocations at startup.

I'll add a comment to this line

test/testsuite/T2_Dircache_attack.c
andylemin: Yes both of the new tests are very valuable for both LRU and ARC.

Test508 was designed to test queue overflow (evictions), and is how I discovered the critical longstanding LRU bug which is triggered when thrashing the LRU cache very quickly.

Test509 was designed to exercise ARC semantics (trigger evictions into B1/B2 ghost lists, promotions into T1/T2 MRU/MFU lists, and scan/sweep resistance).

Test509 has minimal benefit over 508 for LRU however as the LRU is just FIFO (so 509 looks like a bigger stress test of the same, but that's still useful considering the bug I found).

So I think they should both always be enabled?

test/testsuite/T2_Dircache_attack.c
andylemin: Well spotted 🙂
This should be removed as I wrote 509 afterwards which deals with ARC semantics properly, and this is only overflow testing now.

libatalk/util/queue.c
andylemin: Well spotted! I'll fix.

I also deliberated over the inline comments vs line before.

There are many inline comments which I added as they fit better with their use case and structure.

Happy to parse over them again

doc/manual/Configuration.md
```diff
@@ -703,97 +703,102 @@ or ACL_GROUP stay intact.
...
+frequent entries and uses 2× memory (cache + ghosts). Both are production-ready;
+choose based on memory budget and workload characteristics.
 
-Maximum possible entries in the directory cache. The cache stores
-directories and files. It is used to cache the full path to directories
-and CNIDs which considerably speeds up directory enumeration.
-Given value is rounded up to nearest power of 2. Each entry takes about
-100 bytes, which is not much, but remember that every afpd child process
-for every connected user has its own cache.
+See the [Directory Cache Developer Documentation](dircache.md) for detailed
```
rdmark: I cannot test it right now, but it is unlikely that the resulting hyperlink will work as you indent it to, since dircache.md is not in the same hierarchy as the manual.

How about having a hard link to the page as published on the website, i.e. https://netatalk.io/developer/md_developer_2dircache ?

etc/afpd/dircache.c
rdmark: why not! :)

etc/afpd/dircache.c
```diff
@@ -927,39 +1804,145 @@ void log_dircache_stat(void)
...
+            arc_cache.stats.p_decreases);
+        /* ARC operations breakdown - detailed operation counters */
+        LOG(log_info, logtype_afpd,
+            "ARC operations: cache_hits(T1=%llu, T2=%llu), "
+            "promotions(T1→T2=%llu), evictions(T1=%llu, T2=%llu)",
+            arc_cache.stats.hits_t1,
+            arc_cache.stats.hits_t2,
+            arc_cache.stats.promotions_t1_to_t2,
+            arc_cache.stats.evictions_t1,
+            arc_cache.stats.evictions_t2);
```
rdmark, line 1916: maybe I'm overthinking it, but isn't this extremely chatty logs for INFO level logging? DEBUG or even MAXDEBUG seems more appropriate to me.

etc/afpd/directory.h
```diff
@@ -1,24 +1,24 @@
...
- * implied warranties of any kind.
- *
- *	Research Systems Unix Group
- *	The University of Michigan
- *	c/o Mike Clark
- *	535 W. William Street
- *	Ann Arbor, Michigan
- *	+1-313-763-0525
- *	netatalk@itd.umich.edu
+  Copyright (c) 1990,1993 Regents of The University of Michigan.
```
rdmark: how come the copyright date changes here? did you pull in code from another file?

etc/afpd/directory.h
```diff
@@ -1,24 +1,24 @@
 /*
- * Copyright (c) 1990,1991 Regents of The University of Michigan.
- * All Rights Reserved.
```
rdmark, line 3: nitpick: the stylistic convention in this project is the *-indented code comments (as documented in the coding style guide) so it seems counter intuitive to change it ;)

etc/afpd/directory.h
```diff
@@ -1,24 +1,24 @@
...
- *
- *	Research Systems Unix Group
- *	The University of Michigan
- *	c/o Mike Clark
- *	535 W. William Street
- *	Ann Arbor, Michigan
- *	+1-313-763-0525
- *	netatalk@itd.umich.edu
+  Copyright (c) 1990,1993 Regents of The University of Michigan.
+  Copyright (c) 2025 Andy Lemin (andylemin)
```
rdmark: did you consider claiming copyright in directory.c as well? you arguably changed a lot more in that file.

as a side note, I got some advice from Debian project members that header files are generally not considered copyrightable (following the outcome of Google v Oracle, and the whole Java API debacle, I believe) so I personally stopped putting my name in header files for simplicity's sake

I totally support your choice here, but just wanted to share my perspective!

test/testsuite/T2_Dircache_attack.c
rdmark: maybe you want to add a line or two in the test508 description below about expected behavior when running it with LRU? when the "Test Requirements" says that ARC is a precondition, I would expect the test to fail without it if I didn't know better!

test509 has this explained nicely already.

etc/afpd/dircache.c
andylemin: Well spotted! Yes these should be turned back down now I have completed extensive validation testing

doc/manual/Configuration.md
andylemin: Perfect 👍

etc/afpd/directory.h
andylemin: Was trying to normalise it with directory.c?

etc/afpd/directory.h
andylemin: Interesting.. I guess if there already one there, then it's fine to keep it. Otherwise we don't have to add it to new headers?

rdmark: Right, this is the de facto policy I have been operating under. If someone absolutely wants to claim copyright of headers I wouldn’t protest strongly though. 

If it helps I can make it an official project policy.

test/testsuite/T2_Dircache_attack.c
andylemin: Agreed. The first test was a bit scrappy, while 509 was properly planned.
doc strings updated

etc/afpd/dircache.c
andylemin: Actually these should stay as info!
These logs are only printed once when the afpd process shuts down (when the user logs out).

However you are correct there are other other logs which I left at info log level - I have now reduced those, but intentionally left these at info.

etc/afpd/dircache.c
```diff
@@ -1,15 +1,15 @@
...
-  This program is free software; you can redistribute it and/or modify
-  it under the terms of the GNU General Public License as published by
-  the Free Software Foundation; either version 2 of the License, or
-  (at your option) any later version.
-
-  This program is distributed in the hope that it will be useful,
-  but WITHOUT ANY WARRANTY; without even the implied warranty of
-  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-  GNU General Public License for more details.
+ * Copyright (c) 2010 Frank Lahm <franklahm@gmail.com>, 2025 Andy Lemin (andylemin)
```
rdmark: can you please put your name on a separate line? not least to keep the Debian copyright parser happy. ;)

etc/afpd/dircache.h
```diff
@@ -1,16 +1,16 @@
...
-   it under the terms of the GNU General Public License as published by
-   the Free Software Foundation; either version 2 of the License, or
-   (at your option) any later version.
-
-   This program is distributed in the hope that it will be useful,
-   but WITHOUT ANY WARRANTY; without even the implied warranty of
-   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-   GNU General Public License for more details.
- */
+ * Copyright (c) 2010 Frank Lahm <franklahm@gmail.com>, 2025 Andy Lemin (andylemin)
```
rdmark: same as above

etc/afpd/directory.h
rdmark: it seems this wasn't reverted yet?

etc/afpd/directory.h
rdmark: I would argue that normalization is not required in this case, since the .c and .h files aren't necessarily modified at the same pace. Let's keep the UMich copyright years as-is, if you don't mind.

rdmark:
the code looks great now, please just touch up the copyright headers, and we should be good to ship this! 👍

distrib/docker/entrypoint_netatalk.sh
```diff
@@ -257,8 +257,9 @@ cnid mysql host = $AFP_CNID_SQL_HOST
 cnid mysql user = $AFP_CNID_SQL_USER
 cnid mysql pw = $AFP_CNID_SQL_PASS
 cnid mysql db = $AFP_CNID_SQL_DB
-dircachesize = ${AFP_DIRCACHESIZE:-8192}
+dircachesize = ${AFP_DIRCACHESIZE:-65536}
 dircache files = ${AFP_DIRCACHE_FILES:-no}
+dircache mode = ${AFP_DIRCACHE_MODE:-lru}
```
rdmark, line 262: I think this new env variable is worth documenting in distrib/docker/DOCKER.md – this is something we want to encourage users of the container to experiment with, right?

andylemin: Very well spotted!

etc/afpd/directory.c
```diff
@@ -1,6 +1,16 @@
...
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
```
rdmark: just to confirm your intention here: you deliberately chose to change the license of this file from UMich to GNU GPL v2?

if you look at COPYRIGHT in the root of the repo you can see the text of the original UMich license, which as per Debian is a variant of the MIT license with slightly different wording

as a quick history lesson (sorry if you heard this before) this codebase was distributed under the UMich license until 2001, when the maintainers / community at the time decided to change the default license to GPL v2

however all historical code from the 90s is still under the old license which is why we still distribute the COPYRIGHT file (UMich) alongside COPYING (GPL)

the FSF considers MIT to be GPL compatible so this is not a problem as I understand it, however whether we can change the license of historical files without a full rewrite is more of a grey area

IANAL but I *think* it's fine to change it here given that we keep the name of the copyright holders and still have the COPYRIGHT file in the repo

what do you think?

rdmark:
@andylemin ...please pardon the never ending copyright header nitpicks! I think this should be the last one for this PR, I promise.

etc/afpd/directory.c
andylemin: Hi. Interesting! I was just trying to modernise it and make it consistent following significant changes.

Have done some research and it seems that while maintainers can change the license, we have to keep the original license notice in place in the case of MIT.

If we want to fully change it then we need to get the approval of all the original contributors to fully update the license notices.

In our scenario we are introducing dual licensing and can simply supplement the MIT notice with a GPL notice.

rdmark: Sounds like a safe middle road, using both copyright notices together in the affected files.

rdmark (approved):

Issue #2669, opened 2026-01-29
4.4.1 build is broken on 32-bit by usage of legacy sync built-ins


barracuda156:
**Describe the bug**

Since 4.4.x the code started using legacy 8-byte `__sync*` built-ins, which are not supported on 32-bit platforms.

**To Reproduce**

Run the build on ppc (perhaps mips and arm too).

**Expected behavior**

Legacy `__sync*` built-ins should not be used at all – modern `__atomic*` ones work fine via `libatomic`.

**Environment**
- Client OS: macOS 10.6.8
- Netatalk Version: 4.4.1

**Configuration**

Share the (anonymized) contents of your afp.conf file,
plus any other relevant configuration files

**Logs**

```
FAILED: [code=1] etc/afpd/afpd

[...]

PR #2670, opened 2026-01-30
dircache.c: replace legacy __sync* built-ins with C11 __atomic* ones

barracuda156:
Fixes https://github.com/Netatalk/netatalk/issues/2669

### DIFF
--- a/etc/afpd/dircache.c
+++ b/etc/afpd/dircache.c
@@ -257,7 +257,7 @@ static unsigned long queue_count_max = 0;
 static int should_validate_cache_entry(void)
 {
     /* Thread-safe increment using compiler builtins */
-    uint64_t count = __sync_fetch_and_add(&validation_counter, 1);
+    uint64_t count = __atomic_fetch_add(&validation_counter, 1, __ATOMIC_SEQ_CST);
 
     /* Validate every Nth access to detect external changes */
     if (dircache_validation_freq == 0) {
@@ -930,7 +930,7 @@ void log_dircache_stat(void)
 
         if (dircache_validation_freq > 0) {
             /* Thread-safe read of validation counter */
-            uint64_t counter_value = __sync_fetch_and_add(&validation_counter, 0);
+            uint64_t counter_value = __atomic_load_n(&validation_counter, __ATOMIC_SEQ_CST);
             validation_ratio = ((double)counter_value / (double)dircache_validation_freq) /
                                (double)dircache_stat.lookups * 100.0;
         }
@@ -952,8 +952,8 @@ void log_dircache_stat(void)
         dircache_stat.lookups,
         dircache_stat.hits, hit_ratio,
         dircache_stat.misses,
-        dircache_validation_freq > 0 ? __sync_fetch_and_add(&validation_counter,
-                0) / dircache_validation_freq : 0, validation_ratio,
+        dircache_validation_freq > 0 ? __atomic_load_n(&validation_counter,
+                __ATOMIC_SEQ_CST) / dircache_validation_freq : 0, validation_ratio,
         dircache_stat.added,
         dircache_stat.removed,
         dircache_stat.expunged,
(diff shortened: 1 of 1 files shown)


rdmark:
@barracuda156 thanks for filing the PR! first off, a few formalities: since netatalk is a very long-lived project, we care a lot about the quality of commit messages. you should be able to bisect the commit history and immediately understand the context of each commit. therefore, please write at least a paragraph that explains why and what you are fixing for posterity. please remember that the GitHub issue tracker may not live forever, or may have breaking changes in the future, so you cannot rely on only issue ticket references.

secondly, and this is a nitpick, prefix the commit message with "afpd:" rather than "dircache.c:" – see https://netatalk.io/contributing for more details about commit and coding standards

about the logic changes: I'd love to get @andylemin's eyes on this, since he's the one who edited those lines last

andylemin:
Hi @barracuda156 Thanks so much for this!

These changes look great to me, sorry for breaking 32bit builds!
This area is under active development at the moment.

I see all the replacements use __ATOMIC_SEQ_CST, which has the strongest ordering and matches the full-barrier semantics of the legacy __sync_* builtins. This is definitely correct and safe. For a simple counter like this __ATOMIC_RELAXED would probably suffice for the loads, and __ATOMIC_ACQ_REL for the increment - but using SEQ_CST is conservative and definitely correct.

`__sync_fetch_and_add` → `__atomic_fetch_add`
Both return the previous value and atomically add. The C11 __atomic_fetch_add is the direct replacement. The __ATOMIC_SEQ_CST memory ordering matches __sync_fetch_and_add's implicit full barrier semantics.

`__sync_fetch_and_add(&x, 0)` → `__atomic_load_n`
The old code used the hack of adding zero to atomically read the value. Replacing with __atomic_load_n is semantically cleaner - it's a pure load with no write side-effect. So this is also an improvement.

`__sync_lock_test_and_set` → `(void)__atomic_exchange_n`
__sync_lock_test_and_set is technically an atomic swap (returns old value, stores new). Using __atomic_exchange_n with a (void) cast to discard the return value is the equivalent replacement.

The current `main` now has 2 additional __sync_fetch_and_add calls which should also be changed.
`__sync_fetch_and_add(&validation_counter, 0)` — in log_dircache_stat() LRU branch
`__sync_fetch_and_add(&validation_counter, 0)` — in log_dircache_stat() ARC branch

Are you able to rebase and update the others? Thanks again for this..

andylemin:
Issue #2669 mentions needing libatomic on 32-bit platforms. This PR doesn't add -latomic linking in meson.build.
On GCC for 32-bit targets `__atomic_*` operations typically require linking against libatomic. Without that, the build may still fail with a different set of undefined symbols (`__atomic_*_8 instead of __sync_*_8`).

I think the meson build system would need a check like:
```
if cc.has_function('__atomic_load_8', args: '-latomic')
  add_project_link_arguments('-latomic', language: 'c')
endif
```

barracuda156:
@andylemin Thank you for the review. Yes, I can rebase, will deal with this over the weekend.

You are right re linking, I just don’t know how to do it in meson correctly (i.e. configure test for atomics). In ports we can always pass needed flags via portfile code, which is what I usually did outside of CMake builds.

rdmark:
@barracuda156 meson is well documented and has expressive pseudo-python syntax so I would say that the learning curve is low; in this case, if you look in the Libraries section in meson.build in the repo root, you'll find the current convention for adding linker flags to the project. I think it simply should be a matter of conditionally appending the flag to the `netatalk_common_link_args` list, becuase we call `add_global_link_arguments(netatalk_common_link_args, language: 'c')` later which is then used when linking afpd.

barracuda156:
@rdmark Passing the flag is not a problem, the problem is to decide at configure time if it is needed.

rdmark:
Andy's suggested conditional `if cc.has_function('__atomic_load_8', args: '-latomic')` should work, or a variant of it at least. Some experimentation may be needed.

barracuda156:
@rdmark This is probably not what we need. With CMake we can test if `-latomic` is needed and available. Something like this: https://github.com/uxlfoundation/oneTBB/pull/987

rdmark:
sure, building a snippet of code as a test can be done in a similar fashion; here's one example from our meson.build

```
    if cc.compiles(
        '''
        #include <rpcsvc/rquota.h>
        int main(void) {
        enum qr_status foo;
        foo = Q_OK;
        return 0;
        }
        ''',
        dependencies: tirpc,
    )
        cdata.set('HAVE_RQUOTA_H_QR_STATUS', 1)
    endif
```

rdmark:
discarding this in favor of https://github.com/Netatalk/netatalk/pull/2756 -- thanks again @barracuda156 for the report and patch

PR #2672, opened 2026-01-31
cnid: fix sqlite backend bugs, add post init validation and logging

rdmark:
the wrong statement bindings were cleared in cnid_sqlite_execute, which could lead to stale data to persist

use single quotes instead of double quotes for looking up database table names, treating them as string literals instead of identifiers; using double quotes here caused table creation to fail with certain versions of the SQLite library

add validation logic after initializing a CNID database table, plus various additional debug logging


rdmark:
I think this is worth backporting to the stable 4.4 branch as well

andylemin (approved):

PR #2676, opened 2026-02-02
remove redundant project version setting for SonarQube Cloud

rdmark:
the project version is used to label scans on sonarcloud.io and mostly a cosmetic thing, so rather than having the management overhead of keeping this in sync with the project version, I propose we remove it altogether from this configuration file

### DIFF
--- a/sonar-project.properties
+++ b/sonar-project.properties
@@ -4,7 +4,6 @@ sonar.organization=netatalk
 sonar.sources=.
 sonar.sourceEncoding=UTF-8
 sonar.projectName=netatalk
-sonar.projectVersion=4.4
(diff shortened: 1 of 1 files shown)


andylemin:
Any chance we can have a single definition somewhere, which all relevant files can import the current development version from?

I realise that would be quite tricky to do, but I'm sure big projects have done that somehow.

rdmark:
It might be possible, but not through the usual mechanism which is the meson build system. Alternatively we can remove it from this sonar config file as it is mostly a cosmetic thing to label the static analysis scanning history. Maybe Sonar had another way to achieve the same, worth researching a bit.

PR #2683, opened 2026-02-02
Remove superseded dircache metadata window/threshold parameters

andylemin:
Depends on #2668

Remove superseded dircache metadata window/threshold parameters
    
The `dircache metadata window` and `dircache metadata threshold` parameters were a thoughtful optimization when initially implemented to avoid unnecessary cache invalidation from metadata-only directory changes (permissions, xattrs).

However, subsequent improvements have superseded this approach:

1. **Robust invalid-on-use detection** - dircache_report_invalid_entry() now catches all stale entries during actual use, providing comprehensive safety

2. **Direct inode/ctime validation** - Simple, reliable change detection that catches ALL modifications without time-window ambiguity

3. **Improved cached permissions** - accessmode() properly recalculates rights when needed, handling metadata changes transparently

The time-based heuristic added overhead (extra time() syscall per validation) while trying to optimize a case now better handled by these newer mechanisms.
More importantly, the heuristic could miss changes that fell outside its time windows, whereas the simplified approach ensures correctness by detecting ALL modifications.

Changes:
    - Removed cache_entry_externally_modified() function (77 lines)
    - Simplified validation to inline inode/ctime comparison
    - Removed 2 config parameters (window, threshold)
    - Updated dircache_set_validation_params() signature (3→1 param)
    - Cleaned Docker entrypoint, CI workflows, and Webmin module

Benefits:
    - Improves correctness - detects all changes without time-window ambiguity
    - Eliminates time() syscall overhead
    - More cache invalidations detected, but updated entries re-cache immediately
    - Simpler configuration (only dircache_validation_freq remains)
    - More predictable behavior

Testing: All 301 spec tests pass
    
Bonus: Added missing dircache mode (lru/arc) option to Webmin UI

*With this PR the dircache refactoring is finally considered complete, and unblocks higher level performance changes - Eg, AFP_Enumerate can now trust and use the cache instead of calling stat on every file and directory during every enumeration, and we can implement full support for file caching*


rdmark:
Reducing the configuration complexity for end users is an admirable goal! I’m all for this. 

The big ARC PR has been merged now so this branch can be rebased.

contrib/webmin_module/edit_global_section.cgi
rdmark: since you changed the canonical name of this option to `dircache size` in the previous PR I think it's good to change it in the webmin module too

rdmark (changes requested):
please remove the documentation of AFP_DIRCACHE_METADATA_WINDOW and AFP_DIRCACHE_METADATA_THRESHOLD from distrib/docker/DOCKER.md and we should be good to go!

edit: plus the respective paragraphs from doc/manpages/man5/afp.conf.5.md and doc/manual/Configuration.md as well

rdmark:
btw github is having a global Actions outage right now, since about an hour ago:

https://www.githubstatus.com/

curiously github hosted macOS runners are still, well, running, while Ubuntu runners aren't 🤷

andylemin:
> please remove the documentation of AFP_DIRCACHE_METADATA_WINDOW and AFP_DIRCACHE_METADATA_THRESHOLD from distrib/docker/DOCKER.md and we should be good to go!
> 
> edit: plus the respective paragraphs from doc/manpages/man5/afp.conf.5.md and doc/manual/Configuration.md as well

Well spotted 🙂

andylemin:
I fixed all these yesterday, but looks like I did not push. I'll push when next at computer.

andylemin:
@rdmark ready for final review and merge

PR #2685, opened 2026-02-02
decouple the dbd binary from cnid_dbd and make it better

rdmark:
break out the dbd source code from cnid_dbd and move it into its own module, removing superfluous sources, dependencies, and header includes

at one point in the past this tool was probably thightly coupled with the Berkeley DB database interface, but now it uses the CNID backend interface

this allows you to build dbd even when Berkeley DB isn't available, because it can, unlike what the name suggests, operate on any CNID backend

since the wholesale move of the code, SonarQube flagged all code quality bugs as new, so I took the opportunity to fix all of them except three const to non const conversion which would require modification of the CNID API

### DIFF
--- a/doc/manpages/man1/dbd.1.md
+++ b/doc/manpages/man1/dbd.1.md
@@ -10,13 +10,16 @@ dbd — CNID database maintenance utility
 
 # Description
 
-**dbd** scans all file and directories of AFP volumes, updating the CNID
-database of the volume. It must be run with appropriate permissions i.e.
-as root.
+**dbd** scans all file and directories of an AFP volume and updates the CNID
+database when discrepancies are found between the database and the filesystem.
+It must be run with appropriate permissions, i.e. as root.
 
 When run with the **-f** parameter, it will completely wipe the existing
 volume table in the database and create a new one from scratch.
-This can be used to for example convert from one CNID scheme to another.
+This can be used for example to convert a volume from one CNID scheme to another.
+
+Despite sharing a name with the **dbd** (Database Daemon) CNID backend module,
+this tool can be used with any CNID backend.
 
 # Options
 
@@ -67,7 +70,7 @@ modifying files.
 
 # See also
 
-nad(1), cnid_metad(8), cnid_dbd(8)
+nad(1), afpd(8), cnid_dbd(8)
 
 # Author
 


rdmark:
as a side note, it has always bothered me that _dbd_ the tool has the same name as _dbd_ the CNID backend.

I have been toying with the idea of retcon the former to be short for "Database Doctor" but maybe that's too much :)

andylemin:
Are you saying that cnid_dbd provides a generic CNID backend client interface for the upper layer?

And the newly split out DBD code has a CNID backend server layer?

If yes, this is surprising as it must mean all the other cnid_x modules include very similar CNID interface code? And we could do well in the long term to standardise/normalise and rename this to cnid_client or something and use it as the client for all the backends?

Or did I completely misunderstand 😉

rdmark:
cnid_dbd in in fact _not_ generic interfaces for CNID backends, but is hard coded for Berkeley DB. But in my reading of the code it could potentially be rejigged for any database backend, and it's not impossible it could be rearchitected to serve as a general purpose interface for arbitrary CNID backends too.

I can see two paths forward: We either deprecate cnid_dbd when Berkeley DB finally disappears from downstream distribution. Or, we repurpose it for other databases as discussed above.

The argument for deprecation is that the multi-deamon architecture is complex and introduces (small amounts of) latency with IPC for all database operations. It also arguably increases the attack surface of netatalk, something that I felt first hand when I designed the SELinux policies for netatalk a few months back.

The prime argument for keeping and adapting this structure is the extra layer of protection against database corruption with SQLite specifically. The SQLite library serializes write operations by multiple clients but reading the docs I get the impression that they don't guarantee that this scales well with a very large number of concurrent clients. So cnid_dbd + SQLite could potentially be a good combination for very large deployments. But then again, maybe such deployments should be using MySQL instead. :)

cnid_dbd design decisions and tradeoffs are discussed at a high level in https://netatalk.io/developer/md_developer_2cnid

And no, the dbd app that we are splitting out here is _not_ doing any database operations by itself but rather uses the proper CNID API.

andylemin (approved):

andylemin:
Great, this makes sense.

I'm not concerned about the SQLite concurrency limitations as users should indeed be using MySQL.

But I think we have a responsibility to provide guidance. 

Something like "SQLite is suitable for small deployments upto 20 users. Larger installations should use MySQL"

@rdmark

Issue #1739, opened 2024-11-08
s390x / big-endian test failure: set unix access privilege two users


rdmark:
File system permissions issue exclusively on IBM s390x / big-endian (via debci test runners.)

```
############## entering test358 ##############
[FPOpenVolFull] Open Vol test1 bitmap 21
[FPCreateDir] Create Directory Vol 1 did : 0x2 <t358 dir>
directory ID 0x14c
[FPCreateFile] Create File SOFT Vol 1 did : 0x14c <t358 file>
[FPCreateFile] Create File SOFT Vol 1 did : 0x14c <t358 file user 2>
[FPGetFileDirParams] GetFileDirParams Vol 1 did : 0x14c <>
[FPGetFileDirParams] GetFileDirParams Vol 1 did : 0x14c <>
[FPSetFilDirParam] Set FilDirParam 0x14c <> bitmap 0x1001
[FPGetFileDirParams] GetFileDirParams Vol 1 did : 0x14c <>
[FPEnumerateFull] Enumerate Vol 1 did : 0x14c <> sindex: 1 reqcnt: 5
[FPEnumerateFull] Enumerate Vol 1 did : 0x14c <> sindex: 1 reqcnt: 5
header.dsi_code       -5000	AFPERR_ACCESS  
[FPSetDirParms] Set DirParam 0x14c  bitmap 0x1000
[FPDelete] FPDelete conn 17972370 Vol 1 did : 0x14c <t358 file>
[FPDelete] FPDelete conn 17972370 Vol 1 did : 0x14c <t358 file user 2>
[FPDelete] FPDelete conn 17972370 Vol 1 did : 0x14c <>
header.dsi_code       -5032	AFPERR_OLOCK   
[../test/testsuite/FPSetFileDirParms.c:950] FPDelete(Conn, vol, dir , "")
	FAILED
FPSetFileDirParms:test358: set unix access privilege two users - FAILED
```

PR #2688, opened 2026-02-05
testsuite: initialize filedir params to avoid garbage data in the bitmap

rdmark:
test358 was reading unitialized data into the DIRPBIT_ACCESS bit which caused flaky behavior later when attempting to delete the test dir on big endian platforms

eliminate the retry workaround that was masking the actual underlying bug

apply the same initialization fix consistently across the testsuite code to preempt similar issues elsewhere

### DIFF
--- a/test/testsuite/Error.c
+++ b/test/testsuite/Error.c
@@ -270,7 +270,7 @@ afp_enumerate
 static void cname_test(char *name)
 {
     int  ofs =  3 * sizeof(uint16_t);
-    struct afp_filedir_parms filedir;
+    struct afp_filedir_parms filedir = { 0 };
     uint16_t bitmap = (1 <<  DIRPBIT_LNAME) | (1 << DIRPBIT_PDID) |
                       (1 << DIRPBIT_DID) |
                       (1 << DIRPBIT_ACCESS);
@@ -477,7 +477,7 @@ STATIC void test100()
     int dir;
     char *name = "t100 no obj error";
     char *name1 = "t100 no obj error/none";
-    struct afp_filedir_parms filedir;
+    struct afp_filedir_parms filedir = { 0 };
     uint16_t bitmap = (1 << DIRPBIT_ATTR);
     unsigned int ret;
     uint16_t vol = VolID;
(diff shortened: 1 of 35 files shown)


rdmark:
This is hopefully the final fix required for https://github.com/Netatalk/netatalk/issues/1739

I added a [section to the Testing wiki](https://github.com/Netatalk/netatalk/wiki/Testing#test-multiarch) about how to easily test on exotic architectures such as s390x using multiarch container images.

andylemin:
Hi @rdmark 
I must admit that I don't like this fix, I feel that this is potentially hiding an underlying race issue and this is not how real clients interact?
Can we try delete_directory_tree instead?

rdmark:
I see your point, didn't think of it that way. From that perspective, I think it's best to revisit the retry logic further down in the same test. This was added by me through PR https://github.com/Netatalk/netatalk/pull/1797 to address the same issue: flakiness on Debian s390x

Curiously it's only this particular test that is susceptible to this race condition issue on s390x

rdmark:
@andylemin this test behaves very weirdly on Debian s390x. The retry loop for deleting the dir _always_ succeeds on the first try, but if I remove the loop and replace it with seemingly _anything_ else (call to FAIL macro, store return code and check in if statement, create another conditional statement that emulates the while loop, add sleeps inbetween) we get error AFPERR_OLOCK when attempting to delete the dir.

But there's some other odd nondeterminism that's going on too, because sometimes a particular simplified code passes, but then when I try the exact same code after recompiling it is back to failing.

I'm at a loss how to troubleshoot this properly! We have a workaround in place that seems to be fully consistent, but it's kind of gross.

rdmark:
@andylemin it was so simple, a case of accessing uninitialized data in the bitmap, which was masked on little endian because the byte swapping later in directory.c:setdirparams()

for good measure I applied the same initialization across the board in the testsuite

andylemin:
Hi @rdmark Fantastic. Makes perfect sense and this makes me happy to see it fixed properly, and the fix applied consistently.. Thank you for pushing on 🙂

PR #2689, opened 2026-02-05
introduce a default global 'cnid scheme' afp.conf option

rdmark:
rather than allowing a compile time default, which is less transparent to users, introduce 'cnid scheme' as a global afp.conf option which can be overridden at the volume level


andylemin (approved):
Nice

PR #2690, opened 2026-02-05
docs: improve descriptions of CNID backends, set/save password options

rdmark:
- outline the pros and cons of the various CNID backends with somewhat higher fidelity
- touch up descriptions of the set and save password options

### DIFF
--- a/doc/manual/Configuration.md
+++ b/doc/manual/Configuration.md
@@ -368,7 +368,7 @@ encryption, separate password storage)
 
 - "DHX2" UAM (successor of DHCAST128)
 
-There exist other optional UAMs as well:
+With Kerberos support enabled at compile time, Netatalk also supports:
 
 - "Client Krb v2" UAM (Kerberos V, suitable for "Single Sign On" Scenarios
 with macOS clients – see below)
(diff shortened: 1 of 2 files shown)


rdmark:
@andylemin is this what you had in mind about providing guidance to users?

doc/manual/Configuration.md
```diff
@@ -138,20 +138,29 @@ CNID database is restricted to the **cnid_dbd** daemon process.
...
 
 A performant and lean CNID database backend that uses the SQLite v3
 embedded database engine.
 
-### mysql
-
-CNID backend using a MySQL server. The MySQL server has to be
-provisioned by the system administrator, and Netatalk configured
-to connect to it over the network.
+This is the fastest and simplest option as it lacks the dependencies on controller daemons as with dbd,
```
andylemin: SQLite is not the fastest? But SQLite is faster than dbd while being equally as simple to use as dbd.

"it lacks the dependencies on controller daemons as with dbd" - this applies to both MySQL and SQLite

andylemin: "or an external database as with mysql" - the position of this wording is slightly confusing. This is not a performance overhead, only an admin overhead.
A properly configured MySQL backend is the fastest option and the most scalable, but requires administration to configure enough resident memory (for the CNID DB to reside in memory) and appropriate page flushing intervals for resilience.

andylemin:
@rdmark Thank you! Great improvement, just a couple comments

rdmark:
@andylemin I cannot see your comments yet :)

andylemin:
@rdmark Sorry, I started but did not complete the review! you should be able to see it now?

doc/manual/Configuration.md
rdmark: fair points! I have rewritten the descriptions, hopefully for the better.

rdmark:
great, thank you for the constructive feedback

please approve the PR :)

edit: re-approval wasn't needed, only a closing of the last open thread

PR #2692, opened 2026-02-06
update dircache docs to align with recent changes

rdmark:
a few straggler places that refer to dircachesize needed updating


rdmark:
@andylemin I think the **Directory Cache Tuning** section in Configuration.md needs to be updated to explain ARC cache as well

andylemin:
Hi @rdmark Thank you so much for spotting these! Admittedly I did not do a search for any remaining dircachesize mentions.
I will update Configuration.md to provide an ARC cache explanation 👍

rdmark:
great, thanks for doing that. if you like you can push to this branch, you should have permissions to write to other project members' branches

andylemin:
@rdmark How does this look?
It is a little long relative to the rest of the doc, but its the minimum length that still enables the user to make an informed choice?

rdmark:
LGTM!

I think you were able to compress it just enough to fit into the style of this manual chapter :)

PR #2693, opened 2026-02-07
afpd dircache: fix realloc leak, validate ghosts, improve error logging

andylemin:
- Fix realloc memory leak in dircache_remove_children() using temp variable
- Remove ghost-specific validation skip in search_by_did and search_by_name
- ARC ghost entries now validated identically to regular cached entries
- Remove redundant AFP_ASSERT(dir->qidx_node) in arc_case_i()
- Upgrade arc_replace() curdir-protection logs from log_debug to log_warning
- Remove overly strict ghost capacity post-condition assert
- Add NULL callback check in queue_destroy() to prevent latent crash
- Remove dead code t1_size==0 check inside t1_size>=1 branch in arc_replace()
- Use dirlookup_strict() in afp_setfildirparams, afp_createfile, afp_createdir
- Update dirlookup/dirlookup_strict doxygen comments

### DIFF
--- a/etc/afpd/dircache.c
+++ b/etc/afpd/dircache.c
@@ -402,9 +402,6 @@ static void arc_ensure_ghost_capacity(arc_list_t target_list)
             dir_free(ghost);
         }
     }
-
-    /* Post-condition: ghost directory has room for one more entry */
-    AFP_ASSERT((arc_cache.b1_size + arc_cache.b2_size) < arc_cache.c);
 }
 
 /*!
@@ -433,8 +430,8 @@ static void arc_replace(int in_b2)
              (arc_cache.t1_size > arc_cache.p))) {
         /* Evict LRU from T1 → B1 (move to ghost list) */
 
-        /* Check if T1 is empty */
-        if (arc_cache.t1_size == 0 || arc_cache.t1->next == arc_cache.t1) {
+        /* Defensive: check T1 queue empty (counter desync guard) */
+        if (arc_cache.t1->next == arc_cache.t1) {
             LOG(log_debug, logtype_afpd, "arc_replace: T1 is empty");
             return;
         }
@@ -450,8 +447,9 @@ static void arc_replace(int in_b2)
         if (victim == curdir) {
             /* Check if we have more than one entry */
             if (arc_cache.t1_size < 2 || arc_cache.t1->next->next == arc_cache.t1) {
-                LOG(log_debug, logtype_afpd,
-                    "arc_replace: cannot evict curdir from T1");
+                LOG(log_warning, logtype_afpd,
+                    "arc_replace: cannot evict curdir from T1 (T1=%zu, T2=%zu, c=%zu)",
+                    arc_cache.t1_size, arc_cache.t2_size, arc_cache.c);
                 return;
             }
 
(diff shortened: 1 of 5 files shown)


rdmark (approved):
would you say that any of these fixes should be backported to the stable 4.4 branch?

andylemin:
> would you say that any of these fixes should be backported to the stable 4.4 branch?

Hi @rdmark, these are only edge case fixes for rare situations.
For example if someone tries to create a file or directory via the command line after another user has deleted a parent folder.

There is a critical legacy LRU fix in the recent ARC implementation I need to backport though..

andylemin:
Hi @rdmark A little update on progress just for fun..
As of the above PR, the below shows the performance comparison between "LRU and Freq=1" vs "ARC and FREQ=100" on the same code.

## Performance Comparison (5 iterations average, ms)
| Test | AFP Ops | LRU f=1 (ms) | ARC f=100 (ms) | Δ ms | Δ% |
|------|---------|---------|-----------|------|-----|
| File read (1000 files) | 8,000 | 4493 | 4323 | -170 | -3.8% |
| Write 100MB | 103 | 97 | 85 | -12 | -12.4% |
| Read 100MB | 102 | 38 | 36 | -2 | -5.3% |
| Lock/Unlock 10K | 20,000 | 795 | 783 | -12 | -1.5% |
| Create 2000 files | 4,000 | 4964 | 5102 | +138 | +2.8% |
| **Enumerate 2000** | ~51 | 1474 | 1082 | **-392** | **-26.6%** |
| Delete 2000 files | 2,000 | 3858 | 3257 | **-601** | **-15.6%** |
| Create 1000 dirs | 1,110 | 1881 | 1727 | -154 | -8.2% |
| **Cache hits** | 11,000 | 3890 | 3163 | **-727** | **-18.7%** |
| **Mixed cache ops** | 820 | 1008 | 694 | **-314** | **-31.2%** |
| **Deep path traversal** | 3,500 | 2016 | 1068 | **-948** | **-47.0%** |
| Cache validation | 30,000 | 8529 | 9286 | +757 | +8.9% |
| **Total** | **80,686** | **33,043** | **30,606** | **-2,437** | **-7.4%** |
| **Avg per AFP op** | | **0.410 ms** | **0.379 ms** | **-0.031** | **-7.6%** |

## Dircache Statistics
| Metric | LRU f=1 | ARC f=100 | Improvement |
|--------|---------|-----------|-------------|
| Hit rate | 93.1% | 96.2% | +3.1 % |
| Ghost hits | N/A | 0 (0.0%) | — |
| True misses | 6.9% | 3.8% | -3.1 % |
| Validations | ~607,459 (103%) | ~6,049 (1.0%) | **99% reduction** |
| Expunged | 19,817 | 671 | **96.6% reduction** |
| Invalid on use | 18,435 | 425 | **97.7% reduction** |
| Added | 26,035 | 23,488 | -9.8% |
| Removed | 26,034 | 23,487 | -9.8% |
| Max entries | 2,013 | 2,004 | Similar |

Time spent on cache hits is lower because there are fewer lookups, and there are fewer lookups because the initial lookups have a higher hit+valid success rate. For the same reason, fewer entries were added and removed as the MRU and MFU entries stay in the ARC cache, rather than being evicted and re-added again later with LRU FIFO

rdmark:
Seems to neatly validate your original hypothesis! Especially when it comes to enumeration and deep path traversal. As an end user browsing complex volumes must be much more snappy now.

Is this worth popping into the Benchmarks wiki page now? https://github.com/Netatalk/netatalk/wiki/Benchmarks

andylemin:
Yes I want to create a series of performance comparisons comparing all the different milestones and features.
I need to do this on my server where performance measurements are more consistent compared to my laptop which has background processes that add noise to measurements.

PR #2694, opened 2026-02-07
detect big-endian architecture correctly, remove testsuite workaround

rdmark:
replace the Autotool-ism for endianness macro that we had before with one that is correct for Meson; we can now build with big-endian byte order support on f.e. s390x architecture

since the underlying bug that caused test failures on big-endian systems has been fixed, we can remove the -X override for the testsuite spec test that skipped tests that were failing due to endianness data corruption

### DIFF
--- a/doc/manpages/man1/afp_spectest.1.md
+++ b/doc/manpages/man1/afp_spectest.1.md
@@ -4,7 +4,7 @@ afp_spectest — AFP specification compliance test suite
 
 # Synopsis
 
-**afp_spectest** [-1234567aCEimVvX] [-h *host*] [-H *host2*] [-p *port*] [-s *volume*] [-c *path to volume*]
+**afp_spectest** [-1234567aCEimVv] [-h *host*] [-H *host2*] [-p *port*] [-s *volume*] [-c *path to volume*]
 [-S *volume2*] [-u *user*] [-d *user2*] [-w *password*] [-f *test*]
 
 **afp_spectest** -l
@@ -95,9 +95,6 @@ Single tests or entire testsets can be executed with the **-f** option.
 
 **-w** *password*
 : Password for authentication with AFP server
-
-**-X**
-: Skip tests that aren't big endian compatible
 
 # Usage
 
--- a/test/testsuite/FPSetFileDirParms.c
+++ b/test/testsuite/FPSetFileDirParms.c
@@ -889,11 +889,6 @@ STATIC void test358()
     const DSI *dsi2;
     ENTER_TEST
 
-    if (Bigendian) {
-        test_skipped(T_BIGENDIAN);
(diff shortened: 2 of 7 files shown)


andylemin:
This is very cool!
After some searching it looks like the same UB pattern exists in the `LREV` macro (include/atalk/byteorder.h:196):
```c
#define LREV(x) ((IREV(x)<<32) | (IREV((x)>>32)))
```

I think we can add the same `(uint64_t)` cast?
```c
// Fixed:
#define LREV(x) (((uint64_t)IREV(x)<<32) | (IREV((x)>>32)))
```

rdmark:
great catch, pushed a new commit now!

as a side note, our byteorder.h was forked from Samba ages ago, and their version has changed significantly since then; might be worth seeing if there are other improvements to port over

https://github.com/samba-team/samba/blob/master/lib/util/byteorder.h

andylemin:
> great catch, pushed a new commit now!
> 
> as a side note, our byteorder.h was forked from Samba ages ago, and their version has changed significantly since then; might be worth seeing if there are other improvements to port over
> 
> https://github.com/samba-team/samba/blob/master/lib/util/byteorder.h

@rdmark Great idea!

Here is a quick analysis of which fixes in samba's byteorder.h apply to Netatalk's byteorder macros;

1) IREV type safety — Samba casts the argument to uint32_t to prevent 64-bit values leaking through when called from LREV:

// Current Netatalk:
`#define IREV(x) ((SREV(x)<<16) | (SREV((x)>>16)))`
// Samba's fix:
`#define IREV(x) ((SREV((uint32_t)(x))<<16) | (SREV(((uint32_t)(x))>>16)))`

2) int16 → int16_t — 7 occurrences in SVALS, SSVALS, SVALS_NC macros use bare int16 which has no typedef in Netatalk. Samba replaced all with int16_t.

3) Investigate the removal of !CAREFUL_ALIGNMENT direct-cast path (lines 160–191) — only triggers on __i386__, uses strict-aliasing UB (*(uint16_t*)buf). Samba removed this entirely; modern compilers optimise the shift-based macros equally well. Saves ~30 lines but needs perf validation - probably best for another PR

rdmark:
neat, thanks for parsing the header with your keen eyes :-D

actually, I'm thinking of pulling the bit shift macro fixes out of this changeset altogether and put up a new PR for this

FTR big-endian gcc will throw warnings of potential overflow when compiling spotlight_marshalling.c before the LVAL fix; the macro is not used anywhere else in the codebase

rdmark:
@andylemin since I rebased the branch, please re-approve this PR :)

edit: never mind, let me just merge it. I didn't change any of the approved commits, just moved the byteorder.h fixes into a separate PR.

PR #2696, opened 2026-02-07
papd: harden config parsing with graceful error recovery

rdmark:
- replace fatal exit(1) on malformed printer entries with skip-and-continue pattern
- normalize error reporting (fprintf->LOG, perror->LOG)
- fix unsafe memory patterns (malloc+strcpy->strdup, strcpy->strlcpy)
- add calloc and free_printer_fields() for clean per-entry teardown
- strip leading whitespace in the termcap parser to handle indented/blank lines
- fix uamlist global leak
-  docs: papd.conf, remove fictitious 'au' flag, document actual 'ca' and 'sp' options

### DIFF
--- a/doc/manpages/man5/papd.conf.5.md
+++ b/doc/manpages/man5/papd.conf.5.md
@@ -39,14 +39,16 @@ terminated with colons. The available options and flags are:
 **am=(uams list)**
 
 > The **am** option allows specific UAMs to be specified for a particular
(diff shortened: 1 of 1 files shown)


rdmark:
@NJRoadfan heads-up, please let me know if you have opinions on these changes

I've tested it working with a basic CUPS use case

Issue #2691, opened 2026-02-05
PANIC's due to did lower than 17 with mysql CNID backend


BarSebastian:
**Describe the bug**

Occasionall  afp processes panic.

**To Reproduce**

Not sure - happens occasionally

**Expected behavior**

no panics :)

**Environment**

- Server OS: Ubuntu 25.10
- Client OS: macOS Sequoia
- Netatalk Version: 4.4.1

I use mariadb as CNID database and tuned dircache configurations.

**Configuration**

```ini
[Global]
; Global server settings
        afpstats = yes
        name = ****
        unix charsert = UTF8
        vol charset = UTF8
        uam list = uams_dhx2.so
        mimic model = RackMac

        cnid scheme = mysql
        cnid mysql host = localhost
        cnid mysql user = root
        cnid mysql pw = ***
        cnid mysql db = cnid

        dircache validation freq = 100
        dircache metadata window = 3600
        dircache metadata threshold = 1800
        dircachesize = 131072
        dircache files = no
        dircache validation freq = 100

        log file = /var/run/afpd.log
        dsireadbuf = 100
        afp listen = *.*.*.*
        zeroconf = yes
        ea = ad
        disconnect time = 48
        sleep time = 48
...
```

**Logs**

```
Feb 05 11:11:05.474736 afpd[92655] {enumerate.c:427} (error:AFPDaemon): enumerate(vid:4, did:294630, name:'.'): error adding dir: 'ZAPROSZENIA'
Feb 05 11:11:05.478848 afpd[92655] {dircache.c:608} (error:AFPDaemon): dircache_add(): did:0 is less than the allowed 17. Try rebuilding the CNID database for: ".DS_Store"
Feb 05 11:11:05.481324 afpd[92655] {fault.c:93} (severe:Default): PANIC: ntohl(dir->d_did) >= CNID_START
Feb 05 11:11:05.481334 afpd[92655] {fault.c:94} (severe:Default): BACKTRACE: 10 stack frames:
Feb 05 11:11:05.481340 afpd[92655] {fault.c:100} (severe:Default):  #0 /usr/local/lib/x86_64-linux-gnu/libatalk.so.19(netatalk_panic+0x32) [0x7f1961bce5d2]
Feb 05 11:11:05.481469 afpd[92655] {fault.c:100} (severe:Default):  #1 /usr/local/sbin/afpd(dircache_add+0x24c) [0x564ba2dd869c]
[...]

PR #2697, opened 2026-02-07
afpd: validate CNID from get_id() before calling dircache_add()

rdmark:
Database connection failures cause get_id() to return CNID_INVALID (0), which triggered a panic when passed to dircache_add(). Now validate and return early before attempting cache operations.

### DIFF
--- a/etc/afpd/dircache.c
+++ b/etc/afpd/dircache.c
@@ -1250,7 +1250,8 @@ int dircache_add(const struct vol *vol,
 
     if (ntohl(dir->d_did) < CNID_START) {
         LOG(log_error, logtype_afpd,
-            "dircache_add(): did:%u is less than the allowed %d. Try rebuilding the CNID database for: \"%s\"",
+            "dircache_add(): did:%u is less than the allowed %d for: \"%s\". "
+            "This indicates a caller passed an invalid CNID (possible database error).",
             ntohl(dir->d_did), CNID_START,
             dir->d_u_name ? cfrombstr(dir->d_u_name) : "(null)");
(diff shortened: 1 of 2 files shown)


andylemin:
This looks good to me.
Thanks for removing "Try rebuilding the CNID database for", this is an old legacy comment which does not make sense.
I'm personally not a fan of including comments like "/* Defense-in-depth: This should never trigger after fix for #2691, but kept as failsafe */". We could make statements like this about almost every single ASSERT, as this is implied by the purpose of AFP_ASSERT.

I like the guard, but my instinct would be get_id() should return -1 (explicit failure) rather than 0? And the guard should check it is >= CNID_START?

rdmark:
pushed a few fixes, thanks for checking!

I'm not sure about your point "get_id() should return -1" – the current logic of get_id() either gives us the CNID or CNID_INVALID (0), but you're saying we should introduce a different kind of error handling in this function?

andylemin:
> pushed a few fixes, thanks for checking!
> 
> 
> 
> I'm not sure about your point "get_id() should return -1" – the current logic of get_id() either gives us the CNID or CNID_INVALID (0), but you're saying we should introduce a different kind of error handling in this function?

Just suggesting that CNID_INVALID is a positive failure, while -1 is a negative failure (something failed internally in the function).

But you are right, this adds little value, and the checks you added should do the trick

rdmark:
FYI, I'm holding this in draft status for now, want to run some proper tests and try to reproduce the issue first.

andylemin (approved):

PR #2698, opened 2026-02-07
Enable OpenIndiana build in GitHub Actions

rdmark:
Removed condition that prevents the OpenIndiana build from running.


rdmark:
still takes 20min just to download 260MB worth of dependencies, so not realistic to activate this yet I think

PR #2699, opened 2026-02-08
uams: fix password corruption when changing password in Cleartxt PAM UAM

rdmark:
Fix stale pointer write through PAM_password into the DSI command buffer, and fix the PAM conversation function returning the wrong password when pam_chauthtok() prompts for the current password.

Copy old and new passwords into stack buffers upfront, and teach PAM_conv() to return the correct password for each chauthtok prompt based on whether we are running as root or not.

These fixes allows for an AFP client such as afpfs-ng to change the PAM password for an authenticated user remotely without corrupting the resulting password.

### DIFF
--- a/etc/uams/uams_pam.c
+++ b/etc/uams/uams_pam.c
@@ -37,6 +37,9 @@ static pam_handle_t *pamh = NULL;
 static const char *hostname;
 static char *PAM_username;
 static char *PAM_password;
+static char *PAM_oldpassword;
+static int PAM_chauthtok_mode;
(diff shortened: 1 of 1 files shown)


rdmark:
I've tested all the other PAM based UAMs (and Randnum): password changing works correctly there already, no bug fixing needed.

The non-PAM passwd/shadow based UAMs do not have logic for changing passwords, however, but all are coded to fail with AFPERR_BADUAM when you attempt it.

I think this is acceptable.

andylemin:
Nice fixes and easier to read

PR #2705, opened 2026-02-08
GitHub CI: enable OpenIndiana build job in workflow

rdmark:
enable the OpenIndiana build step using the new 202510-build tag that comes with build-essentials by default


rdmark:
thanks to upstream fix in https://github.com/vmactions/openindiana-vm/issues/2 we can now realistically run the OpenIndiana job in CI, as the required dependencies to install for each run is reduced to <30MB rather than closer to 270MB

PR #2712, opened 2026-02-08
touch up the byte order macros and remove i386 optimization

rdmark:
this changeset ports over a number of historical Samba patches, while doing some local cleanup as well

- typecast input to 64 bit int for bit shift macros that shift 32 bits
- fix undefined bitshift (by Joseph Sutton)
- remove 32bit macros breaking strict aliasing (by Andreas Schneider)
- use int16_t instead of int16 (by Stefan Metzmacher)
- rewrite the code comment that describes the purpose of the macros 
- move unmodified macros out of the WORDS_BIGENDIAN conditional
- fix bigendian word assembly order in 32-bit and 64-bit macros
- create a simple test for cross-platform byteorder logic

### DIFF
--- a/include/atalk/byteorder.h
+++ b/include/atalk/byteorder.h
@@ -46,14 +46,14 @@
 
 #define SVAL(buf,pos) (PVAL(buf,(pos)+1)|PVAL(buf,pos)<<8)
 #define SVALS(buf,pos) ((int16_t)SVAL(buf,pos))
-#define IVAL(buf,pos) (SVAL(buf,pos)|SVAL(buf,(pos)+2)<<16)
+#define IVAL(buf,pos) (SVAL(buf,(pos)+2)|SVAL(buf,pos)<<16)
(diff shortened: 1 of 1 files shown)


rdmark:
if you're curious, here are the original Samba commits

https://github.com/samba-team/samba/commit/443d70ee58f3dec05ddede22c33fec0860f52864 (from 2023)
https://github.com/samba-team/samba/commit/af69cb2a78810e608ccff115b433801a58a749e4 (from 2013)
https://github.com/samba-team/samba/commit/f88bf54c7f6d1c2ef833047eb8327953c304b5ff (from 2007)

andylemin:
What impact does removing the 32 bit macros have? Are the handle correctly now without them

rdmark:
for starters I wrote a set of simple tests to validate the correctness of the macros on 32bit (i386), 64bit (aarch64), little-endian (aarch64), and big-endian (s390x); this revealed a novel bug on big-endian in fact, for which I pushed a separate commit.

the tests are passing everywhere now!

PR #2733, opened 2026-02-21
Enumerate uses cache, AD support in cache and inter-process cache sync

andylemin:
Since our recent PRs <list PRs> which provided a more robust and coherent dircache, through lots of bug fixes, periodic stat validation, invalid-on-use detection, and a complete ARC implementation.
As a result the time has come where we can finally start leveraging the cache as a true fully fledged meta cache..

The first commit in this PR enables Enumerate to use the dircache instead of stat’ing every file/dir (this just required storing a few extra ostat fields) - providing a ~5-8% performance gain for Enumerate.

However Netatalk still needs to read the metadata for every file/dir - this is slow as it triggers a lot of disk IO (it takes around 8 seconds on my test machine to fully Enumerate 10,000 files/dirs, and our `lantest` performance utility shows an average time per AFP op at ~0.35ms on the same machine (average over all lantests, not just Enumerate).

In Netatalk all object metadata is stored in AppleDouble structs - and as the dircache is intended to cache metadata, AD data is absolutely in scope.. Just caching the stats is great, but there so much more potential to be realised.

Therefore the second commit in this PR introduces a complete AD metadata framework which extends the existing dircache entries to support storing AD struct data. There was a lot of work done to bolt on and integrate the whole AD structure into the dircache. Without going into details, the result was, every single metadata call across the entire codebase is now fully cached.

```mermaid
flowchart TD
    A[AFP Operation<br/>getfilparams / getdirparams / etc] --> B{Open fork<br/>exists?}
    B -->|Yes| C[Use fork-owned adouble<br/>Opportunistically populate cache]
    B -->|No| D[ad_metadata_cached]
    
    D --> E{strict mode?}
    E -->|false<br/>enumerate, catsearch| F{dcache_rlen ≥ 0?}
    E -->|true<br/>FPGetFileDirParms, delete| G[ostat + ctime/inode check]
    
    F -->|Yes| H[ad_rebuild_from_cache<br/>Serve from cache ✓]
    F -->|dcache_rlen == -2| I[Return ENOENT<br/>No AD exists ✓]
    F -->|dcache_rlen == -1| J[disk_read:<br/>ad_metadata from disk]
    
    G --> K{ctime/inode<br/>match cached?}
    K -->|Yes, rlen ≥ 0| H
    K -->|Yes, rlen == -2| I
    K -->|Changed| L[dir_modify DCMOD_STAT<br/>Invalidate AD → rlen = -1]
    L --> J
    K -->|rlen == -1| J
    
    J --> M[ad_metadata from disk]
    M -->|Success| N[ad_store_to_cache<br/>Populate cache fields]
    M -->|ENOENT| O[Set dcache_rlen = -2]
    N --> P[ad_close internal<br/>Return adouble to caller]
    
    style H fill:#4a4,color:#fff
    style I fill:#aa4,color:#000
    style J fill:#48a,color:#fff
```

The first enumerate is still long (closer to 6-7 seconds (down from 8) for 10,000 files after various bug fixes/improvements - most notably a fix to honour `convert appledouble = no` in Enumerate).
BUT every single subsequent Enumerate on 10,000 files/dirs now takes just ~100ms instead of 6-7 seconds on my machine!

Aside; Performance gains will vary depending on the speed of your storage. For example, _your gains will be proportional to the time that was spent on the avoided IO operations_.

As the metadata caching was integrated transparently into the low level ad_metadata functions, the benefits also touch almost every other AFP operation (either directly or indirectly) - as a result the overall system performance improved too. The total runtime for our lantest performance utility is now 2-3 times faster, and the average time per AFP op is now ~0.1ms (down from ~0.35 on my test hardware)!

However, before we get too excited, it is not just that simple.. Identifying out of date stat data is easy (if the stat is stale, object access fails, and graceful recovery provides elegant invalid-on-use). But it is not possible to replicate invalid-on-use for metadata as it is not involved in the “open” of a file/dir!
For example, stat is internal to Netatalk, while metadata is user data - if an entry is cached with metadata, when requested the metadata will be returned to the client - however we don’t know if it’s out of date or not.

We would have to perform an ostat to compare the ctime’s to validate if the metadata is stale or not - which would be a backwards step as we would be stat’ing everything again (However this is still much faster than reading from disk every time, and this is exactly what happens when you need to co-locate Netatalk with Samba - and you have `dircache validation freq = 1`)!

For a single user this is a non-issue as Netatalk has cache lifecycle coherency for all stat+inode+AD data now (every AFP operation keeps the cache in sync, so the cache is always valid) - but the cache for other users would become stale as a user changes file/dir metadata (each user has their own cache)..

Aside; as each user has their own cache, so the ARC is optimised for their individual usage patterns - Netatalk is the only OSS file sharing service where each user has their own ARC!

This brings us to the third commit in this PR. Netatalk has never had per-user IPC dircache synchronisation - instead it relied on the CNID database which only stores object navigation data. This third commit implements a cache hint signalling system which took inspiration from FCE (File Change Event) signalling, which provides multi-user coherence for ALL cached data.

For every AFP operation (that changes/deletes a file/dir object), we send a cache hint (vol+CNID) to the parent Netatalk afpd process, which relays the hint to all other per-user processes. The other user processes receive these hints and trigger a local cache refresh for the hinted object (only if it is in their ARC).

To achieve this we use IPC and Pipe FDs which provide POSIX atomic messaging for these small 22byte hints, combined with Event based actions allowing all the other per-user processes to refresh their local ARC entries when they are not busy serving AFP commands. There is some latency (hundreds of microseconds) but it's fast enough that the chances of another user reading metadata from the same file/dir which has just been changed, within a 100us window is extremely small.

```mermaid
graph LR
    subgraph U1["afpd - User 1 (writer)"]
        C1["ARC Cache<br/>stat + AD"]
    end
    U1 -->|"ipc_send_cache_hint<br/>(vid, cnid, event)"| N
    N["netatalk parent<br/>afpd daemon<br/>(hint relay)"]
    N -->|"hint relay"| U2
    N -->|"hint relay"| U3
    subgraph U2["afpd - User 2"]
        C2["ARC Cache<br/>stat + AD"]
    end
    subgraph U3["afpd - User 3"]
        C3["ARC Cache<br/>stat + AD"]
    end
```

```mermaid
sequenceDiagram
    participant Client1 as AFP Client 1
    participant AFPD1 as afpd (User 1)
    participant Pipe as IPC Hint Pipe<br/>(parent netatalk)
    participant AFPD2 as afpd (User 2)
    participant AFPD3 as afpd (User 3)
    
    Client1->>AFPD1: FPSetFileParms (modify file)
    AFPD1->>AFPD1: setfilparams()<br/>ad_flush() + dir_modify(STAT|AD)
    AFPD1->>Pipe: ipc_send_cache_hint<br/>(vid, cnid, REFRESH)
    
    Note over Pipe: 22-byte hint message<br/>relayed to all siblings
    
    Pipe->>AFPD2: hint arrives on pipe fd
    Pipe->>AFPD3: hint arrives on pipe fd
    
    Note over AFPD2: After current AFP cmd completes
    AFPD2->>AFPD2: process_cache_hints()<br/>ostat() + dir_modify(STAT|AD_INV|NO_PROMOTE)
    
    Note over AFPD3: After current AFP cmd completes
    AFPD3->>AFPD3: process_cache_hints()<br/>ostat() + dir_modify(STAT|AD_INV|NO_PROMOTE)
```

Therefore we now finally have a fully coherent multi-user ARC cache which supports all metadata. The only data not cached are ResourceForks and ancillary data such as Icon\r icns (folder cover images) which are stored as separate files rather than metadata. The result; Netatalk is orders of magnitude faster without compromising integrity.

Last but not least, this PR also introduces a suite of new integration tests to validate our new AD caching and multi-user cache coherence.

This PR represents a lot of changes, so I'm proposing we apply these to a Beta release and encourage our community to test and report any issues.

*NB: If you are running Netatalk alongside other processes which change files outside of Netatalk (eg Samba), you must ensure `dircache validation freq` remains set to 1 (the default), to ensure Netatalk stats/validates all objects every time (it is still much quicker to stat an object and serve all metadata from cache, than it is to read from disk every time).*

If Netatalk is the only process operating on a volume/share/mount, you can set `dircache validation freq=100` and get the maximum performance gains provided by these changes (by fully trusting the cache - all operations through Netatalk keep all ARCs in sync).

```
=== 4.4.1 Release ===
Test                                                                Time_ms  Time± AFPD_R AFPD_R± AFPD_W AFPD_W± CNID_R CNID_R± CNID_W CNID_W±   MB/s
------------------------------------------------------------------ -------- ------ ------ ------- ------ ------- ------ ------- ------ ------- ------
Open, stat and read 512 bytes from 1000 files [8,000 AFP ops]          3455 1210.9   6000     0.2   7000     0.5      0     0.0      1     0.0      0
Writing one large file [103 AFP ops]                                     69    7.7      0     0.0    299     0.0      0     0.0      0     0.0   1449
Reading one large file [102 AFP ops]                                     38    1.6    100     0.0    100     0.0      0     0.0      0     0.0   2631
Locking/Unlocking 10000 times each [20,000 AFP ops]                     822   48.0      0     0.0  20000     0.0      0     0.0      0     0.0      0
Creating dir with 2000 files [4,000 AFP ops]                           4522  504.6   2000     0.0  10000     0.0      4     1.0   6131     7.1      0
Enumerate dir with 2000 files [~51 AFP ops]                            1059  392.0   1960     0.0     49     0.0      0     0.0      0     0.0      0
Deleting dir with 2000 files [2,000 AFP ops]                           2769  547.2   4000     0.0   4000     0.0      4     1.0   6179     4.7      0
Create directory tree with 1000 dirs [1,110 AFP ops]                   1769   71.4      0     0.0   4440     0.0      2     1.3   2277    26.7      0
Directory cache hits (100 dirs + 1000 files) [11,000 AFP ops]          2787  233.8  10000     0.0  11100     0.0      0     0.0    100     0.0      0
Mixed cache operations (create/stat/enum/delete) [820 AFP ops]          482   46.5    820     0.0   1620     0.0      2     0.0   1217    23.1      0
Deep path traversal (nested directory navigation) [3,500 AFP ops]       813  149.5   2500     0.0   3550     0.0      0     0.0     50     0.0      0
Cache validation efficiency (metadata changes) [30,000 AFP ops]        7438   80.8  30000     0.0  30100     0.0      0     0.0    100     0.0      0
------------------------------------------------------------------ -------- ------ ------ ------- ------ ------- ------ ------- ------ ------- ------
Sum of all AFP OPs = 80686                                            26023         57380          92258             12          16055               

Aggregates Summary:
------------------------------------------------------------------
Average Time per AFP OP: 0.323 ms
Average AFPD Reads per AFP OP: 0.711
Average AFPD Writes per AFP OP: 1.143
See afp_lantest manpage for more information: https://netatalk.io/manual/en/afp_lantest.1

Dircache Statistics (/var/log/afpd.log):
------------------------------------------------------------------
Feb 26 03:25:33.747766 afpd[43] {dircache.c:944} (info:AFPDaemon): dircache statistics: (user: atalk1) entries: 0, max_entries: 2001, config_max: 131072, lookups: 2510304, hits: 2417772 (96.3%), misses: 91842, validations: ~24184 (1.0%), added: 91845, removed: 91845, expunged: 690, invalid_on_use: 91845, evicted: 0, validation_freq: 100

=== cached-enumerate ===
Test                                                                Time_ms  Time± AFPD_R AFPD_R± AFPD_W AFPD_W± CNID_R CNID_R± CNID_W CNID_W±   MB/s
------------------------------------------------------------------ -------- ------ ------ ------- ------ ------- ------ ------- ------ ------- ------
Open, stat and read 512 bytes from 1000 files [8,000 AFP ops]          1405  258.0   4001     1.2   7000     0.0      0     0.0      0     0.0      0
Writing one large file [103 AFP ops]                                     66    5.9      0     0.0    300     0.2      0     0.0      0     0.0   1515
Reading one large file [102 AFP ops]                                     37    2.7    100     0.0    100     0.0      0     0.0      0     0.0   2702
Locking/Unlocking 10000 times each [20,000 AFP ops]                     790   19.8      0     0.0  20000     0.0      0     0.0      0     0.0      0
Creating dir with 2000 files [4,000 AFP ops]                           2628   86.0      0     0.0  10000     0.0      4     1.0   4131     7.1      0
Enumerate dir with 2000 files [~51 AFP ops]                               8    1.2      0     0.0     49     0.0      0     0.0      0     0.0      0
Deleting dir with 2000 files [2,000 AFP ops]                           2756  216.3   4000     0.0   8000     0.0      4     1.0   6179     4.7      0
Create directory tree with 1000 dirs [1,110 AFP ops]                   1673  112.3      0     0.0   5550     0.0      2     1.3   2277    26.7      0
Directory cache hits (100 dirs + 1000 files) [11,000 AFP ops]           464   25.4      0     0.0  11000     0.0      0     0.0      0     0.0      0
Mixed cache operations (create/stat/enum/delete) [820 AFP ops]          544   42.3    400     0.0   2020     0.0      2     0.0   1017    23.1      0
Deep path traversal (nested directory navigation) [3,500 AFP ops]       146    7.5      0     0.0   3500     0.0      0     0.0      0     0.0      0
Cache validation efficiency (metadata changes) [30,000 AFP ops]        1262   17.7      0     0.0  30000     0.0      0     0.0      0     0.0      0
------------------------------------------------------------------ -------- ------ ------ ------- ------ ------- ------ ------- ------ ------- ------
Sum of all AFP OPs = 80686                                            11779          8501          97519             12          13604               

Aggregates Summary:
------------------------------------------------------------------
Average Time per AFP OP: 0.146 ms
Average AFPD Reads per AFP OP: 0.105
Average AFPD Writes per AFP OP: 1.209
See afp_lantest manpage for more information: https://netatalk.io/manual/en/afp_lantest.1

Dircache Statistics (/var/log/afpd.log):
------------------------------------------------------------------
Feb 26 09:52:38.335309 afpd[43] {dircache.c:1978} (info:AFPDaemon): dircache statistics (ARC): (user: atalk1) entries: 0, ghost_entries: 0, max_entries: 2001 (343 KB), config_max: 131072, lookups: 3619264, hits: 3526735 (97.4%), ghost_hits: 0 (0.0%), total_hits: (97.4%), misses: 92529 (2.6%), validations: 35274 (1.0%), added: 91841, removed: 91841, expunged: 688, invalid_on_use: 688, evicted: 0, validation_freq: 100
Feb 26 09:52:38.335313 afpd[43] {dircache.c:2007} (info:AFPDaemon): ARC ghost performance: (user: atalk1) ghost_hits: 0 (0.0%), ghost_hits(B1=0, B2=0), learning_benefit: (0.0%)
Feb 26 09:52:38.335316 afpd[43] {dircache.c:2016} (info:AFPDaemon): ARC table state: (user: atalk1) T1=0 (recent/cached), T2=0 (frequent/cached), B1=0 (recent/ghost), B2=0 (frequent/ghost), total_cached=0/131072, total_ghosts=0, freq_bias: (0.0%)
Feb 26 09:52:38.335320 afpd[43] {dircache.c:2026} (info:AFPDaemon): ARC adaptation: (user: atalk1) p=0/131072 (0.0% target for T1), p_range=[131072,0], adaptations=0 (increases=0, decreases=0)
Feb 26 09:52:38.335322 afpd[43] {dircache.c:2037} (info:AFPDaemon): ARC operations: (user: atalk1) hits(T1=91841, T2=3457744), promotions(T1→T2=91841), evictions(T1=0, T2=0)
Feb 26 09:52:38.335325 afpd[43] {dircache.c:2076} (info:AFPDaemon): dircache statistics (hints): (user: atalk1) sent: 272762, received: 0, acted_on: 0, no_match: 0
Feb 26 09:52:38.335328 afpd[43] {dircache.c:2092} (info:AFPDaemon): dircache statistics (AD): (user: atalk1) hits: 1040071, misses: 2570, no_ad: 0, hit_ratio: 99.8%
```

**NB:** The IO monitoring reads syscr and syscw from /proc/<pid>/io before and after each test, which counts _all_ read()/write() system calls - including socket and pipe operations, not just disk IO - this means the new cache change hint signalling is included in the IO numbers (hence the increase in AFPD_W for tests which mutate/delete files/dirs - the extra writes are the hints to other users).

**Test configuration:** 20 iterations, IO monitoring enabled, real APFS volumes via Docker bind mount
**Hardware:** macOS Docker Desktop on Apple Silicon (case-sensitive APFS)
**Cache config:** dircachesize=131072, validation_freq=100, mode=arc, files=yes, convert_appledouble=no

## Runtime Comparison (averages over 20 iterations)

| Test | 4.4.1 (ms) | cached-enum (ms) | Speedup | Notes |
|------|--------:|--------:|--------:|-------|
| Open+stat+read 1000 files | 3,455 | 1,405 | **2.5x** ⭐ | AD cache eliminates metadata re-reads |
| Write 100MB | 69 | 66 | 1.0x | IO-bound, no cache benefit |
| Read 100MB | 38 | 37 | 1.0x | IO-bound, no cache benefit |
| Lock/Unlock 10k | 822 | 790 | 1.0x | Pure AFP protocol, no stat/AD |
| Create 2000 files | 4,522 | 2,628 | **1.7x** ⭐ | dir_modify avoids full rebuild |
| **Enumerate 2000 files** | **1,059** | **8** | **132x** ⭐ | AD cache serves all metadata |
| Delete 2000 files | 2,769 | 2,756 | 1.0x | IO-bound delete operations |
| Create 1000 dirs | 1,769 | 1,673 | 1.1x | IO-bound directory creation |
| **Cache hits (100d+1000f)** | **2,787** | **464** | **6.0x** ⭐ | AD cache hit ratio 99.8% |
| Mixed cache ops | 482 | 544 | 0.9x | Create/delete overhead |
| **Deep path traversal** | **813** | **146** | **5.6x** ⭐ | Cached stat eliminates traversal IO |
| **Cache validation 30k** | **7,438** | **1,262** | **5.9x** ⭐ | AD cache avoids 30k disk reads |
| **Total** | **26,023** | **11,779** | **2.2x** ⭐ | |
| **Avg per AFP op** | **0.323 ms** | **0.146 ms** | **2.2x** | |

## Syscall IO Comparison (AFPD syscr/syscw per test)

| Test | 4.4.1 Reads | cached Reads | 4.4.1 Writes | cached Writes | Read Δ |
|------|--------:|--------:|--------:|--------:|----------:|
| Open+stat+read 1000 files | 6,000 | 4,001 | 7,000 | 7,000 | **-33%** ⭐ |
| Write 100MB | 0 | 0 | 299 | 300 | — |
| Read 100MB | 100 | 100 | 100 | 100 | — |
| Lock/Unlock 10k | 0 | 0 | 20,000 | 20,000 | — |
| Create 2000 files | 2,000 | 0 | 10,000 | 10,000 | **-100%** ⭐ |
| **Enumerate 2000 files** | **1,960** | **0** | **49** | **49** | **-100%** ⭐ |
| Delete 2000 files | 4,000 | 4,000 | 4,000 | 8,000 | — |
| Create 1000 dirs | 0 | 0 | 4,440 | 5,550 | — |
| **Cache hits (100d+1000f)** | **10,000** | **0** | **11,100** | **11,000** | **-100%** ⭐ |
| Mixed cache ops | 820 | 400 | 1,620 | 2,020 | **-51%** ⭐ |
| **Deep path traversal** | **2,500** | **0** | **3,550** | **3,500** | **-100%** ⭐ |
| **Cache validation 30k** | **30,000** | **0** | **30,100** | **30,000** | **-100%** ⭐ |
| **Total** | **57,380** | **8,599** | **92,258** | **97,519** | **-85%** ⭐ |

### CNID IO Comparison

| Test | 4.4.1 CNID_R | cached CNID_R | 4.4.1 CNID_W | cached CNID_W |
|------|--------:|--------:|--------:|--------:|
| Open+stat+read 1000 files | 0 | 0 | 1 | 0 |
| Create 2000 files | 4 | 4 | 6,131 | 4,131 |
| Delete 2000 files | 4 | 4 | 6,179 | 6,179 |
| Create 1000 dirs | 2 | 2 | 2,277 | 2,277 |
| Cache hits | 0 | 0 | 100 | 0 |
| Mixed cache ops | 2 | 2 | 1,217 | 1,017 |
| Deep path traversal | 0 | 0 | 50 | 0 |
| Cache validation 30k | 0 | 0 | 100 | 0 |
| **Total** | **12** | **12** | **16,055** | **13,604** |

### Notable IO Findings

- **Total AFPD reads reduced 85%** (57,380 → 8,599) — AD metadata cache and optimized stat passthrough eliminate nearly all metadata read syscalls
- **Enumerate: 1,960 → 0 reads** — all metadata served entirely from AD cache with zero disk reads
- **Cache hits: 10,000 → 0 reads** — previously required AD metadata disk reads, now fully cached
- **Cache validation: 30,000 → 0 reads** — AD cache completely eliminates disk read overhead for repeated access
- **Deep traversal: 2,500 → 0 reads** — cached stat data eliminates filesystem traversal reads
- **Create 2000: 2,000 → 0 reads** — creation path no longer reads metadata during file creation
- **Open+stat+read: 6,000 → 4,001 reads (-33%)** — remaining reads are genuine data fork open/read operations
- **CNID writes reduced 15%** (16,055 → 13,604) — eliminated unnecessary hint-related CNID writes
- **AFPD writes increased 5.7%** (92,258 → 97,519) — IPC hint sends for cache coherency on delete/create operations
- **Reads per AFP op: 0.711 → 0.105** (6.7× fewer read syscalls per operation)

## Dircache Statistics Comparison

| Metric | 4.4.1 | cached-enumerate |
|--------|------:|------:|
| Cache hit rate | 96.3% | 97.4% |
| Total lookups | 2,510,304 | 3,619,264 |
| Total hits | 2,417,772 | 3,526,735 |
| invalid_on_use | **91,845** | **688** |
| Expunged | 690 | 688 |
| Evictions | 0 | 0 |
| Validation freq | 100 | 100 |

### New cached-enumerate Metrics

| Metric | Value |
|--------|------:|
| **AD cache hit ratio** | **99.8%** |
| AD cache hits | 1,040,071 |
| AD cache misses | 2,570 |
| ARC T1→T2 promotions | 91,841 |
| ARC T1 hits | 91,841 |
| ARC T2 hits | 3,457,744 |
| IPC hints sent | 272,762 |

## Key Takeaways

1. **Enumerate is 132x faster** with real disk volumes — AD metadata cache eliminates virtually all disk reads for metadata
2. **Overall 2.2x faster** across all operations — dominated by IO-bound create/delete which can't be cached
3. **Total reads reduced 85%** (57,380 → 8,599) — AD cache and optimized stat passthrough eliminate nearly all metadata read syscalls
4. **invalid_on_use dropped from 91,845 to 688** (133x reduction) — dir_modify in-place refresh prevents full entry destruction
5. **AD cache hit ratio 99.8%** — only 2,570 out of 1,042,641 metadata requests hit disk
6. **IPC hint system working** — 272,762 hints sent for cross-process cache coherency (single-user test, so 0 received)
7. **Read-heavy workloads benefit most** — enumerate, cache hits, deep traversal, and validation all show 5-132x improvements with zero read IO
8. **Write-heavy workloads see modest gains** — create files 1.7x faster, delete approximately same speed
9. **Reads per AFP op: 0.711 → 0.107** — 6.6× fewer read syscalls per AFP operation on average

### DIFF
--- a/etc/afpd/ad_cache.c
+++ b/etc/afpd/ad_cache.c
@@ -89,7 +89,7 @@ void ad_store_to_cache(struct adouble *adp, struct dir *cached)
  * Sets up the entry directory in adp so that ad_getattr(), ad_getdate(),
  * ad_entry() all work normally — downstream code is completely unchanged.
  *
- * PRECONDITION: ad_init() must have been called on adp to set up
+ * @pre ad_init() must have been called on adp to set up
  * ad_eid[] offsets and valid_data_len via ad_init_offsets().
  * As of E-006, ad_init() now calls ad_init_offsets() internally.
  */
@@ -158,7 +158,6 @@ int ad_metadata_cached(const char *name, int flags, struct adouble *adp,
             stp = &st_buf;
         }
 
-        /* Compare cached vs fresh to detect external changes */
         if (dir->dcache_ino != stp->st_ino || dir->dcache_ctime != stp->st_ctime) {
             /* External change detected — dir_modify handles coherence:
              * updates stat fields, refreshes CNID on inode change,
@@ -184,7 +183,17 @@ int ad_metadata_cached(const char *name, int flags, struct adouble *adp,
     }
 
     if (!strict && dir && dir->dcache_rlen >= 0) {
-        /* Non-strict: AD data populated — rebuild adouble from cache */
+        /* If caller provided a recent stat, validate ctime/inode. */
+        if (recent_st
+                && (dir->dcache_ctime != recent_st->st_ctime
+                    || dir->dcache_ino != recent_st->st_ino)) {
+            dir_modify(vol, dir, &(struct dir_modify_args) {
+                .flags = DCMOD_STAT,
+                .st = recent_st,
+            });
+            goto disk_read;
+        }
+
         adp->ad_adflags = flags;
         ad_rebuild_from_cache(adp, dir);
         ad_cache_hits++;
@@ -192,7 +201,19 @@ int ad_metadata_cached(const char *name, int flags, struct adouble *adp,
     }
 
     if (dir && dir->dcache_rlen == (off_t) -2) {
-        /* Confirmed no AD exists — return failure without syscall */
+        /* If caller provided a recent stat and ctime/inode changed,
+         * the AD file may have been created since — re-check on disk. */
+        if (recent_st
+                && (dir->dcache_ctime != recent_st->st_ctime
+                    || dir->dcache_ino != recent_st->st_ino)) {
+            dir_modify(vol, dir, &(struct dir_modify_args) {
+                .flags = DCMOD_STAT,
+                .st = recent_st,
+            });
+            goto disk_read;
+        }
+
+        /* Confirmed no AD exists and ctime unchanged — still no AD */
         ad_cache_no_ad++;
         errno = ENOENT;
         return -1;
--- a/etc/afpd/afp_dsi.c
+++ b/etc/afpd/afp_dsi.c
@@ -87,7 +87,6 @@ static void afp_dsi_close(AFPObj *obj)
     close(obj->ipc_fd);
     obj->ipc_fd = -1;
 
-    /* Close dircache hint pipe read end */
     if (obj->hint_fd >= 0) {
         close(obj->hint_fd);
         obj->hint_fd = -1;
@@ -560,11 +559,7 @@ void afp_over_dsi(AFPObj *obj)
              * buffer — if data is in DSI's application-level buffer,
              * poll() blocks even though dsi_stream_receive() would succeed. */
             if (dsi->start != dsi->eof) {
-                /* DSI buffer has data — drain hints then proceed to receive */
-                if (obj->hint_fd >= 0) {
-                    process_cache_hints(obj);
-                }
-
+                /* DSI buffer has data — proceed directly to receive */
                 break;
             }
 
@@ -574,7 +569,7 @@ void afp_over_dsi(AFPObj *obj)
                     process_cache_hints(obj);
                 }
 
-                /* Fall through to dsi_stream_receive() + existing state handling */
+                /* Fall through to dsi_stream_receive() + AFP state handling */
                 break;
             }
 
@@ -845,7 +840,6 @@ void afp_over_dsi(AFPObj *obj)
 
         pending_request(dsi);
         fce_pending_events(obj);
-        process_cache_hints(obj);
     }
 
     /* error */
--- a/etc/afpd/catsearch.c
+++ b/etc/afpd/catsearch.c
@@ -224,7 +224,8 @@ static struct adouble *adl_lkup(struct vol *vol, struct path *path,
                                  path->u_name, strnlen(path->u_name, CNID_MAX_PATH_LEN));
 
         if (ad_metadata_cached(path->u_name, (isdir ? ADFLAGS_DIR : 0),
-                               adp, vol, cached, false, NULL) < 0) {
+                               adp, vol, cached, false,
+                               (path->st_errno == 0) ? &path->st : NULL) < 0) {
             adp = NULL;
         }
 
--- a/etc/afpd/directory.c
+++ b/etc/afpd/directory.c
@@ -941,8 +941,8 @@ struct dir *dir_new(const char *m_name,
     dir->dcache_uid = st->st_uid;
     dir->dcache_gid = st->st_gid;
     dir->dcache_size = st->st_size;
-    dir->dcache_rlen = (off_t)
-                       - 1; /* Not yet loaded — triggers ad_metadata() on first access */
+    /* Not yet loaded — triggers ad_metadata() on first access */
+    dir->dcache_rlen = (off_t) - 1;
 
     if (!S_ISDIR(st->st_mode)) {
         dir->d_flags = DIRF_ISFILE;
@@ -1096,7 +1096,6 @@ int dir_modify(const struct vol *vol, struct dir *dir,
             dir->d_u_name = (strcmp(mname, args->new_uname) == 0)
                             ? dir->d_m_name : bfromcstr(args->new_uname);
 
-            /* Update UCS2 mac name */
             if (dir->d_m_name_ucs2) {
                 free(dir->d_m_name_ucs2);
                 dir->d_m_name_ucs2 = NULL;
@@ -2113,7 +2112,8 @@ int getdirparams(const AFPObj *obj,
                   (1 << DIRPBIT_FINFO))) {
         ad_init(&ad, vol);
 
-        if (!ad_metadata_cached(upath, ADFLAGS_DIR, &ad, vol, dir, false, NULL)) {
+        if (!ad_metadata_cached(upath, ADFLAGS_DIR, &ad, vol, dir, false,
+                                (s_path->st_errno == 0) ? &s_path->st : NULL)) {
             ad_available = 1;
             /* The ad_metadata_cached function handles ad_close() internally. */
         }
--- a/etc/afpd/enumerate.c
+++ b/etc/afpd/enumerate.c
@@ -376,8 +376,6 @@ static int enumerate(AFPObj *obj _U_, char *ibuf, size_t ibuflen _U_,
         struct dir *cached = dircache_search_by_name(vol, curdir, sd.sd_last, len);
 
         if (cached) {
-            /* Cache hit (includes ghosts) - use cached metadata, skip stat()! */
-            /* Note: dircache_search_by_name() already performed validation if needed */
             /* Reconstruct stat from cached fields */
             s_path.st.st_ino = cached->dcache_ino;
             s_path.st.st_ctime = cached->dcache_ctime;
@@ -389,13 +387,11 @@ static int enumerate(AFPObj *obj _U_, char *ibuf, size_t ibuflen _U_,
             /* Populate remaining system struct stat fields with safe defaults.
              * These are NOT used by AFP enumeration (getfilparams/getdirparams)
              * but must be initialized since s_path.st is a full struct stat */
-            s_path.st.st_dev =
-                0;        /* not cached; only needed by ad_convert CNID path */
-            s_path.st.st_nlink =
-                1;      /* safe default (dirs have >=2, but nlink is unused) */
-            s_path.st.st_rdev = 0;       /* is not a device file */
-            s_path.st.st_blocks = 0;     /* AFP uses st_size, not st_blocks */
-            s_path.st.st_blksize = 0;    /* not used by AFP */
+            s_path.st.st_dev = 0;      /* not cached; only needed by ad_convert */
+            s_path.st.st_nlink = 1;    /* dirs have >=2, but nlink is unused */
+            s_path.st.st_rdev = 0;     /* is not a device file */
+            s_path.st.st_blocks = 0;   /* AFP uses st_size, not st_blocks */
+            s_path.st.st_blksize = 0;  /* not used by AFP */
             s_path.st.st_atime = cached->dcache_mtime;  /* approximate; not used by AFP */
             s_path.st_valid = 1;
             s_path.st_errno = 0;
@@ -408,8 +404,7 @@ static int enumerate(AFPObj *obj _U_, char *ibuf, size_t ibuflen _U_,
         } else {
             /* Cache miss - fall back to original stat() behavior */
             if (of_stat(vol, &s_path) < 0) {
-                /* stat failed - entry doesn't exist or is inaccessible */
-                /* so the next time it won't try to stat it again
+                /* The next time it won't try to stat it again
                  * another solution would be to invalidate the cache with
                  * sd.sd_did = 0 but if it's not ENOENT error it will start again
                  */
--- a/distrib/docker/DOCKER.md
+++ b/distrib/docker/DOCKER.md
@@ -289,6 +289,7 @@ Set this environment variable to a specific value or string.
 | **Other Configuration** |                                                                              |
 | TZ                              | The [timezone](https://nodatime.org/TimeZones) to use in the container |
 | AFP_CONFIG_POLLING              | Poll afp.conf for changes every N seconds (for Webmin integration)    |
+| AFP_CONVERT_APPLEDOUBLE         | Convert AppleDouble files on the volume: *yes* (default) or *no*       |
 
 #### Boolean Type
 
--- a/doc/developer/dircache.md
+++ b/doc/developer/dircache.md
@@ -439,10 +439,10 @@ but performance is improved across virtually all workload patterns.
 
 #### ARC Cache Hits
 
-- **cache_hits**: T1/T2 cached entries returning immediately (fast)
+- **hits**: T1/T2 cached entries returning immediately (fast)
 - **ghost_hits**: B1/B2 ghost entries found and returned instantly with learning benefit (fast)
-- **total_hits**: Combined percentage of cache_hits + ghost_hits
-- **true_misses**: Complete misses requiring filesystem/CNID access (slower)
+- **total_hits**: Combined percentage of hits + ghost_hits
+- **misses**: Complete misses requiring filesystem/CNID access (slower)
 
 #### ARC Ghost Performance
 
@@ -468,10 +468,10 @@ but performance is improved across virtually all workload patterns.
 **Example ARC log output:**
 
 ```txt
-dircache statistics (ARC): (user: jdoe) cache_entries: 98234, ghost_entries: 32838,
-max_entries: 131072, config_max: 131072, lookups: 2458716, cache_hits: 1954327 (79.5%),
-ghost_hits: 322351 (13.1%), total_hits: (92.6%), true_misses: 182038 (7.4%),
-validations: ~21365 (0.9%), added: 152341, removed: 54107, expunged: 7234,
+dircache statistics (ARC): (user: jdoe) entries: 98234, ghost_entries: 32838,
+max_entries: 131072, config_max: 131072, lookups: 2458716, hits: 1954327 (79.5%),
+ghost_hits: 322351 (13.1%), total_hits: (92.6%), misses: 182038 (7.4%),
+validations: 21365 (0.9%), added: 152341, removed: 54107, expunged: 7234,
 invalid_on_use: 187, evicted: 53873, validation_freq: 100
 
 ARC ghost performance: ghost_hits: 322351 (13.1%), ghost_hits(B1=198423, B2=123928),
@@ -484,7 +484,7 @@ total_ghosts=32838, freq_bias: (66.6%)
 ARC adaptation: p=54823/131072 (41.8% target for T1), p_range=[28341,82934],
 adaptations=142387 (increases=76234, decreases=66153)
 
-ARC operations: cache_hits(T1=587234, T2=1367093), promotions(T1→T2=423871),
+ARC operations: hits(T1=587234, T2=1367093), promotions(T1→T2=423871),
 evictions(T1=28934, T2=24939)
 ```
 
--- a/doc/manual/Configuration.md
+++ b/doc/manual/Configuration.md
@@ -850,6 +850,71 @@ Default: lru.
 for "dircache statistics:" lines when afpd shuts down gracefully (user
 disconnects).
 
+## Interoperability with Other Network File Systems
+
+Netatalk and Samba can share the same volume and store Extended Attribute
+metadata in a mutually compatible format, enabling access
+via AFP or SMB.
+
+> ***WARNING:*** The **ea = samba** and **ea = sys** (Netatalk default) metadata
+> formats are **not** compatible, and there is currently no automated way to
+> convert between the two. If you have an existing Netatalk volume using
+> the default settings and want to add Samba sharing, the recommended
+> migration path is to copy files from one volume to another using an AFP
+> client.
+
+When sharing a volume with other processes (Samba, NFS, local applications),
+keep `dircache validation freq` at **1** (the default) so Netatalk
+detects external changes on every access. If Netatalk is the **only**
+process on the volume, you can set `dircache validation freq = 100` for
+maximum performance.
+
+### Netatalk configuration
+
+Use `ea = samba` to store Extended Attributes in the Samba-compatible format.
+
+    [Global]
+        vol preset = my default values
+        dircache validation freq = 1
+        ea = samba
+
+    [Homes]
+        basedir regex = /home
+
+    [Test Volume]
+        path = /export/test1
+
+    [My Time Machine Volume]
+        path = /export/timemachine
+        time machine = yes
+
+### Samba configuration
+
+The Samba **catia**, **fruit**, and **streams_xattr** VFS modules provide
+compatibility for Apple SMB clients and interoperability with
+Netatalk. Use **hide files** (not **veto files**) to hide Netatalk's
+invisible files from Windows clients.
+
+    [global]
+        ea support = yes
+        vfs objects = catia fruit streams_xattr
+
+        fruit:encoding = native
+        streams_xattr:prefix = user.
+        streams_xattr:store_stream_type = no
+
+        hide files = /.DS_Store/Network Trash Folder/TheFindByContentFolder/TheVolumeSettingsFolder/Temporary Items/.TemporaryItems/.VolumeIcon.icns/Icon?/.FBCIndex/.FBCLockFolder/
+
+        read only = no
+
+    [homes]
+
+    [Test Volume]
+        path = /export/test1
+
+    [My Time Machine Volume]
+        path = /export/timemachine
+
 ## Filesystem Change Events
 
 Netatalk includes a nifty filesystem change event (FCE) mechanism where
--- a/etc/afpd/acls.c
+++ b/etc/afpd/acls.c
@@ -2025,7 +2025,6 @@ int afp_setacl(AFPObj *obj, char *ibuf, size_t ibuflen _U_, char *rbuf _U_,
                     target_entry->d_rights_cache = 0xffffffff;
                 }
 
-                /* Send hint to afpd siblings — ACL/permissions changed */
                 cnid_t hint_cnid = target_entry ? target_entry->d_did : CNID_INVALID;
 
                 if (hint_cnid == CNID_INVALID && s_path->u_name && curdir) {
@@ -2086,7 +2085,6 @@ int afp_setacl(AFPObj *obj, char *ibuf, size_t ibuflen _U_, char *rbuf _U_,
                     target_entry->d_rights_cache = 0xffffffff;
                 }
 
-                /* Send hint to afpd siblings — ACL/permissions changed */
                 cnid_t hint_cnid2 = target_entry ? target_entry->d_did : CNID_INVALID;
 
                 if (hint_cnid2 == CNID_INVALID && s_path->u_name && curdir) {
--- a/etc/afpd/dircache.c
+++ b/etc/afpd/dircache.c
@@ -1977,9 +1977,9 @@ void log_dircache_stat(void)
                                    (double)dircache_stat.lookups) * 100.0 : 0.0;
         LOG(log_info, logtype_afpd,
             "dircache statistics (ARC): (user: %s) "
-            "cache_entries: %zu, ghost_entries: %zu, max_entries: %lu (%lu KB), config_max: %zu, "
-            "lookups: %llu, cache_hits: %llu (%.1f%%), ghost_hits: %llu (%.1f%%), total_hits: (%.1f%%), true_misses: %llu (%.1f%%), "
-            "validations: ~%llu (%.1f%%), "
+            "entries: %zu, ghost_entries: %zu, max_entries: %lu (%lu KB), config_max: %zu, "
+            "lookups: %llu, hits: %llu (%.1f%%), ghost_hits: %llu (%.1f%%), total_hits: (%.1f%%), misses: %llu (%.1f%%), "
+            "validations: %llu (%.1f%%), "
             "added: %llu, removed: %llu, expunged: %llu, invalid_on_use: %llu, "
             "evicted: %llu, validation_freq: %u",
             username,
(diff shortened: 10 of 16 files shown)


etc/afpd/afp_dsi.c
```diff
@@ -745,6 +845,8 @@ void afp_over_dsi(AFPObj *obj)
 
         pending_request(dsi);
         fce_pending_events(obj);
+        process_cache_hints(
+            obj);  /* Process cross-process dircache invalidation hints */
```
rdmark: pet peeve: please put a longer comment above the line of code so that it doesn't trigger an awkward line break by astyle

or maybe remove the comment as the code is kind of self-explanatory?

andylemin: Yea I have a bunch of clean up still to do

andylemin:
Hi @rdmark I started real life testing with clients over 5 and 6 GHz WiFi yesterday and the results are impressive!

Once the meta cache is warm, it responds surprisingly close to a local SSD! (I was restarting the local file browser to flush local caches, and also using other file browsers, and the speed up is very noticeable).

A large directory on my server (HDD) that usually takes 4-5 seconds to load now loads in ~1s, while a local SSD (without Netatalk) with the same test takes ~500ms. 

Removing the extra roundtrip times caused by metadata query recursion seems to have had a larger impact on remote TCP clients! Ie, These remote client improvements seem better than the server-local improvements which is interesting.

If this finding is shown to be correct I suspect it is probably due to the userland server process responding more quickly/consistently to TCP (less response jitter), which may be helping TCP calculations. I'll test and try to prove it..

Anyway it appears AD metadata caching is significant enough, so after this PR is polished and merged, I'll look at a Tier2 AD cache (SQLite/MySQL only) so the perf gains are maintained across logins/reboots etc.


Anyway I'll continue running user experience tests, and start running lantest across different commits (both locally and remotely) to build performance result tables.

You can also test this PR yourself now? 🙏

NJRoadfan:
This will need to be thoroughly tested on macOS. With FinderInfo being stored natively, there is a possibility that a server side change (ex: host OS modifies FinderInfo directly) can happen without afpd knowing. On that note, this can also be possible when running both Samba and Netatalk on the same share.

andylemin:
> This will need to be thoroughly tested on macOS. With FinderInfo being stored natively, there is a possibility that a server side change (ex: host OS modifies FinderInfo directly) can happen without afpd knowing. On that note, this can also be possible when running both Samba and Netatalk on the same share.

Hi @NJRoadfan Great question!
This should already be covered as the design of this PR is based around the POSIX specification that any changes should update ctime, therefore we cache and use ctime from stat to validate that the cache entry is up to date. If ctime or inode are different, the cache is refreshed from disk;

The POSIX rule governing `st_ctime` behavior is defined in **IEEE Std 1003.1-2017 (POSIX.1-2017)**, specifically:

1. **Section 4.9 "File Times Update"** in the Base Definitions volume — This section specifies the rules for when each of the three file timestamps (`st_atime`, `st_mtime`, `st_ctime`) are "marked for update." It states that `st_ctime` is marked for update by any operation that changes the file status (inode metadata etc).

2. **`<sys/stat.h>` header definition** — Defines `st_ctime` as the **"time of last status change"**, meaning any change to the inode data (as opposed to `st_mtime` which tracks file *content* modification).

The key language from Section 4.9 is that `st_ctime` shall be marked for update by functions that modify:
- File content (write also updates `st_ctime`, not just `st_mtime`)
- File permissions (`chmod`)
- File ownership (`chown`)
- Link count (`link`, `unlink`, `rename`)
- Extended attributes / any other inode metadata

This is the rule that the cached-enumerate branch relies on: since AppleDouble (AD) metadata for a file is stored alongside or as part of the file's state (hence native mode is also covered), any change to AD data will cause the filesystem to update `st_ctime`. The branch therefore uses `st_ctime` as a reliable indicator that "something changed" about a file (content or metadata), enabling efficient cache invalidation without needing to inspect every individual attribute.

**Canonical reference URL:** https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap04.html#tag_04_09

However we currently do NOT have a macos spectest workflow to prove this coverage, and we should.. https://github.com/Netatalk/netatalk/pull/2741 🙂

One of the new tests introduces a full AD FinderInfo change lifecycle where AD is set by user1, and then user2 updates it, and then user1 reads back the AD and verifies it gets the version set by user 2.

To cover the "external changes" use case (co-existing with Samba etc), the user just needs to leave `dircache validation freq=1`, which means that stat is always called and cached AD/stats are always validated 😁
The freq value is the only control the user needs to define how tolerant they are to external changes

PS; ZFS also stores all this data (EAs/ADs/ACLs) "natively" as well, and follows the same POSIX rules to update ctime when any are changed.

andylemin:
Depends on #2741 being merged first so these changes can run through the new macOS spectest workflow.

@rdmark this PR is ready for review.
Still need to do a final parse and clean up some temporary comments, and a few pet peeves still lying around.. sorry
But all of the core logic is complete.

I've been stress-testing with real life workloads for several days (my scripts have done over 4 million AFP operations as of this morning, and there is no further memory growth once the cache is full).

NB; There used to be a small mem leak, this appears confirmed as solved - I have been unable to crash the server (long term issue since early 3.x days).

Dircachesize set to 131072 (smaller than working set to overload the cache) resulting in just 189Kbytes virtual memory and 139Kbytes resident memory.

Successfully overloaded the DSI buffer - no crash, just advisory log reporting readahead buffer full with suggestion to increase dsireadbuf.

Successfully triggered a network Reconnect during active AFP operations, completed successfully without client userland noticing, and triggering AFP Replay Cache match on the server. No corruption on client.

These confirm the DSI over TCP read loop changes (to support event based cache hint signalling) are working correctly.

distrib/docker/entrypoint_netatalk.sh
```diff
@@ -278,6 +278,7 @@ path = /mnt/afpshare
 valid users = $AFP_VALIDUSERS1
 volume name = ${SHARE_NAME:-File Sharing}
 $AFP_RWRO = $AFP_VALIDUSERS1
+convert appledouble = ${AFP_CONVERT_APPLEDOUBLE:-yes}
```
rdmark, line 281: do you think we should document $AFP_CONVERT_APPLEDOUBLE in DOCKER.md?

etc/afpd/ad_cache.c
```diff
@@ -0,0 +1,223 @@
...
+    }
+}
+
+/*!
+ * @brief Populate struct adouble from struct dir cache fields
+ *
+ * Sets up the entry directory in adp so that ad_getattr(), ad_getdate(),
+ * ad_entry() all work normally — downstream code is completely unchanged.
+ *
+ * PRECONDITION: ad_init() must have been called on adp to set up
```
rdmark: doxygen has a `@pre` notation for preconditions! ;)

etc/afpd/catsearch.c
```diff
@@ -205,15 +206,29 @@ static struct adouble *adl_lkup(struct vol *vol, struct path *path,
     isdir  = S_ISDIR(path->st.st_mode);
 
     if (!isdir && (of = of_findname(vol, path))) {
+        /* Fork-owned adouble — use directly, skip cache.
+         * Opportunistically populate cache if not yet loaded. */
```
rdmark, line 210: note to self: it would be good to spin up an OS9 client and test that catsearch works!

etc/afpd/directory.c
```diff
@@ -929,6 +935,14 @@ struct dir *dir_new(const char *m_name,
...
     dir->dcache_ctime = st->st_ctime;
     dir->dcache_ino = st->st_ino;
+    /* Populate additional stat fields for enumerate optimization */
+    dir->dcache_mode = st->st_mode;
+    dir->dcache_mtime = st->st_mtime;
+    dir->dcache_uid = st->st_uid;
+    dir->dcache_gid = st->st_gid;
+    dir->dcache_size = st->st_size;
+    dir->dcache_rlen = (off_t)
+                       - 1; /* Not yet loaded — triggers ad_metadata() on first access */
```
rdmark: please put the comment above the line of code here :)

include/atalk/directory.h
```diff
@@ -53,7 +53,8 @@
 #define DIRF_CNID	   (1<<5) /*!< renumerate id */
 #define DIRF_ARC_GHOST (1<<6) /*!< ARC ghost entry (in B1/B2) */
 
-struct dir {
+struct dir { // NOSONAR (max 20 fields) — fields are intentionally grouped for cache locality
+    /* Fields requiring 8-byte alignment: pointers, time_t, ino_t, off_t */
```
rdmark, line 57: I have now turned off "Structures should not have too many fields" (c:S1820) in our quality profile -- these "not too many" of something rules are often a bit too opinionated, IMHO

test/testsuite/T2_FPOpenFork.c
```diff
@@ -1529,15 +1529,53 @@ STATIC void test431()
...
+    {
+        struct stat st_post;
+
+        if (stat(cmd, &st_post) == 0) {
+            unlink(cmd);
+            snprintf(cmd, sizeof(cmd), "%s/.AppleDouble", Path);
+            rmdir(cmd);
+            snprintf(cmd, sizeof(cmd), "%s/%s", Path, name);
+            unlink(cmd);
+            test_skipped(T_V2CONV);
```
rdmark, line 1574: note to self: I like this graceful skipping of the test if the precondition isn't met. We should arguably do the same for the extmap tests -- rather than failing when Classic Mac OS extmap configurations are disabled.

rdmark:
> If you are running Netatalk alongside other processes which change files outside of Netatalk (eg Samba), you must ensure dircache validation freq remains set to 1 (the default)

We should document this, perhaps in the Configuration manual chapter! Might be worth creating an "Interoperability with other network file systems" section in the chapter, and folding the information from https://github.com/Netatalk/netatalk/wiki/Interoperability-with-Samba -- alternatively keeping the wiki page and fleshing it out. Thoughts?

distrib/docker/entrypoint_netatalk.sh
andylemin: Absolutely.. Done

etc/afpd/ad_cache.c
andylemin: Nice, I didn't know about this. Done

etc/afpd/catsearch.c
andylemin: I don't have any old hardware to test things like this sadly

etc/afpd/directory.c
andylemin: Well spotted, I just fixed up lots of the comments but missed this one

include/atalk/directory.h
andylemin: Thank you! Yes, generally it is poor design to have large structs, but for the cache it's a requirement, and these are the minimum. You might have also noticed I re-aligned the struct (largest to smallest) which saved 12 wasted bytes of padding per-entry.

test/testsuite/T2_FPOpenFork.c
andylemin: I had a similar thought when I looked at how we are skipping others - I wanted to make the tests more dynamic based on the environment as I'm lazy 😉

andylemin:
> > If you are running Netatalk alongside other processes which change files outside of Netatalk (eg Samba), you must ensure dircache validation freq remains set to 1 (the default)
> 
> We should document this, perhaps in the Configuration manual chapter! Might be worth creating an "Interoperability with other network file systems" section in the chapter, and folding the information from https://github.com/Netatalk/netatalk/wiki/Interoperability-with-Samba -- alternatively keeping the wiki page and fleshing it out. Thoughts?

Hi @rdmark 
Rebased, cleaned up, and applied your feedback 🫡 Also added a new "Interoperability with other network file systems" section in the Configuration manual. Let me know what you think?

include/atalk/directory.h
rdmark: Nicely done! Fun times hand-packing structs. Shall I reactivate the rule with a higher limit, then? It's set to 20 but can be bumped to an arbitrary higher number.

test/testsuite/T2_FPOpenFork.c
rdmark: so that we don't forget this particular idea: https://github.com/Netatalk/netatalk/issues/2743

andylemin:
Hi @NJRoadfan @rdmark 
The macos spectest PR has been merged, and just rebased this.
I can see all 309 tests (10 new cache coherency and metadata round-trip tests) are passing on macos.
https://github.com/Netatalk/netatalk/actions/runs/22421606770/job/64920327399?pr=2733

Does this answer your questions NJRoadfan? regarding testing on macos and proving the "native" xattr storage on macos obeys the same POSIX rules?
Thanks

include/atalk/directory.h
andylemin: You can set it to 30. When designing this I investigated adding a second array for storing ResourceForks (to have complete metadata coverage) and that only needs a single pointer in the dircache struct. So 30 should give us more than enough headroom.

andylemin:
Hi. Just some final minor changes following the performance analysis/review - updated the last commit with the details rather than fixing up an older commit.

Also added the formal performance testing data to the summary. This is now finalised.

PS; I also ran the updated speedtest tool which shows some minor improvements, just for fun I also did a sweep of quantum size and found some interesting results - interesting enough for a wiki page showing how to test your own setup for optimal quantum and readahead buffer values 😉 Just need to write a script to make it easier to parse the resulting sweep of sweeps..

include/atalk/directory.h
rdmark: done!

etc/afpd/enumerate.c
```diff
@@ -370,23 +370,74 @@ static int enumerate(AFPObj *obj _U_, char *ibuf, size_t ibuflen _U_,
...
+            s_path.st.st_uid = cached->dcache_uid;
+            s_path.st.st_gid = cached->dcache_gid;
+            s_path.st.st_size = cached->dcache_size;
+            /* Populate remaining system struct stat fields with safe defaults.
+             * These are NOT used by AFP enumeration (getfilparams/getdirparams)
+             * but must be initialized since s_path.st is a full struct stat */
+            s_path.st.st_dev =
+                0;        /* not cached; only needed by ad_convert CNID path */
+            s_path.st.st_nlink =
+                1;      /* safe default (dirs have >=2, but nlink is unused) */
```
rdmark: spotted two more unwanted astyle induced line breaks; here you could split them out into two lines if you like to keep the neat column alignment, f.e.:

```c
            s_path.st.st_nlink = 1;      /* safe default (dirs have >=2,
                                            but nlink is unused) */
```

andylemin: Well spotted! Easy for these to creep in..

andylemin:
@rdmark The changes to the static analyser settings has re-fired the warnings for `typedef struct AFPObj` and `struct afp_options`. As we are using a struct for the afp_options, I think we can increase the threshold to 100 - and a few options can already be retired.

rdmark (approved):
the sonar rule has been updated!

PR #2734, opened 2026-02-22
turn 'convert appledouble' option off by default

rdmark:
the upgrade path from netatalk v2 to v4 is expected to be rare now, so this is a good time to disable the automatic appledouble conversion, since this does add some minor processing overhead for file operations

fixes one case of hard-coded call to ad_convert() in the getvolparams() function

### DIFF
--- a/etc/afpd/ad_cache.c
+++ b/etc/afpd/ad_cache.c
@@ -18,6 +18,7 @@
 
 #include <errno.h>
 #include <stdbool.h>
+#include <stdlib.h>
 #include <string.h>
 #include <sys/stat.h>
 


rdmark:
the failing tests are testing the appledouble conversion, so we need to deliberately enable it in the test environment now

the PR in https://github.com/Netatalk/netatalk/pull/2733 is already adding this, so I will hold off on the current PR

andylemin:
Hi @rdmark 
Yes I added code to skip the AD conversion spectest if 'convert appledouble' is not set.

I did this after I discovered (and fixed) the Enumerate code was not checking the same afp.conf option (AD conversion was always hard enabled).

I think a code search is needed to see where else AD conversion is happening and check it is using the config option?

rdmark:
@andylemin I found a single case of unconditional call to ad_convert, namely in the getvolparams function

andylemin:
> @andylemin I found a single case of unconditional call to ad_convert, namely in the getvolparams function

Thanks, forgot about this.
#2749

.github/workflows/containers.yml
```diff
@@ -305,6 +307,7 @@ jobs:
               -e AFP_EXTMAP=1 \
               -e AFP_DIRCACHESIZE=131072 \
               -e AFP_DIRCACHE_VALIDATION_FREQ=100 \
+              -e AFP_CONVERT_APPLEDOUBLE=yes \
```
andylemin: Should some of these containers be set to no? Setting them all to yes when our default is no seems interesting

rdmark: I was thinking the same thing on the way home from grocery shopping just now. Will mix it up!

rdmark:
I kept the conversion in four of the container test jobs!

andylemin (approved):

PR #2746, opened 2026-02-26
libatalk: log a warning when deprecated apf.conf option is being used

rdmark:
- flip the logic for dircache size option, so that we prefer the new
      format and warn when using the old format
- introduce a check for the previously remove zeroconf name option
      and log a warning here too
- favor the new dircache size option in the macOS CI job
- add section to Upgrade in the manual on changes to afp.conf options between v3 and v4


rdmark:
I interacted with a user recently who still had `zeroconf name` in their afp.conf despite using a recent v4 version, which inspired me to make this change

PR #2749, opened 2026-02-27
afpd: honor 'convert appledouble' config in getvolparams

andylemin:
ad_convert() no longer called unconditionally every time getvolparams() runs, checks 'convert appledouble = no'.


rdmark:
I have a slightly different fix to the same thing in https://github.com/Netatalk/netatalk/pull/2734 -- my version checks the return value and logs a warning if it fails. Which one is better you think?

andylemin:
> I have a slightly different fix to the same thing in https://github.com/Netatalk/netatalk/pull/2734 -- my version checks the return value and logs a warning if it fails. Which one is better you think?

Yours is better, so I merged it.

PR #2751, opened 2026-02-28
meson: consistently define install target tags

rdmark:
this creates a tag for all install targets in the meson build system, allowing for fine grained control of installations per file category


rdmark:
this got builds working on NetBSD again -- we were running into errors like this on NetBSD specifically

```
Found ninja-1.13.2 at /usr/pkg/bin/ninja
Failed to guess install tag for /usr/local/share/man/man1/afpldaptest.1
Failed to guess install tag for /usr/local/share/man/man1/afppasswd.1
Failed to guess install tag for /usr/local/share/man/man1/afpstats.1
Failed to guess install tag for /usr/local/share/man/man1/addump.1
Failed to guess install tag for /usr/local/share/man/man1/asip-status.1
...
```

PR #2756, opened 2026-03-02
afpd: use C11 atomic operations in dircache, conditional check for libatomic

rdmark:
replace the legacy sync built-ins that won't work on many 32 bit systems

add conditional check for atomic keywords in meson and link with libatomic if needed

based on patch by Sergey Fedorov


rdmark:
@barracuda156 can you take a look at this solution? if it works, feel free to steal it for your PR!

andylemin:
Thank you!

rdmark:
I was able to verify the fix on ARMv7 32bit (RPi Zero W)

PR #2767, opened 2026-03-05
[4.4] afpd: use C11 atomic operations in dircache, conditional check for libatomic

rdmark:
replace the legacy sync built-ins that won't work on many 32 bit systems

add conditional check for atomic keywords in meson and link with libatomic if needed

based on patch by Sergey Fedorov

### DIFF
--- a/etc/afpd/dircache.c
+++ b/etc/afpd/dircache.c
@@ -257,7 +257,7 @@ static unsigned long queue_count_max = 0;
 static int should_validate_cache_entry(void)
 {
     /* Thread-safe increment using compiler builtins */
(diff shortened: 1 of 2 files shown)


rdmark:
@andylemin after merging this I plan to tag a 4.4.2 patch release, since I think we have a critical mass of important bugfixes in the branch now. anything else you would like to see in the bugfix release?

andylemin:
Hi @rdmark only one legacy LRU cache fix which causes a crash when LRU is full and cache is thrashing (free'ing Q is too slow). I'll dig it out

PR #2769, opened 2026-03-06
rephrase container documentation to emphasize runtime agnostic operation

rdmark:
use podman consistently as the primary example, since this is what's available by default in open source distributions, while calling out other options

prefer managed volumes over bind mounts as the solution with long term consistency and less risk of trouble with permissions


rdmark:
I'm on the fence whether to rename distrib/docker to distrib/container or not... thoughts?

andylemin (approved):

PR #2775, opened 2026-03-10
afpd: synthesize virtual Icon\r file in volume root

rdmark:
Generate a virtual Icon\r file with ICN#, icl4, and icl8 resource fork data at the root of each shared volume, which is recognized by Classic Mac OS clients as a custom volume icon.

The file is entirely in-memory and served by intercepting FPEnumerate, FPGetFilDirParms, FPOpenFork, and FPRead. Write operations are rejected. The volume root's Finder info has hasCustomIcon set so the Finder knows to look for the file.

We use CNID 16 internally to represent the virtual icon file, which is within the reserved range of the CNID scheme.

This feature is effectively a workaround for the limitation in the AFP protocol of only being able to serve a monochrome ICN# icon resource to clients.

The virtual icon feature is enabled by the 'legacy icon' option in afp.conf, with the default behavior being unchanged that is serves the monochrome ICN# data over the wire only. At the same time, the 'legacy icon' option has been turned into a hybrid volume and global level option.

As a bonus, there is a brand new icon design included, called 'nas', special thanks to techknight for the idea. In addition, the default icon has been tweaked and renamed to 'fileserver' and can now be explicitly configured as a legacy icon.


andylemin:
Hi @rdmark. Here is some feedback.

Memory Leak in `afp_openfork` Virtual Path — `vad` and `vad->ad_name`

**File:** [`fork.c`](etc/afpd/fork.c:400) (virtual open path)

When `of_alloc()` succeeds, a `malloc`'d `struct adouble *vad` and its `strdup`'d `ad_name` are handed to the ofork. However, in [`of_closefork`](etc/afpd/ofork.c:456), the virtual path calls `of_dealloc(ofork)` which frees the ofork struct, but does **not** free `ofork->of_ad` or `ofork->of_ad->ad_name`.

For normal forks, `ad_close()` handles cleanup. For virtual forks, those are skipped. The `vad` and `vad->ad_name` allocated in `afp_openfork` are leaked on every virtual fork close.

Memory leak per open/close cycle. On a busy Classic Mac, the Finder may open `Icon\r` repeatedly.

**Suggested fix:** In the virtual fork close path in `of_closefork`, free `ofork->of_ad->ad_name` and `ofork->of_ad` before calling `of_dealloc`:

```c
if (ofork->of_flags & AFPFORK_VIRTUAL) {
    free(ofork->of_ad->ad_name);
    free(ofork->of_ad);
    of_dealloc(ofork);
    return 0;
}
```

andylemin:
`real_icon_exists()` Uses Process CWD, Not Volume Root

**File:** [`virtual_icon.c`](etc/afpd/virtual_icon.c:248)

```c
int real_icon_exists(void)
{
    struct stat st;
    return (lstat(VIRTUAL_ICON_NAME, &st) == 0);
}
```

This does `lstat("Icon\r", &st)` relative to the **process global current working directory** of the afpd process, not relative to the volume root. The CWD of afpd changes depending on which directory the client is currently navigating via movecwd(). This function is called from `directory.c` (getdirparams), `enumerate.c`, and `filedir.c` — in some of those contexts, `curdir` has been set and `afp_movecwd` has been called, but in others (like `getdirparams` for the root), CWD may or may not be the volume root.

The check may produce false negatives (virtual icon shown even when a real one exists) or false positives (virtual icon suppressed in wrong directory), depending on timing and CWD state.

**Suggested fix:** Accept the volume as a parameter and construct the full path:

```c
int real_icon_exists(const struct vol *vol)
{
    char path[MAXPATHLEN + 1];
    struct stat st;
    snprintf(path, sizeof(path), "%s/%s", vol->v_path, VIRTUAL_ICON_NAME);
    return (lstat(path, &st) == 0);
}
```
All callers need updating to pass `vol`.

andylemin:
`of_closefork` Auto-Delete Relies on Unreachable `forkpath`

**File:** [`ofork.c`](etc/afpd/ofork.c:555) (the new auto-delete block)

The auto-delete logic for emptied `Icon\r` files checks `&& forkpath` as a guard, but `forkpath` is set from `bformat("%s/%s", bdata(dir->d_fullpath), of_name(ofork))` which requires `dir` to be non-NULL. When `dir` is NULL (dirlookup failure), `forkpath` is NULL and the block is skipped — that's fine. But there's also a use of `cached` inside the block:

```c
if (cached) {
    dircache_remove(ofork->of_vol, cached, DIRCACHE | DIDNAME_INDEX | QUEUE_INDEX);
    cached = NULL;
}
```

The variable `cached` is set earlier in the function based on `AFPFORK_DIRTY | AFPFORK_MODIFIED` flags. But after the new code sets `cached = NULL`, the existing post-close cache refresh code further down also checks `cached`:

```c
if (cached) {
    struct stat post_st;
    ...
}
```

Setting `cached = NULL` in the delete_icon path prevents the stale-use, which is correct, but the comment `/* prevent stale use below */` should be more explicit about what it's preventing.

Logic is correct just fragile.. Future refactors could easily introduce a stale pointer bug.

andylemin:
`EMPTY_RFORK_HEADER_SIZE` Magic Number

**File:** [`ofork.c`](etc/afpd/ofork.c:558)

```c
#define EMPTY_RFORK_HEADER_SIZE 286
```

This is defined inside the function body, which is unusual? More importantly, the value 286 appears to be the size of an empty resource map header, but the actual Mac empty resource fork is 286 bytes only under specific circumstances. Some Finder versions may produce slightly different sizes. The heuristic `ad_rlen <= EMPTY_RFORK_HEADER_SIZE` could be too aggressive — it might delete a file that actually has small but valid custom icon data.

Should probably move to virtual_icon.h with a comment explaining the derivation. And consider explicitly checking if the resource fork actually has zero resources rather than relying on size thresholds.

andylemin:
Enumerate Virtual Entry Injection — Off-by-One Risk

**File:** [`enumerate.c`](etc/afpd/enumerate.c:556) (the injection block)

The condition:

```c
sindex + actcnt == curdir->d_offcnt + 1
```

relies on `d_offcnt` accurately reflecting the real directory entry count. If `d_offcnt` is stale (e.g., another client added a file between enumerations), the virtual entry could be emitted at the wrong position or skipped entirely. This is mitigated by `sd_sindex` tracking, but the interaction with stale directory caches is subtle and could do with some comments at least.

ITs unlikely to cause data corruption, but it could cause the virtual icon to not appear in some race conditions.

andylemin:
Write-Open Materialisation Creates File Without Finder Info

**File:** [`fork.c`](etc/afpd/fork.c:370) (write access materialization)

When write access is requested, the code creates a real empty file and seeds the resource fork xattr, but does **not** set up the AppleDouble/Finder info metadata (type=`icon`, creator=`MACS`, invisible flag). The subsequent `break` falls through to the normal open path which may or may not populate Finder info correctly.

The Finder may not recognise the materialised file correctly until it writes its own Finder info.

andylemin:
Type Narrowing: `cc` and `err` Changed from `ssize_t` to `int`

**File:** [`fork.c`](etc/afpd/fork.c:1063)

```c
-    ssize_t      cc;
-    ssize_t      err;
+    int          cc;
+    int          err;
```

On LP64 systems, `ssize_t` is 64-bit. Changing `cc` and `err` to `int` could truncate values if any codepath passes values > INT_MAX. The `cc` variable held return values from `dsi_readinit()`/`dsi_read()` which return `ssize_t`. A new variable `dsi_cc` of type `ssize_t` was introduced for those, but `cc` is still used elsewhere in error paths and `err` holds AFP error codes (which fit in int), so this is probably fine — but the change is unnecessary?

andylemin:
Icon Data in `icon.h` — ~25KB of Static Data in a Header

**File:** [`icon.h`](etc/afpd/icon.h:257)

The PR adds approximately 1,400 lines of static byte arrays to `icon.h`. Since this header is included by both `virtual_icon.c` and `status.c`, all these arrays are compiled into every translation unit that includes `icon.h`. The arrays are `static`, so duplicate copies exist in each object file.

**Suggestion:** Move the new icl4/icl8 arrays into `virtual_icon.c` (or a separate `icon_data.c`) and expose them via `extern` declarations. This saves ~25KB of duplicated static data per translation unit.

andylemin:
Missing `#include <limits.h>` Purpose

**File:** [`enumerate.c`](etc/afpd/enumerate.c:11)

The PR adds `#include <limits.h>` but the only new usage of limits constants is `UINT16_MAX` and `UINT8_MAX` in the entry length overflow check. These constants are actually defined in `<stdint.h>` which is already included transitively. The include is harmless but potentially unnecessary.

rdmark:
@andylemin after sleeping on this, I am leaning towards making the feature opt-in through the `legacy icon` option. The argument being that not all AFP clients filter the `Icon\r` file - afpfs-ng doesn't for instance (although we can certainly fix that). Most users who aren't vintage Mac users would not want a random immutable 0 byte file in their AFP share root.

Thoughts?

rdmark:
> Hi @rdmark. Here is some feedback.
> 
> Memory Leak in `afp_openfork` Virtual Path — `vad` and `vad->ad_name`
> 
> **File:** [`fork.c`](etc/afpd/fork.c:400) (virtual open path)
> 
> When `of_alloc()` succeeds, a `malloc`'d `struct adouble *vad` and its `strdup`'d `ad_name` are handed to the ofork. However, in [`of_closefork`](etc/afpd/ofork.c:456), the virtual path calls `of_dealloc(ofork)` which frees the ofork struct, but does **not** free `ofork->of_ad` or `ofork->of_ad->ad_name`.
> 
> For normal forks, `ad_close()` handles cleanup. For virtual forks, those are skipped. The `vad` and `vad->ad_name` allocated in `afp_openfork` are leaked on every virtual fork close.
> 
> Memory leak per open/close cycle. On a busy Classic Mac, the Finder may open `Icon\r` repeatedly.
> 
> **Suggested fix:** In the virtual fork close path in `of_closefork`, free `ofork->of_ad->ad_name` and `ofork->of_ad` before calling `of_dealloc`:
> 
> ```c
> if (ofork->of_flags & AFPFORK_VIRTUAL) {
>     free(ofork->of_ad->ad_name);
>     free(ofork->of_ad);
>     of_dealloc(ofork);
>     return 0;
> }
> ```

Did you see the free logic in of_dealloc()? In the conditional on this line:

https://github.com/Netatalk/netatalk/blob/60a50c47e50ed80bd979b9b5a01198a726962ee1/etc/afpd/ofork.c#L440-L443

There is only ever going to be a single virtual fork, so the path that I could trace looks like:

1. afp_openfork (fork.c:418) mallocs a vad, calls ad_init(vad, vol) which does memset(ad, 0, ...) — so ad_refcount starts at 0.
2. of_alloc (ofork.c:220-246) receives this vad as the ad parameter. Since ad is non-NULL, it takes the else branch at line 242 and calls ad_ref(ad), incrementing ad_refcount to 1.
4. of_dealloc (ofork.c:439) decrements: ad_refcount-- → back to 0.
6. The condition ad_refcount <= 0 is true (0 ≤ 0), so the free happens. 

Do you by any chance see another path here where the refcount conditional wouldn't trigger in the end?

rdmark:
> `of_closefork` Auto-Delete Relies on Unreachable `forkpath`
> 
> **File:** [`ofork.c`](etc/afpd/ofork.c:555) (the new auto-delete block)
> 
> The auto-delete logic for emptied `Icon\r` files checks `&& forkpath` as a guard, but `forkpath` is set from `bformat("%s/%s", bdata(dir->d_fullpath), of_name(ofork))` which requires `dir` to be non-NULL. When `dir` is NULL (dirlookup failure), `forkpath` is NULL and the block is skipped — that's fine. But there's also a use of `cached` inside the block:
> 
> ```c
> if (cached) {
>     dircache_remove(ofork->of_vol, cached, DIRCACHE | DIDNAME_INDEX | QUEUE_INDEX);
>     cached = NULL;
> }
> ```
> 
> The variable `cached` is set earlier in the function based on `AFPFORK_DIRTY | AFPFORK_MODIFIED` flags. But after the new code sets `cached = NULL`, the existing post-close cache refresh code further down also checks `cached`:
> 
> ```c
> if (cached) {
>     struct stat post_st;
>     ...
> }
> ```
> 
> Setting `cached = NULL` in the delete_icon path prevents the stale-use, which is correct, but the comment `/* prevent stale use below */` should be more explicit about what it's preventing.
> 
> Logic is correct just fragile.. Future refactors could easily introduce a stale pointer bug.

fair points, I can see the potential for future fragility, however I suggest we keep the logic for now improve the code comment to be more precise

rdmark:
> `EMPTY_RFORK_HEADER_SIZE` Magic Number
> 
> **File:** [`ofork.c`](etc/afpd/ofork.c:558)
> 
> ```c
> #define EMPTY_RFORK_HEADER_SIZE 286
> ```
> 
> This is defined inside the function body, which is unusual? More importantly, the value 286 appears to be the size of an empty resource map header, but the actual Mac empty resource fork is 286 bytes only under specific circumstances. Some Finder versions may produce slightly different sizes. The heuristic `ad_rlen <= EMPTY_RFORK_HEADER_SIZE` could be too aggressive — it might delete a file that actually has small but valid custom icon data.
> 
> Should probably move to virtual_icon.h with a comment explaining the derivation. And consider explicitly checking if the resource fork actually has zero resources rather than relying on size thresholds.

good feedback -- I am updating the logic to scan for resource fork data rather than relying on the magic file size number

rdmark:
> Enumerate Virtual Entry Injection — Off-by-One Risk
> 
> **File:** [`enumerate.c`](etc/afpd/enumerate.c:556) (the injection block)
> 
> The condition:
> 
> ```c
> sindex + actcnt == curdir->d_offcnt + 1
> ```
> 
> relies on `d_offcnt` accurately reflecting the real directory entry count. If `d_offcnt` is stale (e.g., another client added a file between enumerations), the virtual entry could be emitted at the wrong position or skipped entirely. This is mitigated by `sd_sindex` tracking, but the interaction with stale directory caches is subtle and could do with some comments at least.
> 
> ITs unlikely to cause data corruption, but it could cause the virtual icon to not appear in some race conditions.

Do you think this is worth a deeper improvement to the logic here to make the offset count more reliable? I asked Claude and got this proposal:

Capture d_offcnt in the savedir struct at buffer-read time, so the injection check uses a value that's consistent with the buffer contents.

1. Add sd_offcnt to struct savedir
2. Set it at line 319 when the buffer is populated
3. Adjust it at line 415 (stat failure path, same as d_offcnt--)
4. Use sd.sd_offcnt instead of curdir->d_offcnt at line 575

rdmark:
> Write-Open Materialisation Creates File Without Finder Info
> 
> **File:** [`fork.c`](etc/afpd/fork.c:370) (write access materialization)
> 
> When write access is requested, the code creates a real empty file and seeds the resource fork xattr, but does **not** set up the AppleDouble/Finder info metadata (type=`icon`, creator=`MACS`, invisible flag). The subsequent `break` falls through to the normal open path which may or may not populate Finder info correctly.
> 
> The Finder may not recognise the materialised file correctly until it writes its own Finder info.

I am no longer confident that icon/MACS is correct type/creator for the Icon files. Did you by any chance have a reference that confirms this is correct as per some reliable resource?

Empirically, Mac OS 7.6.1 creates Icon files with no type/creator, but with the invisible Finder flag. This is on both AFP volume root as well as when assigning icons to folders underneath the root.

```sh
nad ls -al afpicon2
 ---------v- ------ --- ---- ----          22   Icon
```

rdmark:
> Type Narrowing: `cc` and `err` Changed from `ssize_t` to `int`
> 
> **File:** [`fork.c`](etc/afpd/fork.c:1063)
> 
> ```c
> -    ssize_t      cc;
> -    ssize_t      err;
> +    int          cc;
> +    int          err;
> ```
> 
> On LP64 systems, `ssize_t` is 64-bit. Changing `cc` and `err` to `int` could truncate values if any codepath passes values > INT_MAX. The `cc` variable held return values from `dsi_readinit()`/`dsi_read()` which return `ssize_t`. A new variable `dsi_cc` of type `ssize_t` was introduced for those, but `cc` is still used elsewhere in error paths and `err` holds AFP error codes (which fit in int), so this is probably fine — but the change is unnecessary?

My intention with this change was to avoid type conversion, since the function itself has return type int and both cc and err can be used as the return value (which in itself is a bit messy and could arguably be refactored.)

Happy to revert the change and do explicit type conversion when returned instead.

rdmark:
Note that I moved the developer readme from this PR to a wiki page, since it's arguably very very niche and not likely the concern of most developers

https://github.com/Netatalk/netatalk/wiki/Virtual-Icon

Also, all icon data is now in a new icon.c source file -- both new and old! Much cleaner code if I may say so myself.

The `legacy icon` option has been turned into a volume level option with a fallback to read from the global section, which meant that the vol structure now stores the icon data rather than a global variable.

Finally, to commemorate this feature I drew a brand new `nas` icon, meant to depict a NAS unit with three drive bays (thanks to *techknight* for the idea and inspiration), while the `tcp` icon has been renamed to `fileserver` and redrawn slightly to remove the tiny globe that was cluttering up the design (a remnant from netatalk 2.x where afpd could distinguish between TCP and AppleTalk connections when mounting a volume.)

Screenshot 2026-03-15 at 17 40 45

NJRoadfan:
I think its time that the entire suite of color icons is posted.

Regarding FinderInfo: @andylemin I noticed that most versions of Finder will not attempt to write any FinderInfo (including flags and labels) to files that have null values for filetype/creator. Natively, Macintosh applications generally won't create files with NULL fields.

andylemin:
Hi @rdmark sorry for the code feedback in comments. Was in a rush the other day and didn't have time to add them to the relevant places in the commit. I won't have much time this week but will try and do another full review this weekend.

PS; I definitely agree with tieing this into an existing config option. I don't think I would want it for modern macs, and I agree with your comment "not all clients will support it".

andylemin:
@NJRoadfan interesting observation!

andylemin:
@rdmark
Just run another review, looking better now 🙂
Re; Memory leak - I agree I missed of_dealloc() freeing the other. can resolve comment.
VIRTUAL_ICON_RFORK_MAXLEN defined but never used.
Otherwise looking good.

Nitpik, could you please include some mermaid diagrams for this design to help with the high level architectural understanding? I think adding diagrams like on #2783 makes it easier to digest and highlight which functions/areas to scrutinise when we have large features?

andylemin:
> > Enumerate Virtual Entry Injection — Off-by-One Risk
> > **File:** [`enumerate.c`](etc/afpd/enumerate.c:556) (the injection block)
> > The condition:
> > ```c
> > sindex + actcnt == curdir->d_offcnt + 1
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > relies on `d_offcnt` accurately reflecting the real directory entry count. If `d_offcnt` is stale (e.g., another client added a file between enumerations), the virtual entry could be emitted at the wrong position or skipped entirely. This is mitigated by `sd_sindex` tracking, but the interaction with stale directory caches is subtle and could do with some comments at least.
> > ITs unlikely to cause data corruption, but it could cause the virtual icon to not appear in some race conditions.
> 
> Do you think this is worth a deeper improvement to the logic here to make the offset count more reliable? I asked Claude and got this proposal:
> 
> Capture d_offcnt in the savedir struct at buffer-read time, so the injection check uses a value that's consistent with the buffer contents.
> 
> 1. Add sd_offcnt to struct savedir
> 2. Set it at line 319 when the buffer is populated
> 3. Adjust it at line 415 (stat failure path, same as d_offcnt--)
> 4. Use sd.sd_offcnt instead of curdir->d_offcnt at line 575

After thinking about this, I think it should never happen as real_icon_exists() is called on every FPEnumerate of the volume root. The window for this issue to occur is incredibly small its not worth the overhead.

andylemin:
> icon/MACS

I agree. As @NJRoadfan says, I have confirmed the exact same thing. There is no filetype/creator on Icon\r when set with Ventura (all I have to hand to check). But it seems to no matter.

rdmark:
> @rdmark Just run another review, looking better now 🙂 Re; Memory leak - I agree I missed of_dealloc() freeing the other. can resolve comment. VIRTUAL_ICON_RFORK_MAXLEN defined but never used. Otherwise looking good.
> 

orphaned macro removed, thanks!

> Nitpik, could you please include some mermaid diagrams for this design to help with the high level architectural understanding? I think adding diagrams like on #2783 makes it easier to digest and highlight which functions/areas to scrutinise when we have large features?

Is this in line with what you are looking for? I got a pretty good sense of the logic flows! :)

https://github.com/Netatalk/netatalk/wiki/Dev-Docs-Virtual-Icon#virtual-icon-diagrams

rdmark:
> I think its time that the entire suite of color icons is posted.

Screenshot 2026-03-16 at 21 28 27

andylemin:
> > @rdmark Just run another review, looking better now 🙂 Re; Memory leak - I agree I missed of_dealloc() freeing the other. can resolve comment. VIRTUAL_ICON_RFORK_MAXLEN defined but never used. Otherwise looking good.
> 
> > 
> 
> 
> 
> orphaned macro removed, thanks!
> 
> 
> 
> > Nitpik, could you please include some mermaid diagrams for this design to help with the high level architectural understanding? I think adding diagrams like on #2783 makes it easier to digest and highlight which functions/areas to scrutinise when we have large features?
> 
> 
> 
> Is this in line with what you are looking for? I got a pretty good sense of the logic flows! :)
> 
> 
> 
> https://github.com/Netatalk/netatalk/wiki/Dev-Docs-Virtual-Icon#virtual-icon-diagrams

This looks great! I was mostly asking to help me better understand any collisions between this PR and mine 😉

andylemin (approved):

PR #2779, opened 2026-03-13
[4.4] dircache: fix LRU cache eviction crash and dangling pointers

rdmark:
Move cache eviction to after duplicate removal but before hash_alloc_insert, preventing overflow crashes when the LRU cache is full and no duplicates exist to free slots. Use >= instead of == for the fullness check as a defensive guard.

Clear qidx_node after all dequeue operations to prevent dangling pointer use-after-free. Guard against NULL qidx_node in dircache_remove to prevent crash on orphan entries.

### DIFF
--- a/etc/afpd/dircache.c
+++ b/etc/afpd/dircache.c
@@ -370,6 +370,7 @@ static void dircache_evict(void)
         }
 
         queue_count--;
+        dir->qidx_node = NULL;  /* Clear queue pointer after dequeue */
 
         if (curdir == dir) {                          /* 2 */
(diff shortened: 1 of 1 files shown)


rdmark:
@andylemin are these two LRU fixes that you meant to backport from main to 4.4?

andylemin:
Yes, these are indeed the ones. The ordering one was the most easy to trigger when thrashing the legacy LRU cache.
Thanks for picking this up, I have not had time

rdmark:
great, then I feel we should be in good shape for a bugfix release shortly!

PR #2781, opened 2026-03-14
ci: remove Spotlight dependencies from DragonFlyBSD job

rdmark:
tracker3 has fallen off the DragonFlyBSD ports repository, so let's remove the Spotlight stack for now


rdmark:
this should unblock the pipeline for the other jobs!

I don't see any sign that DragonFlyBSD has a localsearch port lined up

PR #2783, opened 2026-03-15
dircache: Added Tier-2 ResourceFork caching framework with expanded rfork spectests, and removed obsoleted 'dircache files'

andylemin:
Added Tier 2 resource fork data caching support to the dircache. 
Small resource forks (e.g. Icon\r custom icons) are cached in memory after first read, eliminating repeated open/pread/close syscalls. Memory-budgeted with dedicated rfork LRU eviction. Controlled by two new afp.conf options: 'dircache rfork budget' for total rfork cache per afpd process (KB, default 0=disabled) and 'dircache rfork maxsize' (KB per entry, default 1024).

Enhanced test544 with deterministic write patterns (0xAA/0xBB) and memcmp content verification. Added test546: multi-user resource fork lifecycle testing write, overwrite, and truncation with cross-connection content verification via Conn2.

Removed the now obsolete 'dircache files' config option — dirlookup always returns file entries from the cache if requested.

Updated man pages, Configuration.md, dircache.md, Docker entrypoint, CONTAINERS.md, and Webmin module to reflect the new rfork options and removal of 'dircache files'.

All 311 spectests pass (1 new resource fork full lifecycle test).

This flowchart shows the complete decision tree for every FPRead on a resource fork,
from cache lookup through population, serving, and fallback to disk I/O:

WARNING: This blocked on #2775. PR created for testing in the meantime whilst that matures.

---

Resource Fork Cache Data Flow

This flowchart shows the complete decision tree for every FPRead on a resource fork,
from cache lookup through population, serving, and fallback to disk I/O:

```mermaid
flowchart TD
    START([FPRead on resource fork]) --> CHK_ENABLED{rfork caching<br/>enabled?<br/>dircache_rfork_budget > 0}
    CHK_ENABLED -->|No: budget=0| FALLBACK[Fall through to<br/>normal ad_read/sendfile]
    CHK_ENABLED -->|Yes| CHK_WRITE{Fork opened<br/>for write?}
    CHK_WRITE -->|Yes: AFPFORK_ACCWR| FALLBACK
    CHK_WRITE -->|No: read-only| LOOKUP[Resolve dircache entry:<br/>dirlookup parent did<br/>+ dircache_search_by_name<br/>using strnlen for safety]

    LOOKUP --> CHK_FOUND{cached_entry<br/>found?}
    CHK_FOUND -->|No: not in dircache| FALLBACK
    CHK_FOUND -->|Yes| CHK_BUF{dcache_rfork_buf != NULL<br/>AND dcache_rlen > 0<br/>AND dcache_rlen == ad_size?}

    CHK_BUF -->|Yes: cache HIT| TMPLOCK[ad_tmplock if<br/>OPTION_AFP_READ_LOCK]
    TMPLOCK --> SERVE[rfork_cache_serve_from_buf<br/>writes DSI header + data<br/>from cache at offset]
    SERVE --> PROMOTE[Promote in<br/>rfork LRU]
    PROMOTE --> DONE([Return data<br/>to client])

    CHK_BUF -->|No: cache MISS| DISK_READ[Normal disk read:<br/>ad_read / sendfile]
    DISK_READ --> CHK_POPULATE{Post-read: entry exists<br/>AND buf == NULL<br/>AND rlen > 0<br/>AND rlen <= maxentry?}
    CHK_POPULATE -->|No: not cacheable| DONE
    CHK_POPULATE -->|Yes: cacheable| BUDGET{rfork_cache_used<br/>+ rlen <= budget?}
    BUDGET -->|No| EVICT[Evict from rfork<br/>LRU head until<br/>under budget]
    EVICT --> BUDGET
    BUDGET -->|Yes: space available| POPULATE[rfork_cache_store_from_fd:<br/>malloc rlen bytes<br/>ad_read entire fork<br/>into buffer]

    POPULATE --> CHK_READ{ad_read returned<br/>exactly rlen bytes?}
    CHK_READ -->|No: short read| INVAL[Invalidate ALL<br/>AD metadata:<br/>rlen=-1, clear FI/FD/AFI]
    INVAL --> DONE
    CHK_READ -->|Yes| STORE[Store in entry:<br/>dcache_rfork_buf = buf<br/>enqueue to rfork LRU tail<br/>rfork_cache_used += rlen]
    CHK_READ -->|enqueue malloc fail| CLEANUP[free buf, revert budget]
    CLEANUP --> DONE
    STORE --> DONE

    FALLBACK --> DISK_FALLBACK[ad_read / sendfile<br/>from fd]
    DISK_FALLBACK --> DONE

    style START fill:#1a5a1a,stroke:#4CAF50,color:#fff
    style DONE fill:#1a5a1a,stroke:#4CAF50,color:#fff
    style SERVE fill:#1a3a5c,stroke:#2196F3,color:#fff
    style STORE fill:#5a1a5a,stroke:#9C27B0,color:#fff
    style EVICT fill:#5a3a1a,stroke:#FF9800,color:#fff
    style FALLBACK fill:#3a3a3a,stroke:#9E9E9E,color:#fff
    style DISK_FALLBACK fill:#3a3a3a,stroke:#9E9E9E,color:#fff
    style DISK_READ fill:#3a3a3a,stroke:#9E9E9E,color:#fff
    style INVAL fill:#5a1a1a,stroke:#f44336,color:#fff
    style CLEANUP fill:#5a1a1a,stroke:#f44336,color:#fff
```

---

Resource Fork Cache Lifecycle State Machine

Every `struct dir` entry's rfork cache state follows this state machine. Each transition
maps to a specific function and is verified against the source tree.

```mermaid
stateDiagram-v2
    state "EMPTY<br/>buf=NULL, rlen=-1 or -2" as S0
    state "POPULATED<br/>buf=malloc, rlen>=0, in LRU" as S1
    state "HIT-SERVING<br/>same as POPULATED, promoted" as S2
    state "EVICTED-BUDGET<br/>buf=NULL, rlen>=0 valid" as S3
    state "INVALIDATED<br/>buf=NULL, rlen=-1" as S4
    state "FREED<br/>struct dir deallocated" as S5

    [*] --> S0 : dir_new via calloc
    S0 --> S1 : store_from_fd - malloc + ad_read + LRU add
    S0 --> S0 : miss - too large or budget full
    S1 --> S2 : cache hit - memcpy + LRU promote
    S2 --> S2 : further FPRead hits
    S1 --> S3 : evict_to_budget - free buf keep rlen
    S2 --> S3 : evict_to_budget - free buf keep rlen
    S3 --> S1 : next FPRead re-populates
    S1 --> S4 : DCMOD invalidation - free buf reset rlen
    S2 --> S4 : DCMOD invalidation - free buf reset rlen
    S4 --> S1 : next FPRead after Tier 1 re-read
    S1 --> S5 : dir_free - ARC eviction or dir_remove
    S2 --> S5 : dir_free - ARC eviction or dir_remove
    S3 --> S5 : dir_free - ARC eviction or dir_remove
    S4 --> S5 : dir_free - ARC eviction or dir_remove
    S0 --> S5 : dir_free - never cached
    S5 --> [*]
```

---

Tier 2 Cache Structure

The Tier 2 cache adds two fields to `struct dir`: a heap buffer pointer and an rfork LRU node pointer. The buffer size is always `dcache_rlen` (the resource fork length already cached in Tier 1). No separate size field is needed — `dcache_rlen` is the authoritative source for both the reported fork length and the cached buffer size.

```mermaid
graph TB
    subgraph StructDir[struct dir - with Tier 2 fields]
        subgraph Existing[Existing Fields]
            L0[Layer 0: Stat Cache - 48B inline]
            T1[Tier 1: AD Cache - 60B inline<br/>Includes dcache_rlen]
        end
        subgraph T2Ptr[Tier 2 Fields - 16B]
            RFD[dcache_rfork_buf - void* ptr to malloc buffer]
            RLRU[rfork_lru_node - qnode_t* position in rfork LRU]
        end
    end

    subgraph RFBuf[Resource Fork Data Buffer - heap allocated]
        DATA[Cached rfork content: dcache_rlen bytes]
    end

    RFD -->|points to| RFBuf

    subgraph RforkLRU[Rfork LRU List - dedicated tracking]
        HEAD[LRU head - eviction candidates]
        TAIL[MRU tail - recently accessed]
    end

    RLRU -->|position in| RforkLRU
    
    subgraph Budget[Memory Budget Tracking]
        TOTAL[rfork_cache_used - global counter]
        MAX[rfork_cache_budget - config limit]
    end

    style L0 fill:#2d5a2d,stroke:#4CAF50,color:#fff
    style T1 fill:#1a3a5c,stroke:#2196F3,color:#fff
    style T2Ptr fill:#5a1a5a,stroke:#9C27B0,color:#fff
    style RFBuf fill:#4a1a4a,stroke:#E040FB,color:#fff
    style Budget fill:#3a3a1a,stroke:#FFC107,color:#fff
    style RforkLRU fill:#1a3a3a,stroke:#00BCD4,color:#fff
```

---

Rfork LRU Mechanics

The rfork LRU is a doubly-linked list of `qnode_t` nodes, where each node's
`data` pointer references back to the `struct dir` that owns the rfork buffer.
The `struct dir` also holds a pointer to its LRU node (`rfork_lru_node`), creating
a bidirectional link. This allows O(1) operations in both directions:
- **From LRU → struct dir**: follow `node->data` to get the owning entry
- **From struct dir → LRU**: follow `dir->rfork_lru_node` to remove/promote

```mermaid
graph LR
    subgraph RforkLRU[Rfork LRU List - sentinel-based doubly-linked]
        S[sentinel<br/>rfork_lru]
        N1[qnode A<br/>data→]
        N2[qnode B<br/>data→]
        N3[qnode C<br/>data→]
        S -->|next| N1
        N1 -->|next| N2
        N2 -->|next| N3
        N3 -->|next| S
        S -->|prev| N3
        N3 -->|prev| N2
        N2 -->|prev| N1
        N1 -->|prev| S
    end

    subgraph Dircache[Dircache - ARC or LRU lists]
        DA["struct dir A<br/>(Icon\r in /Photos)<br/>rfork_lru_node→<br/>dcache_rfork_buf→"]
        DB["struct dir B<br/>(Icon\r in /Music)<br/>rfork_lru_node→<br/>dcache_rfork_buf→"]
        DC["struct dir C<br/>(Icon\r in /Docs)<br/>rfork_lru_node→<br/>dcache_rfork_buf→"]
        DX["struct dir X<br/>(readme.txt)<br/>rfork_lru_node=NULL<br/>dcache_rfork_buf=NULL"]
    end

    subgraph Heap[Heap - malloc'd rfork data]
        HA[120KB icns data]
        HB[85KB icns data]
        HC[200KB icns data]
    end

    DA -.->|rfork_lru_node| N1
    DB -.->|rfork_lru_node| N2
    DC -.->|rfork_lru_node| N3
    N1 -.->|data| DA
    N2 -.->|data| DB
    N3 -.->|data| DC
    DA -->|dcache_rfork_buf| HA
    DB -->|dcache_rfork_buf| HB
    DC -->|dcache_rfork_buf| HC

    style S fill:#1a3a3a,stroke:#00BCD4,color:#fff
    style N1 fill:#1a3a3a,stroke:#00BCD4,color:#fff
    style N2 fill:#1a3a3a,stroke:#00BCD4,color:#fff
    style N3 fill:#1a3a3a,stroke:#00BCD4,color:#fff
    style DA fill:#1a3a5c,stroke:#2196F3,color:#fff
    style DB fill:#1a3a5c,stroke:#2196F3,color:#fff
    style DC fill:#1a3a5c,stroke:#2196F3,color:#fff
    style DX fill:#3a3a3a,stroke:#9E9E9E,color:#fff
    style HA fill:#4a1a4a,stroke:#E040FB,color:#fff
    style HB fill:#4a1a4a,stroke:#E040FB,color:#fff
    style HC fill:#4a1a4a,stroke:#E040FB,color:#fff
```

---

Budget eviction walk (when a new rfork needs space):

```mermaid
flowchart TD
    START([rfork_cache_store_from_fd called<br/>needs 150KB]) --> CHK{rfork_cache_used<br/>+ 150KB > budget?}
    CHK -->|No: fits| STORE([Store new rfork<br/>add to LRU tail/MRU])
    CHK -->|Yes: over budget| HEAD[Get head of rfork LRU:<br/>head = rfork_lru->next]

    HEAD --> GET_DIR[Follow head->data<br/>to get struct dir C]
    GET_DIR --> FREE_BUF["free(dir_C->dcache_rfork_buf)<br/>dir_C->dcache_rfork_buf = NULL<br/>rfork_cache_used -= 200KB"]
    FREE_BUF --> REMOVE_NODE["Remove qnode C from LRU<br/>free(qnode C)<br/>dir_C->rfork_lru_node = NULL"]
    REMOVE_NODE --> NOTE["Note: struct dir C stays<br/>in dircache with valid<br/>stat + AD metadata.<br/>Only rfork DATA freed."]
    NOTE --> CHK

    style START fill:#5a1a5a,stroke:#9C27B0,color:#fff
    style STORE fill:#1a5a1a,stroke:#4CAF50,color:#fff
    style HEAD fill:#5a3a1a,stroke:#FF9800,color:#fff
    style FREE_BUF fill:#5a1a1a,stroke:#f44336,color:#fff
    style REMOVE_NODE fill:#5a1a1a,stroke:#f44336,color:#fff
    style NOTE fill:#3a3a1a,stroke:#FFC107,color:#fff
```

---

Population Path

Resource fork data is cached on the **first FPRead** of a cacheable resource fork.
Instead of reading only the client's requested chunk, we **upgrade the read to a
full-fork `ad_read()`** into the cache buffer, then serve the requested portion from cache.

This avoids caching partial data when AFP clients read in chunks (e.g., 32KB at a
time for a 100KB fork). The single full-fork read replaces what would have been a
partial read — same syscall count, but captures the entire fork.

**Note**: `rfork_cache_store_from_fd()` uses `ad_read()` (NOT raw `pread()`).
`ad_read()` handles all storage formats transparently — EA-based (`ea = sys`),
macOS native xattr, and AppleDouble v2 sidecar files (`ea = ad`). For AD v2,
the resource fork is at an offset inside the sidecar file; `ad_read()` adjusts
the pread offset internally via `ad_getentryoff()`. Performance is identical
to raw `pread()` — it IS `pread()` with the correct offset.

**Concurrent R+W note**: Two different AFP connections (two afpd child processes)
can have the same file open — one for read, one for write. Process A's cache
population via `ad_read()` may read data that Process B is concurrently writing.
This is **identical to the current non-cached behavior** — Process A's `ad_read()`
would read the same data regardless. The cache preserves that snapshot until the
IPC hint from Process B's `of_closefork()` arrives and triggers `DCMOD_AD_INV`.

```mermaid
sequenceDiagram
    participant C as AFP Client
    participant OF as afp_openfork
    participant RF as read_fork
    participant DC as dircache
    participant FS as Filesystem

    C->>OF: FPOpenFork RSRC RDONLY
    OF->>FS: ad_open ADFLAGS_RF
    Note over OF: fd opened normally

    C->>RF: FPRead offset=X count=N
    RF->>DC: dirlookup + dircache_search_by_name
    alt rfork data cached
        DC-->>RF: memcpy from dcache_rfork_buf at offset X
        Note over RF: ZERO I/O - served from cache
    else not cached AND cacheable
        Note over RF: Upgrade: ad_read entire fork into cache buf
        RF->>FS: ad_read(adp, ADEID_RFORK, 0, buf, dcache_rlen)
        FS-->>RF: entire fork data
        RF->>DC: rfork_cache_store entry and data
        Note over DC: malloc dcache_rlen bytes + memcpy + update rfork_cache_used + add to rfork LRU
        RF-->>RF: memcpy requested portion from cache_buf at offset X
    else not cached AND not cacheable
        RF->>FS: normal ad_read for requested portion
        FS-->>RF: data
    end
    RF-->>C: data

    C->>OF: FPCloseFork
    OF->>FS: ad_close ADFLAGS_RF
    Note over OF: fd closed. Cache survives in struct dir.
```


NJRoadfan:
Testing Notes:

Netatalk has four methods of storing resource fork data via its VFS back end that need to be tested.

1. Classic AppleDouble v2 format: Resource fork is stored in a hidden directory called .AppleDouble with the rest of a files metadata.
2. EA AppleDouble: Resource fork is stored in a ._ prefixed file using OS X style AppleDouble files.
3. Resource Fork stored in EA (Solaris hosts only): On Solaris machines running ZFS, the resource fork is stored entirely in an extended attribute called `org.netatalk.ResourceFork`
4. Resource Fork natively stored in file system (macOS hosts only): Under macOS, Netatalk currently takes advantage of the host OS's ability to natively store resource forks on HFS and APFS partitions.

andylemin:
> Testing Notes:
> 
> Netatalk has four methods of storing resource fork data via its VFS back end that need to be tested.
> 
> 1. Classic AppleDouble v2 format: Resource fork is stored in a hidden directory called .AppleDouble with the rest of a files metadata.
> 2. EA AppleDouble: Resource fork is stored in a ._ prefixed file using OS X style AppleDouble files.
> 3. Resource Fork stored in EA (Solaris hosts only): On Solaris machines running ZFS, the resource fork is stored entirely in an extended attribute called `org.netatalk.ResourceFork`
> 4. Resource Fork natively stored in file system (macOS hosts only): Under macOS, Netatalk currently takes advantage of the host OS's ability to natively store resource forks on HFS and APFS partitions.

Hi @NJRoadfan Thank you for the nudge 🙏
The good news! The rfork cache operates at a layer above ad_read(), which is the unified VFS abstraction. The store function rfork_cache_store_from_fd() calls ad_read(adp, eid, 0, buf, rlen) which handles all four backends transparently. All paths converge at adf_pread() → pread(fd, buf, count, offset). The cached bytes are the actual resource fork content after all backend-specific offset translation - so this does not touch or care about the storage semantics.

Safety is ensured by:
- AD_RSRC_OPEN(adp) guard before store — fd must be open
- Short/long-read detection invalidates entire AD cache — self-healing if fork size changed
- Write-mode forks (AFPFORK_ACCWR) bypass the cache entirely
- DCMOD_AD_INV invalidation on any rfork write clears cached data
- Leaving validation freq = 1 still works the same, any rfork changes are picked up immediately by the POSIX compliant (covers resource forks) ctime check, so this continues to be resistant to external changes

The plan is to test these anyway.

PS;
- "3. Resource Fork stored in EA (Solaris hosts only)" - this is no longer for Solaris only. Resource Forks are now stored natively in EA's on all/any operating system running ZFS since the ZFSoL project officially pulled ZFS out of Solaris/Illumos/FreeBSD (circa 2.x).
- "4. Resource Fork natively stored in file system (macOS hosts only)" - this is also no longer macOS only. ZFS also stores EAs natively in SAs (system attributes) the same as macOS.
ZFS is actually now more complete/native than even macOS HFS, as ZFS also supports storing NFSv4 & POSIX ACLs natively as EAs in SAs, which macOS does not.. Ie, ZFS has full parity with APFS!

1/2 - @rdmark we should have a container configured to use these classic storage formats. 3/4 are already covered by the new macOS container (which has APFS, therefore tests the same as ZFS) - plus I also run ZFS in my lab (all my development work is done with Netatalk on ZFS).

andylemin:
Hi @rdmark @NJRoadfan
I have just updated some containers so we have better test AD/RF coverage. I also noticed the workflows with `(ea:ad)` were missing `-e AFP_EA="ad"` so they were not actually testing the `ad` backend.. now fixed

doc/developer/dircache.md
```diff
@@ -531,25 +529,11 @@ it's not doing it at all.
...
-- **Cache lookup**: When a file is found in cache, checks `AFPobj->options.dircache_files`
-  before returning it — if disabled, returns AFPERR_BADTYPE to reject the cached file entry
-- **CNID lookup**: When a file is resolved via CNID, sets the `DIRF_ISFILE` flag
-  which will cause cache rejection if file caching is disabled
-- **Parent recursion**: Uses `strict` parameter to control whether parents must be directories,
-ensuring parent-chain rebuilds work correctly (database design guarantee)
-
-This configuration-driven control separates caching policy from core lookup logic, allowing
-administrators to tune memory vs. performance tradeoffs based on workload characteristics.
+If you have Samba or other processes operating on the volumne this must remain set to 1.
```
rdmark: typo: volumne -> volume

NJRoadfan:
@andylemin The storage of resource forks in EAs is not available on other platforms with ZFS. This is due to other platforms lacking the enhanced streams style extended attributes API that can handle large sized EAs. This appears to have been added to FreeBSD: https://www.freebsd.org/status/report-2025-04-2025-06/named-attributes/

It sounds similar to Apple's method of accessing resource forks using a special directory path. Problem is adding it would break existing installs on FreeBSD/OpenZFS.

andylemin:
> @andylemin The storage of resource forks in EAs is not available on other platforms with ZFS. This is due to other platforms lacking the enhanced streams style extended attributes API that can handle large sized EAs. This appears to have been added to FreeBSD: https://www.freebsd.org/status/report-2025-04-2025-06/named-attributes/
> 
> 
> 
> It sounds similar to Apple's method of accessing resource forks using a special directory path. Problem is adding it would break existing installs on FreeBSD/OpenZFS.

Hi @NJRoadfan Yes I agree this is a subtle distinction - just because a kernel has ZFS does not mean the APIs et al are available.

However I don't understand your last comment? Adding what to what would break what?

From Netatalk perspective if it sees the EA API header files during config/build it uses them, else falls back to AD files?
How the OS/VFS stores the EAs underneath is transparent to Netatalk?

I agree this area is very complex and inconsistent across operating systems. Every time I think I have the complete picture, I discover another gotcha 😂

andylemin:
> @andylemin The storage of resource forks in EAs is not available on other platforms with ZFS. This is due to other platforms lacking the enhanced streams style extended attributes API that can handle large sized EAs. This appears to have been added to FreeBSD: https://www.freebsd.org/status/report-2025-04-2025-06/named-attributes/
> 
> 
> 
> It sounds similar to Apple's method of accessing resource forks using a special directory path. Problem is adding it would break existing installs on FreeBSD/OpenZFS.

Hi. I read this again and I think I understand it now. This is only so NFS can use the larger EAs which have been available on ZFS for a very long time.

System Attributes can only store 128k (sometimes only 64k depending on OS), so if the EA fits it's stored as an SA. If it's larger it is stored as a hidden directory structure (discussed in that post). I believe this is providing NFS specifically with access to the larger EAs via the Named Attribute interface. The underlying storage is unchanged, hence why it's a FreeBSD change rather than a ZFS change.

As I don't think this relates to us, what was it you were thinking?

Thanks again for the feedback, I find this topic quite interesting and nuanced 🙂

NJRoadfan:
This goes into more detail: https://man.freebsd.org/cgi/man.cgi?query=named_attribute&sektion=7&format=html

Basically, the native EA API calls on Solaris allowed for EAs of greater size. A somewhat compatible interface was added to FreeBSD 15 for both NFS and ZFS. I will continue this under Issues as its something to look into, although not a priority.

andylemin:
Will submit the performance regression fix found with flamegraph first and then I'll comeback to this

andylemin:
@rdmark since testing this on top of the idle worker, it is working so well that I had to quadruple my DSI readahead buffer!
An entire catalogue of resource forks now loads on remote clients in less than 1 second (without batch loading as long as DSI is large enough) - this was the last significant bottleneck to providing a remote share experience that almost replicates local disk performance! 🚀

Just a few small things to clean up before it's ready for final review

etc/afpd/dircache.c
```diff
@@ -173,6 +173,28 @@ static struct dircache_stat {
...
+ * Non-static: accessed from ad_cache.c (rfork_cache_store_from_fd, etc.).
+ * Process-safe: netatalk uses fork-per-connection — each child has its own
+ * address space and dircache. No locking needed. */
+size_t rfork_cache_used = 0;        /* current total bytes in rfork cache */
+size_t rfork_cache_budget =
+    0;      /* configurable limit (bytes), 0 = disabled */
+size_t rfork_max_entry_size = 0;    /* per-entry limit (bytes), 0 = disabled */
+unsigned int rfork_lru_count = 0;   /* entries currently in rfork LRU list */
+q_t   *rfork_lru =
+    NULL;           /* dedicated LRU list for rfork cache entries */
```
rdmark: nitpick: we end up with two of these awkward line breaks here!

rdmark (approved):
impressive stuff!

we just need a bit of education for users to configure the resource fork cache appropriately

NJRoadfan:
Also, does any of this work or conflict with AFP AppleTalk connections? I'm noticing a few tie-ins to afp_dsi.c (with this and the dircache enhancements in general), which isn't going to be called (ie: `dircache_rfork_shutdown()`) when an AppleTalk connection is made (afp_asp.c is run instead). Hopefully there isn't any cache coherency issues if both types of connections are active.

andylemin:
> Also, does any of this work or conflict with AFP AppleTalk connections? I'm noticing a few tie-ins to afp_dsi.c (with this and the dircache enhancements in general), which isn't going to be called (ie: `dircache_rfork_shutdown()`) when an AppleTalk connection is made (afp_asp.c is run instead). Hopefully there isn't any cache coherency issues if both types of connections are active.

Hi @NJRoadfan This particular change was designed to only work with TCP DSI clients, as the rfork cache feeds the DSI directly (the rfork layer is fairly low down the stack compared to all the other dircache changes).

So it should not be in the ASP code path. However as the last set of PRs have been large and added a significant number of complex optimisations - whilst I put in an extreme amount of effort to make everything just right and bug free, I asked @rdmark if we could make the next release a Beta release.

I have been pushing to reach this major milestone (remote performance feels close to local performance), so this is the right point to stop and reflect, and allow our user base time to digest and test this stack of enhancements.

I have a small PR with various house keeping and clean up which I'm going to submit in the next few days. Once this and that are merged, we should be good for the Beta release. At which point it would be great if we can ask the AppleTalk community to test and submit issues.

I'm planning to concentrate on any reported bug fixes before we release a full production release.

I also wanted all the changes to be complete before @rdmark meets up with the community IRL soon, so he can encourage a wave of testing there.

@NJRoadfan What do you think? Anything else you can suggest to start hardening the release?

---

For this PR in particular, in my own stress testing this is working perfectly (successfully filled the cache with 10GB (config max) of rforks and they are correctly evicted relative to access patterns rather than just oldest). Just one spurious log message, line wrapping, and documentation to fix.

We also need to consider increasing the default DSI readahead buffer now, even with rfork caching disabled by default, the rest of the dircache changes have removed enough bottlenecks that the default DSI values are now a bottleneck. We can probably set this dynamically based on system memory if not defined explicitly.

On 10GbE I need 256MB DSI readahead to not exhaust the DSI buffer. On 1GbE and WiFi7 I need 64-128MB. A very conservative default should be 32 (up from 12 which is tiny now). However this is tricky as it depends on the number of expected concurrent users..

Last but not least, to celebrate this milestone, after the beta release, I'm going to create a local filesystem version of lantest and speedtest, so I can run realistic workload performance testing against local mounts (enabling us to perform identical testing on Samba, Netatalk, NFS and Local disk mounts) to refresh the performance wiki page.

NJRoadfan:
Just voicing concern as the unit testing doesn't support AppleTalk ASP connections at all. These optimizations wouldn't work with it anyway due to the nature of the protocol.

Speaking of Samba, I'm interested to see how much faster Netatalk will be against it. Maybe the Samba folks could learn a thing or two to speed things up. Too bad the caching might not play well with a Netatalk share that is also shared with Samba.

andylemin:
> we just need a bit of education for users to configure the resource fork cache appropriately

@rdmark It was clearly time to perform a complete update and overhaul of Configuration.md as there was a lot of obsolete information and inconsistent writing styles.

andylemin:
@NJRoadfan Correction.. I just found what I think you were concerned about. The dircache must be initialised for both, however I mistakenly thought its DSI integration would make it a no-op for ASP, you are correct, this is wrong..

Removing dircache_init() from the ASP path entirely would break ASP sessions. The rfork tier-2 init is only a small section, but is still part of the dircache stack. Without dircache_init(), ASP clients would obviously have no directory cache at all and lookups like dirlookup(), dir_add(), etc would crash. Yet we cannot just remove the rfork cache from the dircache for ASP clients as the structs had to be changed..
Sorry it took me a while to get there 🙂 Thanks for your patience

So for ASP users I am now forcing zero values for the rfork feature 🫡

etc/afpd/dircache.c
andylemin: Thanks for spotting this..

andylemin:
@rdmark @NJRoadfan Ready for the final pre-merge review 🎉 I tested and tried as hard as I could to break it.. 🙂

andylemin:
PS; Now I'm in bug hunting mode, I have been working my way through the flamegraph top hitters, and trying to stress test and harden Netatalk - and I figured out some really nice optimisations! 3 to be exact. The tenets of the last few changes (cleaning orphaned entries + IPC hints for cache coherence, and the idle thread) are the right solutions, but the tweaks I am working on both harden them from attacks and provide some nice boosts too.. PRs to follow shortly. No new features now, just fixes and optimisations

rdmark:
> > we just need a bit of education for users to configure the resource fork cache appropriately
> 
> @rdmark It was clearly time to perform a complete update and overhaul of Configuration.md as there was a lot of obsolete information and inconsistent writing styles.

wow, you really doubled down on touching up that Configuration chapter, it definitely reads much better now overall

our userbase generally dislikes the Configuration chapter, I've gotten a lot of, hrm, harsh feedback over the year that it's dense and inscrutable, completely useless for a layman who want to learn how to configure netatalk for basic deployments... which I generally agree with

one thing that's held me back from overhauling the chapter is the labor overhead of redoing the Japanese localization, but we now have access to very potent translation tools I'm less afraid of the consequences

not something for this PR, but I have been thinking about taking it one step further and splitting it up into multiple chapters, one per heading basically; what do you think?

andylemin:
@rdmark Yea it definitely needs breaking up. It has issues with flow, concept grouping, section sizing, and clear recommendations.

My changes were just simple grammar changes and error corrections, to make it easier to get through in the meantime before decomposing.

I think people complain as the first few sections are overweight and not relevant for most users, and you have to hunt for the first valuable sections

PR #2784, opened 2026-03-16
debug: Added flamegraph profiling container and tooling

andylemin:
Added a dedicated debug container for generating CPU flamegraphs of
afpd during test suite execution.

New files:
- distrib/docker/debug_alp.Dockerfile: Alpine-based build with
  -Dbuildtype=debugoptimized (-O2 -g) and -fno-omit-frame-pointer
  for accurate perf stack unwinding. Included linux perf and
  pre-cloned FlameGraph tools.
- distrib/docker/debug_entrypoint_netatalk.sh: Entrypoint with
  perf record/script/flamegraph.pl pipeline. Used fd redirection
  to emit SVG to stdout while logs went to stderr, enabling clean
  capture with: docker run ... > flamegraph.svg
- distrib/docker/env_setup_netatalk.sh: Extracted ~340 lines of
  shared environment setup from both entrypoint scripts into a
  common script sourced via `. /env_setup.sh`. Eliminates code
  duplication between the production and debug entrypoints.
- doc/developer/profiling.md: Developer documentation.

Entrypoint deduplication:
- entrypoint_netatalk.sh: Reduced from 387 to 97 lines.
- debug_entrypoint_netatalk.sh: Reduced from 537 to 240 lines.
- Both now source env_setup_netatalk.sh for all common setup
  (users, volumes, permissions, config, MariaDB, AppleTalk, etc.).
- ERROR/WARNING messages now directed to stderr.
- Introduced DISTRO_ALPINE constant to avoid repeated string literal.

Dockerfile improvements:
- All 4 Dockerfiles now COPY env_setup_netatalk.sh into container.
- Added missing config_watch.sh COPY to debug_alp, testsuite_alp,
  and testsuite_deb Dockerfiles (was only in netatalk.Dockerfile).
- dircache rfork budget/maxsize options now in all containers via
  shared afp.conf generation (defaults safe: budget=0 = disabled).

CI integration (containers.yml):
- Added build-container-debug job to build debug_alp.Dockerfile
  and push to ghcr.io on every PR.
- Added flamegraph-spectest job that runs spectest with FLAMEGRAPH=1
  in --privileged mode (ARC cache, cache=1024 for stress profiling).
- Captures SVG and uploads as 6-month artifact.
- Posts/updates a PR comment with comment-tag for SonarQube-style 
replacement on force-push.
- All new actions SHA-pinned per project supply-chain security policy.

Supported two profiling modes:
- On-CPU (FLAMEGRAPH=1): Sampled at 999 Hz to show CPU hotspots.
  Best for spectest and lantest workloads.
- Off-CPU (FLAMEGRAPH_OFFCPU=1): Traced sched:sched_switch to
  show where afpd blocked/slept. Blue-colored SVG output.

NB; _Github workflows are executed on amd64/intel containers, so GH workflow generated Flamegraphs show x86 instruction times. The below examples were executed locally on macOS/M2, and show ARM64 instruction times._

# Spectest (ARM64)
flamegraph5

# Lantest (ARM64)
flamegraph_lan


andylemin:
I now know _exactly_ why renaming files seems slower now 😉 I'll submit a PR for that performance regression soon

andylemin:
@rdmark Can you please approve those security concerns? - this is just a debug container (and uses the same pattern as the production ones)..

rdmark:
I love this! Actually learned about flamegraphs recently (in the context of OpenJDK) and it's great to see it applied this effectively on afpd here.

The four security hotspots have been flagged as "safe" now with the justification that this is purely a test environment. There are four maintainability issues raised here as well that you may want to take a look at. They're all trivial fixes.

distrib/docker/debug_entrypoint_netatalk.sh
```diff
@@ -0,0 +1,537 @@
...
+# Based on entrypoint_netatalk.sh with added flamegraph profiling support.
+#
+# This script profiles afpd with Linux perf during test execution and
+# generates an interactive SVG flamegraph. Use with debug_alp.Dockerfile.
+#
+# See doc/developer/profiling.md for full documentation.
+#
+# Copyright (C) 2023  Eric Harmon
+# Copyright (C) 2024-2026  Daniel Markstedt <daniel@mindani.net>
+# Copyright (C) 2025-2026  Andy Lemin @andylemin
```
rdmark: I like the @ mark notation for your github handle! perhaps put it in parentheses as well? I have a memory of seeing this style in other projects.

we could perhaps standardize on the `(@username)` notation consistently in this codebase?

as I mentioned before, having consistency in copyright headers will make Debian packaging easier later

distrib/docker/debug_entrypoint_netatalk.sh
```diff
@@ -0,0 +1,537 @@
...
+#   docker run ... > flamegraph.svg
+#
+# All progress/log messages still appear on the terminal via stderr.
+# --------------------------------------------------------------------------
+if [ -n "$FLAMEGRAPH" ]; then
+    exec 3>&1 # save original stdout as fd 3 (SVG will go here)
+    exec 1>&2 # redirect stdout → stderr (all echo/logs go to terminal)
+fi
+
+echo "*** Setting up environment (debug/profiling entrypoint)"
```
rdmark, line 43: long term, do we want to create a separate utility script for the environment setup that we can call here and in the other entrypoint scripts? do you know what's considered good practice in complex container scripts?

I've been thinking of splitting up the entrypoint scripts for the production container and test containers at some point. It's getting quite complex with all the test conditions and I'm worried we'll break something for people who use it in production eventually.

distrib/docker/debug_entrypoint_netatalk.sh
```diff
@@ -0,0 +1,537 @@
...
+
+    echo "*** Flamegraph generated: $FLAMEGRAPH_SVG"
+    echo "*** flamegraph.svg size: $(du -h "$FLAMEGRAPH_SVG" | cut -f1)"
+    echo "*** Folded stacks: $(wc -l < "$PERF_FOLDED") unique stacks"
+
+    # Emit the SVG to fd 3 (the original stdout, saved at script start).
+    # This is the ONLY data that goes to stdout, so the user can capture
+    # it cleanly with:  docker run ... > flamegraph.svg
+    echo "*** Writing flamegraph SVG to stdout (fd 3)"
+    cat "$FLAMEGRAPH_SVG" >&3
```
rdmark, line 188: random thought: do we want to run this as a GitHub action and then publish the resulting svg on GitHub Pages? :)

distrib/docker/debug_entrypoint_netatalk.sh
```diff
@@ -0,0 +1,537 @@
...
+        # Off-CPU: blue color palette and distinct title to distinguish
+        # from on-CPU flamegraphs.
+        "$FLAMEGRAPH_DIR/flamegraph.pl" \
+            --title "Netatalk afpd — ${TESTSUITE} OFF-CPU flamegraph" \
+            --subtitle "buildtype=debugoptimized, sched:sched_switch tracepoint" \
+            --color io \
+            "$PERF_FOLDED" > "$FLAMEGRAPH_SVG"
+    else
+        "$FLAMEGRAPH_DIR/flamegraph.pl" \
+            --title "Netatalk afpd — ${TESTSUITE} flamegraph" \
```
rdmark: seems like the mdash here gets mangled by the Perl script -- maybe a simple dash is easier than messing with encoding

Image

distrib/docker/debug_entrypoint_netatalk.sh
andylemin: Yes I felt it's getting overweight too, which is why I created a copy for debug builds like this.

However when I looked at what could be stripped from the production one, it is all critical stuff..

As there is a fair bit of duplication between this and the production one, I definitely get the idea for having a modular script for common environment setup.

However I think that needs a separate PR as I can see some discussions around this

distrib/docker/debug_entrypoint_netatalk.sh
andylemin: That's an amazing idea to get fast feedback! Love it..

I have no idea how that would work with GitHub actions however?
I mean how would we publish it on a per PR basis?

From your comment I'm guessing you are imagining using some GitHub call to attach the render to the ticket as a comment?

distrib/docker/debug_entrypoint_netatalk.sh
andylemin: Yea this is coming from the flamegraph rendering script.

Agreed, I'll tweak the script to use a different title

distrib/docker/debug_entrypoint_netatalk.sh
rdmark: the easiest option that I can think of would be using `actions/upload-artifact` to store it as an artifact that you can go in to the workflow log and download there, not the most convenient however

number two would be to publish it to GitHub Pages with `actions/configure-pages` and `actions/upload-pages-artifact` -- I'm thinking we could potentially remove or modify static.yml because it's not providing much value at the moment

thirdly, I did a quick google if there's an even better way and found this gist, that supposedly will post a PR comment with the flamegraph which would be the most convenient for certain!

https://gist.github.com/sassman/0a094b942e5bba43da50308654b83f7e

distrib/docker/debug_entrypoint_netatalk.sh
rdmark: absolutely a bigger topic, I created https://github.com/Netatalk/netatalk/issues/2785 now for future follow-up

andylemin: I have just implemented this - added a common env setup script with makes the entry point scripts much cleaner 🙂

andylemin:
This now fixes #2785

distrib/docker/debug_entrypoint_netatalk.sh
andylemin: I think I have got this working. Let's wait and see 😁

I did some digging and discovered how Sonarqube does it (updates any existing comment so there is only one, and force pushes should update it) 🤞

rdmark:
I took the liberty of 1) accepting the dockerfile running as root (test container) and 2) disabling the "File permissions should not be set to world-accessible values" rule altogether for shell scripts -- the next Sonar run should pass

andylemin:
@rdmark It looks like Artifacts are always bundled in zip files. So we will need to use github Pages and link it from there.
The existing static.yml uses artifact-based Pages deployment. To add flamegraphs to Pages without breaking the manual, we can use peaceiris/actions-gh-pages to push to a gh-pages branch under flamegraphs/ with keep_files: true. This requires a one-time repo settings change to switch Pages source to "Deploy from branch: gh-pages".

The one-time repo setting: Settings → Pages → Source → "Deploy from a branch" → Set Branch: gh-pages / / (root). I don't have permissions for Settings tab?

However this will conflict with the existing static.yml which uses actions/deploy-pages (artifact-based deployment). The two mechanisms are mutually exclusive in GitHub. We would need to migrate static.yml to also push to gh-pages branch - I'll have a go at this now

rdmark:
I'm ok with discarding the current static.yml -- publishing the local html manual to Pages hasn't proven useful for anything really

it's an option anyways!

rdmark:
> The one-time repo setting: Settings → Pages → Source → "Deploy from a branch" → Set Branch: gh-pages / / (root). I don't have permissions for Settings tab?

let me look into how I can grant you the permissions

andylemin:
Hi @rdmark
I finally got this working 🙂 It was a lot of trial and error!
No need to change any Settings and it only keeps Flamegraphs for 6 months (we might need to cap by count instead of time, as this is now compatible with force-pushes - it replaces the new Flamegraph comment when the commit is updated).

It required me to remove `static.yml` entirely and add the code to push to Pages within `containers.yml` instead - Using static.yml meant graphs would only be pushed to Pages when the code is finally merged into `main`..

I don't know whats up with those 2 Sonarqube warnings? We want both tags and digest (tags for clear versions, and digest for pining)?

rdmark:
> I don't know whats up with those 2 Sonarqube warnings? We want both tags and digest (tags for clear versions, and digest for pining)?

indeed, I see what sonarqube wants to say here that the version tag is ignored by the container runtime; but I disagree with the rule, in particular since we use dependabot to automatically and consistently update the version tag and digest together

this is akin to the version number comment for actions, it's only there for human readability, but has no programmatic function

e.g.

```
actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
```

the rule has been disabled in our quality profile now!

doc/developer/profiling.md
```diff
@@ -0,0 +1,177 @@
+!Flamegraph Profiling
```
rdmark, line 1: what's up with the exclamation point? :)

andylemin: lol, well spotted.
After working as a DBA for years after Uni I have a hardwired hash-bang habit 🙂

I'll remove in the next PR. Want to merge now as I have too much (performance regression fixes which the original PoC for this highlighted) backed up behind this.

PR #2794, opened 2026-03-17
ci: explicitly skip rebuild for build jobs that use sshfs

rdmark:
the inconsistent time stamp on build files when using sshfs to sync the build job file systems cause intermittent failures when rebuilds are triggered, so let's prevent said rebuilds which should hopefully lead to more consistent FreeBSD / NetBSD / OpenBSD workflow builds

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -537,8 +537,8 @@ jobs:
               -Dwith-tests=true \
               -Dwith-testsuite=true
             meson compile -C build
-            meson test -C build
-            meson install -C build
+            meson test -C build --no-rebuild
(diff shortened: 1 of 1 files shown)


rdmark:
I have verified that this fixes the consistent NetBSD failure in https://github.com/Netatalk/netatalk/pull/2789

andylemin:
Fantastic! Has been a bit annoying. Well done 🙂

I noticed if you re-run the NetBSD build with debug enabled it nearly always works which pointed to something nuanced like this

andylemin (approved):

rdmark:
it worked once at least, let's hope it holds up under repeated scrutiny!

PR #2796, opened 2026-03-18
introduce signal handlers and refactor away unreachable code

rdmark:
a number of endless or long running loops lacked signal handlers to protect against fail states, so this introduces a number of them for papstatus, papd, fce_listen, and cnid_dbd

cleaned up all cases of unreachable code flagged by SonarQube, which was made up mostly of old school defensive coding which I think is less relevant with the tooling we have today


andylemin (approved):
Nice clean up and the SIGnal handling looks good to me

PR #2799, opened 2026-03-21
Revise README for clarity and updated information

rdmark:
Updated the README to improve clarity and accuracy of information about Netatalk, including changes to descriptions of compatibility and community resources.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,27 +16,31 @@ Supply Chain Security:
 
 ## About Netatalk
 
-Netatalk is a Free and Open Source file server for Unix-like operating systems that implements
+Netatalk is a Free and Open Source file server suite for Unix-like operating systems that implements
 the [Apple Filing Protocol](https://en.wikipedia.org/wiki/Apple_Filing_Protocol) (AFP) over TCP/IP and AppleTalk.
-AFP is the native file sharing protocol used on Apple II, Classic Mac OS, early Mac OS X,
-as well as one of several supported protocols on macOS.
-Netatalk is also compatible with many 3rd party AFP clients, including *afpfs-ng* and *afp-perl*.
+AFP is the native file sharing protocol on Apple II, Classic Mac OS, early Mac OS X,
+as well as one of several natively supported protocols on macOS.
+Netatalk is also compatible with many 3rd party AFP clients, including [GNOME Files (Nautilus)](https://apps.gnome.org/Nautilus/),
+[afpfs-ng](https://github.com/Netatalk/afpfs-ng) and [afp-perl](https://github.com/demonfoo/afp-perl).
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -16,27 +16,31 @@ Supply Chain Security:
...
+
+A Netatalk AFP server allows you to share, collaborate on, and back up files remotely.
 The latest macOS at the time of writing (macOS 26 Tahoe) comes with a built-in AFP client,
-so Netatalk can act as a seamless bridge between new and old Macs.
+so Netatalk can act as a seamless bridge between new and old Macs, and 3rd party AFP clients.
 
-Compared to cross-platform file sharing protocols like NFS and FTP, Netatalk delivers a Mac-like user experience,
+Compared to common file transfer protocols like NFS and FTP, Netatalk delivers a Mac-like user experience,
 with seamless integration of Mac filesystem metadata - notably Extended Attributes on macOS and resource forks
 on Classic Mac OS - as well as compatibility with modern macOS features such as Zeroconf (Bonjour) service discovery,
```
andylemin: Resource forks were introduced in Classic Mac OS? Not limited too?

README.md
```diff
@@ -16,27 +16,31 @@ Supply Chain Security:
...
 Time Machine backups, and Spotlight indexed search.
 
 Compared to [Samba](https://www.samba.org/) (SMB),
-Netatalk has [demonstrably faster transfer speeds](https://netatalk.io/docs/Benchmarks),
-and stronger backwards compatibility with OS X 10.8 Mountain Lion clients and earlier.
+Netatalk has [demonstrably faster transfer speeds](https://netatalk.io/docs/Benchmarks)
```
andylemin, line 42: Once I've fixed the performance regressions I want to re-benchmark.
We should be notably faster now.

I'm thinking bonnie++ on the client against the various protocols for more structured and well known tests.

But also need something which will exercise create/rename/move/delete operations etc - a test tool which can do more realistic workload testing

README.md
rdmark: resource forks are used by Carbon applications on early Mac OS X as well -- the summary in this readme is deliberately a simplified view, but if you have a suggestion I'm happy to tweak it

README.md
rdmark: such a methodology would be superior to the stop watch measurements I did at that time :)

README.md
andylemin: How about "resource forks were introduced in Classic macOS"? This is more inclusive for all later versions? What do you think?

rdmark: I went down a slightly different path. What do you think about the current wording?

andylemin: Yea that's good 👍

PR #2803, opened 2026-03-22
nad: centralize volume validation and fix inconsistent error handling

rdmark:
Add a user-facing error message in openvol() when a path is not inside a Netatalk volume, replacing the patchwork of per-command messages that were inconsistent, incorrect, or missing entirely.

Fix all openvol() call sites to check the return value:
- nad_ls: was silently ignoring failures (3 unchecked calls)
- nad_set: was calling exit(1) with no error message
- nad_cp: had a redundant unchecked call before the loop
- nad_mv: had swapped source/destination error labels, plus an unchecked call in the multi-file path
- nad_rm: checked vol.vol == NULL but not the return value
- nad_find: had a redundant error message and a memory leak on the error path

### DIFF
--- a/bin/nad/nad_cp.c
+++ b/bin/nad/nad_cp.c
@@ -358,20 +358,15 @@ int ad_cp(int argc, char *argv[], AFPObj *obj)
     /* Inhereting perms in ad_mkdir etc requires this */
     ad_setfuid(0);
 #endif
+
     /* Load .volinfo file for destination*/
-    openvol(obj, to.p_path, &dvolume);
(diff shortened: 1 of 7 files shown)


rdmark:
@NJRoadfan the nad error handling was quite the mess and inconsistent between commands; can you please test if this works to your satisfaction?

NJRoadfan:
This appears to have solved the problem. Some of the scripts in A2SERVER were relying on `nad ls` returning an error when outside of a volume as a sanity check to ensure they were indeed not in a Netatalk volume.

PR #2805, opened 2026-03-22
cnid: fix permissions for multi-user access to SQLite db files

rdmark:
When root (afpd, dbd) creates the SQLite database, the WAL and SHM auxiliary files inherit root ownership with restrictive permissions, preventing non-root clients such as nad from accessing the database.

Set the CNID directory to sticky bit mode (01777) so any user can create files but not delete others' files. Chmod the WAL and SHM files to 0666 after enabling WAL mode, matching the main db file. Also fix up permissions on pre-existing directories during open.

### DIFF
--- a/libatalk/cnid/sqlite/cnid_sqlite.c
+++ b/libatalk/cnid/sqlite/cnid_sqlite.c
@@ -1279,44 +1279,31 @@ struct _cnid_db *cnid_sqlite_open(struct cnid_open_args *args)
     become_root();
     is_root = true;
 
-    if (mkdir(dirpath, 01777) != 0 && errno != EEXIST) {
-        LOG(log_error, logtype_cnid, "Failed to create CNID DB directory '%s': %s",
-            dirpath, strerror(errno));
-        EC_FAIL;
-    }
+    if (mkdir(dirpath, 01777) != 0) {
+        if (errno == EEXIST) {
+            int dirfd = open(dirpath, O_RDONLY | O_DIRECTORY | O_NOFOLLOW);
 
-    int dirfd = open(dirpath, O_RDONLY | O_DIRECTORY);
+            if (dirfd < 0) {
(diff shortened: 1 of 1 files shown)


rdmark:
@NJRoadfan this is a yet untested fix for the sqlite db file permissions issue: kept as a draft for now since I'm not yet 100% convinced that this is the best solution... there's a tension between secure permissions and ease of use here.

NJRoadfan:
This won't crop up during normal usage, but you have to run `dbd -f` as root, which folks migrating to this backend will have to do. Once a client connects to the volume the permissions issue is fixed, but until then, you can't use any of the `nad` tools as a normal user on the volume.

rdmark:
I ran a few tests and confirmed that this fix allows nad to perform operations on a "virgin" volume.

@andylemin I'd love your opinion on this, what is reasonable / common practice for a world-writable directory

libatalk/cnid/sqlite/cnid_sqlite.c
```diff
@@ -1266,14 +1267,24 @@ struct _cnid_db *cnid_sqlite_open(struct cnid_open_args *args)
...
+            if (dirfd < 0) {
                 LOG(log_error, logtype_cnid, "'%s' exists but is not a directory", dirpath);
                 EC_FAIL;
             }
+
+            /* Ensure existing directories get updated to the sticky bit permissions,
+             * so that non-root clients such as 'nad' can create SQLite WAL/SHM files. */
+            struct stat st;
+
+            if (fstat(dirfd, &st) == 0 && (st.st_mode & 01777) != 01777) {
```
andylemin, line 1295: Should these be 777?

andylemin:
> @andylemin I'd love your opinion on this, what is reasonable / common practice for a world-writable directory

Hi @rdmark I now understand the reason why we need world access, and I see you already set the sticky bit. My eyes went straight to the 777 😂 
Resolved 👍

andylemin:
Good to merge

PR #2806, opened 2026-03-22
Add support for "cnid dev = no" to MySQL and SQLite backends.

### LINKED ISSUE #2802: MySQL and SQLite CNID backends do not respect "cnid dev = no" setting.
**Describe the bug**

When using the MySQL or SQLite CNID backends, the `cnid dev = no` option in `afp.conf` has no effect.

**To Reproduce**

-Install Netatalk.
-Create a volume using the MySQL or SQLite CNID backend.
-Add `cnid dev = no` option to that volume.
-Dump the database for that volume. The value of `DevNo` will be the actual value returned by the OS, not `0`.

**Expected behavior**

The value of the `DevNo` field for files in the volume should be `0` not any other value when `cnid dev = no` is set. The legacy dbd CNID backend respects this.

**Environment**

- Client OS: N/A
- Netatalk Version: 4.4.1

NJRoadfan:
Adds support for the `cnid dev = no` option to the MySQL and SQLite CNID back ends. If this option is set to "no", the DevNo field in the database should be set to "0". Fixes #2802

### DIFF
--- a/libatalk/cnid/mysql/cnid_mysql.c
+++ b/libatalk/cnid/mysql/cnid_mysql.c
@@ -289,7 +289,12 @@ int cnid_mysql_update(struct _cnid_db *cdb,
         EC_FAIL;
     }
 
-    uint64_t dev = st->st_dev;
+    uint64_t dev = 0;
+
+    if (!(cdb->cnid_db_flags & CNID_FLAG_NODEV)) {
+        dev = st->st_dev;
+    }
+
     uint64_t ino = st->st_ino;
 
     do {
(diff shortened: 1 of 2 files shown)


rdmark:
@NJRoadfan thanks for the PR, the logic changes look sound to me

the description of `cnid dev` in the afp.conf man page is pretty vague, but from the code it seems like DevNo is used together with InodeNo to detect if a file changed on disk and do appropriate operations as a result, so I was thinking we should flesh out the description of `cnid dev` to give an indication what the tradeoffs for turning it is

thoughts?

(BTW, disregard the comment from augmentcode -- I'm trying out a code review bot)

NJRoadfan:
The problem I was running into was when I was using a FUSE overlay file system (ciopfs). On reboot, DevNo would sometimes change, leading to Netatalk rebuilding the CNID database even though there was no changes to the files. Setting cnid dev to "no" solved the problem as it now remains fixed to a value of 0.

I don't even see a case where a volume would have multiple DevNos where you would have a collision of inode numbers (which will always be unique on a given filesystem and is a 64bit value), but I'm guessing it was a problem somewhere.

PR #2808, opened 2026-03-23
dircache: add idle worker thread for background cache maintenance

andylemin:
Introduce a generic idle worker thread that runs non-critical dircache housekeeping during poll() idle periods, moving expensive operations off the AFP command hot path.

Worker thread architecture:
- Atomic coordination (is_idle/bg_running) with temporal separation instead of locks — worker and main thread never access dircache data concurrently
- Condvar sleep/wake with signal blocking around start/poll/stop
- Compile-time lock-free atomic check (ATOMIC_INT_LOCK_FREE)
- Graceful fallback to synchronous behavior if thread init fails

New files:
- etc/afpd/idle_worker.h: public API (init/start/stop/shutdown/stats)
- etc/afpd/idle_worker.c: worker thread implementation

Deferred child cleanup:
- Add dircache_remove_children_defer() — O(1) enqueue with static circular buffer (MAX_DEFERRED_CLEANUPS=128), synchronous fallback when queue full or worker inactive
- Add dircache_process_deferred_chain() — incremental hash chain walk with batch limit (DEFERRED_CHAIN_BATCH=16) per iteration
- Add dircache_flush_deferred_for_vol() — synchronous drain on volume close to prevent stale volume references
- Convert all 8 dircache_remove_children() call sites to _defer (directory.c x4, filedir.c x2, dircache.c x1, fork.c x1)

Hash table resize safety:
- Add DIRCACHE_NOSHRINK flag to dircache_remove() — uses hash_scan_delfree() instead of hash_delete_free() for both dircache and index_didname tables during worker iteration
- Add hash_chain_head() static inline accessor in dircache.c

Direct entry removal:
- Add dir_remove_and_free() — removes from cache indexes and frees immediately without enqueue into invalid_dircache_entries, used by worker during temporal separation and volume close flush

DSI loop integration:
- idle_worker_start() before poll(), idle_worker_stop() after poll()
- Signal blocking (SIGURG/SIGALRM/SIGTERM/SIGQUIT) around start/poll/stop to protect pthread_mutex_unlock from siglongjmp
- idle_worker_stop() in sigsetjmp recovery path
- idle_worker_stop_signal_safe() in afp_dsi_close() (async-signal-safe)
- idle_worker_init() after dircache_init()
- Guard dir_free_invalid_q() with idle_worker_is_active()
- idle_worker_log_stats() at session close
- Section marker comments [A] through [I]

Build system:
- Add idle_worker.c to etc/afpd/meson.build and test/afpd/meson.build

Test adjustment:
- T2_FPGetFileDirParms test127: add sleep(1) before checking deferred child cleanup result to allow idle worker processing time (empty AFP command buffer)

Documentation:
- Add threading contract comment to invalid_dircache_entries
- Add "Part 8: Idle Worker Thread" section to doc/developer/dircache.md

---

Tested and working well. Instead of blocking AFP commands, we now defer clean up until idle time, allowing AFP commands to complete as fast as possible and read the next AFP command immediately.

This has been so successful that we can also consider optimising `invalid_dircache_entries` queue and FCE event handling.

Total spectest runtime is identical, as the same work is done, but the time to complete ops has reduced

## Results

The lantest results are positive, no performance regressions - only gains, but sadly we do not have a test in here which explicitly deletes/renames deep directory trees to really highlight the difference this fix targets.

**Platform:** Bare metal Linux server, no other processes running (no virtualization overhead)
**Iterations:** 10 per branch
**Config:** ARC mode, dircache=131072, validation_freq=100

### Performance Comparison

| Test | AFP Ops | Baseline (ms) | Baseline ± | Idle Worker (ms) | Worker ± | Delta |
|------|---------|--------------|-----------|-----------------|---------|-------|
| Open, stat and read 1000 files | 8,000 | 127 | 4.6 | 125 | 4.5 | -1.6% |
| Writing one large file (100 MB) | 103 | 41 | 1.1 | 41 | 1.3 | 0.0% |
| Reading one large file (100 MB) | 102 | 35 | 2.6 | 35 | 2.0 | 0.0% |
| Locking/Unlocking 10000x each | 20,000 | 241 | 6.8 | 236 | 5.0 | -2.1% |
| Creating dir with 2000 files | 4,000 | 675 | 17.8 | 673 | 24.7 | -0.3% |
| Enumerate dir with 2000 files | ~51 | 3 | 0.7 | 3 | 0.8 | 0.0% |
| Deleting dir with 2000 files | 2,000 | 636 | 27.1 | 600 | 16.4 | **-5.7%** |
| Create directory tree (1000 dirs) | 1,110 | 367 | 25.7 | 354 | 9.9 | **-3.5%** |
| Directory cache hits | 11,000 | 167 | 5.8 | 166 | 6.0 | -0.6% |
| Mixed cache operations | 820 | 133 | 5.6 | 132 | 4.5 | -0.8% |
| Deep path traversal | 3,500 | 71 | 4.6 | 70 | 4.1 | -1.4% |
| Cache validation efficiency | 30,000 | 444 | 14.5 | 436 | 12.3 | -1.8% |
| **Total** | **80,686** | **2,940** | | **2,871** | | **-2.3%** |
| **Average per AFP OP** | | **0.036 ms** | | **0.036 ms** | | **-2.3%** |

### Summary

- **2.3% overall improvement** (2,940ms → 2,871ms for 80,686 AFP ops)
- **All tests equal or faster** — no regressions
- **Delete 2000 files: -5.7%** (636ms → 600ms) — deferred child cleanup reduces hot-path latency
- **Create 1000 dirs: -3.5%** (367ms → 354ms) — deferred `dir_free_invalid_q()` reduces per-command overhead
- **Standard deviations consistently lower** with idle worker (more stable performance)
- Zero overhead on cache-hit hot path (cache hits: 167ms → 166ms, within noise)

---

### Idle Worker Thread Lifecycle

```mermaid
stateDiagram-v2
    [*] --> Sleeping: idle_worker_init()

    Sleeping --> Working: idle_worker_start()<br/>[is_idle=1, condvar signal]
    Working --> Sleeping: Work complete<br/>[bg_running=0]
    Working --> Interrupted: idle_worker_stop()<br/>[is_idle=0]
    Interrupted --> Sleeping: Worker sees is_idle==0<br/>[bg_running=0]

    Sleeping --> Terminated: idle_worker_shutdown()<br/>[shutdown_flag=1]
    Working --> Terminated: idle_worker_shutdown()<br/>[shutdown_flag=1]

    Sleeping --> ProcessExit: idle_worker_stop_signal_safe()<br/>[is_idle=0, no spin → exit()]
    Working --> ProcessExit: idle_worker_stop_signal_safe()<br/>[is_idle=0, no spin → exit()]

    note right of Working
        Worker checks is_idle per hash chain.
        bg_running=1 while active.
    end note

    note right of Sleeping
        Worker holds sleep_mutex,
        blocked in pthread_cond_wait().
        bg_running=0.
    end note

    note right of ProcessExit
        Signal handler context only.
        All paths end in exit().
        Kernel terminates worker thread.
    end note
```

### DSI Loop Integration Points

```mermaid
flowchart TD
    A["[A] DSI buffer check"] -->|data buffered| BREAK1[break — NO idle]
    A -->|no data| B["[B] State check"]
    B -->|special state| BREAK2[break — NO idle]
    B -->|normal| C["[C] Setup poll fds"]
    C --> SIGBLOCK["Block signals<br/>(SIGURG/SIGALRM/SIGTERM/SIGQUIT)"]
    SIGBLOCK --> START["idle_worker_start()"]
    START --> SIGRESTORE["Restore signal mask"]
    SIGRESTORE --> D["[D] BLOCK IN poll()"]
    D --> STOP["idle_worker_stop()"]
    STOP --> E["[E] poll error handling"]
    E -->|EINTR| A
    E -->|error| BREAK3[break]
    E -->|ok| F["[F] Hint pipe handling"]
    F --> G["[G] Socket error handling"]
    G --> H["[H] Client data ready?"]
    H -->|POLLIN| BREAK4[break — process command]
    H -->|no| I["[I] Hint only — loop back"]
    I --> A

    style SIGBLOCK fill:#FF9800,color:#fff
    style START fill:#4CAF50,color:#fff
    style SIGRESTORE fill:#FF9800,color:#fff
    style STOP fill:#F44336,color:#fff
```

### Deferred Job: dircache_remove_children

```mermaid
flowchart LR
    subgraph "AFP Command (main thread)"
        RC[dircache_remove_children_defer] -->|queue full?| SYNC[Synchronous fallback]
        RC -->|queue has room| ENQ["Enqueue: strdup path,<br/>store vid, chain_idx=0"]
    end

    subgraph "Idle Worker (worker thread)"
        CHK[Check deferred_count > 0] --> VOL{getvolbyvid?}
        VOL -->|NULL| DISCARD[Free path, advance head]
        VOL -->|valid| CHAIN["Walk one hash chain<br/>collect up to 16 matches"]
        CHAIN --> REM["dir_remove_and_free()<br/>for each match"]
        REM --> ADV{Batch full?}
        ADV -->|yes| CHK
        ADV -->|no| NEXT[chain_idx++]
        NEXT --> DONE{All chains done?}
        DONE -->|yes| COMPLETE[Free path, advance head]
        DONE -->|no| CHK
    end

    ENQ -.->|"temporal separation<br/>(poll idle)"| CHK
```

### DIFF
--- a/doc/developer/dircache.md
+++ b/doc/developer/dircache.md
@@ -512,6 +512,62 @@ Cache is undersized for working set. Increase `dircache size`.
 
 ---
 
+## Part 8: Idle Worker Thread — Background Cache Maintenance
+
+Netatalk includes an idle worker thread that performs non-critical dircache
+housekeeping during `poll()` idle periods, when the main AFP command thread
+is blocked waiting for client data.
+
+### Architecture
+
+The idle worker uses **temporal separation** rather than locks: the worker
+thread and main thread never access shared dircache data concurrently.
+
+- `idle_worker_start()` — called just before `poll()`, sets `is_idle=1`
+  and wakes the worker via condvar
+- `idle_worker_stop()` — called immediately after `poll()` returns, sets
+  `is_idle=0` and spins until the worker finishes its current unit of work
+- The worker checks `is_idle` per hash chain, yielding instantly when the
+  main thread reclaims access
+
+### Deferred Work Types
+
+1. **Invalid entry queue draining** — `dir_remove()` enqueues freed entries
+   into `invalid_dircache_entries`; the worker drains this queue during idle
+   periods instead of at the end of every AFP command
+2. **Deferred child cleanup** — `dircache_remove_children_defer()` enqueues
+   an O(1) descriptor; the worker walks hash chains incrementally, removing
+   stale children in batches of 16 per chain
+
+### Graceful Degradation
+
+If the worker thread fails to start (e.g., `pthread_create()` failure or
+platforms without lock-free atomics), all operations fall back to synchronous
+behavior transparently. The `idle_worker_is_active()` guard ensures the
+`dir_free_invalid_q()` call is preserved in the DSIFUNC_CMD handler when
+the worker is not running.
+
+### Signal Safety
+
+`afp_dsi_close()` can be called from signal handler context. It uses
+`idle_worker_stop_signal_safe()` which performs only a single atomic store
+(no spin, no mutex). Since all `afp_dsi_close()` paths end in `exit()`,
+the kernel terminates the worker thread.
+
+### Statistics
+
+Worker statistics are logged at session close:
+
+```txt
+idle_worker stats: cycles=1234 completed=1100 interrupted=134
+```
+
+- **cycles**: Total idle cycles entered
+- **completed**: Cycles that finished all pending work
+- **interrupted**: Cycles cut short by `poll()` returning (client data arrived)
+
+---
+
 ## Conclusion
 
 The real benefit of this optimization isn't just the eliminated stat() calls—
--- a/etc/afpd/afp_dsi.c
+++ b/etc/afpd/afp_dsi.c
@@ -50,6 +50,7 @@
 #include "dircache.h"
 #include "directory.h"
 #include "fork.h"
+#include "idle_worker.h"
 #include "switch.h"
 
 #ifndef SOL_TCP
@@ -84,6 +85,12 @@ static void afp_dsi_close(AFPObj *obj)
 {
     DSI *dsi = obj->dsi;
     sigset_t sigs;
+    /* Sets is_idle=0 — no spin. Async-signal-safe (single atomic store only).
+     * Worker stops at next is_idle check. No spin needed because all
+     * afp_dsi_close() paths end in exit() which terminates the worker. */
+    idle_worker_stop_signal_safe();
+    /* Log idle worker stats AFTER stop but BEFORE log_dircache_stat() */
+    idle_worker_log_stats();
     close(obj->ipc_fd);
     obj->ipc_fd = -1;
 
@@ -512,6 +519,9 @@ void afp_over_dsi(AFPObj *obj)
         afp_dsi_die(EXITERR_SYS);
     }
 
+    /* Start idle worker thread — failure is non-fatal (synchronous fallback) */
+    (void)idle_worker_init();
+
     /* set TCP snd/rcv buf */
     if (obj->options.tcp_rcvbuf) {
         if (setsockopt(dsi->socket,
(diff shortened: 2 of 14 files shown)


andylemin:
The `dircache_remove_children` calls are no longer visible in flamegraph! 🚀
dircache_remove_children() is still in the top 10 leaf functions, but you cannot see it in the graph! (this is expected).
ie, total runtime should be the same, as overall the same work is done - entries are still cleaned, however they are now deferred until idle or after all tests (when no AFP commands in buffer) - spectest never pauses..

_Runtime will increase by 1s due to required sleep in test127 (tests external file change detection) - FlameGraph shows runtime increased by exactly 1 second, so no performance regression._
However user level testing shows latency has been fixed as AFP commands are not blocked.

andylemin:
PS; @rdmark none of the GitHub workflows were historically running with debug build options, so the AFP_ASSERT invariants would never have fired - which was a test gap.

The new flamegraph workflow does use debug so will now also correctly fail if any invariants fire..

rdmark:
@andylemin what do you make of the performance degradation in the "Writing one large file (100 MB)" test? was it consistently slower across multiple runs?

andylemin:
> @andylemin what do you make of the performance degradation in the "Writing one large file (100 MB)" test? was it consistently slower across multiple runs?

Hi @rdmark that test is IO bound, and on my M2 laptop IO has a lot of jitter as Docker virtualises the entire IO layer rather than emulating. To be fair I also have CPU jitter as well, but not as much and I can more easily smooth out CPU jitter with lots of test iterations.

Apple has too many background processes.. so I agree this needs deeper testing on a dedicated server without docker or resource contention.

This PR will allow me to kick that off, I'll get both on-CPU as well as off-CPU graphs, for all spectest, lantest and speedtest with high iteration counts 👍

My latest changes to speedtest entirely supersedes that test now

andylemin:
Hi @rdmark I updated the results with tests from a baremetal server with no other processes running.

% gains are much tighter when the total runtime for lantest is already less than 3 seconds on this machine, so the hash scan overhead is very low already - but the gains are consistently there.

Doing manual user testing between the baseline without this PR and with the PR you can tell the difference - this non-blocking AFP command change is all about latency rather than throughput. Should be quite noticeable on slower old machines.

Thanks! This one took a lot of research.. multi-threading is always tricky, especially with Netatalk using various Signals under normal operations - I learnt a lot writing this one 🙂

PR #2810, opened 2026-03-23
docs: more helpful description of the 'cnid dev' option

rdmark:
describe concrete consequences and preconditions when using this option

### DIFF
--- a/doc/manpages/man5/afp.conf.5.md
+++ b/doc/manpages/man5/afp.conf.5.md
@@ -1265,8 +1265,13 @@ earlier did not notify kCaseSensitive flag to the client. Starting with
 
 cnid dev = *BOOLEAN* (default: *yes*) **(V)**
 
-> Whether to use the device number in the CNID backends. Helps when the
-device number is not constant across a reboot, e.g. cluster, ...
+> Whether to store the unique device number for each file or directory in the CNID backend.
+This is used to detect when a file is moved across devices,
+which allows Netatalk to update the CNID database accordingly.
+>
+> When using this option, make sure that the device number of the volume is consistent across reboots,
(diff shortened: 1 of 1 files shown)


rdmark:
@NJRoadfan I had something like this in mind; thoughts?

doc/manpages/man5/afp.conf.5.md
```diff
@@ -1265,8 +1265,13 @@ earlier did not notify kCaseSensitive flag to the client. Starting with
 
 cnid dev = *BOOLEAN* (default: *yes*) **(V)**
 
-> Whether to use the device number in the CNID backends. Helps when the
-device number is not constant across a reboot, e.g. cluster, ...
+> Whether to store the device number for each file or directory in the CNID backend.
+This is used to detect when a file is moved across devices,
+which allows Netatalk to update the CNID database accordingly.
+>
```
andylemin, line 1271: A device number is a unique identifier assigned to a volume or disk device, helping the system distinguish different resources.

A part of me wants to explain that this is not limited to fancy clusters, many file systems use dynamic ids

For example the user should configure configure stable identifiers instead of volatile device numbers when configuring systems that must survive reboots:
Filesystems: use LABEL= or UUID= in /etc/fstab (or equivalent) rather than /dev/sdX.
Mount-management tools: systemd mount units use UUID/LABEL or /dev/disk/by-*/ symlinks.

Netatalk: point shares at mount points or use the stable device identifiers (UUID/LABEL) so Netatalk sees a consistent path*

But this is too long for here

NJRoadfan:
Looks fine to me.

doc/manpages/man5/afp.conf.5.md
rdmark: I like the idea of having a description that is more actionable and less resigned to fate :)

pushed a new revision -- how about this wording?

PR #2822, opened 2026-03-28
afpd: log when the root directory is name locked or rename fails

rdmark:
the 'name locked' FinderFlag prevents both the renaming of the AFP volume as well as the changing of its Finder icon

adding logging for both the flag and a failure to rename to help users pinpoint why their volumes appear locked

thanks to Scott Small for the report and analysis

### DIFF
--- a/etc/afpd/filedir.c
+++ b/etc/afpd/filedir.c
@@ -628,6 +628,8 @@ int afp_rename(AFPObj *obj, char *ibuf, size_t ibuflen _U_, char *rbuf _U_,
         }
     } else {
         if (sdir->d_did == DIRDID_ROOT) {   /* root directory */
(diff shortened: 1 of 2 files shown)


andylemin:
@rdmark if this has been raised then I assume we are not retuning a useful error code to the user?

The logs are good, but an error code that causes a popup might be the most useful?

Not near computer to check properly, but the log diff looks fine

andylemin (approved):

rdmark:
@andylemin the challenge is that the Finder (Mac OS 7.6, 9.2) silently disallows name or icon operations when the *name locked* flag is set, specifically when it comes to icons the Get Info dialog disallows you from selecting the volume icon for replacement

I can take one more look at the AFP spec and see if there's some logic that we're missing

PR #2826, opened 2026-03-29
docs, fixes: update profiling, compilation, manpage, meson options, dircache max size and other minor issues

andylemin:
- dircache: fix MAX_DIRCACHE_SIZE from 2M to 1M entries; ARC mode already doubles internally for ghosts (c cached + c ghosts = 2c total hash entries), and MAX_DIRCACHE_SIZE only used for user config parsing.
- afp.conf.5.md: update dircache size maximum to 1M
- Configuration.md: update dircache size maximum to 1048576
- afp_dsi: add idle_worker_shutdown() calls before all session close/exit paths to ensure clean worker thread termination
- idle_worker: log stats during shutdown while state is still valid
- containers.yml: clarify flamegraph github workflow with "(AFP_ASSERT active)" - this is the only workflow where AFP_ASSERT's will fire, so if it fails we should make it clear what the difference is
- COMPILATION.md: add optional FreeBSD `libsunacl` package for ACL support
- profiling.md: fix markdown heading syntax
- meson_options.txt: fix with-debugging description (Enable -> Disable) to clarify that when debugging is enabled with-debugging = true, both tickles and sigalarms are disabled.

### DIFF
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -49,7 +49,7 @@ option(
     'with-debugging',
     type: 'boolean',
     value: false,
-    description: 'Enable SIGALARM timers and DSI tickles (eg for debugging with gdb/dbx/...)',
+    description: 'Disable SIGALARM timers and DSI tickles (useful for debugging with gdb/dbx/...)',
 )
 option(
     'with-docs-install-path',
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -520,6 +520,7 @@ jobs:
               iniparser \
               libevent \
               libgcrypt \
+              libsunacl \
               localsearch \
               meson \
               mysql91-client \
--- a/.github/workflows/containers.yml
+++ b/.github/workflows/containers.yml
@@ -195,7 +195,7 @@ jobs:
           tags: ${{ steps.metadata.outputs.tags }}
 
   flamegraph-spectest:
-    name: Flamegraph (spectest AFP 3.4)
+    name: Flamegraph (spectest AFP 3.4) (AFP_ASSERT active)
     needs: build-container-debug
     runs-on: ubuntu-latest
     timeout-minutes: 15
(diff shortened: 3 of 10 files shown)


COMPILATION.md
```diff
@@ -366,6 +366,12 @@ Install required packages
 pkg install -y avahi bison cmark db5 flex iniparser libevent libgcrypt localsearch meson mysql91-client openldap26-client p5-Net-DBus perl5 pkgconf sqlite3 talloc
 ```
 
+Optional: Install `libsunacl` to enable ACL support
+
+```shell
+pkg install -y libsunacl
+```
+
```
rdmark: FYI this markdown file is automatically created by `contrib/scripts/make_compile_docs.pl` so this change will be overwritten the next time we run the script (which happens during release preparation)

one way to achieve a similar effect would be to add this to the FreeBSD job in `.github/workflows/build.yml`

```yaml
            pkg install -y \
              ...
              talloc
            # Optional: Install `libsunacl` to enable ACL support
            pkg install -y libsunacl
```

alternatively, why not add *libsunacl* to the standard list of ports to install? the other build jobs install ACL libraries where applicable!

for historical context, we used to have a suite of wiki pages with detailed installation instructions for each platforms, which descriptions of mandatory and optional packaged mapped to specific features; however, this was a major overhead to keep up to date, which it why I switched to this hybrid solution

meson_options.txt
```diff
@@ -49,7 +49,7 @@ option(
     'with-debugging',
     type: 'boolean',
     value: false,
-    description: 'Enable SIGALARM timers and DSI tickles (eg for debugging with gdb/dbx/...)',
+    description: 'Disable SIGALARM timers and DSI tickles (useful for debugging with gdb/dbx/...)',
```
rdmark, line 52: not sure about SIGALARM, but I think it's accurate that this disables DSI tickles

andylemin: Hi @rdmark 

The SIGALRM timer setup is guarded by #ifndef DEBUGGING in two files:

etc/afpd/afp_dsi.c:480–489: The tickle handler's SIGALRM signal and ITIMER_REAL timer are skipped when DEBUGGING is defined:
```
#ifndef DEBUGGING
    /* SIGALRM - tickle handler */
    action.sa_handler = alarm_handler;
    if ((sigaction(SIGALRM, &action, NULL) < 0) ||
            (setitimer(ITIMER_REAL, &dsi->timer, NULL) < 0)) {
        afp_dsi_die(EXITERR_SYS);
    }
#endif /* DEBUGGING */
```

libatalk/dsi/dsi_tcp.c:155–170: The connection timeout SIGALRM handler and ITIMER_REAL timer are skipped when DEBUGGING is defined:
```
#ifndef DEBUGGING
    /* install an alarm to deal with non-responsive connections */
    newact.sa_handler = timeout_handler;
    ...
    if ((sigaction(SIGALRM, &newact, &oldact) < 0) ||
            (setitimer(ITIMER_REAL, &timer, NULL) < 0)) {
        ...
    }
#endif
```

COMPILATION.md
andylemin: Thanks, I have moved libsunacl to .github/workflows/build.yml 👍

meson_options.txt
rdmark: never mind, my brain could parse the logic now 😅

PR #2827, opened 2026-03-29
dircache optimisation: replace strncmp with blength pre-check + memcmp in child scan

andylemin:
Replace strncmp() with (size_t)blength() pre-check and memcmp() in three descendant-path matching sites: dircache_remove_children(), dircache_flush_deferred_for_vol(), and dircache_process_deferred_chain().

The blength() pre-check is an O(1) integer comparison that rejects entries with paths shorter than or equal to the parent allowing a quick skip by avoiding string comparison entirely.
memcmp() enables SIMD-optimized comparison without per-byte NUL checking overhead which is much faster than the string compare.

I have just compared a before and after flame graph;

## Key Results

| Metric | Before | After | Improvement |
|--------|--------|-------|-------------|
| `strncmp` CPU share | **4.24%** (73M samples) | **0.05%** (1M) | **99% eliminated** |
| Child-removal total CPU | **5.63%** (97M) | **3.76%** (72M) | **33% reduction** |

### Baseline

| Symbol | Samples | % of Total |
|--------|---------|------------|
| `strncmp` (in `dircache_remove_children`) | 56,056,056 | 3.25% |
| `strncmp` (in `dircache_process_deferred_chain`) | 16,016,016 | 0.93% |
| `strncmp@plt` | 1,001,001 | 0.06% |
| **Total `strncmp`** | **73,073,073** | **4.24%** |

### Optimized (this)

| Symbol | Samples | % of Total |
|--------|---------|------------|
| `strncmp` (unrelated path) | 1,001,001 | 0.05% |
| `memcmp` | 1,001,001 | 0.05% |
| **Total comparison** | **2,002,002** | **0.10%** |

The `blength()` O(1) pre-check filters out most cache entries before any byte comparison occurs, reducing comparison overhead from 73M to 2M samples — a **97% reduction**.

This change should significantly harden Netatalk against refresh attacks and reduce CPU - this cleaning was already deferred to idle times, while not handling AFP user commands, but we still want it to be faster.

The deferred scan's 3.76% is dominated by irreducible O(N) hash traversal (`hash_chain_head`, loop overhead), not string comparison.

Confirmed on ARM64 as well;

flamegraph_arm64_memcmp_spectest

### DIFF
--- a/etc/afpd/dircache.c
+++ b/etc/afpd/dircache.c
@@ -1715,7 +1715,8 @@ int dircache_remove_children(const struct vol *vol, struct dir *dir)
 
             /* Check if entry's path starts with parent's path followed by '/' */
             if (parent_path && entry_path && parent_len > 0 &&
-                    strncmp(entry_path, parent_path, parent_len) == 0 &&
+                    (size_t)blength(entry->d_fullpath) > parent_len &&
+                    memcmp(entry_path, parent_path, parent_len) == 0 &&
                     entry_path[parent_len] == '/') {
                 /* Expand to_remove array */
                 if (remove_count >= remove_capacity) {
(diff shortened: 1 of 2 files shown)


andylemin:
A casual visual comparison of the flamegraph with previous PRs nicely shows the DSI/TCP stacks are starting to dominate now 🙂

rdmark (approved):

PR #2829, opened 2026-03-30
bump to bstring subproject v1.1.0

rdmark:
release notes for bstring v1.1.0: https://github.com/msteinert/bstring/releases/tag/v1.1.0

immediately relevant for netatalk is primarily the memory safety bugfixes

it might be worthwhile to closer at the new UTF-8 string manipulation functions for future use

### DIFF
--- a/subprojects/bstring.wrap
+++ b/subprojects/bstring.wrap
@@ -1,8 +1,8 @@
 [wrap-file]
-directory = bstring-1.0.3
-source_url = https://github.com/msteinert/bstring/releases/download/v1.0.3/bstring-1.0.3.tar.xz
-source_filename = bstring-1.0.3.tar.xz
-source_hash = 90db08fd33e9494aea3f00f9b71cdcf3114c65457ee35558e8274df6ebac43f3
+directory = bstring-1.1.0
+source_url = https://github.com/msteinert/bstring/releases/download/v1.1.0/bstring-1.1.0.tar.xz
+source_filename = bstring-1.1.0.tar.xz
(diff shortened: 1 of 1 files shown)


rdmark:
@andylemin I did a visual check on the flamegraph compared to the most recent previous one in https://github.com/Netatalk/netatalk/pull/2828 and could observe some variations, most notably a higher peak of the big central pillar (like a narrow but tall branch at the very right hand side) -- are these normal variations on these GitHub runners would you say?

andylemin:
> @andylemin I did a visual check on the flamegraph compared to the most recent previous one in https://github.com/Netatalk/netatalk/pull/2828 and could observe some variations, most notably a higher peak of the big central pillar (like a narrow but tall branch at the very right hand side) -- are these normal variations on these GitHub runners would you say?

Hi.
These are not workflow variations, my graphs are exactly the same every time, they change when you change the code. As long as the runtime is the same then it was a like for like (if the runtime is higher then the workflow CPU might have been contented and a function might have been sampled a few more times, but should never change heights, only the widths by fractions of a percentage).

I have been able to very reliably see impact to change by change. For example in the new IPC relay changes, one of the code fixes caused a performance regression and I could see it immediately.

The higher the stack the more functions calls that were made. So if changing to the new bstring version makes the flames higher, it just means they have added more function calls. This is fine unless they are recursive calls (which would make it wider) without inline'ing functions.

What we care about are the widths.. the graph is all percentage based, and we want the flames relating to Netatalk operations to be narrower.

andylemin:
Interesting that there are new internal UTF-8 string operations! With MAC-UTF8 I agree there could be an opportunity there

PR #2830, opened 2026-03-30
idle_worker: replace condvar signal with timedwait self-wake

andylemin:
Replace the pthread_cond_signal() call in idle_worker_start() with a 1ms pthread_cond_timedwait() self-wake loop in the worker thread, instead of explicitly starting and stopping the idle worker thread before and after every poll().

idle_worker_start() now uses a single atomic_store (async-signal-safe) eliminating the futex(FUTEX_WAKE) and try_to_wake_up kernel calls that consumed 2.84% CPU and the pthread_sigmask calls that consumed 1.24% CPU during spectest profiling.

Changes:
- idle_worker.c: Replace cond_wait with cond_timedwait (1ms interval)
- idle_worker.c: Strip condvar signal from idle_worker_start()
- afp_dsi.c: Remove idle_block sigset setup and pthread_sigmask calls

Tradeoff: worker wakes every 1ms (~0.05% CPU) instead of on-demand. Background cleanup latency increases from immediate to ≤1ms, which is acceptable for non-latency-sensitive dircache garbage collection.

The primary benefit is the removal of the kernel scheduler overhead making the idle worker thread truly non-blocking, reducing AFP command latency. The CPU reduction also helps

---

| Metric | Baseline | Optimized | Change |
|--------|----------|-----------|--------|
| `idle_worker_start` total samples | 46,046,046 | **0** | **Eliminated** |
| `idle_worker_start` CPU % | **2.84%** | **0%** | **-2.84%** |
| `pthread_sigmask` samples | 20,020,020 | **0** | **Eliminated** |
| `pthread_sigmask` CPU % | **1.24%** | **0%** | **-1.24%** |
| `pthread_cond_signal` | 1,001,001 (0.06%) | **0 (0%)** | **Eliminated** |
| `try_to_wake_up` | 6,006,006 (0.37%) | 2,002,002 (0.12%) | **-66% (residual is unrelated)** |

As idle_worker_start is now a single `atomic_store(&is_idle, 1)` (~5ns) it is too fast for the 999 Hz perf sampler to capture.

The `pthread_sigmask` calls in the DSI loop existed solely to protect the mutex inside `idle_worker_start()` from `siglongjmp()` corruption. With no mutex, they were removed.

Total CPU Recovered

| Overhead Source | Baseline CPU % | Optimized CPU % | Recovered |
|----------------|---------------|-----------------|-----------|
| `idle_worker_start` body | 2.84% | 0% | 2.84% |
| `pthread_sigmask` (2 calls) | 1.24% | 0% | 1.24% |
| **Total** | **4.08%** | **0%** | **4.08%** |

Worker Thread: Increased Useful Work

The worker thread now self-wakes every 1ms via `pthread_cond_timedwait()`, continuously (while main thread in poll) discovering and processing deferred cleanup work instead of being awakened only on-demand.

| Metric | Baseline | Optimized | Change |
|--------|----------|-----------|--------|
| `idle_worker_main` total | 46,046,046 (2.85%) | 125,125,125 (7.88%) | +5.03% |
| `idle_worker_has_work` | 0 (0%) | 9,009,009 (0.57%) | New (predicate polling) |
| `idle_worker_stop` | 1,001,001 (0.06%) | 1,001,001 (0.06%) | Unchanged |

The `idle_worker_main` increase is expected and beneficial — the worker is processing more dircache cleanup work per spectest run because it now checks for work during the entire main `poll()` duration, rather than only processing work that existed at the instant `idle_worker_start()` was called.

Net Impact Summary

| Aspect | Before | After |
|--------|--------|-------|
| `idle_worker_start()` cost per call | ~2000-6000ns | ~5ns |
| Main thread overhead (start + sigmask) | 4.08% CPU | 0% CPU |
| Worker thread useful work | 2.85% CPU | 7.88% CPU |
| Worker idle overhead | 0% (sleeping) | ~0.05% (timedwait polling) |
| Signal masking in DSI loop | Required | Eliminated |
| `idle_worker_start()` async-signal-safe | No | Yes |

ARM64 Flamegraph;

flamegraph_idle_worker_optimised

### DIFF
--- a/etc/afpd/idle_worker.c
+++ b/etc/afpd/idle_worker.c
@@ -21,6 +21,7 @@
 #include <signal.h>
 #include <stdatomic.h>
 #include <string.h>
+#include <time.h>
 
 #include <atalk/logger.h>
 #include <atalk/queue.h>
@@ -29,6 +30,11 @@
 #include "directory.h"
 #include "idle_worker.h"
 
+/* Self-wake interval for the idle worker thread. 1ms ~0.05% CPU overhead */
+#define IDLE_WORKER_WAKE_MS 1
+_Static_assert(IDLE_WORKER_WAKE_MS < 1000,
+               "Use a while loop for nanosecond normalization if IDLE_WORKER_WAKE_MS >= 1s");
+
 /* Atomic coordination flags.
  * Using default memory_order_seq_cst for all atomic operations. Could use
  * acquire/release for ~25ns savings on ARM, but seq_cst is simpler and the
@@ -36,9 +42,12 @@
 static atomic_int is_idle = 0;
 static atomic_int bg_running = 0;
 
-/* Condvar for sleep/wake — mutex only held during condvar wait */
+/* Condvar + mutex for worker timedwait loop and shutdown signal.
+ * The mutex is required by pthread_cond_timedwait() and provides
+ * a memory barrier for the non-atomic queue reads in idle_worker_has_work().
+ * idle_worker_shutdown() signals the condvar for immediate wakeup. */
 static pthread_mutex_t sleep_mutex = PTHREAD_MUTEX_INITIALIZER;
-static pthread_cond_t  sleep_cond  = PTHREAD_COND_INITIALIZER;
+static pthread_cond_t  sleep_cond;
 static atomic_int      shutdown_flag = 0;
 
 static pthread_t worker_tid;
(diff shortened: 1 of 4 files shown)


etc/afpd/idle_worker.c
```diff
@@ -210,18 +226,7 @@ void idle_worker_start(void)
...
-     * by main thread during AFP command processing (before this point).
-     * If no work exists, worker stays sleeping */
-    if ((invalid_dircache_entries &&
-            invalid_dircache_entries->next != invalid_dircache_entries) ||
-            dircache_has_deferred_work()) {
-        pthread_mutex_lock(&sleep_mutex);
-        pthread_cond_signal(&sleep_cond);
-        pthread_mutex_unlock(&sleep_mutex);
-    }
+    /* No condvar signal — worker self-wakes via timedwait */
```
rdmark: do you think future contributors / readers of the code will get value out of this code comment?

I'm a little wary of leaving remnants of refactorings unless they describe a some kind of unusual design decision or constraint!

PR #2831, opened 2026-03-31
IPC-Hints: dircache hints relay now batches IPC writes

andylemin:
Replace synchronous per-hint write() relay with batched accumulation and a dedicated flush thread. Hints are buffered under a single mutex and flushed every 50ms or when the 128-entry buffer fills, using PIPE_BUF-safe chunked writes with priority sorting.

- Add flush thread with 50ms timer and buffer-full signal
- Add per-child rate limiter (1000 hints/sec attack guard)
- Add single-session skip (servch_count <= 1)
- Add server_child_add() capacity enforcement
- Replace per-hint read() receiver with bulk framed read
- Add 1s sleep to test545 for batched hint delivery timing

Parent CPU for hint relay: 21.45% -> 3.61% (83% reduction).

---

Before vs After — Spectest Workload

| Function | Baseline (%) | Batching (%) | Change | Reduction |
|---|---|---|---|---|
| `ipc_relay_cache_hint()` | **20.52%** | **0.10%** | −20.42pp | **99.5%** |
| `ipc_server_read()` | **21.45%** | **1.10%** | −20.35pp | **94.9%** |
| `hint_flush_thread()` | N/A | **2.51%** | +2.51pp | (new) |
| **Total IPC hint overhead** | **21.45%** | **3.61%** | −17.84pp | **83.2%** |

1. **`ipc_relay_cache_hint()`** dropped from 20.52% to 0.10% — the function now only does a buffer append under a ~100ns mutex hold instead of N×`write()` syscalls per hint.

2. **`ipc_server_read()`** dropped from 21.45% to 1.10% — the parent's IPC read path is no longer dominated by synchronous hint relay.

3. **`hint_flush_thread()`** appears at 2.51% — this is the dedicated flush thread performing batched `write()` calls. The cost of writing is now amortized across many hints per `write()`.

4. **Net reduction**: Total IPC hint overhead went from 21.45% → 3.61% of parent CPU, a **83.2% reduction** in parent-process CPU consumed by hint relay.

**_pipe_write stack almost entirely eradicated from Flamegraph_!** 🎉

**ARM64 BEFORE**
flamegraph_idle_worker_optimised

**ARM64 AFTER**
flamegraph_ipc_batching

The x86_64 github workflow Flamegraph in the comments below also shows a significant improvement towards most time being spent on TCP now 🚀

### DIFF
--- a/etc/afpd/auth.c
+++ b/etc/afpd/auth.c
@@ -742,11 +742,15 @@ int afp_login(AFPObj *obj, char *ibuf, size_t ibuflen, char *rbuf,
     }
 
     i = afp_uam->u.uam_login.login(obj, &pwd, ibuf, ibuflen, rbuf, rbuflen);
+    LOG(log_debug, logtype_afpd, "afp_login: UAM login returned %d, pwd=%p", i,
+        pwd);
 
     if (!pwd || (i != AFP_OK && i != AFPERR_PWDEXPR)) {
+        LOG(log_error, logtype_afpd, "afp_login: UAM login failed, sending reply");
         return send_reply(obj, i);
     }
 
+    LOG(log_debug, logtype_afpd, "afp_login: UAM login OK, calling login()");
     return send_reply(obj, login(obj, pwd, afp_uam->u.uam_login.logout,
                                  ((i == AFPERR_PWDEXPR) ? 1 : 0)));
 }
--- a/etc/uams/uams_pam.c
+++ b/etc/uams/uams_pam.c
@@ -153,7 +153,7 @@ static int login(void *obj, char *username, int ulen,  struct passwd **uam_pwd,
 
     if (uam_afpserver_option(obj, UAM_OPTION_CLIENTNAME,
                              (void *) &hostname, NULL) < 0) {
-        LOG(log_info, logtype_uams,
+        LOG(log_error, logtype_uams,
             "uams_pam.c :PAM: unable to retrieve client hostname");
         hostname = NULL;
     }
@@ -174,21 +174,26 @@ static int login(void *obj, char *username, int ulen,  struct passwd **uam_pwd,
         return AFPERR_NOTAUTH;
     }
 
-    LOG(log_info, logtype_uams, "cleartext login: %s", username);
+    LOG(log_warning, logtype_uams, "cleartext login: %s", username);
     PAM_username = username;
     PAM_password = ibuf; /* Set these things up for the conv function */
     err = AFPERR_NOTAUTH;
+    LOG(log_debug, logtype_uams, "PAM: calling pam_start for %s", username);
     PAM_error = pam_start("netatalk", username, &PAM_conversation,
                           &pamh);
 
     if (PAM_error != PAM_SUCCESS) {
+        LOG(log_error, logtype_uams, "PAM: pam_start failed: %d", PAM_error);
         goto login_err;
     }
 
+    LOG(log_debug, logtype_uams, "PAM: pam_start OK, calling pam_set_item");
     pam_set_item(pamh, PAM_TTY, "afpd");
     pam_set_item(pamh, PAM_RHOST, hostname);
     /* use PAM_DISALLOW_NULL_AUTHTOK if passwdminlen > 0 */
+    LOG(log_debug, logtype_uams, "PAM: calling pam_authenticate");
     PAM_error = pam_authenticate(pamh, 0);
+    LOG(log_debug, logtype_uams, "PAM: pam_authenticate returned: %d", PAM_error);
 
     if (PAM_error != PAM_SUCCESS) {
         if (PAM_error == PAM_MAXTRIES) {
@@ -198,7 +203,9 @@ static int login(void *obj, char *username, int ulen,  struct passwd **uam_pwd,
         goto login_err;
     }
 
+    LOG(log_debug, logtype_uams, "PAM: calling pam_acct_mgmt");
     PAM_error = pam_acct_mgmt(pamh, 0);
+    LOG(log_debug, logtype_uams, "PAM: pam_acct_mgmt returned: %d", PAM_error);
 
     if (PAM_error != PAM_SUCCESS) {
         /* Password change required */
@@ -220,19 +227,24 @@ static int login(void *obj, char *username, int ulen,  struct passwd **uam_pwd,
 #ifndef PAM_CRED_ESTABLISH
 #define PAM_CRED_ESTABLISH PAM_ESTABLISH_CRED
 #endif /* PAM_CRED_ESTABLISH */
+    LOG(log_debug, logtype_uams, "PAM: calling pam_setcred");
     PAM_error = pam_setcred(pamh, PAM_CRED_ESTABLISH);
+    LOG(log_debug, logtype_uams, "PAM: pam_setcred returned: %d", PAM_error);
 
     if (PAM_error != PAM_SUCCESS) {
         goto login_err;
     }
 
+    LOG(log_debug, logtype_uams, "PAM: calling pam_open_session");
     PAM_error = pam_open_session(pamh, 0);
+    LOG(log_debug, logtype_uams, "PAM: pam_open_session returned: %d", PAM_error);
 
     if (PAM_error != PAM_SUCCESS) {
         goto login_err;
     }
 
     *uam_pwd = pwd;
+    LOG(log_debug, logtype_uams, "PAM: login complete for %s", username);
 
     if (err == AFPERR_PWDEXPR) {
         return err;
@@ -436,7 +448,7 @@ static int pam_printer(char *start, char *stop, char *username,
     data = (char *)malloc(stop - start + 1);
 
     if (!data) {
-        LOG(log_info, logtype_uams, "Bad Login ClearTxtUAM: malloc");
+        LOG(log_error, logtype_uams, "Bad Login ClearTxtUAM: malloc");
         return -1;
     }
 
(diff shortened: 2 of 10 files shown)


andylemin:
One of the fixes for one of the augmentcode issues introduced a regression. I'll fix over easter weekend.. 👍

andylemin:
Performance regression fixed, was due to an overly strict defensive coding for an impossible edge case to silence the augmentcode reviewer. Will resolve the macos logins now

andylemin:
@rdmark Fixed the macos login issue.
The root cause was that the parent is multithreaded when it fork's. On macOS, this causes the child's pam_opendirectory.so to SIGABRT because macOS detects a fork from a multithreaded parent and restricts Mach/XPC APIs that Open Directory PAM uses. I tried stopping and restarting the thread during fork, and confirmed macOS does not reset the `__is_threaded` flag when threads are stopped.
This explains why afpstats is hard disabled in code on macOS, as afpstats also uses a thread in the parent..

The Netatalk fork design is not compatible with threads in the parent. Therefore I simplified the design (removed the writer thread). So now this change just just adds batching and sorting in the main process loop, to write all hints to all users in atomic-sized write's.

Just some final debugging to do now

rdmark:
@andylemin that's right, I spent a day or two bashing my head against the afpd threading model on macOS, trying to get afpstats to not crash. Ultimately I landed at the conclusion that it would take a complete redesign, and hard coded afpstats to be disabled on macOS. It was a learning experience!

andylemin:
> @andylemin that's right, I spent a day or two bashing my head against the afpd threading model on macOS, trying to get afpstats to not crash. Ultimately I landed at the conclusion that it would take a complete redesign, and hard coded afpstats to be disabled on macOS. It was a learning experience!

Funny that we both went though the same experience and came to the same conclusion 😉 Painful 2 person verification! My conclusion was spawning child processes for user sessions would need to be done in a dedicated side-car process, and not done by the central orchestrator/sync/parent process.

I do want to solve this but much farther down the road.. Once Netatalk is only left with micro-optimisations (almost there) and I have spent time with afpfs-ng/OpenAFPfs (is rfork friendly) and is also down to only micro-optimisations left - I want to try the multi-TCP stream idea (legacy clients open only one connection, while OpenAFPfs could open more) - we will need the sidecar to route additional tertiary connections to the correct session child. It takes a minimum of 4 TCP sessions to fill a wifi channel or a long-fat pipe, just an immutable fact of TCP on links with high bandwidth delay products and/or high RTT jitter..

andylemin:
@rdmark after spending a lot of time on the fork'ed threading dead-end, this is now looking good. About 50% CPU time is networking, and ~30% is storage. Just ~20% left in Netatalk internally.

Maybe one or two small optimisations left before I can have a go at OpenAFPfs 🙂

PS; Interesting to think that much of the ~50% Network time is spent waiting, and much of the ~30% IO time is also spent waiting.. This is a great "problem" to have 😉 Now the real fun starts as we are close to exploring experimental solutions

etc/afpd/auth.c
```diff
@@ -741,12 +741,16 @@ int afp_login(AFPObj *obj, char *ibuf, size_t ibuflen, char *rbuf,
         return send_reply(obj, i);
     }
 
+    LOG(log_info, logtype_afpd, "afp_login: calling UAM login");
     i = afp_uam->u.uam_login.login(obj, &pwd, ibuf, ibuflen, rbuf, rbuflen);
+    LOG(log_info, logtype_afpd, "afp_login: UAM login returned %d, pwd=%p", i, pwd);
 
     if (!pwd || (i != AFP_OK && i != AFPERR_PWDEXPR)) {
+        LOG(log_info, logtype_afpd, "afp_login: UAM login failed, sending reply");
```
rdmark: should this be warning level logging?

etc/afpd/auth.c
```diff
@@ -741,12 +741,16 @@ int afp_login(AFPObj *obj, char *ibuf, size_t ibuflen, char *rbuf,
         return send_reply(obj, i);
     }
 
+    LOG(log_info, logtype_afpd, "afp_login: calling UAM login");
     i = afp_uam->u.uam_login.login(obj, &pwd, ibuf, ibuflen, rbuf, rbuflen);
+    LOG(log_info, logtype_afpd, "afp_login: UAM login returned %d, pwd=%p", i, pwd);
```
rdmark: shall these be debug level logging?

etc/afpd/auth.c
```diff
@@ -741,12 +741,16 @@ int afp_login(AFPObj *obj, char *ibuf, size_t ibuflen, char *rbuf,
...
 
     if (!pwd || (i != AFP_OK && i != AFPERR_PWDEXPR)) {
+        LOG(log_info, logtype_afpd, "afp_login: UAM login failed, sending reply");
         return send_reply(obj, i);
     }
 
+    LOG(log_info, logtype_afpd, "afp_login: UAM login OK, calling login()");
```
rdmark: also feels more like debug logging

etc/uams/uams_pam.c
```diff
@@ -178,17 +178,22 @@ static int login(void *obj, char *username, int ulen,  struct passwd **uam_pwd,
     PAM_username = username;
     PAM_password = ibuf; /* Set these things up for the conv function */
     err = AFPERR_NOTAUTH;
+    LOG(log_info, logtype_uams, "PAM: calling pam_start for %s", username);
```
rdmark: most of the logging in this file also looks like debug logging! do you intend to clean it up before merging?

libatalk/dsi/dsi_getsess.c
```diff
@@ -76,6 +76,8 @@ int dsi_getsession(DSI *dsi, server_child_t *serv_children, int tickleval,
         return -1;
     }
 
+    LOG(log_info, logtype_dsi, "dsi_getsess: about to fork child (proto_open)");
```
rdmark: the logging in this file may need clean-up as well!

rdmark:
@andylemin it would be neat to have a diagram to demonstrate how the cache hint relay works in the big picture!

perhaps updating some of the Dev Docs with this, f.e. https://github.com/Netatalk/netatalk/wiki/Dev-Docs-Data-Flow

etc/afpd/auth.c
andylemin: Thanks for the reminder - yes I need to clean up all the logging now I've got everything working

andylemin:
> @andylemin it would be neat to have a diagram to demonstrate how the cache hint relay works in the big picture!
> 
> perhaps updating some of the Dev Docs with this, f.e. https://github.com/Netatalk/netatalk/wiki/Dev-Docs-Data-Flow

Done.. And working through the other Dev Docs pages too..

@rdmark This is now ready for final review

https://github.com/Netatalk/netatalk/wiki/Dev-Docs-Caching-Architecture#cross-process-cache-invalidation-ipc-hint-system

etc/uams/uams_pam.c
```diff
@@ -174,21 +174,26 @@ static int login(void *obj, char *username, int ulen,  struct passwd **uam_pwd,
         return AFPERR_NOTAUTH;
     }
 
-    LOG(log_info, logtype_uams, "cleartext login: %s", username);
+    LOG(log_warning, logtype_uams, "cleartext login: %s", username);
```
rdmark, line 177: I don't really see the argument for making this a warning level log?

seems like the code path of "normal" clrtxt UAM operation, but maybe I'm wrong!

etc/uams/uams_pam.c
```diff
@@ -561,7 +573,7 @@ static int pam_printer(char *start, char *stop, char *username,
     /* Login successful, but no need to hang onto it,
        so logout immediately */
     append(out, loginok, strlen(loginok));
-    LOG(log_info, logtype_uams, "Login ClearTxtUAM: %s", username);
+    LOG(log_warning, logtype_uams, "Login ClearTxtUAM: %s", username);
```
rdmark, line 576: not sure about this being a warning level log either

andylemin: I think a warning for a cleartext login makes sense?
It should only be used in labs or rare situations, and you should ideally be reminded you are running a risky insecure config?

etc/uams/uams_pam.c
andylemin: I think a warning for a cleartext login makes sense?
It should only be used in labs or rare situations, and you should ideally be reminded you are running a risky insecure config?

andylemin: Most systems do not even allow this at all?

andylemin:
> neat concept! have you tested with a larger number of concurrent users beyond two? just thinking how the IPC will scale when several afpd processes operate on the same volume, what the chance of synchronization issues is.

Hi @rdmark Funny you should say that 😉 https://github.com/Netatalk/netatalk/pull/2853

etc/uams/uams_pam.c
andylemin: If it is ok with you, I want to be pushy on this one and say we should make this UAM noisy - they are insecure so there should be some price to pay

rdmark: the clrtxt UAM is the only option for certain vintage computing setups, notable when netbooting an older Mac over AFP, or using Apple II ProDOS AFP clients

it's also an option for System 6 / System 7 Macs when you don't want to jump through the hoops to configure RandNum ;)

not opposed to keeping the warning level logging -- perhaps the wording of the message can be made more helpful for the user instead?

also, we ought to add the equivalent logging to etc/uams/uams_passwd.c, the non-PAM variant of the same UAM

rdmark: will file a new ticket for this!

PR #2850, opened 2026-04-04
ci: build on OpenBSD without search indexing stack

rdmark:
the localsearch/tinysparql stack on OpenBSD 7.8 pulls in pretty much all of the GNOME desktop environment, taking up unnecessary GBs of space on a headless server

let's save some time and storage by not building with Spotlight support in the CI build


rdmark:
this reduced the OpenBSD build job runtime from 14m to 4m!

andylemin (approved):

PR #2851, opened 2026-04-04
afpd: force-close stale forks during FPDelete to unblock deletion

andylemin:
Non-compliant AFP clients (e.g. Preview.app) sometimes leave open forks after completing operations, causing FPDelete to return AFPERR_BUSY when trying to delete after viewing.

Add of_close_stale_forks() which iterates all open forks matching the target file (by dev/ino) and force-closes any fork that passes both safety checks:
  - Not DIRTY (no pending AD header metadata flush)
  - No outstanding locks on the shared adouble

AFPFORK_MODIFIED is intentionally not checked since the file is being deleted and any data written via FPWrite is already on disk.

In afp_delete(), when of_findname() detects open forks, attempt stale fork cleanup before returning AFPERR_BUSY. If all forks are closed, re-verify with of_findname() and proceed to deletion. If any fork remains active (dirty or locked), return AFPERR_BUSY as before.

This is safe because:
  - No data loss: file is being deleted, written data is on disk
  - No lock violations: explicit lock check before force-close
  - Per-FD refcounting respected: of_closefork() properly decrements
  - Scope limited to FPDelete: normal fork lifecycle unaffected
  - Post-close behavior identical to session disconnect cleanup

Extensive investigation and debugging (2 weeks of testing with logs and Wireshark - I found it hard to believe Apple are the ones who violate the spec sometimes) was performed to validate that this is NOT a Netatalk issue.
Tracing AFP operations from some clients (such as macOS Ventura) using specific applications, it was proven that the AFP spec is sometimes violated by clients. 
The spec states if a client opens a file for Read, and opens it for Writing later, the client must still close both open's - not just the most privileged open.

_As there is only a single file descriptor on the client, it is understandable that sometimes this tracking fails and only one close is received for the client's fd (instead of an exact close for each original open), leaving a stale fork on the server - subsequently blocking deletions._

This forced-cleanup does not violate the AFP specification.

### DIFF
--- a/etc/afpd/ofork.c
+++ b/etc/afpd/ofork.c
@@ -203,7 +203,9 @@ of_alloc(struct vol *vol,
     lastrefnum = refnum;
 
     if (i == nforks) {
-        LOG(log_error, logtype_afpd, "of_alloc: maximum number of forks exceeded.");
+        LOG(log_error, logtype_afpd,
+            "of_alloc: maximum number of forks (%d) reached for "
+            "\"%s\"", nforks, path);
         return NULL;
     }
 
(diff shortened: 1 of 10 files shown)


etc/afpd/ofork.c
```diff
@@ -670,6 +672,90 @@ int of_closefork(const AFPObj *obj, struct ofork *ofork)
...
+            LOG(log_warning, logtype_afpd,
+                "of_close_stale_forks: force-closing stale fork refnum:%u "
+                "name:\"%s\" flags:0x%x access:%s (not dirty, no locks)",
+                of->of_refnum, of_name(of), of->of_flags,
+                (of->of_flags & AFPFORK_ACCWR) ? "write" :
+                (of->of_flags & AFPFORK_ACCRD) ? "read" : "none");
+            /* Note: of_closefork() may emit FCE_FILE_MODIFY for MODIFIED
+             * stale forks. This is harmless — the subsequent deletefile()
+             * emits FCE_FILE_DELETE, and well-behaved FCE consumers handle
+             * MODIFY→DELETE sequences correctly.
```
rdmark, line 772: I didn't know there are live FCE consumers out there, well-behaved or not :-D

it's a hypothetical statement I presume?

rdmark (approved):
does Preview.app have an actual built-in AFP client?

do you know of other macOS apps that do as well?

etc/afpd/ofork.c
andylemin: Yea pretty much, but we should still assume there are.

I've also been meaning to move it into the idle worker and make it a build conditional as it's in the hot path

andylemin:
> does Preview.app have an actual built-in AFP client?
> 
> do you know of other macOS apps that do as well?

Well this is one of those changes that has affected me personally for a long time.

It seems to be related to the Ventura (and some other versions - not tested all) kernel module itself, with any application that peeks or previews (quick view) data (including Preview, image magick CLI on terminal, and Photoshop) - many tools used by photographers.

I suspect these kinds of apps open read only, and the kernel module upgrades to open an extra Write mode on behalf of the app, but does not close all when the app closes.

Needs some testing before merge

PR #2863, opened 2026-04-06
dbd: add -i option to invalidate AppleDouble CNID hints

rdmark:
When nested volumes leave behind stale CNID hints in AppleDouble files, dbd -f alone cannot fix them because cnid_add() still receives the stale hint. The new -i flag skips reading AD hints entirely, letting the database assign fresh CNIDs which are then written back to the AppleDouble files.

Also fix two bugs in check_cnid() with DBD_FLAGS_STRIP_AD set: on AD_VERSION_EA volumes, files without a meta EA would incorrectly return CNID_INVALID because the adfile_ok probe was skipped; and the CNID overwrite path lacked a DBD_FLAGS_SCAN guard, causing filesystem writes in read-only scan mode.

Finally, remove vestigal -r option argument that has no logic

### DIFF
--- a/bin/dbd/cmd_dbd.c
+++ b/bin/dbd/cmd_dbd.c
@@ -140,7 +140,7 @@ int main(int argc, char **argv)
     const char *username = NULL;
     int c;
 
-    while ((c = getopt(argc, argv, ":cfF:irstu:vV")) != -1) {
+    while ((c = getopt(argc, argv, ":cfF:istu:vV")) != -1) {
         switch (c) {
         case 'c':
             flags |= DBD_FLAGS_V2TOEA;
@@ -156,10 +156,6 @@ int main(int argc, char **argv)
 
         case 'F':
             obj.cmdlineconfigfile = strdup(optarg);
-            break;
-
-        case 'r':
-            /* the default */
             break;
 
         case 's':
--- a/bin/dbd/cmd_dbd.c
+++ b/bin/dbd/cmd_dbd.c
@@ -95,7 +95,7 @@ static void set_signal(void)
 
 static void usage(void)
 {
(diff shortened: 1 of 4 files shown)


andylemin:
This is useful, however I'm not convinced -f cannot fix them?

On rebuild Dbd -f will see that the CNID hint does not match the database, and the database wins, and the CNID ID from the DB is used overwriting the CNID hint.

If the DB has no entry it saves the CNID hint into the DB checking the unique constraint on the ID key.
So any secondary duplicate CNID hint will have a key collision at which point it falls back to generating a fresh ID.

Am I missing something?
I understand this might be related to the different vol's but the CNIDs are unique per vol, so after the vol's are separated the rebuild still should work?

Approving as I like the start fresh ability anyway, just want to confirm my understanding

andylemin (approved):

rdmark:
this is fair feedback, I wrote this as a hypothetical measure for the scenario where an invalid CNID was stored as AD hint, but I only tested with "normal" CNID hint repair scenarios

let me bring this back to draft status and gather some evidence that the new logic is actually required

andylemin:
> this is fair feedback, I wrote this as a hypothetical measure for the scenario where an invalid CNID was stored as AD hint, but I only tested with "normal" CNID hint repair scenarios
> 
> 
> 
> let me bring this back to draft status and gather some evidence that the new logic is actually required

I still like this 😉 resetting all CNIDs is feel-good

rdmark:
augment review

rdmark:
> I still like this 😉 resetting all CNIDs is feel-good

right, I like it too! the idea is to have another tool in the toolbox when a user reports some corner case fail state with their CNID db and appledouble metadata in a weird state – just run the command and start fresh

PR #2874, opened 2026-04-09
idle_worker: replace condvar with nanosleep to eliminate high cadence syscall and futex

andylemin:
Replace pthread_cond_timedwait() + mutex with nanosleep() for the idle worker's 1ms polling loop. On musl libc, condvar timedwait uses futex() internally, causing intermittent 18% overhead in on-CPU flamegraphs when perf's sampling frequency synchronized with the 1ms wake interval.

nanosleep() is a direct SYS_nanosleep syscall (no hidden futex), so the worker thread sleeps via hrtimer and appears off-CPU during profiling. This also removes the clock_gettime() call per iteration, reducing per-wake syscall count from 2 to 1.

Removes: pthread mutex, condvar, condattr init, HAVE_PTHREAD_CONDATTR_SETCLOCK ifdefs, clock selection logic, and nanosecond normalization code.

Validated: 311/311 spectest pass, flamegraph confirms 0 futex stacks (previously up to 18% when perf timer aligned with condvar wake cycle).

### DIFF
--- a/doc/developer/dircache.md
+++ b/doc/developer/dircache.md
@@ -524,7 +524,7 @@ The idle worker uses **temporal separation** rather than locks: the worker
 thread and main thread never access shared dircache data concurrently.
 
 - `idle_worker_start()` — called just before `poll()`, sets `is_idle=1`
-  and wakes the worker via condvar
+  (the worker self-wakes every 1ms via `nanosleep()` and checks this flag)
 - `idle_worker_stop()` — called immediately after `poll()` returns, sets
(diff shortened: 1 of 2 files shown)


andylemin:
Hi @rdmark sorry for the crazy push noise on the other PRs!
I was trying to solve the random futex stacks that were appearing in the flamegraphs..
Required a couple of changes (perf freq value change to avoid idle worker synchronisation, and this PR which was triggering a futex at the same time as perf freq).

All 4 open PRs are finalised and finally ready for review

rdmark (approved):

PR #2876, opened 2026-04-10
add SRP user authentication method

### LINKED ISSUE #2842: reverse engineer the SRP UAM
if the SRP UAM can be reverse engineered, it can potentially provide a stronger encryption than DHX2 with macOS clients

----

original summary from https://sourceforge.net/p/netatalk/feature-requests/68/

Apple added an undocumented UAM called "SRP" which is used by TCs.
SRP probably stands for [Secure Remote Password](http://srp.stanford.edu/).

TC device password

Image

TC drive password

Image

TC account password

Image

rdmark:
Implement the SRP (Secure Remote Password) UAM for AFP, as used by Apple Time Capsule.
The protocol uses SRP-6a with SHA-1, MGF1 KDF, and RFC 5054 group #2 (1536-bit).

afppasswd is modified to operate on SRP storage file by default, which stores per-user salts and verifiers, while retaining the legacy RandNum functionality activated with the -r flag.

An additional breaking change is that afppasswd -a now takes username as argument rather than previous positional username arugment.

The containers now use SRP for authentication by default.


rdmark:
~~while SRP auth works afpfs-ng -> netatalk, it doesn't work macOS -> netatalk yet so we have a bug somewhere~~

false alarm, it works when you use the correct password 😅

NJRoadfan:
Tested and working here.

andylemin:
Very cool! Well done indeed.. I'll do a proper code review in the next couple of days

andylemin:
QR review shows this is a high quality PR 🙂 This looks to be a well-engineered implementation of the SRP-6a protocol.
The cryptographic primitives are correct, sensitive data is properly cleaned up, the code follows established codebase patterns, error handling is thorough, and the documentation is comprehensive. RFC 5054 N prime constants look good.

Minor comments;
Remove unused SRP_AUTH_FAILURE macro if not planned for future use.
Consider constant-time comparison for M1 verification (memcmp → custom ct_memcmp) as a defense-in-depth measure.

LGTM - Impressive work! 🚀

andylemin (approved):

rdmark:
@andylemin thanks for the review!

>  Remove unused SRP_AUTH_FAILURE macro if not planned for future use.

I have been going back and forth about this; `SRP_AUTH_FAILURE` (-6754) is what I observed the Time Capsule AFP server returned when the M1 proof is invalid (both incorrect crypto as well as incorrect user password), however this error code is not defined in he AFP spec. Right now I'm using `AFPERR_NOTAUTH` (-5023) consistently to match the other UAMs which the macOS AFP client doesn't seem to mind.

Do you have any advice here, whether to emulate Time Capsule fully or attempt to adhere to the AFP spec where appropriate?

andylemin:
> @andylemin thanks for the review!
> 
> > Remove unused SRP_AUTH_FAILURE macro if not planned for future use.
> 
> I have been going back and forth about this; `SRP_AUTH_FAILURE` (-6754) is what I observed the Time Capsule AFP server returned when the M1 proof is invalid (both incorrect crypto as well as incorrect user password), however this error code is not defined in he AFP spec. Right now I'm using `AFPERR_NOTAUTH` (-5023) consistently to match the other UAMs which the macOS AFP client doesn't seem to mind.
> 
> Do you have any advice here, whether to emulate Time Capsule fully or attempt to adhere to the AFP spec where appropriate?

I think we have seen enough examples of Apple not complying with the AFP spec themselves, so I think we should go with what is required for Time Capsule and just acknowledge any spec violations?

rdmark:
> I think we have seen enough examples of Apple not complying with the AFP spec themselves, so I think we should go with what is required for Time Capsule and just acknowledge any spec violations?

let's go with this! I updated the PR so that srp_logincont() returns SRP_AUTH_FAILURE when the M1 comparison fails (e.g. password doesn't match)

> Consider constant-time comparison for M1 verification (memcmp → custom ct_memcmp) as a defense-in-depth measure.

done! added a ct_memcmp() function for this purpose. does this match what you were imagining? @andylemin

rdmark:
as a side note, I acquired a 2010 MacBook running OSX 10.7.5 and can confirm that the AFP client is able to authenticate with netatalk's SRP UAM, so we know now that SRP is at least included since Lion in 2011

PR #2878, opened 2026-04-10
Spotlight: Fix timestamps and add additional attributes

### LINKED ISSUE #2258: Support ctime and atime
**Is your feature request related to a problem? Please describe.**

Clients are only shown the `mtime` (not sure if this is technically correct, but it's usually the closest to "Created" in Finder) for a file, and it is displayed for a file's `ctime` ("Modified") and `atime` ("Last opened") as well.

**Describe the solution you'd like**

`ctime` and `atime` to be displayed to the client, as the server can read them fine.

**Describe alternatives you've considered**

To verify that the `ctime` and `atime` work on AFP, I tested sharing a directory on a Mac running Snow Leopard. These dates display and update fine on the client side when shared natively from a Mac, but I'm running `netatalk4` on FreeBSD for my AFP server.

**Additional context**

My ZFS pool and datasets and configured with `atime=on`, and I can see on the server (using `ls -u`) that the `atime` is updated appropriately after a file is opened, but it is never displayed for the client.

When working through files sequentially, I use the "Last opened" time in Finder to know which file I looked at last in a directory.

NJRoadfan:
Upstream patch from Samba. Fixes: #2258

### DIFF
--- a/etc/afpd/spotlight.c
+++ b/etc/afpd/spotlight.c
@@ -48,6 +48,14 @@
 
 #define MAX_SL_RESULTS 20
 
+struct timeval convert_timespec_to_timeval(const struct timespec ts)
+{
+    struct timeval tv;
+    tv.tv_sec = ts.tv_sec;
+    tv.tv_usec = ts.tv_nsec / 1000;
+    return tv;
+}
+
 struct slq_state_names {
     slq_state_t state;
     const char *state_name;
@@ -346,7 +354,8 @@ static bool add_filemeta(sl_array_t *reqinfo,
 
     for (i = 0; i < metacount; i++) {
         if (strequal(reqinfo->dd_talloc_array[i], "kMDItemDisplayName")
-                || strequal(reqinfo->dd_talloc_array[i], "kMDItemFSName")) {
+                || strequal(reqinfo->dd_talloc_array[i], "kMDItemFSName")
+                || strequal(reqinfo->dd_talloc_array[i], "_kMDItemFileName")) {
             if ((p = strrchr(path, '/'))) {
                 name = dalloc_strdup(meta, p + 1);
                 dalloc_add(meta, name, "char *");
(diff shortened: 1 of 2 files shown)


rdmark:
great patch, I'm pleased to see the spotlight code getting some long-overdue love

I was thinking about attribution since this is ported from Samba; seems like portions were authored by Ralph in https://github.com/samba-team/samba/commit/5d4eceae1579de341e69a090a3983e323591ce8e and https://github.com/samba-team/samba/commit/5d4eceae1579de341e69a090a3983e323591ce8e

the `convert_timespec_to_timeval` structure was authored by Jeremy Allison in 2007 https://github.com/samba-team/samba/commit/4952fe368a40b239140b3035db6075427d237bb9 but maybe it can be considered boilerplate / trivial?

NJRoadfan:
Some (all?) platforms have a macro called `TIMESPEC_TO_TIMEVAL` in `<sys/time.h>`.

rdmark (approved):

rdmark:
heads-up that this patch caused a breakage: when spotlight is active it's not possible to edit a file in-place on an AFP volume, e.g. with TextEdit on macOS

TextEdit warns that "another application has modified the file" and prevents you from saving, probably due to the timestamp changing in a way that it doesn't expect

NJRoadfan:
Steps to reproduce?

TextEdit (at least in Catalina) is not producing the same issue on my end with Spotlight enabled. I would be curious to see if this can be reproduced on a Samba share with Spotlight enabled as it uses the same method of updating the "Last Opened" time on the server side.

I am opening a text file, making changes, then saving in place. I do get a note from TextEdit the first time I save that the file system doesn't support storing revisions, but other then that it works fine.

NJRoadfan:
Give #2896 a try. I noticed a regression where `mtime` was being set to -1 on the filesystem. This is likely causing problems elsewhere.

PR #2879, opened 2026-04-11
remove obsolete crypt.h header inclusion

rdmark:
libcrypt is a 25 year old dependency that is long since obsolete, so sign that any symbols from this library is required anymore (supplanted by libgcrypt)


rdmark:
I was wrong, libcrypt provides f.e. crypt_checkpass

PR #2882, opened 2026-04-12
[4.4] meson: fix bugs that prevent you from disabling dev docs

rdmark:
two silly mistakes introduced when first creating the developer docs build scripts made it so that they were always built regardles of meson options

Reported-by: Kilian Kempf


rdmark:
@andylemin with this I am proposing we cut a 4.4.2 release: we arguably have critical mass of useful patches in the stable branch! thoughts?

andylemin (approved):

PR #2883, opened 2026-04-12
[4.4] meson: detect and configure endianness preprocessor flag

rdmark:
replace the Autotool-ism for endianness macro that doesn't work with Meson, with some surrounding cleanup

we can now build with big-endian byte order support on f.e. s390x architecture


rdmark:
on second thought, there were a handful of other fixes needed for bigendian, let's not backport this

PR #2889, opened 2026-04-14
ci: switch to NetBSD pkgsrc mirror for build job

rdmark:
the main ftp.netbsd.org mirror has been down for most of the day so let's switch over to a mirror for now; using the one of the few that seems to work is the French mirror (Universite de Pierre et Marie Curie)


rdmark:
@andylemin this should unblock the pipeline for us, currently failing due to an extended ftp.netbsd.org outage

PR #2895, opened 2026-04-18
TimeLord: Fix local time offset when DST is in effect.

NJRoadfan:
On platforms lacking `tm_gmtoff`, TimeLord is utilizing the `timezone` variable which does not take Daylight Saving Time into account in its calculation of local time. Check `tm_isdst` and adjust the time if needed.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -352,6 +352,9 @@ endif
 if cc.has_member('struct stat', 'st_birthtimespec', prefix: '#include <sys/stat.h>')
     cdata.set('HAVE_STRUCT_STAT_ST_BIRTHTIMESPEC', 1)
 endif
+if cc.has_member('struct stat', 'st_birthtimespec', prefix: '#include <sys/stat.h>')
+    cdata.set('HAVE_STRUCT_STAT_ST_BIRTHTIMESPEC', 1)
+endif
 
 #################
 # Configuration #
--- a/meson.build
+++ b/meson.build
@@ -352,9 +352,6 @@ endif
 if cc.has_member('struct stat', 'st_birthtimespec', prefix: '#include <sys/stat.h>')
     cdata.set('HAVE_STRUCT_STAT_ST_BIRTHTIMESPEC', 1)
 endif
-if cc.has_member('struct stat', 'st_birthtimespec', prefix: '#include <sys/stat.h>')
-    cdata.set('HAVE_STRUCT_STAT_ST_BIRTHTIMESPEC', 1)
-endif
 
 #################
 # Configuration #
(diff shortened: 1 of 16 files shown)


NJRoadfan:
`augment review`

NJRoadfan:
Looking at the supported OS list (at least the ones in the CI queue for building), all of them support `timegm()`. Solaris was the last holdout, but it looks like it was added in Solaris 11.

Also, we might consider removing this compatibility macro since only NetBSD and Linux support DDP AppleTalk these days and they both support `tm_gmtoff`.

rdmark:
I'm generally in favor of eliminating obsolete compatibility macros to reduce complexity.

when it comes to AppleTalk, I have the vague hope that we will get a general-purpose userland stack some day so I aim to keep the AppleTalk code reasonably cross-platform compatible...

rather than assuming timegm() is supported, do we want a capability check, and then keep the fallback logic that doesn't take DST into consideration when that fails too? to have 3 branches effectively.

rdmark:
@NJRoadfan can you please give thumbs up or down to the augmentcode comments, and then close them?

NJRoadfan:
Another option is to implement a simple `timegm()` function for platforms that lack it. That or we can manually add an hour offset as I did in an earlier revision if `tm_isdst` is set...... which wouldn't work on Lord Howe Island (the only place in the world with a 30 min DST offset!).

rdmark:
I like this idea, if you're up for it

NJRoadfan:
Some background on the problem: https://berthub.eu/articles/posts/how-to-get-a-unix-epoch-from-a-utc-date-time-string/

The PR actually abuses `timegm()` to create a `time_t` value from LOCAL time, not UTC.

rdmark (approved):

rdmark:
that was a great read; I feel inclined to accept your patch now

NJRoadfan:
We can optionally remove the `HAVE_STRUCT_TM_TM_GMTOFF` macro. This method is more portable.

rdmark:
Sure, please go ahead.

meson.build
```diff
@@ -349,8 +349,8 @@ endforeach
 if cc.has_type('fshare_t', prefix: '#include <fcntl.h>')
     cdata.set('HAVE_FSHARE_T', 1)
 endif
-if cc.has_member('struct tm', 'tm_gmtoff', prefix: '#include <time.h>')
-    cdata.set('HAVE_STRUCT_TM_TM_GMTOFF', 1)
+if cc.has_member('struct stat', 'st_birthtimespec', prefix: '#include <sys/stat.h>')
+    cdata.set('HAVE_STRUCT_STAT_ST_BIRTHTIMESPEC', 1)
```
rdmark, line 353: isn't this duplicate logic to the lines just below?

NJRoadfan: Something happened with the rebase. I'll fix it later.

rdmark (approved):

rdmark:
reducing preprocessor macro sprawl; I like it

PR #2896, opened 2026-04-18
Spotlight: Revert some changes to timestamps

NJRoadfan:


NJRoadfan:
Unless a `stat()` call is made to retrieve a file's `mtime` beforehand, you can't update just the `atime` with `utime()`, hence the error with the `mtime` being reset everytime the client sends `kMDItemLastUsedDate` back to the server. We don't want to needlessly trigger a `stat()` call each time a file is accessed, so this revert will have to do for now. Note that some OSes, such as Linux mount their filesystems with the `relatime` option which won't update the `atime` of a file on each actual access to lower overhead, so the "Last Accessed" time in Finder won't always be accurate.

rdmark:
@NJRoadfan can you take a look at this PR that might address the issue without reverting -- https://github.com/Netatalk/netatalk/pull/2897

this resolves the issue I was seeing (host is Linux, client is macOS)

do you see any reason why calling utimensat() every time here would be a bad idea?

NJRoadfan:
It should be fine for this use. Is utimensat() available on all platforms? I'm seeing references that older macOS (<10.13) don't support it and the call isn't used anywhere else in the Netatalk code base.

rdmark:
utimensat is POSIX 1-2008 and contemporaneous with other at-functions that we rely on

my second internalized justification was that the Spotlight stack (Localsearch or recent Tracker etc) is unlikely to be usable on older OSes anyways, so they won't be compiling this code

that said a capability check for utimensat with a portable fallback is definitely the more prudent solution

rdmark:
I attempted a combination of stat() and utime() as fallback but it led to race conditions and other bugs, I'm leaning towards making utimensat mandatory for Spotlight

rdmark:
The other PR had been merged now so let me discard this one

NJRoadfan:
In the future, it may be advisable to eventually retrieve and store file stats out of whatever database is backing Spotlight. Running queries against tracker shows the database already stores this info.

PR #2897, opened 2026-04-18
afpd: use utimensat when storing kMDItemLastUsedDate for Spotlight

rdmark:
The previous utime() call left utimbuf.modtime uninitialised on the stack. utime() writes both atime and mtime, so the file's mtime was clobbered with garbage. Apps such as TextEdit compare the mtime they recorded at open time against the current mtime before saving; seeing an unexpected change they abort the save with "file modified by another application".

Replace the utime() call with utimensat(AT_FDCWD, path, ts, 0) where ts[1].tv_nsec = UTIME_OMIT, which updates only the access time and leaves the modification time strictly unchanged.

Fixes breakage introduced in https://github.com/Netatalk/netatalk/pull/2878


rdmark:
augment review

rdmark:
augment review

andylemin (approved):

PR #2900, opened 2026-04-19
afpd: improved Spotlight RPC logic and bugfixes

rdmark:
a range of improvements and bugfixes that should align closer to macOS Spotlight RPC behavior, based on analysis of Samba RPC logic

- call stat() before add_filemeta() in tracker_cursor_cb
- return Spotlight status 35 for all pre-DONE states in add_results()
- handle fetchAllAttributes:forOIDArray:context: RPC for Spotlight
- handle kMDItemContentModificationDate alias in Spotlight add_filemeta()
- handle kMDItemContentCreationDate in Spotlight add_filemeta()
- enrich fetchPropertiesForContext reply dict for Spotlight
- Spotlight NFC to NFD path normalization in add_filemeta()
- improve Spotlight logging for query tracing
- guard against NULL query_results in tracker_cursor_cb
- afpd: safer buffer length checks in spotlight
- declare dalloc macros without final semicolon

### DIFF
--- a/etc/afpd/spotlight.c
+++ b/etc/afpd/spotlight.c
@@ -387,6 +387,14 @@ static bool add_filemeta(sl_array_t *reqinfo,
                             "kMDItemLastUsedDate")) {
             sl_time = convert_timespec_to_timeval(sp->st_atim);
             dalloc_add_copy(meta, &sl_time, sl_time_t);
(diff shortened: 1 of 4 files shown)


rdmark:
@andylemin any idea why we got such a funky looking flamegraph?

andylemin (approved):

PR #2902, opened 2026-04-19
netatalk: replace gsettings with dconf keyfile for indexer config

rdmark:
The current gsettings set approach to starting the localsearch indexer is unreliable at runtime because gsettings requires a dconf-service on the session bus, which netatalk's private dbus-daemon does not auto-activate.

Replace set_sl_volumes() with a dconf keyfile writer: it creates /etc/dconf/db/netatalk.d/10-spotlight and runs dconf update, which works without any session bus. Add DCONF_PROFILE env so the indexer reads from the netatalk-specific database. Ship config/dconf/netatalk (system-db:netatalk) as the dconf profile, installed to $sysconfdir/dconf/profile/. Detect the dconf binary at build time and gate have_spotlight on it instead of gsettings.

This introduces a new runtime dependency dconf instead of gsettings.

In meson, decouple indexer binary detection from GSettings schema detection:

On Debian Trixie, the localsearch binary is present but ships the Tracker3 GSettings schema rather than LocalSearch3. Decouple binary and schema selection so that any available binary (localsearch preferred, tracker3 as fallback) is paired with whichever schema is actually installed.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -182,6 +182,7 @@ jobs:
             ca-certificates \
             cmark-gfm \
             cracklib-runtime \
+            dconf-cli \
             file \
             flex \
             gcc \
@@ -342,6 +343,7 @@ jobs:
             ca-certificates \
             cmark-gfm \
             cracklib-runtime \
+            dconf-cli \
             file \
             flex \
             gcc \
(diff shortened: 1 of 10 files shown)


rdmark:
augment review

rdmark:
agument review

andylemin (approved):

PR #2905, opened 2026-04-19
ci: remove overwrite option from Upload Pages action

rdmark:
removed the 'overwrite' option from the Upload Pages action step, which is not a supported option for this action


andylemin (approved):
Thanks @rdmark. This was added in an attempt to allow forced rebuilds without a new commit id (pushing the same id breaks the deployment and the workflow), but it didn't work and should have been removed.

PR #2915, opened 2026-04-22
afpd: add support for Spotlight wildcard queries with SPARQL

rdmark:
Previously, '*' wildcard queries were mapped to ssmt_fts (fts:match only), which missed filename-only matches. The new ssmt_fts_or_fname type generates a SPARQL UNION query that combines fts:match with an nfo:fileName regex filter, enabling case-insensitive filename pattern matching for Spotlight searches.

### DIFF
--- a/etc/spotlight/sparql_map.c
+++ b/etc/spotlight/sparql_map.c
@@ -29,7 +29,7 @@
 
 struct spotlight_sparql_map spotlight_sparql_map[] = {
     /* ssm_spotlight_attr               ssm_enabled, ssm_type,   ssm_sparql_attr */
-    {"*",                               true, ssmt_fts,   "fts:match"},
+    {"*",                               true, ssmt_fts_or_fname, "fts:match"},
 
     /* Filesystem metadata */
     {"kMDItemFSLabel",                  true, ssmt_num,   NOTSUPPORTED},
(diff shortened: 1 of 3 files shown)


rdmark:
augment review

andylemin (approved):
Hurrah! Thanks @rdmark 

Would be great if the Spotlight backend could also be set to CNID (MySQL or SqLite - But not DBD) for super-fast and instantly up-to-date filename-only searches (for those who don't want to run DBUS etc) 😉

Seeing as you've done so much work in this area...

Also I think it's time to change `search db = BOOLEAN (default: no) (V)` to default to yes.
The logic behind this option needs checking and the default could be `auto` (check CNID before auto enable)?

Not quite related, but also related 🙃

rdmark:
indeed, I was pleased to see Spotlight search hits in macOS for the first time in 5 years of tinkering with netatalk :)

I agree that it is time to turn `search db` on by default -- we should be trusting the CNID database, and if ends up in a bad state at a regular basis that's a bug that should be surfaced and fixed

why don't you think DBD should be used as the Spotlight backend? it ostensibly supports cnid_find

andylemin:
Hi @rdmark 
I don't have anything against DBD being a backend. I just thought it was not practical or possible with DBD architecture.
Also with SQLite becoming the default I thought it would be a waste of effort. 

But, if cnid_find works then we should use it to be agnostic to the backend type (one change to support them all)?

PR #2923, opened 2026-04-23
afpd: introduce pluggable Spotlight search backend architecture

### LINKED ISSUE #1409: Spotlight search using data in MySQL CNID DB
**Is your feature request related to a problem? Please describe.**
- Setting up Tracker and DBUS etc, for Spotlight is difficult, complex, and fragile, and often does not work. It is a lot of effort just to be able to search using an index rather than sequential scan.

**Describe the solution you'd like**
- have a simple filename-only based Spotlight search using the MySQL CNID DB.

I have seen very few, if any, truly useful cases in production of searching via the metadata that Tracker provides. The vast majority of searches are just filename searches, and the metadata often just adds noise to the results. As a result you often have to configure the tracker to not scan for extra data anyway.

Being able to have a simple name only search using the existing MySQL would be fast, simple, stable and efficient.

It would also allow for instant Spotlight searches without having to wait for the index scans to catch up.

rdmark:
Replace the monolithic localsearch-only Spotlight implementation with a vtable-based backend system. The AFP RPC protocol layer (marshalling, slq_t lifecycle, add_filemeta, command dispatch) remains in spotlight.c and is shared by all backends. Query execution is delegated via a registered sl_backend_ops vtable stored in vol->v_search_backend.

Two backends are defined:
- cnid: uses cnid_find() for LIKE-based filename searches directly in the CNID database; no extra dependencies; enabled by default
- localsearch: the existing LocalSearch/tinysparql implementation, extracted from afpd/spotlight.c into sl_localsearch.c

Key changes:
- spotlight.h: remove LocalSearch/GLib includes and struct sl_ctx; add sl_backend_ops vtable; rename tracker_cursor -> slq_backend_private
- globals.h: AFPObj::sl_ctx becomes void * (opaque backend handle)
- volume.h: add v_search_backend_name and v_search_backend fields
- netatalk_conf.c: parse new per-volume "search backend" afp.conf option
- volume.c: resolve backend name to ops pointer after volume_openDB()
- meson_options.txt: replace with-spotlight boolean with with-search-backends array option (default: ['cnid'])
- All meson.build: have_spotlight -> have_search; GLib/sparql deps scoped to use_search_localsearch; etc/spotlight restructured with localsearch/ and cnid/ subdirs producing libspotlight_backends

New meson option -Dwith-search-backends of array type which takes cnid and localsearch; by default we attempt to build both.

The cnid backend has a single dependency, talloc, while the localsearch backend also requires the whole LocalSearch / TinySPARQL / GLib / D-Bus stack.

The newly added 'search backend' afp.conf option controls which backend that gets referenced at runtime. This option can be set at a global and volume level.

While the cnid backend is light-weight and does not require many support libraries, it can only do filename search
(but it does them very fast.) Any file contents or metadata searches in the Finder will be ignored. The backend will use the same CNID database as defined by the cnid scheme option for the volume in question.

The localsearch backend is full-featured and should handle file content and metadata search to the extent localsearch extractors are active on the host file system and has extracted the data you want to match. Localsearch is a pluggable architecture as well, and you can install additional extractors. See the localsearch documentation.

Also in the changeset:
- libatalk: add portable inline wrappers for stat time field access
- cnid: limit to 100 results from cnid_find() calls to mysql/sqlite 
- tone down log messages for child process termination
- libatalk: turn on spotlight and search db options by default


rdmark:
@andylemin this is ready for your review now!

I have tested that the `cnid` backend for Spotlight works with MySQL, SQLite, and BerkeleyDB, but only on fairly small and artificial volumes. I put a constraint in place that max 100 results will be returned from from cnid_find not to overwhelm the Spotlight subsystem. A fast follow would be to introduce some kind of pagination system to allow for more results.

Also please note that there's a change the Localsearch backend won't work properly: in localsearch v3.8 I consistently get segfault crashes in localsearch itself, due to the SPARQL queries containing UNION that I introduced recently for filename matching. Apparently a known bug that is fixed in more recent versions. At the same time, localsearch v3.11 has done away with the traditional Tracker style of spawning indexer daemons (`localsearch daemon` on the shell) so our glue code for kicking off the indexer in netatalk.c isn't compatible anymore. Right now my plan is to drop support for localsearch v3.9 and earlier, and support only v3.10 and later, which is the breaking point where the new way of spawning indexers was introduced.

andylemin:
Hi @rdmark 
This is great.

I understand why you went with replacing `with-spotlight` for `with-search-backends` if there is a new `search backend` afp.conf option.

However changing build options is hard on downstream, and I thought these only relate to Spotlight?

Would `with-spotlight=cnid` and `spotlight backend` be more intuitive?

Or do these changes affect `search db`?

andylemin:
Thanks for making this new capability the default and default enabled. I think that makes sense.

It might be worth mentioning the other benefit of this is new/renamed objects are instantly searchable unlike with localsearch

andylemin:
Absolutely agree with pagination.
Do we need to port any of the RPC fixes in Samba over to ensure the stream is robust?

I also agree with dropping support for legacy versions of `localsearch` - especially when local search is fragile. We should not import technical debt due to third party issues

rdmark:
> Hi @rdmark This is great.
> 
> I understand why you went with replacing `with-spotlight` for `with-search-backends` if there is a new `search backend` afp.conf option.
> 
> However changing build options is hard on downstream, and I thought these only relate to Spotlight?
> 
> Would `with-spotlight=cnid` and `spotlight backend` be more intuitive?
> 
> Or do these changes affect `search db`?

Fair point about not breaking things for downstream unnecessarily. I have brought back `-Dwith-spotlight` the boolean option now, set to *true* by default. This can be handy if you want to build without this feature even when dependencies are found.

Because of Meson quirks, I prefer to have two separate options for complex features like this, one boolean and one array. It makes the internal logic more legible and less fragile. (In my experience.)

And no, these changes do not affect `search db` at all, because it controls CatSearch behavior, and not Spotlight.

On using Search vs. Spotlight for naming options etc.: This becomes a bit of a longer story and I would like your opinion on this.

After much recent experimentation and research, I can only conclude that what Apple calls **Spotlight Search**, i.e. the search widget activated with Cmd+Space, does *not* support network file system search at all. I have tried with Apple's own SMB and AFP implementations, but there is no evidence whatsoever that Spotlight Search will ever match anything that is not locally indexed with Core Spotlight.

At the same time, **Finder Search** (the search bar in each Finder window) is able to search on both specific mounted remote file systems as well as the entire Network neighborhood. Finder Search uses the same Spotlight RPC technology behind the scenes, but the branding is not Spotlight as it were.

While the *spotlight* nomenclature is correct at a technical implementation level (e.g. implements the `FPSpotlightRPC` command) as an end user I would be confused when I enabled Spotlight in Netatalk, but still cannot search AFP volumes with... Spotlight Search. 😅

Therefore I was leaning towards using the *search* nomenclature more and more as I was writing this code. But I can see the argument that *spotlight* is already an established term and arguably shorthand for search in macOS. I'm on the fence here. What do you think?

rdmark:
> Absolutely agree with pagination. Do we need to port any of the RPC fixes in Samba over to ensure the stream is robust?
> 
> I also agree with dropping support for legacy versions of `localsearch` - especially when local search is fragile. We should not import technical debt due to third party issues

I ported a batch of RPC fixes from Samba in https://github.com/Netatalk/netatalk/pull/2900 but I was looking more closely at additional attributes and return codes. It's worth taking another look.

rdmark:
After typing out the above and thinking about it I made up my mind. Now symbols and user facing options are consistently using the *spotlight* label, sometimes in combination with *search*.

Two reasons: Since we already have user facing spotlight options, it's good to group and a name them consistently. Secondly, *search* is ambiguous since it may also refer to CatSearch.

I added a paragraph to the Search manual page to explain the situation.

NJRoadfan:
Apple seems to also have used the SpotlightRPC calls to access additional metadata on files that isn't supported by the AFP protocol eg: File Last Opened.

andylemin:
Hi @rdmark
Perfect, yes it was the naming collision with cat search I was struggling with.

I agree spotlight search is also confusing, as it is not technically spotlight search but finder search, but as you say, is defined as spotlight search protocol. This has confused me in the past as well.

Keeping it under `with-spotlight` makes sense as the main build option, but using more specific build option to build for specific types looks right.

Thank you for clarifying the manual, to disambiguate the afp.conf settings.
Looking really good now

andylemin (approved):
The PR summary needs a little update 👍

.clusterfuzzlite/build.sh
```diff
@@ -35,7 +35,6 @@ meson setup build \
     -Dwith-ldap=false \
     -Dwith-pam=false \
     -Dwith-quota=false \
-    -Dwith-spotlight=false \
```
andylemin: Is this correct after your last comment?

rdmark:
Cheers, I think this is a good starting point to a) further improve the Localsearch backend and b) introduce additional backends such as Apple Search Kit on macOS.

PR #2927, opened 2026-04-24
ci: update FreeBSD VM action to use release 14.4

rdmark:
the FreeBSD 15.0 VM is in a broken state, so falling back to 14.4


rdmark:
merging without review to unblock pipeline

PR #2928, opened 2026-04-24
ci: explicitly install BerkeleyDB in macOS spectest

rdmark:
something changed in the macOS runner that led to berkeley-db no longer being installed by default, so we now need to explicitly install it otherwise the spectest execution fails when the dbd CNID backend is not found


rdmark:
merging without review to unblock pipeline

PR #2929, opened 2026-04-25
afpd: align invalid reference errors with AFP 3.4

rdmark:
Return AFP 3.4 reference-compliant errors for unknown volume and desktop references across several AFP handlers.

- Validate FPCloseDir VolumeID and return AFPERR_PARAM when unknown
- Validate FPCloseDT desktop ref and return AFPERR_PARAM when unknown
- Check FPCopyFile DestVolumeID before source file access/lock handling
- Return AFPERR_PARAM for unknown VolumeID in extended attribute handlers
- Return AFPERR_PARAM for unknown VolumeID in FPGetFileDirParms
- Return AFPERR_PARAM for unknown VolumeID in ACL handlers
- Return AFPERR_NOOBJ for invalid FPSyncDir volID per AFP 3.4

modernize VolumeID AFP error handling tests in the testsuite to accommodate the above changes

- rework Error:test35 to send spec-shaped requests per command
- cover AFP 3.x VolumeID commands and desktop reference commands
- add expected-result table, including FPSyncDir kFPObjectNotFound
- test both SourceVolumeID and DestVolumeID for FPCopyFile
- make FPMoveAndRename:test73 independent of case sensitivity
- flip skip condition for Mac AppleShare server under test to enable a few tests
- add missing volume attribute bits to afpclient

### DIFF
--- a/etc/afpd/acls.c
+++ b/etc/afpd/acls.c
@@ -1763,7 +1763,7 @@ int afp_access(AFPObj *obj, char *ibuf, size_t ibuflen _U_, char *rbuf _U_,
 
     if (NULL == (vol = getvolbyvid(vid))) {
         LOG(log_error, logtype_afpd, "afp_access: error getting volid:%d", vid);
-        return AFPERR_NOOBJ;
+        return AFPERR_PARAM;
     }
 
     memcpy(&did, ibuf, sizeof(did));
@@ -1821,7 +1821,7 @@ int afp_getacl(AFPObj *obj, char *ibuf, size_t ibuflen _U_, char *rbuf _U_,
 
     if (NULL == (vol = getvolbyvid(vid))) {
         LOG(log_error, logtype_afpd, "afp_getacl: error getting volid:%d", vid);
-        return AFPERR_NOOBJ;
+        return AFPERR_PARAM;
     }
 
     memcpy(&did, ibuf, sizeof(did));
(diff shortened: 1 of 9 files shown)


rdmark:
@andylemin this started out by me noticing that a handful of `if (Mac)` conditionals for skipping tests that are not compatible with Apple's AFP server were flipped to `if (!Mac)` meaning that they were always skipped when testing Netatalk, which wasn't the original intention. The rabbit hole went pretty deep.

This led me to looking closer at `test35` that checks the error handling for AFP commands with invalid Volume ID, which let me looking closer how Netatalk handles that error, and validating our logic against the AFP 3.4 spec. Unless I was completely misguided in my reading of the spec, I think this changeset will make the Netatalk AFP server more standards compliant.

I'm leaning towards keeping this as a draft until after the 4.5.0 stable release. As we agreed elsewhere, the *next* feature version should be the AFP compliance release! There is a possibility that the old logic encoded a previous AFP spec and/or quirks with very old clients (e.g. early AppleShare clients or Apple II ProDOS clients etc.) so I want to crowd source some testing before shipping it.

What do you think?

NJRoadfan:
Many of the changes made were originally documented as such in "Inside AppleTalk", so they shouldn't be breaking anything. Don't know why they deviated from the spec unless there were weird bugs in Apple's client at one point. I think the ACL error codes were an oversight and never changed when ACL support was added to Netatalk.

rdmark:
thanks for validating the changes, @NJRoadfan -- this makes me more confident that I'm on the right track

andylemin (approved):

andylemin:
Hi @rdmark 
Thanks for this one! This is great to see.

I would imagine it was a simple mistake to exclude them, or something quirky as @NJRoadfan pointed out

NJRoadfan:
Running tests against an Apple AFP 2.2 server reveals test36 failing with multiple errors. Investigating why, but it looks like Apple's server returns different errors for some on these commands.

PR #2932, opened 2026-04-25
uams: harden SRP, DHX2, and DHX multi step error handling

rdmark:
DHX2: Zero-pad the DH prime p when serializing for the wire, matching the existing treatment of g and Ma. Add error handling for all gcry_mpi_print calls when serializing.

DHX: Guard logincont and changepw against being called before the shared key K is established, preventing a NULL pointer dereference if the login step was skipped or failed. Also release K immediately after serializing it to the local key buffer, fixing a leak that left key material in memory indefinitely after authentication.

SRP: Check that the SRP session verifier is set before proceeding in srp_logincont, preventing a NULL pointer dereference if the login step was skipped or failed.

Inspired by bug fixes in afp-perl by Derrik Pates


rdmark:
the heavy code repetition is a recurring concern with these UAMs -- we've had a few corner case bugs in the past because of this. I have been pondering alternative solutions: such as a static library with abstractions for the repeated logic, or maybe even merge the respective PAM and passwd libraries into one and make the authentication backend a runtime option.

andylemin:
> the heavy code repetition is a recurring concern with these UAMs -- we've had a few corner case bugs in the past because of this. I have been pondering alternative solutions: such as a static library with abstractions for the repeated logic, or maybe even merge the respective PAM and passwd libraries into one and make the authentication backend a runtime option.

Yes we should try to have shared code if we can - this is going to be prone to issues as you say, which is not a good thing for the front security layer

rdmark:
@andylemin I have reverted this PR back to the pure UAM hardening state, which I think are important to get merged sooner rather than later!

now we end up with the massive code duplication that prompted me to attempt the broader refactor in the first place, however I filed https://github.com/Netatalk/netatalk/issues/3249 as a fast follow

ready for your review :)

andylemin:
> @andylemin I have reverted this PR back to the pure UAM hardening state, which I think are important to get merged sooner rather than later!

Yep fair as we are going for a big multi-product release. The duplication can be resolved later when we have more time as appreciate it will require refactoring.

Also gives us a chance to identify functional issues from refactoring issues.

Give me a few days to try and review this one properly. I may not get to it until the weekend.

etc/uams/uams_dhx2_pam.c
```diff
@@ -328,13 +333,28 @@ static int dhx2_setup(void *obj, char *ibuf _U_, size_t ibuflen _U_,
     memcpy(rbuf, &uint16, sizeof(uint16_t));
     rbuf += 2;
     *rbuflen += 2;
+
     /* p */
-    gcry_mpi_print(GCRYMPI_FMT_USG, (unsigned char *)rbuf, PRIMEBITS / 8, NULL, p);
+    if (gcry_mpi_print(GCRYMPI_FMT_USG, (unsigned char *)rbuf, PRIMEBITS / 8,
+                       &nwritten, p) != 0) {
+        ret = AFPERR_MISC;
+        goto error;
```
andylemin, line 341: Non-blocking minor: these new print-error exits leave `*rbuflen` at the partial count (session ID and g already written), whereas the DHX equivalents in this PR zero `*rbuflen` before bailing. Since `dhx2_setup` also serves as changepw phase 1 via `changepw_1`, the stray bytes would ride along with the error reply there. Worth a `*rbuflen = 0;` inside the `if (ret != AFPERR_AUTHCONT)` branch at the `error:` label to match the DHX treatment? (Applies to the passwd twin too.)

etc/uams/uams_dhx_pam.c
```diff
@@ -45,6 +45,14 @@
 /*! the secret key */
 gcry_mpi_t K;
 
+static void dhx_release_key(void)
```
andylemin, line 48: Non-blocking thought: now that K's lifetime is nicely tightened, should the challenge nonce follow on the early-fail exits? On the sessid-mismatch and print-failure returns, `randbuf` still holds the nonce (the normal flow only zeroes it later during verification). An `explicit_bzero(randbuf, sizeof(randbuf))` on those paths would finish the cleanup — though not in this helper itself, since logincont still needs `randbuf` for the nonce check after releasing K. Far less sensitive than K, so feel free to ignore.

etc/uams/uams_dhx_pam.c
```diff
@@ -675,13 +705,21 @@ static int pam_changepw(void *obj, unsigned char *username,
 
     /* otherwise, it's like logincont but different. */
 
+    /* Make sure dhx_setup actually ran and established the shared key */
+    if (K == NULL) {
+        LOG(log_error, logtype_uams,
+            "DHX: changepw called without completing key exchange");
```
andylemin, line 722: Non-blocking: this looks like the one pre-fix path that was genuinely reachable — a session authenticated via another UAM (e.g. ClearTxt) sending FPChangePassword with the DHCAST128 UAM and a nonzero session ID would previously hit `gcry_mpi_print` on a NULL K. Would a small regression test driving that sequence be worthwhile (afparg might be a fit), perhaps folded into the #3249 follow-up?

andylemin (commented):
Went through the DHX, DHX2, and SRP flows — the guards and the tightened key lifetime all check out (including the changepw phase-1 re-setup and the retry paths). A few non-blocking minors below; none of them should hold this up.

andylemin (approved):

etc/uams/uams_dhx2_pam.c
rdmark: agreed, I see the cleanup improvement here

etc/uams/uams_dhx_pam.c
rdmark: this would indeed be better hygiene! let me bzero the memory throughout this module.

etc/uams/uams_dhx_pam.c
rdmark: I think this should be a great fit for the logintest suite

rdmark: attempted to implement a test, however it would require instrumentation/mocking around gcry_mpi_print, so a bit too large for this PR

PR #2934, opened 2026-04-26
afpd: Update file attribute handling for AFP 1.1 clients

NJRoadfan:
With AFP 2.0, Apple added two additional file attributes, `RenameInhibit` and `DeleteInhibit` while changing the name of the `ReadOnly` flag to `WriteInhibit`. In order to maintain compatibility with older clients, Apple specified rules for handling these flags.

- If the client requests a file with the `RenameInhibit` or `DeleteInhibit` flags set, we now set the `WriteInhibit` flag for AFP 1.1 clients.

- If an AFP 1.1 client sets or clears the `WriteInhibit` flag, we also set or clear the `RenameInhibit` and `DeleteInhibit` flags.

### DIFF
--- a/etc/afpd/file.c
+++ b/etc/afpd/file.c
@@ -411,6 +411,7 @@ int getmetadata(const AFPObj *obj,
             path->m_name = utompath(vol, upath, id, utf8_encoding(vol->v_obj));
         }
     }
+
     /* If this is a ProDOS client, adjust date-time values to match local time */
     if (obj->afp_version < 30 && (bitmap & 1 << FILPBIT_PDINFO)) {
         timeoffset = 1;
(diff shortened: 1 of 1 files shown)


etc/afpd/file.c
```diff
@@ -433,6 +433,15 @@ int getmetadata(const AFPObj *obj,
...
             ashort &= ~htons(vol->v_ignattr);
+
+            /*
+             * If this is an AFP 1.1 client we need to set the WriteInhibit
+             * flag if RenameInhibit and/or DeleteInhibit flags are set.
+             */
+            if (obj->afp_version < 20) {
+                if (ashort & htons(ATTRBIT_NORENAME) || ashort & htons(ATTRBIT_NODELETE))
+                    ashort |= htons(ATTRBIT_NOWRITE);
+            }
```
rdmark: This breaks two stylistic rules: First, nested if statements must be consolidated. Secondly, one liner if statement must always use curly brackets (the outer does it but not the inner.)

Actually this reveals that the Code Formatting CI job isn't fully working. Looking at the job log for file.c you can see that it was flagged as "formatted" but it does not translate to a job failure for some reason.

Also, you can run `./contrib/scripts/codefmt.sh -s c` to do a global reformatting for C code!

etc/afpd/file.c
```diff
@@ -1303,6 +1312,15 @@ int setfilparams(const AFPObj *obj, struct vol *vol,
...
             oshort = bshort;
 
+            /*
+             * AFP 1.1 clients need to also set or clear RenameInhibit
+             * and DeleteInhibit bits alongside WriteInhibit.
+             */
+            if (obj->afp_version < 20) {
+                if (ntohs(ashort) & ATTRBIT_NOWRITE)
+                    ashort |= htons(ATTRBIT_NORENAME | ATTRBIT_NODELETE);
+            }
```
rdmark: same feedback as above

rdmark (changes requested):

NJRoadfan:
All cleaned up.

rdmark (approved):
LGTM!

Issue #1224, opened 2024-07-12
Alternative indexing backend for Spotlight


rdmark:
Netatalk presently uses Gnome Tracker as the backend for file system indexing and search for the Spotlight feature. However, Tracker has the drawback of being [explicitly designed for GUI applications](https://tracker.gnome.org/overview/), notably the Gnome desktop environment. As such, it has a few drawbacks:

- It may pull in many dependencies that aren't relevant for Netatalk or a headless server system in general
- It expects to have a Gnome desktop environment, and [throws many warnings](https://gitlab.gnome.org/GNOME/localsearch/-/issues/232) when missing

Since Netatalk introduced Spotlight support a decade ago, more light-weight and specialized indexing solutions have appeared. Notably Elasticsearch.

In fact, Samba supports [Elasticsearch as a Spotlight indexing backend](https://wiki.samba.org/index.php/Spotlight_with_Elasticsearch_Backend) in addition to Tracker. It should be investigated if the same could be applied to Netatalk.

rdmark:
Alpine Linux version 3.21 has now moved from Tracker to Localsearch (Gnome's rebranding of this technology). When installing the `localsearch` package, 962 MB of packages are installed on a headless Alpine system, including Weyland, the Mesa 3D graphics library, several multimedia codecs, hardware drivers, and so on.

This is yet another data point that proves Localsearch (Tracker) isn't an appropriate backend technology for a server application.

MaddTheSane:
Perhaps using [OpenSearch](https://opensearch.org) for the back-end?

rdmark:
OpenSearch has a few drawbacks, first that it's a Java application (adds heavy dependencies, kind of defeats the purpose of moving away from Localsearch) and secondly that it's not very broadly redistributed (notably missing from Debian, Fedora).

rdmark:
Opus 4.6 has this to say on the topic – the question is how critical metadata attribute search is to this feature; to be frank, I've never seen Spotlight searches work as intended in my test environments.

----

# Spotlight Indexer Replacement Evaluation

## Current Architecture

The Spotlight integration with localsearch/tinysparql is deep but well-contained:

- **Protocol**: D-Bus SPARQL queries via `tracker_sparql_connection_*` API
- **Dependencies**: GLib (event loop), GObject, D-Bus daemon, localsearch/tracker3
[...]

PR #2938, opened 2026-04-27
netatalk: leverage D-Bus session bus auto activation for localsearch

### LINKED ISSUE #2907: method for starting localsearch on d-bus has changed
somewhere between localsearch v3.8.2 (Debian Trixie) and 3.11.0 (Debian unstable) the mechanism for starting the metadata extractor on d-bus has changed

in 3.8 you ran `localsearch daemon -s` to start and `localsearch daemon -t` to stop the extractor daemon, which is what we use in our spotlight code

in 3.11 the `daemon` subcommand has been removed and at a glance it is not obvious what the new mechanism is

rdmark:
we have so far been using the 'daemon' subcommand of tracker/localsearch to manually launch the indexer daemon from the netatalk controller daemon

however this subcommand was completely removed with localsearch v3.10, so this changeset instead relies on the native auto activation on the D-Bus session bus to launch the indexer

the first SPARQL connection from sl_localsearch_init(), our private dbus-daemon sees no provider for org.freedesktop.LocalSearch3, reads the auto-activation file, and forks localsearch; the indexer then claims the bus name and the SPARQL connection succeeds

shutdown is implicit: when Netatalk SIGTERMs dbus_pid, the indexer's bus connection drops and it exits on its own; no need to send an explicit termination command like previously

### DIFF
--- a/doc/manual/Installation.md
+++ b/doc/manual/Installation.md
@@ -186,6 +186,10 @@ functionality.
     strings from source files and merge them with the translations
     stored in PO files.
 
+- talloc
+
+    Samba's talloc library is used for memory management in the Spotlight query parser.
+
 - TCP wrappers
 
     Wietse Venema's network logger, also known as TCPD or LOG_TCP.
@@ -202,9 +206,9 @@ functionality.
     This is mostly relevant for developers or package managers who want to regenerate
     the Unicode source files.
 
-### Spotlight dependencies
+### Spotlight LocalSearch backend dependencies
 
-Netatalk's Spotlight support relies on the following third-party software:
+The LocalSearch backend for Spotlight depends on the following third-party software:
 
 - bison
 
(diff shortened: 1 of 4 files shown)


rdmark:
@andylemin it turned out to be easier than expected to delegate the indexer daemon spawning to the native D-Bus mechanism, and it works with older versions of LocalSearch too (v3.8 at least) so I decided against introducing an arbitrary version number check

OTOH, I have now come to the realization that localsearch/tinysparql v.3.11 is just as unstable as the earlier versions, and segfaults constantly when you throw more complex SPARQL queries at it, and complex queries is precisely what is required for Spotlight

at this point I'm leaning towards flagging this backend as deprecated/unsupported after this PR, and start working on the next generation backends, with the goal of removing localsearch in a future release

what do you think?

andylemin (approved):

andylemin:
Hi @rdmark 
Hmm, that's disappointing it is struggling with these queries.

What are the alternatives?
Should we raise a bug report?

rdmark:
the idea right now is to use libxapian for cross-platform, and Search Kit on macOS -- some analysis in https://github.com/Netatalk/netatalk/issues/1224

filing a bug report with the GNOME project seems reasonable; just need to figure out how to install debug symbols on Debian to get a deobfuscated backtrace :)

PR #2940, opened 2026-04-28
introduce a Spotlight backend built on Xapian

### LINKED ISSUE #1224: Alternative indexing backend for Spotlight

rdmark:
implement a Netatalk-managed Xapian Spotlight backend with filename, plain-text content, and MIME type search; compared to the cnid backend it is more full-featured, supporting all three query modes of macOS Finder search, while compared to the localsearch backend it is much more light-weight, with fewer build and run time dependencies and less scaffolding for running the indexer

Xapian has a C++ API, so this introduces a wrapper with the C interface for Spotlight backends, while handling the detection of a C++ compiler in Meson; we set C++17 as the baseline standard, since this is what Xapian 2.0 uses

dependencies are xapian-core and libmagic (file) for MIME type detection

control this feature at compile time with the -Dwith-spotlight-backends=xapian meson option

at runtime, configure 'spotlight backend = xapian' in afp.conf

----

## implementation notes

The key interface is sl_backend_ops in include/atalk/spotlight.h. Each backend implements init, close, open_query, fetch_results, close_query, and optionally index_event. sl_backends.h exposes the compiled-in ops instances, while volume.c binds a volume to cnid, localsearch, or xapian.

Xapian is split between sl_xapian.c the C backend/vtable adapter and result marshaller; xapian_wrapper.cc the C++ libxapian/libmagic implementation behind the C ABI in sl_xapian.h. The Xapian DB lives under _PATH_STATEDIR/spotlight/xapian/<volume uuid>, seeded by a full reconcile on first query per volume, then kept current through AFP file event hooks where possible.

in the current implementation, the AFP daemon (afpd) for each connected user is responsible for on-demand indexing and reconciliation. I have taken measures to avoid heavy blocking operations, but there is an inherent weakness in this multiple readers + multiple writers structure, long term having a dedicated indexer daemon should be evaluated.

```mermaid
flowchart TD
    XAPIAN_C["Xapian C backend<br/>etc/spotlight/xapian/sl_xapian.c"] -->|"C ABI calls"| WRAP["C++ xapian_wrapper.cc<br/>exception-safe libxapian boundary"]

    XAPIAN_C -->|"first query per volume"| ENSURE["sl_xapian_ensure_seeded"]
    ENSURE -->|"process fast path"| SEED["seeded_volumes list"]
    ENSURE -->|"authoritative check"| META["Xapian metadata<br/>schema_version, volume_uuid/path<br/>last_complete_generation<br/>reconcile_in_progress, dirty"]
    META -->|"valid complete index"| XAPIAN_C
    META -->|"missing, incompatible, dirty"| RECON["sl_xapian_reconcile"]

    RECON -->|"new generation id"| WALK["full volume walk"]
    WALK -->|"upsert docs with generation"| XDB[("Xapian database")]
    RECON -->|"delete docs whose generation<br/>does not match completed scan"| XDB
    RECON -->|"commit complete metadata"| META

    FILEOPS["AFP file and directory events<br/>create, modify, move, delete"] -->|"create or modify"| UPSERT["sl_xapian_upsert_path"]
    FILEOPS -->|"file delete"| DEL["sl_xapian_delete_path"]
    FILEOPS -->|"directory delete"| DELP["sl_xapian_delete_prefix"]
    FILEOPS -->|"file move"| MOVF["delete old path<br/>upsert new path"]
    FILEOPS -->|"directory move"| MOVD["sl_xapian_reindex_subtree<br/>delete old prefix<br/>walk new subtree"]

    MOVD -->|"success"| XDB
    MOVD -->|"failure"| DIRTY["mark metadata dirty<br/>unmark seeded volume"]
    DIRTY --> META

```

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -45,6 +45,7 @@ jobs:
             dbus-dev \
             dconf \
             flex \
+            file-dev \
             gcc \
             glib \
             iniparser-dev \
@@ -67,7 +68,8 @@ jobs:
             sqlite-dev \
             talloc-dev \
             tinysparql-dev \
-            valgrind
+            valgrind \
+            xapian-core-dev
       - name: Configure
         run: |
           meson setup build \
@@ -122,6 +124,7 @@ jobs:
             db \
             dconf \
             flex \
+            file \
             gcc \
             iniparser \
             libev \
@@ -134,7 +137,8 @@ jobs:
             rpcsvc-proto \
             sqlite \
             talloc \
-            tinysparql
+            tinysparql \
+            xapian-core
       - name: Configure
         run: |
           meson setup build \
@@ -187,6 +191,7 @@ jobs:
             dconf-cli \
             file \
             flex \
+            g++ \
             gcc \
             libacl1-dev \
             libavahi-client-dev \
@@ -201,6 +206,7 @@ jobs:
             libiniparser-dev \
             libkrb5-dev \
             libldap2-dev \
+            libmagic-dev \
             libmariadb-dev \
             libpam0g-dev \
             libsqlite3-dev \
@@ -208,6 +214,7 @@ jobs:
             libtirpc-dev \
             libtracker-sparql-3.0-dev \
             libwrap0-dev \
+            libxapian-dev \
             meson \
             ninja-build \
             quota \
@@ -272,12 +279,14 @@ jobs:
             dbus-devel \
             dconf \
             flex \
+            gcc-c++ \
             glib2-devel \
             iniparser-devel \
             krb5-devel \
             libacl-devel \
             libdb-devel \
             libev-devel \
+            file-devel \
             libgcrypt-devel \
             libtalloc-devel \
             localsearch \
@@ -294,7 +303,8 @@ jobs:
             systemd \
             systemtap-sdt-devel \
             tinysparql-devel \
-            valgrind
+            valgrind \
+            xapian-core-devel
       - name: Configure
         run: |
           meson setup build \
@@ -349,6 +359,7 @@ jobs:
             dconf-cli \
             file \
             flex \
+            g++ \
             gcc \
             libacl1-dev \
             libavahi-client-dev \
@@ -362,12 +373,14 @@ jobs:
             libiniparser-dev \
             libkrb5-dev \
             libldap2-dev \
+            libmagic-dev \
             libmariadb-dev \
             libpam0g-dev \
             libtalloc-dev \
             libtirpc-dev \
             libtracker-sparql-3.0-dev \
             libwrap0-dev \
+            libxapian-dev \
             meson \
             ninja-build \
             quota \
@@ -426,10 +439,12 @@ jobs:
             cracklib \
             iniparser \
             libev \
+            libmagic \
             mariadb \
             meson \
             openldap \
-            talloc
+            talloc \
+            xapian
       - name: Configure
         run: |
           meson setup build \
@@ -484,6 +499,7 @@ jobs:
               db5 \
               iniparser \
               libev \
+              file \
               libgcrypt \
               meson \
               mysql80-client \
@@ -494,7 +510,8 @@ jobs:
               py39-sqlite3 \
               py39-tkinter \
               sqlite \
-              talloc
+              talloc \
+              xapian-core
           run: |
             set -e
             meson setup build \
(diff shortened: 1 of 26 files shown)


rdmark:
augment review

andylemin:
Hi @rdmark 

Wow, nicely done. I will test hopefully this weekend.

FYI: I've found some issues with the recent Spotlight changes with MySQL backend on FreeBSD (Finder searches are not working (strange errors in logs) and general Netatalk performance seems notably reduced) - will try to look into this too.

We might need to hold back release until all the variations are understood, or maybe reconsider enabling Spotlight by default.

Do we have a new spectest which is capable of creating spotlight searches? We might need new GitHub workflows for the various backends.

rdmark:
It might have something to do with that CNID backend queries are synchronous while the Spotlight backend architecture was built for async LocalSearch originally. I tried to accommodate for this in the CNID backend code but there are evidently gaps.

Good idea to introduce continuous testing for this. I think we need to develop a test harness for SpotlightRPC.

Agreed that we shouldn't push out the new Spotlight backends in this shape. In fact, I am responsible for this feature creep, so what do you say we revert the Spotlight CNID backend, and aim to get 4.5.0 out the door first? The pluggable Spotlight architecture can stay I think.

andylemin:
> It might have something to do with that CNID backend queries are synchronous while the Spotlight backend architecture was built for async LocalSearch originally. I tried to accommodate for this in the CNID backend code but there are evidently gaps.
> 
> Good idea to introduce continuous testing for this. I think we need to develop a test harness for SpotlightRPC.
> 
> Agreed that we shouldn't push out the new Spotlight backends in this shape. In fact, I am responsible for this feature creep, so what do you say we revert the Spotlight CNID backend, and aim to get 4.5.0 out the door first? The pluggable Spotlight architecture can stay I think.

Hi @rdmark I don't think we need to revert/stash the CNID backend for the release. If you were to take that action you would need to revert all of the new spotlight changes as they will each have some little issues (some known) considering how quirky spotlight search is..

Things are orders of magnitude better with Spotlight right now, just some teething pains - so only reverting Spotlight to disabled by default, is the simplest and right thing to do IMO - and I have confirmed disabling spotlight mitigates the performance regression I was seeing (with my CNID Spotlight backend init failure).

We want users to test and feedback, without any regressions by default - this is not release blocking.

I created Issue #2946

andylemin (approved):

rdmark:
sounds like a plan -- I'm working on a new branch now with some spotlight bugfixes and improvements, let me throw in some documentation tweaks as well to emphasize that both cnid and xapian are experimental spotlight backends

at the same time, I will revert to spotlight being disabled by default, and the localsearch backend as the default, because it's the most stable despite being hardly functional

as a side note, I have been toying with the idea of collapsing the cnid and xapian backends into a single hybrid backend where CNID db lookups are used for filename queries (e.g. kMDItemDisplayName) and Xapian for file content or metadata queries (e.g. kMDItemTextContent) -- for the best of both worlds!

it might introduce complexity and fragility to the architecture, but it's worth a try anyways

andylemin:
> as a side note, I have been toying with the idea of collapsing the cnid and xapian backends into a single hybrid backend where CNID db lookups are used for filename queries (e.g. kMDItemDisplayName) and Xapian for file content or metadata queries (e.g. kMDItemTextContent) -- for the best of both worlds!
> 
> it might introduce complexity and fragility to the architecture, but it's worth a try anyways

Hi @rdmark its an interesting idea and a potentially cheap way to improve the performance!

However please can we maintain the ability to have CNID-only or CNID-Xapian?

For security reasons some users may not want to have a system such as xapian or localsearch scraping storage, when they only need simple filename search.
CNID-only serves an important use case.

PR #2943, opened 2026-04-29
libatalk: avoid misaligned ifreq access in getifaces

rdmark:
BSD-style SIOCGIFCONF results can contain variable-length ifreq records whose starting addresses are not aligned for struct ifreq on arm64. Traversing the buffer with struct ifreq pointers lets UBSAN flag misaligned loads before afpd starts serving.

Walk the ioctl buffer as bytes and copy each record into an aligned local struct before inspecting ifr_name or ifr_addr.sa_len.

### DIFF
--- a/libatalk/util/getiface.c
+++ b/libatalk/util/getiface.c
@@ -84,8 +84,9 @@ static int getifaces(const int sockfd, char ***list)
     return i;
 #else
     struct ifconf	ifc;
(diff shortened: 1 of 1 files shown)


rdmark:
this fix allows you to cleanly run afpd under ASAN on macOS!

andylemin (approved):

PR #2956, opened 2026-05-05
spotlight: add support for localsearch indexing of home directories

### LINKED ISSUE #2908: spotlight not enabled when only home volume defined
since the home volumes are dynamic, the OPTION_SPOTLIGHT flag is set for them, which means that the `if (obj.options.flags & OPTION_SPOTLIGHT)` condition evaluates to false in netatalk.c and the indexer not started

rdmark:
The [Homes] spotlight option (default: inherits from [Global]) controls whether the basedir regex path is submitted to the search indexer, covering all user home directories beneath it.

Because [Homes] volumes are resolved per-user at login and never pass through creatvol() in the master process, OPTION_SPOTLIGHT must be set explicitly during config parsing when Homes spotlight is enabled -- otherwise the localsearch dbus infrastructure is never started.

Add INIPARSER_GETBOOL macro parallel to the other iniparser wrapper macros.

### DIFF
--- a/doc/manpages/man5/afp.conf.5.md
+++ b/doc/manpages/man5/afp.conf.5.md
@@ -732,11 +732,14 @@ of 10000 candidate results. Set a larger nonzero value to raise the
 *xapian* cap. The *cnid* backend does not use this option and is limited
 by its fixed CNID search reply cap.
 
-spotlight = *BOOLEAN* (default: *yes*) **(G)**/**(V)**
+spotlight = *BOOLEAN* (default: *yes*) **(G)**/**(V)**/**(H)**
 
 > Whether to enable Spotlight-compatible searches.
-As a global option, this sets the default for volumes. As a volume option,
-it overrides the global default for that volume.
+As a global option, this sets the default for volumes and home directories.
+As a volume or Homes option, it overrides the global default.
+When enabled for the Homes section, the directory specified by
+**basedir regex** is submitted to the search indexer, covering all user
+home directories beneath it.
 >
 > A volume is searchable only when its effective **spotlight** setting is
 enabled and a supported **spotlight backend** is available for that volume.
--- a/etc/netatalk/netatalk.c
+++ b/etc/netatalk/netatalk.c
@@ -169,6 +169,22 @@ static int set_sl_volumes(void)
         }
     }
 
+    int global_sl = INIPARSER_GETBOOL(obj.iniconfig, INISEC_GLOBAL, "spotlight", 1);
+
+    if (INIPARSER_GETBOOL(obj.iniconfig, INISEC_HOMES, "spotlight", global_sl)) {
+        const char *basedir = INIPARSER_GETSTR(obj.iniconfig, INISEC_HOMES,
+                                               "basedir regex", NULL);
+
+        if (basedir && *basedir) {
+            if (first) {
+                fprintf(fp, "index-recursive-directories=['%s'", basedir);
+                first = false;
+            } else {
+                fprintf(fp, ", '%s'", basedir);
+            }
+        }
+    }
+
     if (first) {
         /* No Spotlight volumes: emit typed empty array so dconf update parses it */
         fprintf(fp, "index-recursive-directories=@as []\n");
--- a/include/atalk/iniparser_util.h
+++ b/include/atalk/iniparser_util.h
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2024-2025 Daniel Markstedt
+ * Copyright (C) 2024-2026 Daniel Markstedt
  * All Rights Reserved.  See COPYING.
  *
  * This program is free software; you can redistribute it and/or modify
@@ -29,6 +29,12 @@
 
 #define INISEC_GLOBAL "global"
 #define INISEC_HOMES  "homes"
+
+#define INIPARSER_GETBOOL(config, section, key, default) ({           \
+    char _option[MAXOPTLEN];                                          \
+    snprintf(_option, sizeof(_option), "%s:%s", section, key);        \
+    iniparser_getboolean(config, _option, default);                   \
+})
 
 #define INIPARSER_GETSTR(config, section, key, default) ({            \
     char _option[MAXOPTLEN];                                          \
--- a/libatalk/util/netatalk_conf.c
+++ b/libatalk/util/netatalk_conf.c
@@ -2779,6 +2779,20 @@ int afp_config_parse(AFPObj *AFPObj, char *processname)
         options->flags |= OPTION_SPOTLIGHT_VOL;
     }
 
+    {
+        /* OPTION_SPOTLIGHT is normally set by creatvol() per named volume, but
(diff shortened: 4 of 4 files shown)


andylemin (approved):
I like this requirement to specify spotlight for each home share separately - I think it is a better choice for individual users, where some users may not want localsearch indexing enabled when enabled globally.

rdmark:
@andylemin do you have a suggestion how to manage a per-user volume setting in practice? we don't have such a mechanism right now IINM. maybe something like a ~/.netatalk/afp.conf that afpd reads after each afpd fork?

andylemin:
Hi @rdmark 

you are right.. there is currently only a single virtual homes share which covers all users.. (I don't use homes so forgot this).

To provide per user override you would need the ability to optionally specify an individual [username] home mount with settings that override the common homes and global config. This would allow Spotlight to be enabled for only a single user if homes does not exist at all, or disabled if [homes] exists and enables.

[username] > [homes] > [global]

Whether you also allow those per username home mount settings to be defined in ~/.netatalk/afp.conf, IMO would just an extension of the same override logic.

However, only implementing the username override in the main afp.conf keeps it simple and focused on value (patterns that are most likely to actually be used = only settable by the admin).

And we still need to sort out Netatalk startup as non-root first.

So definitely keep it simple and only in the main afp.conf for now. Ideally we should allow any property under [homes] to be override-able. There are clear use cases for different permission bitmaps for example.

andylemin:
@rdmark 

PS; Spotlight is working fantastically well with CNID after the bugfix.

Love how fast it is! Less than 1 second to search >250k and display results. And searching >1M is indistinguishable (fraction of a second slower). 🚀

And performance regression is resolved 🫡

Just the 100 results limit that gets in the way - otherwise near perfect!

rdmark:
filed https://github.com/Netatalk/netatalk/issues/2963 to follow up on the granular home dir settings feature!

I think we're very close to a non-root operation mode of netatalk -- what I'm thinking is adding logic to the netatalk controller daemon when detecting non-root uid, attempts to create ~/.netatalk/afp.conf with a suite of default options including server signature, vol uuid, vol dbpath with cnid scheme = sqlite ... and if the file already exists we read it as-is instead of course

rdmark:
> PS; Spotlight is working fantastically well with CNID after the bugfix.
> 
> Love how fast it is! Less than 1 second to search >250k and display results. And searching >1M is indistinguishable (fraction of a second slower). 🚀
> 
> And performance regression is resolved 🫡
> 
> Just the 100 results limit that gets in the way - otherwise near perfect!

this was an unexpected lucky break, I had anticipated hours to debugging :-D

with the PR at hand now indexing all home dirs by default, I'm starting to get cold feet about this idea of using localsearch as the default -- maybe I'm overthinking it but isn't there privacy implications of having a global-readable index that includes both filenames and fragments of file contents that any user with the right know-how can query? this concern goes for both localsearch and xapian.

the xapian backend is arguably safer since it only indexes when a volume is mounted, and it should only be indexing contents of files that are global-readable in the first place (need to double check the logic, but this is what I remember)

the localsearch indexer is running as root and proactively indexing all volumes paths listed in `/etc/dconf/db/netatalk.d/10-spotlight` -- and if you run tinysparql query against the right D-Bus socket you can query to your heart's content...

andylemin:
Hi @rdmark 

Yes, there are absolutely serious privacy risks here, which is where the granular per-user settings proposal was coming from.

And at the very least, we need to ensure user A cannot get results for user B's home index!

The safest solution is to not index any homes for now until we have a plan for per-user compartmentalisation of the index data?

Indexes for shared volumes is an another potential issue.

We also have to consider that shares may be restricted to certain users as well. Which means only those users should be able to get search results for those shares they are allowed to access..

I don't know the data structure for localsearch or xapian (to understand if there is a 'path' in the data to filter/restrict user queries only to their allowed paths)?

This should be easy with the CNID backend as the "index" is already a table per volume, so search can be restricted per user only to tables they are allowed easily?

andylemin:
> filed https://github.com/Netatalk/netatalk/issues/2963 to follow up on the granular home dir settings feature!
> 
> 
> 
> I think we're very close to a non-root operation mode of netatalk -- what I'm thinking is adding logic to the netatalk controller daemon when detecting non-root uid, attempts to create ~/.netatalk/afp.conf with a suite of default options including server signature, vol uuid, vol dbpath with cnid scheme = sqlite ... and if the file already exists we read it as-is instead of course

Why would we try to auto-create the per-user ones?
I suspect that would lead to stale config files in most cases causing problems?

Best to let users create them if they want/need them.

The default [homes] set by the admin should work for most, and these are only for overrides for users who need overrides (which should be the exception). 

I can see situations where users complain because homes are broken for some users and admins don't know why (because they don't know about the auto-created per user config), and that config is now stale after a config version/change or something?

rdmark:
> Why would we try to auto-create the per-user ones? I suspect that would lead to stale config files in most cases causing problems?
> 
> Best to let users create them if they want/need them.
> 
> The default [homes] set by the admin should work for most, and these are only for overrides for users who need overrides (which should be the exception).
> 
> I can see situations where users complain because homes are broken for some users and admins don't know why (because they don't know about the auto-created per user config), and that config is now stale after a config version/change or something?

the use case I had in mind was that a normal user wants to run an ad-hoc AFP server, and runs `netatalk` as their normal user without sudo

this fails for a number of reasons, notably lack of write permissions to the netatalk localstatedir, which means it cannot edit afp_signature.conf or afp_voluuid.conf, or create a subdir in CNID/ to store the database

but I see your point, the better approach is probably to instruct users who want to run an ad-hoc non-privileged AFP server to craft their own afp.conf in their home dir and then launch `netatalk -F ~/path/to/afp.conf`

rdmark:
pulling this back to draft while we figure out the secure way to index home dirs!

NJRoadfan:
Before reinventing the wheel... how does Apple's Spotlight indexing work with a user's shared home folder on a Mac hosted server? Heck, how does Samba handle it?

Looking on my Mac, it appears the index for a user's home is stored in `Users/~/Library/Metadata/CoreSpotlight`.

andylemin:
> Before reinventing the wheel... how does Apple's Spotlight indexing work with a user's shared home folder on a Mac hosted server? Heck, how does Samba handle it?
> 
> 
> 
> Looking on my Mac, it appears the index for a user's home is stored in `Users/~/Library/Metadata/CoreSpotlight`.

Nice, looks like an entirely self contained index - which is the right way to go for both secure access and secure at rest..

@rdmark would this mean multiple localsearch processes (one per user)?

rdmark:
yeah, multiple localsearch processes with multiple dconf db configuration files... potentially also multiple private D-Bus buses

but I agree that for privacy's sake, an individual home dir index should be stored in the respective home dir (while making sure the index isn't indexed, e.g. skip dot files/dirs) -- should be straight forward to achieve with the xapian backend at least

rdmark:
@andylemin I am discarding this in favor of the private database concept that you brought up in https://github.com/Netatalk/netatalk/pull/3277

PR #2961, opened 2026-05-05
docker: harden RandNum UAM in production container

### LINKED ISSUE #2877: docker: encrypt RandNum password on disk
there is opt-in DES encryption for the afppasswd file that RandNum uses, namely by creating a file afppasswd.key containing 16 hex characters (8 bytes) and restrict permissions (for example, owner-readable only)

we should apply this to the production container build for an additional layer of security

rdmark:
two measures to make the RandNum UAM safer in the container

- DES encrypt the afppasswd file with afppasswd.key
- create the afppasswd file only with INSECURE_AUTH

also: crisper description of RandNum password storage in the manual, and hardening of the RandNum UAM key file read bounds and integrity checks

### DIFF
--- a/distrib/docker/env_setup_netatalk.sh
+++ b/distrib/docker/env_setup_netatalk.sh
@@ -91,8 +91,32 @@ fi
 
 echo "$AFP_USER:$AFP_PASS" | chpasswd > /dev/null 2>&1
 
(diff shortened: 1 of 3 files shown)


rdmark:
augment review

andylemin (approved):
Simple but valuable 🙂

PR #2962, opened 2026-05-06
meson: put spooldir creation behind statedir creation flag

rdmark:


rdmark:
@quentinmit can you test if this does the trick for you? the guard is controlled with -Dwith-statedir-creation

quentinmit:
Works for me!

If you have Nix installed (you don't need to be on NixOS), you can reproduce the build with:

```shell
nix build -L $(nix eval --raw --impure github:NixOS/nixpkgs/nixos-unstable#netatalk --apply 'a: (a.overrideAttrs { src = ./.; }).drvPath')^*
```

where `./.` is the path to your `netatalk` repo.

andylemin (approved):

PR #2970, opened 2026-05-08
docker: dbus-daemon not needed in production container

rdmark:
dbus-daemon is only launched by the localsearch search backend which is not active in the netatalk production container


andylemin:
Wow, the flamegraph still occasionally does/does-not sync with the `perf` sampling frequency! Hmm.. I think 1019 was a poor choice, considering 10ms timer / 1.019 freq = 9.81 + ~200 usecs for flushing/work still equals 10ms / high probability of aligned sampling 😂
Need to think about it some more. The 10ms change was valid in its own right though..

andylemin (approved):

PR #2973, opened 2026-05-09
[4.4] uams: warn when Randnum afppasswd key file is missing

rdmark:
Log a security warning when the Randnum UAM falls back to clear-text afppasswd entries because the companion key file is not present or the configured path is too long.

The warning notes that this insecure fallback is deprecated and will become a hard failure in a future release.

Also allow the boundary case where the afppasswd path exactly leaves room for the ".key" suffix plus terminator, avoiding an unnecessary key-path failure.

Additionally: build astyle v3.6.14 from scratch for code formatting job, because they introduced multiple regression bugs in 3.6.15 (reported upstream).

### DIFF
--- a/etc/uams/uams_randnum.c
+++ b/etc/uams/uams_randnum.c
@@ -161,9 +161,25 @@ static int afppasswd(const struct passwd *pwd,
     /* open the key file if it exists */
     strcpy(buf, path);
 
-    if (pathlen < (int) sizeof(buf) - 5) {
+    if (pathlen <= (int) sizeof(buf) - 5) {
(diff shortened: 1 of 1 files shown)


andylemin (approved):

andylemin:
@rdmark LGTM. just some formatting to fix

andylemin:
@dependabot rebase

andylemin:
@dependabot recreate

rdmark:
@andylemin dependabot is only allowed to push to PRs that it created ;)

PR #2975, opened 2026-05-09
[4.4] docker: introduce AFP_UAMS option for fine grained control of UAMs

rdmark:
AFP_UAMS overrides both the default list as well as INSECURE_AUTH

it allows you to specify the list of UAM libraries afpd loads

add logic that checks for uams_randnum.so so that afppasswd gets processed


andylemin (approved):

andylemin:
@dependabot rebase

PR #2976, opened 2026-05-09
docker: introduce AFP_UAMS option for fine grained control of UAMS

rdmark:
AFP_UAMS overrides both the default list as well as INSECURE_AUTH, while setting the default to "uams_dhx2.so uams_srp.so"

it allows you to specify the list of UAM libraries afpd loads

add logic that checks for uams_randnum.so and uams_srp.so so that afppasswd and afppasswd.srp get processed


andylemin:
@rdmark 

This is great to see 🙂

What is the most secure now? uams_dhx2.so or uams_srp.so?

Is SRP still experimental?

andylemin (approved):

rdmark:
@andylemin I'm not a cryptography expert, but by my reading DHX2 provides the equivalent of 1024-bit encryption, while SRP provides 1536-bit encryption. SRP is also highly resistant to man-in-the-middle attacks, with no password data passed over the wire. The SRP password is set separately from the system user password, which is safer too!

I would say that SRP is ready to be used broadly. There is a known corner case where we haven't fully reverse engineered Apple's logic, as documented in the last section of: https://netatalk.io/spec/SRP_UAM

So there's a small chance that SRP authentication with an Apple client can fail occasionally.

PR #2981, opened 2026-05-10
refactor potentially unsafe C library calls flagged by OpenBSD linker

rdmark:
Buffer safety in production code:

- mangle(): sprintf -> snprintf with sizeof(mangle_suffix); strcat ->
  strlcat(m, ..., MAXPATHLEN) for the mangle suffix and fallback "???"
- quota/getquota(): malloc+strcpy pairs replaced with strdup(); sprintf
  for the NFS quotas path replaced with snprintf(buf, sizeof(buf), ...)
- messages/readmessage(): malloc size captured in filenamelen, both
  sprintf calls replaced with snprintf(filename, filenamelen, ...)
- nad/ftw: strcpy in mystpcpy (already bounds-checked) -> strlcpy
- appl/afp_addappl and afp_rmvappl: strcpy(obj->oldtmp, dtf) ->
  strlcpy(..., AFPOBJ_TMPSIZ)
- desktop/setdeskmode and setdeskowner: strcpy/strcat into modbuf[13] ->
  strlcpy/strlcat with sizeof(modbuf)
- desktop/dtfile: strcat(path, ext) -> strlcat(..., sizeof(path))
- desktop/mtoupath: strcpy(upath, ".") -> strlcpy(..., sizeof(upath))

Test code brought to the same standard:

- afpfunc_helpers/cnamewrap: strcpy -> strlcpy(p, name, sizeof(buf)-len)
- afphelper/bitmap2text: all sprintf+strcat pairs -> snprintf+strlcat
- speedtest Write/Copy/ServerCopy/Read: sprintf -> snprintf, strcpy ->
  strlcpy, all with sizeof(temp)
- T2_FPByteRangeLock/test_bytelock: strcpy/strcat chain -> strlcpy/strlcat
- subtests/test001_add_x_dirs: sprintf -> snprintf
- T2_FPResolveID/test129: sprintf -> snprintf
- T2_FPGetFileDirParms/test336: sprintf -> snprintf;
  strcat -> strlcat(..., sizeof(temp))

----

afpd: use full appl buffer size AFPOBJ_TMPSIZ+1 in strlcpy

oldtmp is declared as char oldtmp[AFPOBJ_TMPSIZ + 1], but both strlcpy calls were passing AFPOBJ_TMPSIZ as the size limit, causing a path of exactly MAXPATHLEN bytes to be silently truncated and potentially directing open()/rename() at the wrong temp file.

----

afpd: messages MAXPATHLEN+1 stack buffer for message filename

sizeof(SERVERTEXT) + 15 left only six bytes for the PID decimal representation; a pid_t can require up to ten digits, causing snprintf to truncate and unlink() to operate on the wrong path.

Replace the dynamically-allocated buffer with a stack-local char filename[MAXPATHLEN + 1] and use sizeof(filename) in both snprintf calls, which also eliminates the malloc/free pair.

----

testsuite: fix afp helper length/data mismatch in cnamewrap

The advertised path length was taken from strlen(name) before strlcpy could truncate, producing a malformed AFP request where the uint16_t length field claimed more bytes than were actually present in the buffer.

AFP 3 (UTF-8) paths are length-prefixed, not null-terminated, so strlcpy was also the wrong primitive. Fix by:
- computing the available space after the length field header up front
- clamping namelen to that space before encoding it into the uint16_t field
- using memcpy to write exactly namelen bytes, keeping length and data
  in sync regardless of the input size

----

afpd,testsuite: replace unsafe strcpy/sprintf with bounded equivalents

Replace remaining sprintf/strcpy/strcat calls flagged by the OpenBSD linker with snprintf/strlcpy/strlcat/memcpy throughout afpd and the test suite. Key fixes beyond mechanical substitution:

- appl.c afp_addappl/afp_rmvappl: guard against NULL return from
  dtfile() before passing dtf to strlcpy; rmvappl also closes sdt_fd
  on this new error path to match existing cleanup
- file.c set_name: save strlen(name) into tp_len at the strdup site
  so the restore memcpy uses tp_len+1 instead of a redundant strlen
- filedir.c afp_moveandrename: strcpy(newname, oldname) -> strlcpy with
  AFPOBJ_TMPSIZ+1, matching the size used for the same buffers nearby
- T2_FPDelete.c: all sprintf(temp/temp1, ...) -> snprintf with sizeof

### DIFF
--- a/bin/nad/ftw.c
+++ b/bin/nad/ftw.c
@@ -235,7 +235,7 @@ static char *mystpcpy(char *dest, const char *src, size_t dest_size)
 
         return dest;
     } else {
-        strcpy(dest, src);
+        strlcpy(dest, src, dest_size);
         return dest + src_len;
     }
 }
--- a/etc/afpd/appl.c
+++ b/etc/afpd/appl.c
@@ -273,7 +273,7 @@ int afp_addappl(AFPObj *obj, char *ibuf, size_t ibuflen _U_, char *rbuf _U_,
 
     dtf = dtfile(vol, creator, ".appl.temp");
     tempfile = obj->oldtmp;
-    strlcpy(tempfile, dtf, AFPOBJ_TMPSIZ);
+    strlcpy(tempfile, dtf, AFPOBJ_TMPSIZ + 1);
 
     if ((tfd = open(tempfile, O_RDWR | O_CREAT, 0666)) < 0) {
         return AFPERR_PARAM;
@@ -409,7 +409,7 @@ int afp_rmvappl(AFPObj *obj, char *ibuf, size_t ibuflen _U_, char *rbuf _U_,
         return AFPERR_PARAM;
     }
 
-    strlcpy(tempfile, dtf, AFPOBJ_TMPSIZ);
+    strlcpy(tempfile, dtf, AFPOBJ_TMPSIZ + 1);
 
     if ((tfd = open(tempfile, O_RDWR | O_CREAT, 0666)) < 0) {
         close(sa.sdt_fd);
--- a/etc/afpd/appl.c
+++ b/etc/afpd/appl.c
@@ -272,6 +272,11 @@ int afp_addappl(AFPObj *obj, char *ibuf, size_t ibuflen _U_, char *rbuf _U_,
     }
 
     dtf = dtfile(vol, creator, ".appl.temp");
+
+    if (!dtf) {
+        return AFPERR_PARAM;
+    }
+
     tempfile = obj->oldtmp;
     strlcpy(tempfile, dtf, AFPOBJ_TMPSIZ + 1);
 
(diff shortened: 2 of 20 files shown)


rdmark:
augment review

rdmark:
augment review

rdmark:
@andylemin FYI this became a slippery slope trying to appease the OpenBSD linker!

I'm thinking of breaking out the random string hardening and postpone the rest until another day and another stable release

andylemin:
> @andylemin FYI this became a slippery slope trying to appease the OpenBSD linker!
> 
> 
> 
> I'm thinking of breaking out the random string hardening and postpone the rest until another day and another stable release

Hi @rdmark thats absolutely reasonable. Not surprised it got out of hand quickly

rdmark:
@andylemin this is ready for your review now!

I dropped most of the changes to afpd/desktop.c, on second thought they were overly intrusive and out of scope for this PR

right now in our CI build job, the OpenBSD linker calls out unsafe code only in afpd/directory.c and testsuite/Uft8.c which is overall a lot cleaner than the situation before! however once you start squashing those the linker will invariably find more problems, because in my observation it flags the only first one it finds per compiled object.

an alternative approach to harden the code would be to work through the Security Hotspots that SonarQube flags, because they more or less overlaps with OpenBSD's opinion

also worth noting is that we are *not* covering AppleTalk code here because we don't build that on OpenBSD

at any rate, I think this is "good enough" to ship now!

andylemin:
Hi @rdmark Sorry I missed this, github was not showing it too me. I only noticed by chance!

Nice this is great quality, no issues I can see at all other than a few tiny observations not even worth discussing 🙂

andylemin (approved):

PR #2985, opened 2026-05-11
afpd: fix connection limit off-by-one condition

rdmark:
The previous condition allowed one more connection than configured: f.e. 201 for the default 200

cnx_cnt is captured in the child at fork time, before the parent calls server_child_add — so it counts existing connections, not including the current one. The > check introduced in an earlier commit assumed cnx_cnt already included the current connection and over-corrected, allowing one connection beyond the configured limit. Restore >= so the limit is enforced as configured.

### DIFF
--- a/etc/afpd/auth.c
+++ b/etc/afpd/auth.c
@@ -276,7 +276,7 @@ static int login(AFPObj *obj, struct passwd *pwd, void (*logout)(void),
         return AFPERR_NOTAUTH;
     }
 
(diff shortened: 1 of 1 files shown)


rdmark:
@andylemin FYI I'm working on an updated afp_logintest suite which led to the identification of this corner case bug

will submit the improved test suite in a later PR!

andylemin (approved):

PR #2992, opened 2026-05-14
testsuite: overhaul DSI test client and expand logintest

rdmark:
This changeset enables the AFP 3.x LoginExt method which opens up for broader UAM support in the AFP test client. We fix up the long-disabled test4 that exercises the DSI session max connection limit while adding a handful of new login tests (not a complete set, just a few sanity tests to get started.)

We also fix a handful of latent bugs in DSI session code.

## DSI function namespace cleanup

Drop the my_dsi_* prefix from all test-client DSI I/O functions to align with the libatalk server-side naming convention: my_dsi_stream_read/write/ send/receive, my_dsi_tickle, my_dsi_cmd_receive, my_dsi_data_receive all
become their unprefixed equivalents. Affects call sites and comments across afpclient.c, afpcmd.c, and ~25 test modules. dsi.h is pruned of server-side extern declarations (dsi_init, dsi_setstatus, dsi_kill, dsi_writeinit,
dsi_write, dsi_writeflush, dsi_wrtreply, dsi_readinit, dsi_read, dsi_readdone) that the test client never uses; dsi_clientID migrates from afpclient.h to dsi.h where dsi_serverID already lives.

dsi_stream_read: unconditionally set dsi_code to 0xffffffff on short/EOF reads so callers cannot mistake a zeroed header for a success response.

## AFPopenLoginExt / AFPLoginCont / AFPopenLoginExt_pwd

Rewrite AFPopenLoginExt with a correct wire format: auth_info is now an opaque blob, UserAuthInfo is padded to an even offset, and kFPAuthContinue replies are captured by a new capture_login_cont() helper that saves the UAM continuation ID and payload into three new CONN fields (login_cont_id, login_cont_len, login_cont_data[DSI_CMDSIZ]).

Add AFPLoginCont() implementing FPLoginCont for multi-round UAMs: sends the saved continuation ID (re-encoded big-endian) with an arbitrary auth_info blob, and re-runs capture_login_cont() on another AFPERR_AUTHCONT reply.

Add AFPopenLoginExt_pwd() as a convenience wrapper: routes to AFPopenLogin for Cleartxt (FPLoginExt + Cleartxt is broken in netatalk), passes NULL auth_info for anonymous, and zero-pads the password to PASSWDLEN bytes for all other UAMs.

FPopenLoginExt() in afpcmd.c now calls AFPopenLoginExt_pwd() directly.

## afp_logintest improvements

- connect_server(): zero-initialise the DSI struct and install a 15-second SO_RCVTIMEO to prevent indefinite blocking when the server stalls after accept()
- test4(): expand connection array to 201 (TEST4_MAX_CONNS), drive a full AFP login (FPopenLoginExt / FPopenLogin) rather than raw DSIOpenSession, accept both DSIERR_SERVBUSY (DSI-layer cap) and AFPERR_MAXSESS (AFP-layer cap) as "limit hit", skip DSICloseSession on SERVBUSY (no session was established), and sleep 3 s after teardown to let the server reap workers
- test7() (new): DSI I/O round-trip smoke test — sends an unknown AFP command (0xFF) and verifies a non-zero dsi_code is returned; catches regressions in the renamed send/receive path independently of the login tests
- test8() (new): exercises AFPopenLoginExt_pwd() directly with No User Authent on AFP >= 3.0, bypassing FPopenLoginExt version-branching
- test9() (new): AFPLoginCont smoke test — provokes kFPAuthContinue via DHCAST128 then sends 16 bytes of garbage; verifies the round-trip reaches the UAM layer without crashing the server; skipped if DHCAST128 is absent or server does not return AFPERR_AUTHCONT
- test10() (new): UAM matrix walker over a dispatch table; covers No User Authent and Cleartxt Passwrd with placeholder slots for Randnum Exchange, 2-Way Randnum, DHCAST128, DHX2, and SRP

## Test runner infrastructure

- -f <name>: run a single test by symbol name
- -l: list all registered tests and exit with code 2
- aggressive parameter assertion in the afpclient interface

## libatalk/dsi: fix session-cap enforcement in dsi_getsession

Two bugs when server_child_add returns NULL because the session cap is reached (introduced in GitHub https://github.com/Netatalk/netatalk/pull/2831):

1. Post-fork path fell through to *childp = NULL; return 0, which made the parent process enter the child code path in dsi_start(), calling afp_over_dsi() and exit()ing, killing the listener. Fixed by calling dsi->proto_close() and returning -1 inside the failure block.

2. SERVBUSY was sent with dsi_command=0 and dsi_requestID=0 because the parent never reads the DSI header (the child reads it after the fork). Clients that match responses by request ID would discard the reply and time out. Fixed by adding a pre-fork cap check that accepts the TCP connection without forking, reads the DSIOpenSession header with MSG_DONTWAIT (non-blocking, no DoS risk), and sends a correctly-formed SERVBUSY with the echoed command and request ID before returning -1. Eliminates the fork+SIGKILL cycle for every over-cap connection. The post-fork check in server_child_add remains as defence-in-depth.

In afpd: log DSI session error only when errno is set

### DIFF
--- a/.github/workflows/containers.yml
+++ b/.github/workflows/containers.yml
@@ -1093,7 +1093,7 @@ jobs:
             -e AFP_USER="atalk1" \
             -e AFP_PASS="$CONT_PASSWD" \
             -e AFP_GROUP="afpusers" \
-            -e INSECURE_AUTH="1" \
+            -e AFP_UAMS="uams_guest.so uams_clrtxt.so uams_dhx.so" \
             -e VERBOSE="1" \
             -e TESTSUITE="login" \
             -e AFP_VERSION="7" \
--- a/doc/manpages/man1/afp_logintest.1.md
+++ b/doc/manpages/man1/afp_logintest.1.md
@@ -4,13 +4,15 @@ afp_logintest — AFP authentication and DSI session test suite
 
 # Synopsis
 
-**afp_logintest** [-1234567CmVv] [-h *host*] [-p *port*] [-s *volume*] [-u *user*] [-w *password*]
+**afp_logintest** [-1234567CmVv] [-f *test*] [-h *host*] [-p *port*] [-s *volume*] [-u *user*] [-w *password*]
+
+**afp_logintest** -l
 
 # Description
 
 **afp_logintest** is a testsuite for DSI sessions and authentication from an AFP client.
 It will run a range of happy path and corner case tests for establishing a DSI session
-with an AFP server and then use a subset of available UAMs to authenticate with a user.
+with an AFP server and then use UAMs (User Authentication Methods) to authenticate with a user.
 
 # Options
 
@@ -38,8 +40,14 @@ with an AFP server and then use a subset of available UAMs to authenticate with 
 **-C**
 : Turn off ANSI colors in terminal output
 
+**-f** *test*
+: Run a specific test
+
 **-h** *host*
 : Server hostname or IP address (default: localhost)
+
+**-l**
+: List available tests
 
 **-m**
 : Run tests in AppleShare (Mac) AFP server compatibility mode
@@ -68,6 +76,9 @@ Configure the UAMs in netatalk's afp.conf:
     [Global]
     uam list = uams_clrtxt.so uams_guest.so
 
+Other tests may require additional UAMs to be configured;
+see the test output for details on which tests failed due to missing UAMs.
+
 Additionally, in order to test non-Guest authentication, a username and password must be passed to the test runner.
 
 # Examples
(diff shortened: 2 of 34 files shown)


NJRoadfan:
Do any AFP 3.0 servers (which support FPOpenLoginExt) actually support logging in with the ClearTxt UAM? Another issue with the test suite is that if you want to test against a newer macOS server (starting with 10.5?), they don't support anything but DHX2 login for security reasons.

rdmark:
@NJRoadfan hypothetically yes, if you explicitly enable the insecure UAMs in the AFP client

This knowledge base article talks about Lion specifically, but it would be interesting to test which OSX versions retain which legacy UAMs.

https://support.apple.com/en-us/101137

rdmark:
and, it is my goal with this series of improvements to eventually be able to run the testsuite tests authenticated with any UAM

NJRoadfan:
FInally spun up a MacOS X 10.4 VM and the built-in AFP server supports Clear Text logins. It gives another practical test case for this part of the test suite.

rdmark:
@andylemin I think this is in good shape now and ready for your review -- a foundation for two future improvements:

1. the ability to run any testsuite authenticated with any UAM
2. thorough testing of all UAM happy path and error handling

andylemin (approved):

andylemin:
Wow, a lot here.. as it's all testsuite stuff I see no issues with merging this in the 13th hour..

rdmark:
@andylemin indeed, it is a big one, I justify it with the resolving of tech debt that's been bothering me for 2 years (testing AFP 3.x with the anachronistic FPLogin method)

FTR, I force pushed a small change that cleans up the code comments a bit.

rdmark:
@NJRoadfan I ran the improved afp_logintest against the Mac OS X 10.7 AFP server (AFP 3.4) and all tests except test9 are passing _without_ the Mac compatible flag (`-m`). You may want to test against your test lab of older machines to see if the Mac compatibility code is needed or not.

This is how test9 fails with Apple's AFP server (DHCAST128 UAM enabled); another Netatalk quirk we should look closer at in the future.

```
############## entering test9 ##############
AFPopenLoginExt returned -5019 (AFPERR_PARAM   ), expected AFPERR_AUTHCONT
	NOT TESTED
Logintest:test9: AFPLoginCont primitive round-trip - NOT TESTED
```

PR #3001, opened 2026-05-15
spotlight/testsuite: raise Spotlight result cap to 10000 with paginated cnid_find

andylemin:
Spotlight searches were previously capped at 100 results across all three CNID backends (BerkeleyDB, SQLite, MySQL). Raise the cap to 10000 by paginating the cnid_dbd SEARCH RPC, with a 10-second wall-clock budget bounding total search time on busy volumes.

cnid_find() gains a `more_available` parameter so callers can tell truncated result sets apart from complete ones, and a 400-byte minimum result-buffer requirement that is enforced uniformly across backends. Parameter validation is centralised in the libatalk wrapper so the three backends now behave identically against bad input. The libatalk `soversion` is bumped to reflect the breaking ABI change.

The Spotlight backend now silently ignores filename queries shorter than 3 characters; nad find and FPCatSearch / CatSearchExt continue to accept short prefixes unchanged.

Adds an FPSpotlightRPC spectest covering open / query / drain / close, gated on the existing have_spotlight meson option. The Spotlight testsuite helpers live in their own translation unit to avoid a DSI header-guard collision between the testsuite-local and libatalk dsi.h that would otherwise produce a silent ABI mismatch.

Also bumps the debug-entrypoint perf sampling frequency from 1019 Hz to 1193 Hz to better decouple from the idle-worker poll interval, and dumps the generated afp.conf at container shutdown when SERVER_CONFIG is set, both useful for flamegraph CI runs.

Material security improvements as side effects of this change:

  * dbd_search no longer treats client-induced validation errors (malformed srch_offset, oversize offset, empty search name) as daemon-fatal. Returning -1 from a request handler aborts the single-threaded cnid_dbd via main.c's dispatch loop and recycles CNID service for every volume on the host, which would have been a remote-DoS vector. -1 is now reserved for genuine BDB engine errors; client errors return 1 with RES_ERR_DB.

  * dbd_search log lines that print the wire-supplied search name now use the %.*s precision specifier. The wire payload is bounded by rqst.namelen (CVE-2026-44050) but is not NUL-terminated, so the previous %s formatter would walk past the search name into adjacent receive-buffer memory and into the daemon log file.

  * dbd_search zeroes its static result buffer between requests, so any future write-side bug (e.g. a forgotten rply->namelen update) cannot leak stale CNIDs from one TCP client to the next.

  * Two latent SIGTERM/SIGINT handler leaks in libatalk/cnid/dbd's transmit() are closed: the ECONNREFUSED and MAX_DELAY return paths skipped the sigaction() restore block, leaving the AFP worker with cnid_dbd's signal handler installed for the rest of its lifetime. The new transmit_locked / transmit split routes every exit through a single restore epilogue.

  * dbif_search fails closed on a NULL `more` out-parameter rather than NULL-derefing. cnid_dbd is single-threaded, so a SEGV here would have killed CNID service for every volume on the host.


doc/developer/cnid-dbd-protocol.md
```diff
@@ -0,0 +1,152 @@
...
+
+**Daemon-side reply-code semantics (`SRCH_CNT` vs `SRCH_DONE`)**: the daemon emits `SRCH_CNT` iff
+`dbif_search` confirms a `(srch_offset + DBD_MAX_SRCH_RSLTS + 1)`-th matching entry exists in the
+secondary index — the cursor walk performs a one-step *peek* immediately after the buffer fills,
+and tests the result for `(DB_NOTFOUND ∨ prefix-mismatch ∨ engine-error)` vs `(success ∧
+prefix-match)`. The legacy "answer `SRCH_CNT` whenever `count == DBD_MAX_SRCH_RSLTS`" heuristic is
+**incorrect** because it cannot distinguish the boundary case where the matching range contains
+exactly `srch_offset + DBD_MAX_SRCH_RSLTS` entries (no further matches; correct answer is
+`SRCH_DONE`) from the case with strictly more entries (correct answer is `SRCH_CNT`). See
+[`etc/cnid_dbd/dbif.c::dbif_search`](../../etc/cnid_dbd/dbif.c) for the post-fill peek
```
rdmark: it's better not to make this a markdown hyperlink: Doxygen will identify source code file names and link them when it builds the docs

doc/developer/cnid-dbd-protocol.md
```diff
@@ -0,0 +1,152 @@
...
+> that asserts a strict `got == N` count over paginated results MUST
+> guarantee no concurrent CNID ADD/DELETE on the volume during the
+> query. Concurrent mutation between paginated SEARCH batches can
+> produce duplicates or misses on the DBD backend (stateless cursor
+> re-walk); `test530`/`test531` use static, test-private corpora and
+> long unique filename prefixes for this reason.
+
+## Single-threaded daemon
+
+The cnid_dbd dispatch loop ([`etc/cnid_dbd/main.c`](../../etc/cnid_dbd/main.c)) is single-threaded.
```
rdmark: same as the above!

doc/developer/cnid-dbd-protocol.md
```diff
@@ -0,0 +1,152 @@
...
+```
+
+For same-host topologies the systemd `Type=forking` unit terminates the entire process group;
+`cnid_metad` and all `cnid_dbd` children restart together. Remote `cnid server = <host>`
+deployments require restarting cnid_metad on each remote host.
+
+This release's wire change is **SEARCH-only**. Operators are expected to upgrade all components
+together; the project does not support mixed-version deployments. Any unexpected SEARCH reply code
+from the daemon is a protocol violation and `abort()`s the AFP worker, matching the convention of
+every other `cnid_dbd_*` op.
```
rdmark: I think you should lead this section with explaining the wire format change, then talk about restarting services after that

even better would be a table that breaks down historical wire format changes!

finally a nitpick: seems out of place to put exact init system commands here: 1) this is arguably prerequisite knowledge when you got this far, and 2) such information tend to not be future-proof

etc/cnid_dbd/dbd_search.c
```diff
@@ -22,29 +23,95 @@ int dbd_search(DBD *dbd, struct cnid_dbd_rqst *rqst, struct cnid_dbd_rply *rply)
...
-    LOG(log_debug, logtype_cnid, "dbd_search(\"%s\"):", rqst->name);
-    memset(&key, 0, sizeof(key));
     rply->name = resbuf;
     rply->namelen = 0;
-    key.data = (char *)rqst->name;
-    key.size = rqst->namelen;
+    /* Defence-in-depth: zero `resbuf` between requests so any future
+     * write-side bug (e.g. forgetting to update rply->namelen) cannot
+     * leak stale CNIDs from a prior request to the same TCP client. */
+    memset(resbuf, 0, sizeof(resbuf));
```
rdmark: do we want to use `explicit_bzero` here instead of `memset` to guarantee that it doesn't get optimized away by the compiler?

etc/cnid_dbd/dbif.c
```diff
@@ -967,57 +967,144 @@ int dbif_del(DBD *dbd, const int dbi, DBT *key, uint32_t flags)
...
     char *cnids = resbuf;
     cnid_t cnid;
     char *namebkp = key->data;
-    int namelenbkp = key->size;
+    /* DBT::size is u_int32_t */
+    size_t namelenbkp = (size_t)key->size;
+
+    /* API contract: caller MUST supply a backing bool. cnid_dbd is a
+     * single-threaded long-lived daemon — a NULL deref here would SIGSEGV
+     * the daemon and kill CNID service for every volume on the host. Fail
```
rdmark: would a crash here really take down all cnid_dbd processes owned by other authenticated users?

my assumption has always been that they're isolated from each other, so that only the current user's cnid_dbd dies and everyone else have uninterrupted service

but I could be wrong

etc/cnid_dbd/dbif.c
```diff
@@ -967,57 +967,144 @@ int dbif_del(DBD *dbd, const int dbi, DBT *key, uint32_t flags)
...
-        count++;
+        if (skipped < offset) {
+            skipped++;
+            ret = cursorp->pget(cursorp, key, &pkey, &data, DB_NEXT);
+            continue;
+        }
+
         memcpy(cnids, pkey.data, sizeof(cnid_t));
-        memcpy(&cnid, pkey.data, sizeof(cnid_t));
+        memcpy(&cnid,  pkey.data, sizeof(cnid_t));
```
rdmark: nitpick: awkward extra space added between `&cnid` and `pkey.data` :)

include/atalk/cnid.h
```diff
@@ -49,6 +52,14 @@
 #define CNID_ERR_CLOSE 0x80000004   /*!< the db was not open */
 #define CNID_ERR_MAX   0x80000005
 
+/* cnid_find() requires a result buffer large enough to hold at least one
+ * full DBD-backend batch (CNID_FIND_MIN_RESULTS = 100 CNIDs in network
+ * byte order). All three CNID backends (dbd, sqlite, mysql) enforce this
+ * same minimum so the API behaves identically regardless of which
+ * backend is configured. */
```
rdmark: nitpick: listing up currently supported cnid backends and make assertions about their capabilities is the type of code comment documentation that may get stale in the future

include/atalk/cnid_bdb_private.h
```diff
@@ -1,6 +1,15 @@
 /*!
  * @file
- * Interface to the cnid_dbd daemon that stores/retrieves CNIDs from a database.
+ * Interface to the cnid_dbd daemon. The structs below are the on-the-wire
+ * layout of a private TCP/IP RPC (default localhost:4700; per-volume
+ * `cnid server` / `cnid port` afp.conf options). All fields are written in
```
rdmark: nitpick again: talking about configuration defaults and afp.conf options in a code comment may lead to outdated information in the future -- I personally like to avoid such details in code comments

etc/cnid_dbd/dbd_search.c
andylemin: Nice. Done

etc/cnid_dbd/dbif.c
andylemin: Inspecting the handling and design again and you are right 🙂 The statement should be "every user/session accessing that one volume".

There's one cnid_dbd per volume, shared by all afpd workers (and thus all users) accessing that volume.
So a SIGSEGV in cnid_dbd:
Kills CNID service for that one volume only (other volumes have their own cnid_dbd)
Affects all users accessing that _volume_, not just the one who triggered the bug
Does NOT affect other cnid_dbd processes for other volumes

libatalk/cnid/sqlite/cnid_sqlite.c
```diff
@@ -1086,8 +1086,20 @@ int cnid_sqlite_getstamp(struct _cnid_db *cdb, void *buffer,
...
+ * cdb, name, namelen and buflen are all sane on entry. Detects truncation
+ * by binding LIMIT max_results+1 and observing whether the extra row was
+ * produced; reports it via @p more_available.
+ */
+/* namelen is unused: the SQLite backend builds the LIKE pattern via
+ * asprintf("%%%s%%", name), which already requires a NUL-terminated
+ * `name`. We keep the parameter to satisfy the cnid_db function-pointer
+ * signature shared with the dbd / mysql backends. */
+int cnid_sqlite_find(struct _cnid_db *cdb, const char *name, size_t namelen _U_,
+                     void *buffer, size_t buflen, bool *more_available)
```
rdmark, line 1103: from experience, some Doxygen parsers will error our here; please merge these two comments into one

libatalk/meson.build
```diff
@@ -53,8 +53,8 @@ libatalk = library(
     include_directories: root_includes,
     dependencies: [libatalk_deps, libcnid_deps],
     link_whole: libatalk_libs,
-    version: '19.0.0',
-    soversion: '19',
+    version: '20.0.0',
+    soversion: '20',
```
rdmark, line 57: shall we decide on a project policy for library versioning?

this project used to have granular ABI compatibility tracking but it was abandoned during the 3.1 release series

rdmark:
@andylemin pardon, I fell asleep halfway reviewing this PR last night! first pass completed now. 👍

rdmark (approved):
I can see how the DBD backend caused you headaches -- this is a not insignificant amount of plumbing!

andylemin:
@rdmark Build warnings fixed and resolved the stale and colliding enums and aliases in second commit (this was a satisfying cleanup).
Fully tested now and RPC pagination between client and server is working well now 🙂

rdmark:
excellent, let's ship it!

separately, I feel bad about robbing you of PR #3000 -- this one is much more deserving of the honor ;-)

andylemin:
> excellent, let's ship it!
> 
> separately, I feel bad about robbing you of PR #3000 -- this one is much more deserving of the honor ;-)

@rdmark Hahah funny you mention it as I did notice 3001 actually and though I just missed it 😂

andylemin:
@rdmark I just tested a full search and it ran perfectly to 10000 queries 👍
Yes the DBD backend was the entire commit really.. I thought about limiting it to MySQL and SQLite so many times, but that would have caused a headache for the spectests being backend agnostic....

PR #3005, opened 2026-05-17
update changelog for stable release v4.5.0

rdmark:
pull in and consolidate changelog for v4.4.2 and v4.4.3 for completeness, and fold the v4.5.0beta changelog into v4.5.0 proper

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,71 +1,146 @@
 Netatalk Changelog
 ==================
 
-Changes in 4.5.0beta
---------------------
+Changes in 4.5.0
+----------------
 
-* BREAKING: libatalk: cnid_find() ABI changed — gained a `bool *more_available` out-parameter and a
-  400-byte minimum result-buffer requirement. The libatalk soversion has been bumped accordingly.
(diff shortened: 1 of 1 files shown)


andylemin (approved):

andylemin:
@rdmark actually please don't merge it like this. The beta changes should be part of this release as the beta changes were not a full release (was source only).

All the major open projects I know always list the beta change log again in the main release, as the beta is only an early preview. This is the first time we are introducing all those features in a public release?

It also shows why people should upgrade, otherwise it looks (at a glance) like there's not much in the release.

For an example, take a look at how OpenZFS manage their change notes on their betas and releases

andylemin (changes requested):
Please combine beta change log into release change log

rdmark:
sounds like a good policy, I can see how this is a preferable format for most readers of the changelog

in fact, I have found it hard sometimes to read the historical changelogs of this project when there's a long chain of alpha/beta releases, to wrap your head around what was actually new in a major feature release

andylemin (approved):

PR #3006, opened 2026-05-17
afpd: harden follow symlinks against cross-device targets

rdmark:
Two complementary checks are added to prevent symlinks from crossing filesystem device boundaries when the follow symlinks volume option is in use.

symlink_target_safe() in file.c now stats both the volume root and the resolved symlink target, rejecting the symlink if their st_dev values differ.

A new ochdir_vol() wrapper in directory.c calls ochdir() and then, when symlink following is active, stats the resulting working directory and rejects the chdir if it landed on a different device than the volume. This catches pre-existing or locally-created cross-device symlinks at traversal time, which creation-time validation cannot cover.

Documentation in afp.conf.5.md and doc/manual/Configuration.md is updated to reflect both validations and the remaining caveat about pre-existing symlinks.

### DIFF
--- a/doc/manpages/man5/afp.conf.5.md
+++ b/doc/manpages/man5/afp.conf.5.md
@@ -1348,13 +1348,18 @@ delete any files and directories within the vetoed directory.
 follow symlinks = *BOOLEAN* (default: *no*) **(V)**
 
 > The default setting is false thus symlinks are not followed on the
-server. This is the same behaviour as OS X's AFP server. Setting the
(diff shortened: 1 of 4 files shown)


rdmark:
@andylemin the impetus for this change was the long-standing *will subtly break when the symlinks point across filesystem boundaries* remark in the afp.conf.5 man page -- we don't want anything to subtly break, do we. ;)

andylemin (approved):

PR #3018, opened 2026-05-21
afpd: rework afpstats to drop D-Bus and reveal ASP sessions

### LINKED ISSUE #3017: afpd: Enabling afpstats and using the localsearch Spotlight backend causes conflict.
**Describe the bug**

If afpstats is enabled and one opts to use the localsearch Spotlight backend, a client will lockup if one attempts to do a Spotlight search on the AFP volume. This appears to be some sort of DBus related conflict. The FPSpotlightRPC calls never occur over the wire if afpstats is enabled. The Xapian and CNID Spotlight backends are not affected and work fine. Disabling afpstats allows the localsearch backend to function as normal.

**To Reproduce**

-Compile Netatalk with Localsearch and afpstats support.
-Setup afp.conf with afpstats = yes, spotlight = yes, and use spotlight backend = localsearch for your share.

**Expected behavior**

Spotlight searches should work.

**Environment**

- Server OS: Ubuntu server 26.04 LTS and Ubuntu Server 24.10 LTS
- Client OS: macOS Catalina
- Netatalk Version: 4.5 main tree

rdmark:
Replace the GDBus-based afpstats listener with a plain Unix domain socket. The GDBus listener ran in a GThread inside the afpd master process; every per-session fork() then inherited GDBus's process-global worker-thread and singleton state in a broken form. The first GDBus call in the child -- typically tracker_sparql_connection_bus_new() from the localsearch Spotlight backend -- would deadlock, hanging Spotlight searches for clients of servers built with both features enabled.

Socket details:

* afpd master bind()s $LOCALSTATEDIR/netatalk/afpstats.sock (mode 0660), registers it as a new STATS_FD type in the existing asev poll loop, and unlinks it on shutdown.
* Each client connection is served inline: walk server_children under servch_lock and write one text line per session in the format the afpstats Perl client already parses, then close.
* The Perl CLI switches from Net::DBus to IO::Socket::UNIX (Perl core) and computes column widths dynamically so long hostnames and volume lists no longer bleed into the next column.

While here, plumb the IPC socketpair and dircache-hint pipe through the ASP fork path in libatalk/asp/asp_getsess.c the same way DSI does in dsi_getsess.c. The shared auth.c IPC_LOGINDONE and volume.c IPC_VOLUMES sites already fire for any protocol; they were no-ops for ASP because obj->ipc_fd was -1. ASP sessions now populate afpch_valid/uid/hostname/volumes and appear in afpstats output.

Build wiring:

* Removed the -Dwith-afpstats compile time option; now always built.
* Removed the -Dwith-dbus-sysconf-path option as well, no longer needed
* Runtime option flag renamed OPTION_DBUS_AFPSTATS -> OPTION_AFPSTATS.
* The Darwin afpstats build gate is removed; Unix sockets are portable.


andylemin (approved):
Nice!! This is very cool..

Great to see afpstats finally free of dbus!

Much cleaner for forks, more secure, and more useable! 
I might actually use it now 😂

rdmark:
@andylemin indeed, we can now call afpstats a true drop-in replacement for macusers!

what do you think about the compile time feature macro? shall we retain the ability to turn this feature off with -Dwith-afpstats=false and hide the implementation behind `WITH_AFPSTATS`? we can potentially remove this altogether and always build the feature

the feature is still behind a runtime afp.conf option, disabled by default, for highly security conscious users who don't want a world-accessible socket leaking information about the AFP server

rdmark:
@NJRoadfan can you test if the ASP session user stats code is working end to end? I don't have an AppleTalk setup right now

andylemin:
> what do you think about the compile time feature macro? shall we retain the ability to turn this feature off with -Dwith-afpstats=false and hide the implementation behind `WITH_AFPSTATS`? we can potentially remove this altogether and always build the feature

I'm ok with it being built by default as long as disabling it in the config mitigates any performance overhead and socket security?

rdmark:
augment review

rdmark:
@andylemin the afpstats code is not in any critical path in my reading: a single poll/epoll slot in the parent's event loop, never touched unless someone runs afpstats

in fact, I noted on closer inspection that the server child and server IPC mutexes are no longer needed, as no GLib threads are created anymore -- the end result is very little bespoke afpstats code left in afpd

NJRoadfan:
I will do AppleTalk testing later today. If there is room in the output, maybe an extra field for protocol in the report? I don't know how many people would use it though, but seeing `tcpip` or `atalk` next to a session would be neat.

I'm trying to figure out why DBus was used for this to begin with. Was it a scalability issue? Security? Seems like an awful lot of code and overhead for this feature when I socket clearly does it with less code. Maybe someone wanted to learn DBus?

rdmark:
@NJRoadfan something like this?

```shell
Connected user  PID    Login time       State   Protocol  Hostname             Mounted volumes
nobody          54863  May 21 20:02:58  active  tcpip     Daniels-MacBook-Pro  MyV1
atalk1          54734  May 21 20:02:46  active  tcpip     Daniels-MacBook-Pro  ICON2, ICON1
```

I wonder if Frank&Ralph initially had grand plans for the AFPStats D-Bus protocol? all sorts of fine grained administrative inspection and control...

rdmark:
augment review

NJRoadfan:
ASP logins are NOT working with this PR. I get an error -1068 and a message that the server isn't allowing anymore logins.

```
2026-05-21T19:27:48.628594+00:00 a2server afpd[1825]: afp_login: UAM login failed, sending reply
2026-05-21T19:27:48.636458+00:00 a2server afpd[1825]: login: too many connections, limit: 0
2026-05-21T19:27:48.684621+00:00 a2server afpd[1825]: AFP statistics: 0.05 KB read, 0.01 KB written via ASP
2026-05-21T19:27:48.684911+00:00 a2server afpd[1704]: child[1825]: asev_del_fd: 8
```

NJRoadfan:
Note that I'm getting `afp_login: UAM login failed, sending reply` even when connecting via TCP/IP and with the correct credentials.

rdmark:
if you don't mind, let me postpone the protocol idea until after the release, the code changes are getting too intrusive for a last-minute fix

rdmark:
@NJRoadfan reverted to the previous state where ASP sessions are tracked, but the protocol type is not printed

can you please test again in your environment if it works better?

I have tested on macOS and Debian with appletalk enabled, but only DSI connections, which seems to work

NJRoadfan:
Still not working here with ASP. DSI connections are fine. For some reason the server connection limit appears to be zero for them and denies login because the server thinks it has too many clients.

I would roll this back for now. Maybe lockout afpstats if localsearch is enabled and call it a day.

rdmark:
@NJRoadfan one final try: does this work better with ASP? I agree that reverting is better, but just curious if this gets it functional or not

NJRoadfan:
That did the trick. I can now connect to the server again and see Appletalk users logged in on afpstats. The state shows as "unknown", but really should always be "active" for AppleTalk clients since they don't support session reconnect.

etc/afpd/afpstats.c
```diff
@@ -17,181 +18,175 @@
...
+    (void)fcntl(fd, F_SETFD, FD_CLOEXEC);
 
-static void on_name_lost(GDBusConnection *connection _U_,
-                         const gchar     *name,
-                         gpointer         user_data _U_)
-{
-    LOG(log_debug, logtype_afpd, "on_name_lost: %s", name);
-}
+    /* Remove any stale socket from a previous run */
+    if (unlink(sock_path) != 0 && errno != ENOENT) {
```
andylemin, line 84: unconditionally unlinks any pre-existing socket file, then binds. If two afpd parents are launched against the same _PATH_STATEDIR (e.g., a misconfigured systemd unit or a packaging mishap), the second one silently steals the path from the first and the original parent's listener becomes orphaned (still bound, but no one connects). Compare to how netatalk handles its PID file — there's typically an existing-instance check upstream. Not major, just inconsistent

Either:
document that afpstats relies on the existing afpd PID-file lock to serialize startup, or
check whether the socket has a live listener (connect() probe) before unlinking.
The first is probably sufficient and cheaper; just say so in a comment.

etc/afpd/afpstats.c
```diff
@@ -17,181 +18,175 @@
...
-    g_dbus_connection_unregister_object(bus, registration_id);
-    g_dbus_node_info_unref(introspection_data);
-    g_object_unref(obj);
-    g_main_context_pop_thread_default(ctxt);
-    g_main_context_unref(ctxt);
-    return thread_loop;
-}
+    /* Allow any local user to read session info, mirroring the prior
+     * D-Bus policy that granted unrestricted access on the system bus. */
+    if (chmod(sock_path, 0666) != 0) {
```
andylemin: The comment justifies this as mirroring the prior D-Bus policy:

The previous D-Bus policy at config/netatalk-dbus.conf did grant unrestricted access on the system bus, so behavior-preserving is fair.

*But a UDS in /var/lib/netatalk/ is not the same trust surface as the system bus*.
Any local user can now enumerate all AFP session usernames, PIDs, hostnames, and mounted volumes. On multi-user hosts (which AFP servers historically were), that is more session-metadata disclosure than most operators expect.

Can we consider:
default 0660 with group netatalk (or whatever group afpd runs as?), and a config knob to widen it; or
at minimum, a loglevel=info line at startup announcing the permissive mode so operators see it.
This is a judgment call, not a blocker — but please get explicit sign-off rather than burying it in a comment.

etc/afpd/afpstats.c
```diff
@@ -17,181 +18,175 @@
...
+                    goto unlock;
+                }
+            }
+
+            child = child->afpch_next;
+        }
     }
 
-    return 0;
+unlock:
```
andylemin: Looks like a leftover from the mutex era? it only does close(client_fd). With servch_lock gone, nothing is unlocked. Rename to done: or out: or remove maybe?

andylemin:
doc/Upgrading.md is not updated. Operators upgrading from a release that started a system dbus-daemon for afpstats need update change note.

description still notes add protocol reporting to afpstats by tracking each child session's transport protocol at creation time and exposing it through the stats socket output - think this has been dropped for now?

etc/afpd/afpstats.c
```diff
@@ -17,181 +18,175 @@
...
+                     * keep the session visible with a numeric fallback. */
+                    snprintf(namebuf, sizeof(namebuf), "uid=%u",
+                             (unsigned)child->afpch_uid);
+                    uname = namebuf;
+                }
+
+                strftime(timebuf, sizeof(timebuf), "%b %d %H:%M:%S",
+                         localtime(&login_time));
+
+                if (dprintf(client_fd,
```
andylemin, line 214: concern: if the peer closes mid-write, the kernel raises SIGPIPE on the writing process before dprintf gets a chance to return -1 with EPIPE. The error-handling branch is correct, but it's only reached if SIGPIPE is ignored or blocked — otherwise the parent afpd dies on a misbehaving stats client.

dprintf has no MSG_NOSIGNAL equivalent, a fix could be:
ignore SIGPIPE process-wide at afpd startup (signal(SIGPIPE, SIG_IGN)), or
on BSD/macOS, set SO_NOSIGPIPE on client_fd near the existing SO_SNDTIMEO setsockopt at etc/afpd/afpstats.c:149
Worth checking whether afpd already does SIG_IGN on SIGPIPE somewhere in main.c startup — if so, this is already covered.

Not a critical, should be a very rare edge case

andylemin:
Otherwise looking good in general.

Does the IPC + dircache hint pipes plumbing through ASP, mean ASP clients now support dircache hints?

Answered myself, so while the hint pipe is plumbed through the ASP path and the child receives obj->hint_fd, the ASP session loop never reads it.

While ASP clients do fully utilise the new dircache framework, the ASP loop is missing the pipe handling and does not use the new multi-fd poll() in DSI either.

However it looks like a fairly simple refactor of asp_getrequest to be multi-fd poll driven

andylemin:
Please can you squash the commits once you are happy you have a working final version to enable a last review? Tricky to review at the moment?

etc/afpd/afpstats.c
rdmark: it's fair observation, and we arguably have a similar situation with the local CNID database files, namely BerkeleyDB and SQLite, which are world read and writable by default

I don't think we have a scheme for running netatalk/afpd as a specific group, each process run in the process owners group

if we shore up this and enforce the group, we can harden both this socket and the CNID databases in one fell swoop

the question is how to do this seamlessly without breaking thousands of existing installations that upgrade? ideally, only users that are members of this "netatalk" group should be able to connect to the AFP server in the default configuration, too. the default configuration should be more secure.

rdmark:
> However it looks like a fairly simple refactor of asp_getrequest to be multi-fd poll driven

@andylemin do you see this as a fast follow improvement, or shall I draft this here in this PR?

rdmark:
> That did the trick. I can now connect to the server again and see Appletalk users logged in on afpstats. The state shows as "unknown", but really should always be "active" for AppleTalk clients since they don't support session reconnect.

@NJRoadfan great, then we're on the right track!

having special handling of the state name for ASP will require afpstats to be aware of the protocol of each session, which brings us back to the same situation as before: a new field in the afp_child structure, which changes the ABI in server_child.h slightly.

if you don't mind, let's postpone this improvement until later, together with the printing of the protocol for each session!

NJRoadfan:
We don't actually need a new field for it, just a new state flag. Right now "active" is assigned to DSI_RUNNING. One could add a new state ASP_RUNNING (1 << 10) and derive the protocol from that. Given that the state of an ASP connection will always be active and thus static, its a fair compromise.

You can set the state of the connection in `login()` in auth.c per my testing right after this line: https://github.com/Netatalk/netatalk/blob/c5e2f021f4f41313794dd8a4371d404e4dd720d3/etc/afpd/auth.c#L343

Conceptual code to be inserted here:

```
    if (obj->proto == AFPPROTO_ASP) {
        ipc_child_state(obj, ASP_RUNNING);
    }
```

For the protocol column in afpstats, check the state, if ASP_RUNNING, its AppleTalk, else TCP/IP.

etc/afpd/afpstats.c
rdmark: @andylemin let me table the broader question for now and get back to the net new code in this PR!

how about this for a solution:

- the afpstats socket is created with 0660 -- now only sudoer users can use it because afpd creates the socket owned by the sysadmin's group (admin, staff, etc.)
- introduce a `afpstats group` runtime option that both chowns the socket to this group, and allows members of this group to access it

gives us a safe-by-default configuration, with an opt-in lenient option. thoughts?

etc/afpd/afpstats.c
rdmark: good observation, I opt for option 1 in this case

etc/afpd/afpstats.c
rdmark: good catch, let me lightly refactor this

rdmark:
@NJRoadfan I see your argument, and I do appreciate the unintrusive nature of this solution. My only caveat is that it's kind of an indirect check, but since we're highly unlikely to add a third transport layer in the future it's good enough.

etc/afpd/afpstats.c
rdmark: afpd already ignores SIGPIPE during startup in etc/afpd/main.c, so this stats write path should receive a dprintf() error instead of terminating the parent. I added a short comment near the dprintf() error handling to make that dependency explicit!

good instinct though, it could have been a serious issue!

rdmark:
augment review

NJRoadfan:
So far so good here.

```
Connected user  PID   Login time       State     Protocol   Hostname  Mounted volumes
user1           3431  May 22 18:01:58  active    AppleTalk  a2server  A2FILES, user1's home
nobody          3494  May 22 18:05:31  active    AppleTalk  a2server  A2FILES
user1           3472  May 22 18:03:53  sleeping  TCP/IP     a2server  A2FILES
```

Forgot to note that with the permissions changes you have to run afpstats with root privileges now.

andylemin:
> > However it looks like a fairly simple refactor of asp_getrequest to be multi-fd poll driven
> 
> 
> 
> @andylemin do you see this as a fast follow improvement, or shall I draft this here in this PR?

I think it should be a separate PR as the clock is ticking for the CVE announcements, so we need to get fixes into people hands.

So let's keep this focused on the fix

rdmark:
> So far so good here.
> 
> ```
> Connected user  PID   Login time       State     Protocol   Hostname  Mounted volumes
> user1           3431  May 22 18:01:58  active    AppleTalk  a2server  A2FILES, user1's home
> nobody          3494  May 22 18:05:31  active    AppleTalk  a2server  A2FILES
> user1           3472  May 22 18:03:53  sleeping  TCP/IP     a2server  A2FILES
> ```
> 
> Forgot to note that with the permissions changes you have to run afpstats with root privileges now.

there is an OS dependency here: on macOS for instance, all users with the flag "Allow this user to administer this computer" are on the same default group as the root user (`admin`) so such a user get default access to afpstats

on Debian 13 however, the equivalent group is `staff` which is not equivalent to `sudo`, but if you assign yourself to the former group (which is not an unreasonable thing to do for an admin user) you can use afpstats without sudo

rdmark:
@andylemin this is ready for your review! I removed the unrelated documentation / localization commits to reduce clutter

etc/afpd/afpstats.c
andylemin: Makes perfect sense to me. And having a dedicated afpstats group is the way to go so only the admin has access, and can selectively delegate access to others.

andylemin:
@rdmark Looking great. only minor nit picks such as; _PATH_STATEDIR "afpstats.sock" is repeated three times.
A #define AFPSTATS_SOCKET_PATH _PATH_STATEDIR "afpstats.sock" near the top (or in afpstats.h) would be nice.

PR #3021, opened 2026-05-23
afpd: don't log UAM continuations as failures

### LINKED ISSUE #3020: afpd: "afp_login: UAM login failed, sending reply" error generated with certain UAMs
**Describe the bug**

When logging into a server with the RandNum or DHX UAMs, Netatalk generates a log entry

`May 22 22:45:56 a2server afpd[1002]: afp_login: UAM login failed, sending reply`

These errors DO NOT occur with the Guest, SRP, or DHX2 UAMs.

**To Reproduce**

- Setup Netatalk to use the RandNum and/or DHX UAMs
- Log into a client using one of the above. The error should appear after authenticating with the server.

**Expected behavior**

This error should not occur. This is likely a regression introduced during code refactoring. Issue not present in 4.4.2.

**Environment**

- Client OS: macOS Catalina, MacOS 8.1, MacOS 9.1
- Netatalk Version: 4.5.0 main tree

rdmark:
DHX and RandNum return AFPERR_AUTHCONT from the initial login leg without a passwd result, then complete on AFP_LOGINCONT. Treat that continuation before checking pwd so the expected multi-step auth flow is not logged as a failed login.

This was a minor breakage from https://github.com/Netatalk/netatalk/pull/2831

### DIFF
--- a/etc/afpd/auth.c
+++ b/etc/afpd/auth.c
@@ -764,6 +764,10 @@ int afp_login(AFPObj *obj, char *ibuf, size_t ibuflen, char *rbuf,
     LOG(log_debug, logtype_afpd, "afp_login: UAM login returned %d, pwd=%p", i,
         pwd);
 
+    if (i == AFPERR_AUTHCONT) {
+        return send_reply(obj, i);
+    }
+
     if (!pwd || (i != AFP_OK && i != AFPERR_PWDEXPR)) {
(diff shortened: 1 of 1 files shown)


NJRoadfan:
Tested and working on this end. No more errors.

andylemin (approved):

PR #3022, opened 2026-05-23
cnid: make name search match substrings in DBD backend

rdmark:
DBD cnid_find walked the lowercased name index with DB_SET_RANGE and stopped at the first non-prefix key, making it prefix-only while the public cnid_find contract and the SQLite/MySQL backends use substring matching.

Scan the name index from the beginning and page over entries whose indexed name contains the search term, preserving the existing DBD_MAX_SRCH_RSLTS pagination and more-available signalling.

### DIFF
--- a/etc/cnid_dbd/dbif.c
+++ b/etc/cnid_dbd/dbif.c
@@ -966,13 +966,36 @@ int dbif_del(DBD *dbd, const int dbi, DBT *key, uint32_t flags)
     }
 }
 
+static bool dbif_name_contains(const void *name, size_t namelen,
+                               const void *term, size_t termlen)
+{
+    const char *n = name;
+    const char *t = term;
+
+    if (termlen == 0) {
+        return true;
+    }
+
+    if (namelen < termlen) {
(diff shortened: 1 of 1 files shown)


rdmark:
@andylemin without this fix, you can only get matches with the _start_ of a filename string when using the CNID search backend + DBD -- something that I ran into while testing Spotlight in general!

etc/cnid_dbd/dbif.c
```diff
@@ -1070,6 +1088,11 @@ int dbif_search(DBD *dbd, DBT *key, char *resbuf,
             continue;
         }
 
+        if (count == DBD_MAX_SRCH_RSLTS) {
```
andylemin, line 1091: Relying only on this means we overshoot if the total number of matches is divisible by DBD_MAX_SRCH_RSLTS - ie, we do an extra query that will have 0 results.

However this is benign, and should not be noticeable by the user due to RPC pagination

andylemin (approved):

PR #3023, opened 2026-05-23
afpd: let sleeping sessions handle deferred reconnects

### LINKED ISSUE #3019: afpd: Session reconnect feature appears to be broken
**Describe the bug**

Putting a client to sleep with mounted Netatalk shares then waking it results in the shares being disconnected. When machine wakes, a dialog appears indicating that connection to the server's shares was interrupted.

**To Reproduce**

- Setup Netatalk
- Mount a Netatalk share on a macOS machine
- Put machine to sleep for approximately two minutes to ensure the machine is completely asleep.
- Wake machine. Client should restore connections to mounted Netatalk shares. 

**Expected behavior**

If a machine is put to sleep with mounted shares, the client should seemlessly reconnect to them when woken up.

**Environment**

- Client OS: macOS Catalina
- Netatalk Version: 4.5.0Beta

**Configuration**

Share the (anonymized) contents of your afp.conf file,
plus any other relevant configuration files

**Logs**
```
May 22 19:48:28 a2server afpd[1197]: Login by user1 (AFP3.4)
May 22 19:48:40 a2server afpd[1197]: afp_zzz: entering extended sleep
May 22 19:48:51 a2server afpd[1197]: afp_zzz: waking up from extended sleep
May 22 19:49:00 a2server afpd[1197]: afp_zzz: entering extended sleep
May 22 19:50:49 a2server afpd[1204]: Login by user1 (AFP3.4)
May 22 19:50:49 a2server afpd[1204]: afp_disconnect: trying primary reconnect
May 22 19:50:49 a2server afpd[1177]: Reconnect: transferring session to child[1197]
May 22 19:50:49 a2server afpd[1177]: Reconnect: killing new session child[1204] after transfer
May 22 19:50:51 a2server afpd[1204]: afp_disconnect: primary reconnect succeeded
```
**Additional context**

It appears that the server is properly handling the re-connection of the old session per the logs. Note the new afpstats changes are not the source of the problem. This was broken with the posted 4.5.0Beta release and likely for a long time before that. I don't know if this is a client problem or a server problem at this point.

rdmark:
After the signal-safety refactor, SIGURG only sets transfer_pending and wakes the self-pipe. Sleeping AFP sessions were still bypassing poll() and blocking in dsi_stream_receive(), so the deferred reconnect handler never ran.

Allow sleeping sessions to remain in the poll path so self-pipe wakeups can process primary reconnects, while keeping disconnected and dying sessions on the existing special-state path.

### DIFF
--- a/etc/afpd/afp_dsi.c
+++ b/etc/afpd/afp_dsi.c
@@ -779,8 +779,9 @@ void afp_over_dsi(AFPObj *obj)
                 break;                          /* NO idle — data ready */
             }
 
-            /* [B] State check — handle sleep/disconnected/die */
-            if (dsi->flags & (DSI_EXTSLEEP | DSI_SLEEPING | DSI_DISCONNECTED | DSI_DIE)) {
+            /* [B] State check — disconnected/die bypass poll; sleeping must
+             * still poll so deferred SIGURG reconnects can wake via self-pipe. */
+            if (dsi->flags & (DSI_DISCONNECTED | DSI_DIE)) {
                 if (obj->hint_fd >= 0) {
                     process_cache_hints(obj);
                 }


rdmark:
@NJRoadfan can you test if this makes a difference for you re: reconnection bug?

andylemin (approved):

PR #3029, opened 2026-05-25
afpd: report client address in afpstats

rdmark:
Send the connected client host/address in the login IPC payload instead of the configured server hostname. Resolve TCP clients to reverse-DNS names when available, fall back to normalized IP strings, and report AppleTalk sessions by DDP net.node address.

**Before**

    $ afpstats
    Connected user   PID      Login time        State          Mounted volumes
    alice            452547   Apr 28 21:58:50   active         Test Volume, alice's Home
    bob              451969   Apr 28 21:21:32   active         My AFP Volume

**After**

    $ afpstats
    Connected user  PID     Login time       State     Protocol   Client address     Mounted volumes
    alice           452547  Apr 28 21:58:50  sleeping  TCP/IP     workstation.local  Data Vault, alice's Home
    bob             451969  Apr 28 21:21:32  active    AppleTalk  65280.12           My AFP Volume

### DIFF
--- a/contrib/bin_utils/afpstats.pl
+++ b/contrib/bin_utils/afpstats.pl
@@ -23,7 +23,7 @@ use IO::Socket::UNIX;
 my $DEFAULT_SOCKET = '@localstatedir@/netatalk/afpstats.sock';
 
 sub usage {
-    printf("Usage: %s [-h hostname] [-s socket] [-v]\n", basename($0));
+    printf("Usage: %s [-h client-address] [-s socket] [-v]\n", basename($0));
     exit 1;
 }
 
@@ -67,7 +67,7 @@ sub main {
         exit 1;
     }
 
-    my @header = ('Connected user', 'PID', 'Login time', 'State', 'Protocol', 'Hostname', 'Mounted volumes');
+    my @header = ('Connected user', 'PID', 'Login time', 'State', 'Protocol', 'Client address', 'Mounted volumes');
     my @rows;
 
     while (defined(my $user = <$sock>)) {
--- a/doc/manpages/man1/afpstats.1.md
+++ b/doc/manpages/man1/afpstats.1.md
@@ -4,7 +4,7 @@ afpstats — List Netatalk AFP server statistics
 
 # Synopsis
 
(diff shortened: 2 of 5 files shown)


andylemin (changes requested):
Agree this is a no-brainer. It's not clear why it was originally the server name really.

Just 2 minor static analyser issues.

Could you please also add an example afpstats output to the description for historical purposes 🙂

NJRoadfan:
This is working pretty good. Actually its an improvement over the `macusers` script in that the output is immediate, particularly with the client addresses. When running `macusers` as root, it took forever to resolve the client address and it didn't support AppleTalk addresses at all!

Eventually the Webmin panel for active afpd users should use `afpstats` for building the table, but that can wait for a future release. You may want to add a command line switch to add a delimiter to the output for easier parsing of the list.

rdmark:
wouldn't you say that the *Protocol* column is almost entirely redundant now?

a knowledgeable person would always know how to map a client address to the session protocol

one option for keeping ambiguity low could be to append an abbreviation to the addresses and remove *Protocol* as a separate column; mockup:

    $ afpstats
    Connected user  PID     Login time       State     Client address          Mounted volumes
    alice           452547  Apr 28 21:58:50  sleeping  workstation.local (IP)  Data Vault, alice's Home
    bob             451969  Apr 28 21:21:32  active    65280.12 (AT)           My AFP Volume

andylemin:
> wouldn't you say that the *Protocol* column is almost entirely redundant now?
> 
> a knowledgeable person would always know how to map a client address to the session protocol

For me it confirms that the client address is supposed to look like that. Without it I would believe it to be a corrupted value 😉

Please leave as it is as this provides concrete clarity

andylemin (approved):

NJRoadfan:
My only concern is that long host names and IPv6 addresses may lead to overflow onto the next line. Output options can come in a future version.

andylemin:
> My only concern is that long host names and IPv6 addresses may lead to overflow onto the next line. Output options can come in a future version.

Good call @NJRoadfan.
I would suggest truncating them to a max fixed width, and IPv6 addresses can be shorted with a few common conventions.

PR #3030, opened 2026-05-25
webmin: Add xapian spotlight backend

NJRoadfan:


rdmark (approved):
good catch, thanks

this reminds me that a future improvement area is to introduce something akin to the `%netatalkParameterDefaults` hash in netatalk-lib.pl but for the available select options, so that we don't have to define them multiple times

PR #3032, opened 2026-05-25
spotlight: drain ready localsearch callbacks on fetch

rdmark:
The AFP client can poll fetchQueryResultsForContext immediately after the Tracker query callback has queued the first cursor_next_async(), but before the cursor callback has populated the current page or marked the query done.

Run ready GLib main-context callbacks while the current page is still empty so localsearch returns the first available result page, or DONE for an empty result set, instead of repeated empty pending pages. Also continue cursor iteration when Tracker returns a row without a URI.


rdmark:
this resolves certain race conditions that I have observed with the localsearch backend, the symptoms being that no results are inexplicably being returned

andylemin (approved):

PR #3042, opened 2026-05-29
meson: generate compilation doc on the fly, don't store in git

rdmark:
the COMPILATION.md doc is only meant to be published on the website, so rather than storing a snapshot in git that has to be refreshed regularly, let it be built on the fly only when needed

this saves us some manual overhead for each release


rdmark:
@andylemin this change further streamlines the release process, updating the netatalk.io website is almost entire automated now

andylemin (approved):

PR #3048, opened 2026-05-30
move containers readme to repository root

rdmark:
this makes it so that documentation that is to be distributed and published to the website is always either under the doc/ subdir or in the repository root


rdmark:
@andylemin this came up as I was preparing the website update, and realized that the containers readme had fallen off distribution (a while ago actually)

I think this is a good change structurally, as it prevents similar oversights in the future: out of sight out of mind!

andylemin (approved):

andylemin:
@rdmark Nice! Well spotted..

I'm keeping a close on on the PRs tonight and will approve as soon as I see them so we can get the release out today 🤞

PR #3052, opened 2026-05-30
 nad: introduce support for MacBinary and BinHex conversion

rdmark:
this functionality is built on the historical 'megatron' tool which was shipped with netatalk since v1.2 but removed in v3.0; it uses the same foundational structure, but has been overhauled for robustness and usability

- moved away from symlink invocation to positional commands in `nad`
- make it read afp.conf to be aware of AFP volume paths and filesystem metadata scheme
- leverage libatalk VFS subsystem to support conversion to and from both AppleDouble v2 and EA
- added support for updaing CNID database when files are created in an AFP volume
- added support for BinHex encoding (previously only decoding was supported)
- removed the direct BinHex to MacBinary conversion path (`hqx2bin`): use a two step process with Netatalk metadata as the intermediate step
- make stdout quiet by default, and introduce `--verbose` flag to print diagnostic information
- replaced the original CRC-16 Xmodem encoding lookup table with a minimal implementation from scratch
- removed the ancient AppleSingle format, which libatalk VFS no longer handles
- removed the Shift JIS and EUC-JP character encoding conversion routines
- credit to original creator of megatron [Charles Clark](https://websites.umich.edu/~rsug/netatalk/archive/admins/1992/0032.html) (reportedly introduced with netatalk v1.2)
- added standard UMich copyright headers aligned with the other netatalk source files of that era

modifications to nad:

- allow for file operations in any arbitrary file system location
- skip AFP volume initialization and CNID db operations on non-AFP paths
- naive nad find implementation for non-AFP paths
- consolidate logging routines, new interface is NAD_INFO, NAD_FATAL, NAD_DEBUG

modifications elsewhere:

- set_utc_offset utility function moved from afpd to libatalk/adouble
- consolidate Mac epoch constants from megatron.h to adouble.h

mapping new and old commands:

| action | megatron command | nad command|
| --- | --- | --- |
| convert MacBinary (.bin) to Netatalk EA/AD | `unbin` | `nad unbin` |
| convert BinHex (.hqx) to Netatalk EA/AD | `unhex` | `nad unhex` |
| convert AppleSingle to AppleDouble | `unsingle` | _removed_ |
| convert BinHex (.hqx) to MacBinary (.bin) | `hqx2bin` | `nad unhex; nad bin` |
| convert AppleSingle to MacBinary (.bin) | `single2bin` | _removed_ |
| convert Netatalk EA/AD to MacBinary (.bin) | `macbinary` | `nad bin` |
| convert Netatalk EA/AD to BinHex (.hqx) | - | `nad hex` |
| show MacBinary header | `binheader` | `nad unbin --header` |
| show BinHex header | - | `nad unhex --header` |
| show Netatalk EA/AD header | `nadheader` | `nad bin --header` _OR_ `nad hex --header` |

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -8,6 +8,7 @@ with their contributions. Cheers.
 Thanks a bunch folks, wherever you are.
 
 - Wesley Craig <wesley.craig@umich.edu>
+- Charles Clark <cmclark@umich.edu>: megatron Macintosh file format transformer
 - Adrian Sun <asun@cobaltnet.com>
 - Steve Hirsch <shirsch@adelphia.net>: ProDOS II support/initial Randnum
 support
--- a/bin/nad/crc16.c
+++ b/bin/nad/crc16.c
@@ -1,0 +1,35 @@
+/*
+ * Copyright (c) 2026 Daniel Markstedt <daniel@mindani.net>
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ */
+
+#ifdef HAVE_CONFIG_H
+#include "config.h"
+#endif /* HAVE_CONFIG_H */
+
+#include "crc16.h"
+
+#define CRC16_XMODEM_POLY 0x1021u
+
+uint16_t crc16_xmodem_update(uint16_t crc, const void *buf, size_t len)
+{
+    const uint8_t *ptr = buf;
+
+    while (len-- > 0) {
+        crc ^= (uint16_t) * ptr++ << 8;
+
+        for (unsigned int bit = 0; bit < 8; bit++) {
+            if ((crc & 0x8000) != 0) {
+                crc = (uint16_t)((crc << 1) ^ CRC16_XMODEM_POLY);
+            } else {
+                crc = (uint16_t)(crc << 1);
+            }
+        }
+    }
+
+    return crc;
+}
--- a/bin/nad/crc16.h
+++ b/bin/nad/crc16.h
@@ -1,0 +1,9 @@
+#ifndef CRC16_H
+#define CRC16_H 1
+
+#include <stddef.h>
+#include <stdint.h>
+
+uint16_t crc16_xmodem_update(uint16_t crc, const void *buf, size_t len);
+
+#endif /* CRC16_H */
--- a/bin/nad/hqx.c
+++ b/bin/nad/hqx.c
@@ -1,0 +1,996 @@
+/*
+ * Copyright (c) 1990,1991 Regents of The University of Michigan.
+ * Copyright (c) 2026 Daniel Markstedt <daniel@mindani.net>
+ * All Rights Reserved.  See COPYRIGHT.
+ *
+ * Macintosh file format transformer based on megatron by Charles Clark.
+ */
+
+#ifdef HAVE_CONFIG_H
+#include "config.h"
+#endif /* HAVE_CONFIG_H */
+
+#include <sys/types.h>
+#include <sys/uio.h>
+#include <sys/time.h>
+#include <sys/param.h>
+
+#include <string.h>
+#include <ctype.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <time.h>
+
+#include <unistd.h>
+#ifdef HAVE_FCNTL_H
+#include <fcntl.h>
+#endif /* HAVE_FCNTL_H */
+
+#include <netinet/in.h>
+
+#include <atalk/adouble.h>
+#include <netatalk/endian.h>
+
+#include "nad.h"
+#include "megatron.h"
+#include "nad_adouble.h"
+#include "hqx.h"
+#include "crc16.h"
+
+/*	String used to indicate standard input instead of a disk
+	file.  Should be a string not normally used for a file
+ */
+#ifndef	STDIN
+#	define	STDIN	"-"
+#endif /* ! STDIN */
+
+/*	Yes and no
+ */
+#define NOWAY		0
+#define SURETHANG	1
+
+/*	Looking for the first or any other line of a binhex file
+ */
+#define	FIRST		0
+#define OTHER		1
+
+/*	This is the binhex run length encoding character
+ */
+#define RUNCHAR		0x90
+
+/*	These are field sizes in bytes of various pieces of the
+	binhex header
+ */
+#define	BHH_VERSION		1
+#define	BHH_TCSIZ		8
+#define	BHH_FLAGSIZ		2
+#define	BHH_DATASIZ		4
+#define	BHH_RESSIZ		4
+#define BHH_CRCSIZ		2
+#define BHH_HEADSIZ		21
+
+static struct hqx_file_data {
+    uint32_t		forklen[ NUMFORKS ];
+    unsigned short		forkcrc[ NUMFORKS ];
+    int                 forkdone[ NUMFORKS ];
+    char		path[ MAXPATHLEN + 1];
+    unsigned short		headercrc;
+    int			filed;
+    int                 owns_fd;
+    unsigned char       outbuf[3];
+    int                 outcnt;
+    int                 linecol;
+    int                 writing;
+} 		hqx;
+
+extern char	*forkname[];
+static unsigned char	hqx7_buf[8192];
+static unsigned char	*hqx7_first;
+static unsigned char	*hqx7_last;
+static int	first_flag;
+static const unsigned char hqxchars[] =
+    "!\"#$%&'()*+,-012345689@ABCDEFGHIJKLMNPQRSTUVXYZ[`abcdefhijklmpqr";
+
+static int hqx_write_all(const void *buf, size_t len)
+{
+    const char *p = buf;
+
+    while (len > 0) {
+        ssize_t cc = write(hqx.filed, p, len);
+
+        if (cc <= 0) {
+            return -1;
+        }
+
+        p += cc;
+        len -= cc;
+    }
+
+    return 0;
+}
+
+static int hqx_put_char(unsigned char c)
+{
+    if (hqx.linecol >= 63) {
+        if (hqx_write_all("\n", 1) < 0) {
+            return -1;
+        }
+
+        hqx.linecol = 0;
+    }
+
+    if (hqx_write_all(&c, 1) < 0) {
+        return -1;
+    }
+
+    hqx.linecol++;
+    return 0;
+}
+
+static int hqx_emit_6bit(unsigned char b)
+{
+    hqx.outbuf[hqx.outcnt++] = b;
+
+    if (hqx.outcnt == 3) {
+        unsigned char out[4];
+        out[0] = hqxchars[(hqx.outbuf[0] >> 2) & 0x3f];
+        out[1] = hqxchars[((hqx.outbuf[0] << 4) | (hqx.outbuf[1] >> 4)) & 0x3f];
+        out[2] = hqxchars[((hqx.outbuf[1] << 2) | (hqx.outbuf[2] >> 6)) & 0x3f];
+        out[3] = hqxchars[hqx.outbuf[2] & 0x3f];
+
+        for (int i = 0; i < 4; i++) {
+            if (hqx_put_char(out[i]) < 0) {
+                return -1;
+            }
+        }
+
+        hqx.outcnt = 0;
+    }
+
+    return 0;
+}
+
+static int hqx_emit_byte(unsigned char b)
+{
+    if (hqx_emit_6bit(b) < 0) {
+        return -1;
+    }
+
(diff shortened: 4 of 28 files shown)


andylemin:
@rdmark if we are bringing back a large toolkit, can we think about more professional naming?

nad does well at this.

nadc, nconv, netconv, nadconv? Something that follows existing naming patterns is preferable?

andylemin (changes requested):
Modernise and standardise naming?

rdmark:
@andylemin it is indeed a silly name for a CLI app, you can tell that a computer geek named it in 1991 (sorry if you read this, Charles, I mean it in an endearing way!)

the one big drawback with renaming is that megatron is a well established name in the community, there are historical man pages published all over the web, so we will lose some free recognition and continuity with the old world

I want to experiment with similar-sounding names! perhaps `mtron` or `mtran` - as in Mac (file format) transformer

andylemin:
> @andylemin it is indeed a silly name for a CLI app, you can tell that a computer geek named it in 1991 (sorry if you read this, Charles, I mean it in an endearing way!)
> 
> the one big drawback with renaming is that megatron is a well established name in the community, there are historical man pages published all over the web, so we will lose some free recognition and continuity with the old world
> 
> I want to experiment with similar-sounding names! perhaps `mtron` or `mtran` - as in Mac (file format) transformer

Hi @rdmark
I would contend that actually - as it was dropped quite some time ago the community already knows it's dead, and the new logic is materially different with a different user interface/arguments etc. So using the same name even causes problems when old docs are found instead of latest docs matching the new functionality.

I think we should use the opportunity to rename it, and call out in the manual (and -h help) that it is the new "equivalent" of what used to be known as megatron? Then new searches match the new code, naming is aligned with the product suite (as this is a new cli command), and the name gives you a clue what it is for?

rdmark:
@andylemin I like this approach — rename and leave breadcrumbs in docs and code that it was based on megatron but significantly revamped and improved

My personal goal is to have a command that is comfortably typed and memorable. :)

NJRoadfan:
Was the tool's original intent to offer easy of extracting the resources of a MacBinary file into Netatalk compatible format? If so, than only permitting it to run on Netatalk volumes is fine. If the original tool allowed creation of MacBinary and BinHex from any file on the file system, it could be useful to lift that restriction if another tool doesn't exist to do the job.

The `cppo` Python script with A2SERVER does similar work, but isn't dependent on Netatalk at all. It creates the AppleDouble files (both EA and v2) with the Netatalk specific information nulled out (CNID, inode, etc), which Netatalk will later add when the files are placed in a volume and accessed. It can also expand the resource forks and FinderInfo to separate files.

rdmark:
@NJRoadfan the original megatron could indeed operate on any file on the file system, in fact it wasn't aware of AFP volumes at all. I have removed the restriction now, made it so that when outside an AFP volume unbin/unhex defaults to EA, and a new flag `--adouble` that forces an AppleDouble v2 sidecar to be created instead.

rdmark:
@andylemin I went with my own renaming idea, `mtran`, but open to counter arguments :)

andylemin:
> @andylemin I went with my own renaming idea, `mtran`, but open to counter arguments :)

Can we try a name which keeps all the tools in a consistent aligned/alphabetical namespace? We have nad, netatalk, netatalk_client. So it would be ideal if it start with `na` or `ne`? How about nadtran or ne(t)tran or some variant?

rdmark:
@andylemin I get your point, however I have the opposite preference: a unique name that is memorable and don't share a prefix with other commands in the same suite

for instance, I personally dislike how most of the tools in netatalk_client / afpfs-ng all share an 'afp' prefix, makes it uncomfortable to type when operating rapidly on the command line, and for me hard to consistently remember the exact spelling of the various commands

the 'mtran' name is unique enough that typing 'mt' plus Tab gives me the command on my Linux systems (macOS/BSD have 'mtree' unfortunately)

anyways, this is just how my brain and hands work, everyone are different! again, I absolutely see the pros and cons here

I did actually toy with the idea of baking the megatron functionality into nad proper, add bin/unbin/hex/unhex as commands in that tool; but I decided against it with the Unix design philosophy in mind (small tools that do one thing well)

andylemin:
"I did actually toy with the idea of baking the megatron functionality into nad proper" - This would be the best compromise 🙂

rdmark:
> "I did actually toy with the idea of baking the megatron functionality into nad proper" - This would be the best compromise 🙂

Hehe, I actually created a quick prototype for this in the morning before heading to work. The idea is growing on me.

andylemin:
> > "I did actually toy with the idea of baking the megatron functionality into nad proper" - This would be the best compromise 🙂
> 
> Hehe, I actually created a quick prototype for this in the morning before heading to work. The idea is growing on me.

Hahah. This makes me happy.. My OCD was twitching at the thought of a command randomly located elsewhere 😂

rdmark:
augment review

rdmark:
@andylemin it's been done! I have also revamped nad itself to allow it to operate on any file system path, while introducing a naive implementation of `nad find` for non-AFP volumes

andylemin:
Cool! Let me do a full review on this tonight 🙏

NJRoadfan:
Since this affects a workflow. Will this still return the error `"." is not inside an AFP volume` if `nad ls` is run outside of a Netatalk volume? I am relying on that in the `afptype` script in A2SERVER to determine if the tool is being run on a Netatalk volume or not. Of course I just checked and that check is actually broken anyway. Whoops.

rdmark:
@NJRoadfan the behavior has indeed changed here: commands outside of an AFP volume will quietly succeed. I have two ideas:

- by default, file operations that are outside of an AFP volume throw a warning and has a >0 exit code, *but* the action is executed
- or, keep the new behavior as default, and add a `--strict` flag to nad that has the above effect

thoughts?

NJRoadfan:
I'll have to take a look and see the new behavior.

Somewhat related, I noticed time transformation functions creeped their way into the code. Originally these archives and classic MacOS did not store any timezone information with files so the exact time stamps are unknown. Its best to just assume they are UTC (`cppo` does this), or take advantage of the `utc_offset()` function in Netatalk. Its portable (except for AIX) and takes AppleDouble date stamps already.

andylemin:
Hi @rdmark A couple of timestamp related issues similar to what @NJRoadfan was calling out (his UTC suggestions are spot on I think); 

macbin_ad_date_from_header computes MAC_DATE_TO_UNIX(mac_date) - bin.gmtoff then subtracts AD_DATE_DELTA. The header defines both AD_DATE_FROM_MAC/AD_DATE_TO_MAC (megatron.h:2773-2774) and this hand-rolled path. The hand-rolled version applies a timezone offset (bin.gmtoff) that the macro versions don't. Having two parallel date-conversion mechanisms in the same change is a maintenance hazard — confirm the unused AD_DATE_FROM_MAC/AD_DATE_TO_MAC macros are actually needed, or remove them?

And macbin_ad_date_to_header open-codes 2082844800U (Mac epoch offset) while the header provides MAC_DATE_FROM_UNIX which uses the same constant.

---

And then some minor ones;

Moving closevol(&volume) out of the if (nftw(...) == -1) block fixes a real volume-handle leak on the success path. Good catch — but nad_rm still uses openvol(), not openvol_optional(), unlike its siblings. Was that intentional?

The HEXOUTPUT blocks (writing rawhex.unhex, headerfork, etc. to cwd) are debug-only and gated off. Since this is a fresh rewrite rather than a patch to existing code, consider dropping this scaffolding entirely rather than carrying it forward — it writes files to the current directory and is easy to enable by accident.

OPTION_HEADERONLY (1<<0), OPTION_STDOUT (1<<2), OPTION_VERBOSE (1<<5), OPTION_ADOUBLE (1<<6) — bits 1, 3, 4 are skipped. Likely hangover from megatron's larger option set. Harmless, would be nice to clean up in this refactor?

"This should never happen, dude!" and "Bad ... crc, dude" are preserved from the 1991 original. Endearing, but inconsistent with netatalk's professional diagnostic tone. Your call given the heritage-preservation intent, but worth a deliberate decision.

---

Testing - the most significant gap;

There are no automated tests for the new conversion logic, and SonarQube confirms 0% coverage on new code. The codebase has a test harness and extensive CI spec tests, but nothing exercises bin/hex/unbin/unhex.

Considering the round-trip nature of these operations, this is highly testable and can be added as build time unit tests?

- Round-trip: nad bin → nad unbin and nad hex → nad unhex should reproduce the original data fork, resource fork, and Finder info.
- CRC verification: known-answer test for crc16_xmodem_update (the XMODEM CRC of "123456789" is 0x31C3 — a one-line unit test).
- Cross-format: the documented hqx2bin replacement path (nad unhex; nad bin).
- Header parsing: feed known MacBinary I/II/III and BinHex fixtures, check --header output.

I'd recommend at least a round-trip + CRC known-answer tests, since the hand-rolled CRC and decode state machine are exactly the kind of code that breaks silently.

A lot of great work here! And thanks again for fixing up the naming and nad integration 🙂

andylemin (approved):

rdmark:
@andylemin good points, thanks!

on the testing matter: agreed completely on the suggested test cases, in fact I would argue that all of the `nad` commands deserve some end to end testing coverage

since this would mean introducing a brand new test harness and coverage from scratch, what do you say we do it as a fast-follow and not in this PR?

rdmark:
> I'll have to take a look and see the new behavior.

I have what I think is a good solution lined up in my local branch: restore the AFP volume validation by default, and introduce a `--force` global flag to nad which overrides the validation and allows nad operations anywhere. Thoughts?

> Somewhat related, I noticed time transformation functions creeped their way into the code. Originally these archives and classic MacOS did not store any timezone information with files so the exact time stamps are unknown. Its best to just assume they are UTC (`cppo` does this), or take advantage of the `utc_offset()` function in Netatalk. Its portable (except for AIX) and takes AppleDouble date stamps already.

Great catch. I assume you mean `set_utc_offset()` from etc/afpd/filedir.c ? Seems reasonable to break this out as a utility function and keep it somewhere in libatalk for use by both afpd and nad.

rdmark:
> Hi @rdmark A couple of timestamp related issues similar to what @NJRoadfan was calling out (his UTC suggestions are spot on I think);
> 
> macbin_ad_date_from_header computes MAC_DATE_TO_UNIX(mac_date) - bin.gmtoff then subtracts AD_DATE_DELTA. The header defines both AD_DATE_FROM_MAC/AD_DATE_TO_MAC (megatron.h:2773-2774) and this hand-rolled path. The hand-rolled version applies a timezone offset (bin.gmtoff) that the macro versions don't. Having two parallel date-conversion mechanisms in the same change is a maintenance hazard — confirm the unused AD_DATE_FROM_MAC/AD_DATE_TO_MAC macros are actually needed, or remove them?
> 
> And macbin_ad_date_to_header open-codes 2082844800U (Mac epoch offset) while the header provides MAC_DATE_FROM_UNIX which uses the same constant.
> 

All of this logic is now consolidated and delegated to NJRoadfan's `set_utc_offset` utility function which has been relocated to libatalk/adouble/ad_date.c

The Mac epoch constants have been refactored and moved to include/atalk/adouble.h

> And then some minor ones;
> 
> Moving closevol(&volume) out of the if (nftw(...) == -1) block fixes a real volume-handle leak on the success path. Good catch — but nad_rm still uses openvol(), not openvol_optional(), unlike its siblings. Was that intentional?
> 

Fixed as a side effect of reintroducing AFP volume validation and the `--force` override!

> The HEXOUTPUT blocks (writing rawhex.unhex, headerfork, etc. to cwd) are debug-only and gated off. Since this is a fresh rewrite rather than a patch to existing code, consider dropping this scaffolding entirely rather than carrying it forward — it writes files to the current directory and is easy to enable by accident.
> 

Good call, no need to keep risky debug code sitting around.

> OPTION_HEADERONLY (1<<0), OPTION_STDOUT (1<<2), OPTION_VERBOSE (1<<5), OPTION_ADOUBLE (1<<6) — bits 1, 3, 4 are skipped. Likely hangover from megatron's larger option set. Harmless, would be nice to clean up in this refactor?
> 

Indeed, it was a carryover from a previous set of options. Agreed that there is no interface that we need to maintain compatibility with.

> "This should never happen, dude!" and "Bad ... crc, dude" are preserved from the 1991 original. Endearing, but inconsistent with netatalk's professional diagnostic tone. Your call given the heritage-preservation intent, but worth a deliberate decision.
> 

For user facing strings: Yes, it makes more sense to align with the voice of the rest of the application. I rephrased one instance of "it's cool" as well. Keeping the megatron.c|h filenames around is enough of a nod to the heritage of this module I think. :)

rdmark:
augment review

rdmark:
put some final touches: consolidated logging macros so that nad and megatron don't have separate routines, improved / added Doxygen code comments to macbin and binhex functions, introduced proper calculation of creation/modification/backup dates when printing the headers, and touched up manpage and help text for clarity

I think this is ready to ship!

andylemin:
@rdmark this is looking good now! Well done! Lets ship it 🛳️

PR #3057, opened 2026-06-01
afpd: explicit header include in file.c

rdmark:
the Debian build wrapper has stringent implicit declaration rules, which leads to a hard compiler error without this header


rdmark:
as a side note, we have had this situation at least half a dozen times now: our build systems don't catch this, Debian or Fedora reports it later which leads to cascading downstream patching

I want to try again to configure our CI build jobs to emulate their level of stringency

andylemin (approved):

PR #3058, opened 2026-06-01
build: make localsearch dconf path configurable

erichelgeson:
For nixpkgs/nixos where we cant write to `/etc`.

### DIFF
--- a/config/dconf/meson.build
+++ b/config/dconf/meson.build
@@ -1,5 +1,5 @@
 install_data(
     'netatalk',
-    install_dir: '/etc/dconf/profile',
+    install_dir: dconf_path / 'profile',
     install_tag: 'config',
 )
--- a/meson.build
+++ b/meson.build
@@ -1384,8 +1384,19 @@ if get_option('with-spotlight')
                 )
             endif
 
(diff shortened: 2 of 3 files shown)


andylemin:
Hi @erichelgeson Thanks for the PR 🙂
augmentcode got this one right! `dconf update` is missing the custom db dir?

At meson.build:1417 the DCONF_UPDATE_COMMAND is still just `<dconf> update`, and etc/netatalk/netatalk.c:197 runs it via system() after writing the keyfile to INDEXER_DCONF_DB_DIR/10-spotlight.
`dconf update` with no argument compiles the system default db directory (/etc/dconf/db), not the relocated <dconf_path>/db.

dconf update does accept an optional DBDIR argument. The command should pass the configured db dir when the path is non-default?

Maybe something like;
```
cdata.set(
    'DCONF_UPDATE_COMMAND',
    '"' + dconf.full_path() + ' update ' + dconf_path + '/db"',
)
```

rdmark:
as discussed with Eric in private, rather than attempting to get localsearch working end to end on NixOS, we agreed to drop it in favor of the CNID / Xapian backends

PR #3059, opened 2026-06-01
meson: add error flag for implicit function declaration

rdmark:
this emulates a recurring strictness error in Debian and Fedora build wrappers

this also leads to the need to set platform specific feature macros via compiler arguments: the libc prototype is gated by feature macros, and those need to be present from the compiler command line, before any system header in any file

Remove the defining of the same macros in config.h which is arguably a cargo cult carryover from the Autotools build system.

finally, explicitly include a range of header to appease the new stricter compiler flag

### DIFF
--- a/bin/dbd/cmd_dbd_scanvol.c
+++ b/bin/dbd/cmd_dbd_scanvol.c
@@ -16,6 +16,7 @@
 #include "config.h"
 #endif /* HAVE_CONFIG_H */
 
+#include <arpa/inet.h>
 #include <dirent.h>
 #include <errno.h>
 #include <fcntl.h>
--- a/etc/spotlight/localsearch/sl_localsearch.c
+++ b/etc/spotlight/localsearch/sl_localsearch.c
@@ -16,6 +16,7 @@
 #include "config.h"
 #endif
 
+#include <arpa/inet.h>
 #include <stdbool.h>
 #include <stdint.h>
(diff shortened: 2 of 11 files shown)


rdmark:
@andylemin this turned into a bit of a rabbit hole, but we should now be in a more portable state, and be protected against more stringent build systems

andylemin:
Lol, yes I can see it now - The fallout required moving the Solaris/illumos feature-test macros to the compiler command line and adding missing #include directives across 10 files so the prototypes (ntohs/htons/ntohl, strlcpy/strlcat/strnlen, explicit_bzero) are visible before first use! Nicely done.

Only comments I can see;
The Solaris feature macros are now defined in two places. They remain in cdata (config.h, meson.build:2343) and are also added globally on the command line - harmless and identical, was this intended?

andylemin (approved):

rdmark:
Ah yes, good call on the sunos feature macros. Let me see if I can consolidate them in a cleaner way.

rdmark:
@andylemin after reviewing the build script, my conclusion in that the older cdata.set lines adds those compatibility macros to config.h which I believe is a cargo cult carryover from the Autotools days. The more consistent and idiomatic way to do it in Meson is to add compiler arguments.

andylemin:
@rdmark sounds fair. Consistency is the key for me 🙂

PR #3068, opened 2026-06-05
docs: replace the C font style with CR in cmark manpages

rdmark:
code blocks in markdown is transcoded to the 'C' font style (Courier monospace) in troff when processed with cmark / cmark-gfm

the groff parser treats this style as invalid for a man page and will throw warning messages, becuase it's a non-standard style

this modification to the filter script replaces the 'C' font styles with 'CR' style in formatted man pages, which aligns with pandoc output

### DIFF
--- a/doc/manpages/make_man.sh
+++ b/doc/manpages/make_man.sh
@@ -1,5 +1,7 @@
 #!/bin/sh
 
+set -o pipefail
+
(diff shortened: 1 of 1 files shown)


rdmark:
FTR, I have filed https://github.com/commonmark/cmark/issues/620 with the cmark project

andylemin (approved):

PR #3070, opened 2026-06-06
ci: use OpenBSD native clang, remove nonexistent dependencies

rdmark:
rather than installing gcc, let's build with OpenBSD native clang

remove a few dependencies that are invalid, silently skipped by pkg_add


rdmark:
I have a vague memory that there was a good reason for not building with OpenBSD clang in the ancient past, but it clearly works now!

andylemin:
@rdmark This is a great result! OpenBSD is an important distro

andylemin (approved):

PR #3072, opened 2026-06-06
afpd: remove obsolete dbd backend guard for CatSearch

rdmark:
until recently, cnid_find was only supported in the dbd backend so this guard was correct, but now all current CNID backends has this API so the guard must be removed

reported by NJRoadfan

additional hardening that was revealed after this fix:

catsearch_db converts the search term before passing it to cnid_find, but continued to use strlen(uname) as the length argument

when conversion changes the byte length, dbd could search a truncated term or read past the converted buffer; use the converted length returned by convert_charset instead

### DIFF
--- a/etc/afpd/catsearch.c
+++ b/etc/afpd/catsearch.c
@@ -980,6 +980,7 @@ static int catsearch_db(const AFPObj *obj,
     static int num_matches;
     int ccr, r;
     int result = AFP_OK;
+    size_t namelen;
     struct path path;
     char *rrbuf = rbuf;
     char buffer[MAXPATHLEN + 2];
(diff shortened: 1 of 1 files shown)


andylemin:
@rdmark This is looking good to me. Is this ready? I see two commits which makes me thing it is still WIP?

rdmark:
@andylemin yes this is ready for review!

I kept the separate commits because I saw them as logically separate

happy to squash them into one commit if you think that's cleaner!

andylemin:
> happy to squash them into one commit if you think that's cleaner!

No need, just checking 👍

andylemin (approved):

PR #3075, opened 2026-06-07
spectest: lock/delete/exchange tests for the server-side lock-fix plan

andylemin:
This PR carries tests ahead of their server-side fixes and exists to build a comprehensive lock testing framework to exercise the full spectest suite across all platforms — including strict-POSIX/BSD FreeBSD, DragonflyBSD, NetBSD, OpenBSD, OmniOS, OpenIndiana and Solaris — _**unblocking the upcoming series of planned locking fixes**_.

---

## Spectest tests

**Re-enabled byte-range lock tests (5 entries / 7 test numbers)**

(now gated on the new Locking flag; active only with `afp read locks = yes` + `afp_spectest -L`/`AFP_READ_LOCKS=yes`
test63/64 — fork-locked range blocks a 2nd fork's read/write (DF, RF) → AFPERR_LOCK
test65 — cross-user read/write over another user's locked range → AFPERR_LOCK
test78 — to-EOF lock survives a sibling fork close; cross-user FPWrite_ext → AFPERR_LOCK (temporarily disabled)
test79 — 2nd fork's FPRead over a fork-locked range → AFPERR_LOCK
test66/67 — FPByteRangeLockExt read/write-vs-lock conflict (DF, RF) → AFPERR_LOCK

**NEW lock tests (11 entries)**

Black-box lock tests (FPByteRangeLock.c):
test602 — shared deny-write survives a partial close
test606 — disjoint locks coexist across sessions; overlapping lock refused → AFPERR_LOCK
test630 — a just-freed fork refnum isn't immediately reused
test631 — a fork op on a closed refnum → AFPERR_PARAM

Other:
test511 (rotest.c) — write-intent open on a read-only volume → AFPERR_VLOCK, never AFPERR_NFILE
test625 (FPExchangeFiles.c) — exchange succeeds with metadata present, FinderInfo stays with the name

Testset T2LockAttack (kernel-observed via raw fcntl on the -c path):
test600 — cross-session FPDelete of an open+locked file → AFPERR_BUSY, holder's lock intact
test603 — shared read lock (refcount 2) survives the first close, released on the last
test604 — last fork close releases every byte lock the fork held
test613 — external (non-afpd) fcntl holder → cross-session FPDelete → AFPERR_BUSY
test620 — delete contention in a dir with an open, locked file → AFPERR_DIRNEMPT / AFPERR_BUSY, both succeed after release

## Spectests on non-Linux Platforms

Enabled the performance-tuned ("PerfConfig") AFP afp_spectest tests on FreeBSD, DragonflyBSD, NetBSD, OpenBSD, OmniOS, OpenIndiana and Solaris.

These are all VM platforms that previously only built and smoke-started. The spectest is a new step inside each existing build job (not a separate workflow).

These strict-POSIX/BSD backends behave differently from Linux and have surfaced real bugs (#3081, #3087) that the Linux and non-strict POSIX systems can't see. Config's match the Alpine PerfConfig spectest (ARC dircache, validation freq 100, large rfork budget, cnid=dbd), however only those with ZFS use ea=sys, all others use ea=ad due to missing EA support.

They also provide a cross-section of filesystems for testing, expanding our VFS test coverage across EXT4, UFS, FFS, HAMMER2, and ZFS.

Made env_setup_netatalk.sh portable across all platform (additive only — every Linux/Alpine path is byte-for-byte unchanged) - allowing the container's setup script is now reused by the VM jobs instead of duplicating afp.conf generation in YAML.

## Asserts now run in CI everywhere

NDEBUG was never defined, and flamegraph workflows "(AFP_ASSERT active)" label was misleading. This is now made correct and deliberate:

meson: `b_ndebug=if-release` now gives standard C semantics — release builds define `NDEBUG` (*asserts off*), everything else keeps them on. Also fixed a critical bug where `-DEBUG` was resulting in the junk `EBUG` macro (instead of DEBUG) → `-DDEBUG` now sets `DEBUG` properly.

All CI build/test/spectest jobs and testsuite images now use `debugoptimized`, meaning they are performance optimized and assert-enabled, so internal invariant violations crash loudly in CI instead of corrupting silently — while staying fast enough for performance/regression checks. Exposed a macOS crash (fixed in #3087).

## Performance Dashboard

The Flamegraph workflow has been expanded into 3 performance runners (afp_spectest, afp_speedtest, & afp_lantest) giving us Flame, Performance and Latency Graphs. All combined into a new unified report (PR Comment) by a new utility workflow which pulls the results and publishes them together.

The file cleanup has been restructured so we only keep the performance analysis images for the most recent commit (push) under a single PR, enabling longer term storage of image results (6 months). Previously we were hitting the 1000 files limit in under 60-90 days.

---

Notes:

test625 reproduces the FPExchangeFiles access-mode bug the Linux legs cannot observe (#3087).

test78 is a true reproducer of an unfixed server bug — adf_unlock() releases every fork's byte-range locks on a single fork close, instead of only the closing fork's (per AFP spec). It is expected to fail under the read-locks CI leg until the per-fork-release fix lands, and is kept compiled in (not skipped) so the failure stays a visible signal.

---

Includes - Five DragonflyBSD-specific bugs in afpd, each caused by assuming Linux/FreeBSD behaviour:

- ad_attr.c (ad_setid/ad_getid): dev_t/ino_t were stored using sizeof(dev_t) as the length, but ADEID_PRIVDEV/PRIVINO are fixed 8-byte on-disk fields. DragonFly's dev_t is 4 bytes, so the entry was rejected and the CNID identity (device/inode) was never persisted, breaking FPResolveID and local-fs change detection. Marshal through a uint64_t into the fixed-width field; byte-identical on 8-byte-dev_t platforms.

- meson.build: OPEN_NOFOLLOW_ERRNO defaulted to ELOOP for DragonFly. Like its FreeBSD ancestor, DragonFly returns EMLINK from open(O_NOFOLLOW) on a symlink; add a 'dragonfly' branch setting EMLINK.

- ad_open.c (ad_open_df): symlink emulation matched a platform-specific errno to recognise an O_NOFOLLOW symlink open. DragonFly performs the permission check before the symlink check, so an unprivileged caller gets EACCES, not EMLINK, and the symlink was treated as a permission failure. Detect the symlink with lstat() instead — unambiguous on every platform. open() errno is preserved across the probe so unrelated failures keep their original error.

- ad_open.c (ad_open_hf_v2): a symlink (data fork AD_SYMLINK) has no .AppleDouble sidecar; short-circuit as ad_open_hf_ea() already does, instead of attempting to open a non-existent sidecar.

- file.c (afp_createfile): DragonFly remaps EACCES->EEXIST for a denied O_CREAT|O_EXCL create, so a create denied by directory permissions returned AFPERR_EXIST instead of AFPERR_ACCESS. On EEXIST, lstat() the target; if it does not exist, the create was permission-denied -> AFPERR_ACCESS.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -472,7 +472,7 @@ jobs:
           path: build/meson-logs
 
   build-dflybsd:
-    name: DragonflyBSD
+    name: AFP Spectest - DragonflyBSD (sys:ad, HAMMER2)
     runs-on: ubuntu-latest
     timeout-minutes: 24
     steps:
@@ -519,6 +519,104 @@ jobs:
             meson install -C build
             netatalk -V
             afpd -V
+      # ================================================================
+      # Production AFP spec test (AFP 3.4)
+      #
+      # Runs inside the same VM (kept alive by the action's custom shell)
+      # using the binaries installed by the build step above. Identical in
+      # intent to the FreeBSD job below — DragonflyBSD shares the BSD `pw` user
+      # tooling, so it reuses the same portable env_setup_netatalk.sh branch.
+      # Only the compiled lock path differs (DragonFly has no meson
+      # special-case -> /var/spool/locks).
+      # ================================================================
+      - name: AFP Spectest (production, AFP 3.4)
+        shell: dragonflybsd {0}
+        run: |
+          set -e
+          export AFP_USER=atalk1
+          export AFP_USER2=atalk2
+          export AFP_GROUP=afpusers
+          export AFP_PASS="$(openssl rand -base64 12 | tr -d '/+=' | cut -c1-8)"
+          export AFP_PASS2="$AFP_PASS"
+          export SHARE_NAME=test1
+          export SHARE_NAME2=test2
+          # DragonflyBSD installs under the /usr/local prefix
+          export NETATALK_CONFDIR=/usr/local/etc
+          export NETATALK_SHARE_DIR=/mnt/afpshare
+          export NETATALK_BACKUP_DIR=/mnt/afpbackup
+          export NETATALK_LOG_FILE=/var/log/afpd.log
+          export NETATALK_LOCKDIR=/var/spool/locks
+          # Production max-performance tuning knobs
+          export AFP_DIRCACHESIZE=1048576
+          export AFP_DIRCACHE_MODE=arc
+          export AFP_DIRCACHE_VALIDATION_FREQ=100
+          export AFP_DIRCACHE_RFORK_BUDGET=1048576000
+          export AFP_DIRCACHE_RFORK_MAXSIZE=1048576
+          export AFP_CNID_BACKEND=dbd
+          export AFP_CONVERT_APPLEDOUBLE=no
+          export AFP_EXTMAP=1
+          # HAMMER2, DragonflyBSD's default root filesystem (where the share
+          # /mnt/afpshare lives), does not support extended attributes, so
+          # afpd is built with "EA support: ad" only. Use AppleDouble EA
+          # storage (ea = ad, + afp_spectest -a) to match; ea = sys would make
+          # every FPGetExtAttr test fail.
+          export AFP_ADOUBLE=1
+          # Same cleartext-over-PAM auth path as the Linux/Alpine prod
+          # spectest (see the FreeBSD job for the full explanation).
+          export INSECURE_AUTH=1
+          export DISABLE_TIMEMACHINE=1
+          export SERVER_NAME=Netatalk
+          export AFP_HOST=127.0.0.1
+          export AFP_PORT=548
+          export AFP_VERSION=7
+          # TESTSUITE=spectest makes env_setup populate TEST_FLAGS with the
+          # spectest options: -c <share> for the local-fs tests, -a (AFP_ADOUBLE
+          # is set above, so the client uses AD_V2 semantics and manipulates the
+          # .AppleDouble sidecar). We pass $TEST_FLAGS to afp_spectest exactly
+          # like entrypoint_netatalk.sh.
+          export TESTSUITE=spectest
+
+          [ -f "$NETATALK_CONFDIR/extmap.conf" ] || cp config/extmap.conf "$NETATALK_CONFDIR/extmap.conf"
+
+          # env_setup is written for the container contract (set -e, no
+          # nounset) and references many optional AFP_* vars bare, expecting
+          # them to expand to empty. Disable nounset around the source so it
+          # behaves the same here as in the container.
+          set +u
+          . ./distrib/docker/env_setup_netatalk.sh
+          set -u
+
+          echo "==== Starting netatalk (production config) ===="
+          netatalk
+          sleep 3
+          asip-status "$AFP_HOST:$AFP_PORT" || true
+
+          echo "==== Running AFP spec test (AFP 3.4) ===="
+          # Mirror entrypoint_netatalk.sh: pass $TEST_FLAGS (built by env_setup
+          # from TESTSUITE/AFP_ADOUBLE) so -a and -c are included.
+          # Full suite (no -f) to get the true current pass/fail after the dev_t,
+          # EMLINK and ad_open_hf_v2 fixes.
+          set +e
+          afp_spectest $TEST_FLAGS \
+            -"$AFP_VERSION" \
+            -h "$AFP_HOST" \
+            -p "$AFP_PORT" \
+            -u "$AFP_USER" \
+            -d "$AFP_USER2" \
+            -w "$AFP_PASS" \
+            -s "$SHARE_NAME" \
+            -S "$SHARE_NAME2"
+          SPECTEST_RC=$?
+          set -e
+
+          echo "==== afpd.log (tail) ===="
+          tail -200 "$NETATALK_LOG_FILE" 2>/dev/null || echo "no log file"
+          echo "==== afp.conf (redacted) ===="
+          sed -E 's/^([[:space:]]*[^=#]*(pass|pw|secret|key)[^=]*=).*/\1 ***REDACTED***/I' \
+            "$NETATALK_CONFDIR/afp.conf" 2>/dev/null || true
+
+          killall netatalk afpd cnid_metad 2>/dev/null || true
+          exit $SPECTEST_RC
       - name: Upload meson logs
         if: always()
         uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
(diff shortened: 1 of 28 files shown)


andylemin:
@rdmark 
re-enabled lots of tests that were incorrectly disabled (thought to be broken, when they are just only valid when read locks are enabled). Also added a bunch of new lock tests.

And some which are failing (expected) due to now known bugs.
This is my regression and evidence framework for the series of lock fixes which I am going to start pushing - big improvements coming 🙂

andylemin:
augment review

andylemin:
Hi @rdmark This is ready for review now.

**The new Performance Dashboard changes have created a race condition however with legacy Flamegraph runners** So this needs to be merged first before any other PRs. Other PRs are currently broken at the moment.

As this was intended to collect all the new locking tests until all the planned fixes are in (some of which will fail until they are fixed), I have disabled them (such as test78) making this PR clean and ready to go.

It was meant to be a scaffolding/test-rig PR but it was getting very large anyway..

.github/workflows/build.yml
```diff
@@ -675,14 +975,14 @@ jobs:
               libtalloc \
               mariadb-client \
               meson \
-              openldap-client-2.6.10v0 \
+              openldap-client \
```
rdmark: this introduces ambiguity since there is more than one port with this root in OpenBSD 7.9, from the CI log:

>  Ambiguous: openldap-client could be openldap-client-2.6.13v0-gssapi openldap-client-2.6.13v0

please pick one or the other!

.github/workflows/containers.yml
```diff
@@ -437,15 +381,14 @@ jobs:
               -e AFP_GROUP=afpusers \
               -e SHARE_NAME=test1 \
               -e SHARE_NAME2=test2 \
-              -e VERBOSE=1 \
```
rdmark, line 440: the reason I had these VERBOSE was to aid troubleshooting when there's a failed test -- but perhaps you don't see that this is required here?

test/testsuite/spectest.c
```diff
@@ -353,7 +355,7 @@ char *uam = "Cleartxt Passwrd";
 void usage(char *av0)
 {
     fprintf(stdout,
-            "usage:\t%s [-1234567aCilmnVv] [-h host] [-H host2] [-p port] [-s vol] [-c vol path] [-S vol2] "
+            "usage:\t%s [-1234567aCEiLlmVv] [-h host] [-H host2] [-p port] [-s vol] [-c vol path] [-S vol2] "
```
rdmark, line 358: I think the new -L flag should be documented in the man page!

rdmark (approved):
only minor remarks!

one big picture comment: it might be worth splitting build.yml into pure build jobs and the AFP spectest jobs -- the two types are fundamentally different in nature

and an even bigger picture comment: given that we have such massive amounts of repeated boilerplate across our Ci workflows, it might be time to invest in writing our own GitHub Actions!

finally, this is a very personal note, but I made an effort in recent years to purge all Python scripts from the codebase... and now they make a comeback ^^;

(but it's fine, since they're not in the critical path of building and running the end user software)

rdmark:
btw love the new graphs, very informative and pleasing to the eyes!

.github/workflows/build.yml
andylemin: Well spotted. Thanks. I had to remove it to get OpenBSD working again as they dropped/update the pinned version (notice the log says `openldap-client-2.6.13v0` now).
This is going to be a recurring overhead, which we need to automate like we do for other pinned packages?

.github/workflows/containers.yml
andylemin: I was having a big cleanup and trying to reduce the size of the logs. On one or two, we had forgotten to add the `tail -200` and the logs were hundreds of thousands of lines every single push!
VERBOSE however only enables the afp_ tools logs, so this is acceptable with tail. I'll restore it.

test/testsuite/spectest.c
andylemin: 100% thanks, I thought I did that! 🙂

andylemin:
> only minor remarks!
> 
> one big picture comment: it might be worth splitting build.yml into pure build jobs and the AFP spectest jobs -- the two types are fundamentally different in nature
> 
> and an even bigger picture comment: given that we have such massive amounts of repeated boilerplate across our Ci workflows, it might be time to invest in writing our own GitHub Actions!
> 
> finally, this is a very personal note, but I made an effort in recent years to purge all Python scripts from the codebase... and now they make a comeback ^^;
> 
> (but it's fine, since they're not in the critical path of building and running the end user software)

@rdmark 

1) Yes I already started looking into how to break up the workflows. Splitting by VM build/spectest and Container Spectest jobs makes sense. Just need to double check their dependencies group cleanly too. I could not split the performance test runners as they consume the built container image. So your suggested grouping looks right to me.
I will do this in a later PR.

2) Writing our own GitHub Actions is a good idea too! I will do this later in the same follow up PR.

3) Purge all Python scripts - Lol, what language do you prefer for things like this? I see we prefer perl and shell?
I was considering doing this now, but we need to unblock PRs, so let me do this in the same follow up PR (I will declare that followup PR blocking for the rest of the locking and test plans)

rdmark:
> Purge all Python scripts - Lol, what language do you prefer for things like this? I see we prefer perl and shell?

yeah, Perl and shell scripts that run on the `/bin/sh` interpreter

it's not the sexiest stack but proven to be resilient and portable

andylemin:
> > only minor remarks!
> > one big picture comment: it might be worth splitting build.yml into pure build jobs and the AFP spectest jobs -- the two types are fundamentally different in nature
> > and an even bigger picture comment: given that we have such massive amounts of repeated boilerplate across our Ci workflows, it might be time to invest in writing our own GitHub Actions!
> > finally, this is a very personal note, but I made an effort in recent years to purge all Python scripts from the codebase... and now they make a comeback ^^;
> > (but it's fine, since they're not in the critical path of building and running the end user software)
> 
> @rdmark
> 
> 1. Yes I already started looking into how to break up the workflows. Splitting by VM build/spectest and Container Spectest jobs makes sense. Just need to double check their dependencies group cleanly too. I could not split the performance test runners as they consume the built container image. So your suggested grouping looks right to me.
>    I will do this in a later PR.
> 2. Writing our own GitHub Actions is a good idea too! I will do this later in the same follow up PR.
> 3. Purge all Python scripts - Lol, what language do you prefer for things like this? I see we prefer perl and shell?
>    I was considering doing this now, but we need to unblock PRs, so let me do this in the same follow up PR (I will declare that followup PR blocking for the rest of the locking and test plans)

Resolved in #3093

PR #3076, opened 2026-06-07
testsuite: use libafpclient transport interface

rdmark:


rdmark:
@andylemin putting this up for transparency -- it's an early stage experiment, not ready for review! but it mostly works and demonstrates an AFP test client using afpfs-ng / netatalk-client libafpclient instead of the simplistic testsuite client.

note that the CI jobs do *not* use the new transport: you need to have libafpclient.so and headers installed, and without them falls back to the legacy client. I'm actively working towards a first netatalk-client release tarball before setting this up in the pipeline.

as you're doing your own overhaul of the testsuite right now, I wanted to give you a heads-up that a number of tests needed rework, because of various reasons, such as accessing the raw DSI responses in ways that isn't valid when using the libafpclient. there are still a large number of newly skipped tests that need further investigation.

a strategic decision that we should make: go all in on libafpclient and purge the legacy test client, or keep the legacy client as a fallback and for low level DSI tests. the former is cleaner and more maintainable, while the latter enables a different class of tests.

rdmark:
@NJRoadfan I've tested that this enables the testsuite to run with DHX2 or SRP auth!

andylemin:
@rdmark Nice, thanks for the example.
I do not see any conflict here with the lock bugs + tests & the general AFP Spec coverage tests I'm working on. So I will carry on as planned.
Once netatalk_client is more polished we can swap the auth layer as required. Many of the Spectests will need to continue using the more direct access patterns to perform more complex test cases, but the simple ones can easily be migrated.

rdmark:
superseded by https://github.com/Netatalk/netatalk/pull/3176

Issue #1825, opened 2024-12-25
testsuite: File ID change after local file system operation on macOS hosts


rdmark:
Six testsuite test cases that deal with local file system file IDs fail when running on macOS Sequoia hosts:

* FPGetFileDirParms:test420: FPGetFileDirParms an open file is renamed with local fs - FAILED
* FPMoveAndRename:test302: file renamed by someone else, cnid not updated - FAILED
* FPResolveID:test331: Resolve ID file modified with local fs - FAILED
* FPResolveID:test360: Resolve ID file modified with local fs and a file is opened - FAILED
* FPResolveID:test412: Resolve ID file modified with local fs - FAILED
* FPResolveID:test413: Resolve ID file modified with local fs - FAILED

One example of simplified logic from test360:

1. Create a file for testing `FPCreateFile(Conn, vol,  0, dir , name)`
1. Get parameters `FPGetFileDirParams(Conn, vol,  dir , name, bitmap,0)`
1. Unpack `afp_filedir_unpack(&filedir, dsi->data +ofs, bitmap, 0)`
1. Store the ID `fid = filedir.did`
1. Rename the file `rename(temp, temp1)`
1. Unpack `afp_filedir_unpack(&filedir, dsi->data +ofs, bitmap, 0)`
1. Assert that the ID remains the same `if fid != filedir.did test_failed()`

On Linux, the ID remains the same. on macOS, the ID changes.

rdmark:
```
afp_spectest -7 -h 127.0.0.1 -p 548 -u atalk1 -d atalk2 -w afpafp -s test1 -S test2 -c /tmp/afptest1 -v -f test420
Connecting to host 127.0.0.1:548
[FPopenLoginExt] LoginExt Version: "AFP3.4" uam: "Cleartxt Passwrd" user: "atalk1"
[FPopenLoginExt] LoginExt Version: "AFP3.4" uam: "Cleartxt Passwrd" user: "atalk2"
[FPOpenVolFull] Open Vol test1 bitmap 21
############## entering test420 ##############
[FPCreateDir] Create Directory Vol 256 did : 0x2 <t420 dir>
directory ID 0x5f3
[FPCreateFile] Create File SOFT Vol 256 did : 0x5f3 <t420 file>
[FPOpenFork] Open Fork resource Vol 256 did : 0x5f3 <t420 file> access 33
fork refnum ID 1
[FPByteLock] Byte lock,fork 1, start, lock, offset: 0, size: 100
[...]

PR #3081, opened 2026-06-09
DragonflyBSD: fix AppleDouble CNID, symlink, and create-error bugs

andylemin:
This PR reverts the changes in #1829

### DIFF
--- a/libatalk/adouble/ad_open.c
+++ b/libatalk/adouble/ad_open.c
@@ -1871,11 +1871,7 @@ static bool ad_entry_check_size(uint32_t eid,
         [ADEID_SHORTNAME] = {ADEDLEN_SHORTNAME, false, false},
         [ADEID_AFPFILEI] = {ADEDLEN_AFPFILEI, true, false},
         [ADEID_DID] = {ADEDLEN_DID, true, false},
-#ifdef __APPLE__
-        [ADEID_PRIVDEV] = {ADEDLEN_PRIVDEV, false, false},
-#else
         [ADEID_PRIVDEV] = {ADEDLEN_PRIVDEV, true, false},
-#endif
         [ADEID_PRIVINO] = {ADEDLEN_PRIVINO, true, false},
         [ADEID_PRIVSYN] = {ADEDLEN_PRIVSYN, true, false},
         [ADEID_PRIVID] = {ADEDLEN_PRIVID, true, false},


andylemin:
Ready for review

NJRoadfan:
The issue with dev_t is known. I discovered the problem under macOS and fixed it by loosening the AppleDouble checks on that platform only. The fix didn't come to the other BSDs because I feared breaking compatibility with Samba's vfs_fruit module which uses the same AppleDouble routines as Netatalk.

See: #1829 #1825 and #1841 

Samba shared adouble code: https://gitlab.com/samba-team/samba/-/blob/master/source3/lib/adouble.c?ref_type=heads

If this fix can be implemented without breaking Samba (or a PR sent to the Samba team to fix it on their end). I see no issue doing this. You can reverse the macOS ifdef in #1829 as well since it isn't needed anymore.

andylemin:
> The issue with dev_t is known. I discovered the problem under macOS and fixed it by loosening the AppleDouble checks on that platform only. The fix didn't come to the other BSDs because I feared breaking compatibility with Samba's vfs_fruit module which uses the same AppleDouble routines as Netatalk.
> 
> See: #1829 #1825 and #1841
> 
> Samba shared adouble code: https://gitlab.com/samba-team/samba/-/blob/master/source3/lib/adouble.c?ref_type=heads
> 
> If this fix can be implemented without breaking Samba (or a PR sent to the Samba team to fix it on their end). I see no issue doing this. You can reverse the macOS ifdef in #1829 as well since it isn't needed anymore.

Nice! Thanks @NJRoadfan Yes it's is a hard break for Dragonfly (AppleDouble data borked for dozens of operations).

I'll dig into Samba case to verify - But I have high confidence it should be fine as the result is byte identical for all existing systems - it only changes the size for systems with 32bit dev_t (which were already broken for Netatalk+Samba anyway).

andylemin:
Investigation
#1825 (closed): six testsuite cases failing on macOS — file IDs changing after a local-fs rename; root cause = ad_getid never matched because dev/ino weren't persisted.
#1829 (merged, commit 4f37f70a): Did not touch ad_attr.c. It only loosened the read-side validator ad_entry_check_size() ad_open.c#L1849 so that on `__APPLE__` the `ADEID_PRIVDEV` entry is fixed_size = false (accepts length 4 instead of requiring 8). It left ad_setid still writing sizeof(dev_t) = 4 on macOS.
#1841 (closed): "Loosen checks for DragonFly and OpenBSD" — the follow-up that was never done. This PR is effectively the proper, all-platform resolution of #1841.

This PR replaces `sizeof(dev_t)`/`sizeof(ino_t)` with the fixed on-disk constants `ADEDLEN_PRIVDEV`/`ADEDLEN_PRIVINO` (both 8, adouble.h#L110), marshalling through a uint64_t on write and narrowing on read. The on-disk EA/v2 layout already reserves a fixed 8-byte slot for these entries ad_open.c:75-88, so writing 8 fits the reserved space on every platform - and normalises AD with EA

@NJRoadfan Good news 🙂

Samba interop verdict — the fix improves interop, doesn't break it
From Samba's [source3/lib/adouble.c](https://github.com/samba-team/samba/blob/master/source3/lib/adouble.c), ad_entry_check_size validates:
`[ADEID_PRIVDEV] = {ADEDLEN_PRIVDEV, true, false}`,   // fixed_size = true → must be exactly 8
`[ADEID_PRIVINO] = {ADEDLEN_PRIVINO, true, false}`,   // fixed_size = true → must be exactly 8

Samba also demands exactly 8 bytes (with a `got_len == 0` empty-entry escape; it never populates dev/ino itself — vfs_fruit leaves them empty/zero).

Therefore:
8-byte platforms: netatalk always wrote 8 → already Samba-compatible; this PR changes nothing.
4-byte-dev_t platforms (macOS/OpenBSD/DragonFly): netatalk pre-PR wrote a length-4 PRIVDEV, *which Samba's fixed-size validator rejects*. Post-PR netatalk writes length-8, which Samba accepts.

One genuine nuance: back-compat self-heal on 4-byte platforms - Existing macOS .AppleDouble files written under #1829 carry a length-4 PRIVDEV.

The new read gate is strict equality: `if (adp && ad_getentrylen(adp, ADEID_PRIVDEV) == ADEDLEN_PRIVDEV)`   // == 8
A length-4 entry now fails this gate → ad_getid returns CNID_INVALID. *However it is graceful, not a corruption - file.c#L332 falls back to cnid_add/DB resolution, and the next ad_setid rewrites the entry at the correct 8-byte width.*

This confirms #1829 #ifdef is now safe to drop — once writes are 8 wide, the loosened read no longer has anything to catch.

andylemin:
`PRIVDEV` requires exactly 8 bytes → https://github.com/samba-team/samba/blob/234df17bc44b39c91db1f90dca048486e31d8b98/source3/lib/adouble.c#L305
`PRIVINO` requires exactly 8 bytes → https://github.com/samba-team/samba/blob/234df17bc44b39c91db1f90dca048486e31d8b98/source3/lib/adouble.c#L306

The validator that enforces it:
`ad_entry_check_size()` function start → https://github.com/samba-team/samba/blob/234df17bc44b39c91db1f90dca048486e31d8b98/source3/lib/adouble.c#L280
`got_len == 0` empty-entry escape → https://github.com/samba-team/samba/blob/234df17bc44b39c91db1f90dca048486e31d8b98/source3/lib/adouble.c#L314
the `expected_len != got_len` rejection for fixed-size entries → https://github.com/samba-team/samba/blob/234df17bc44b39c91db1f90dca048486e31d8b98/source3/lib/adouble.c#L330

(line numbers fixed)

NJRoadfan:
Most of these issues originate from the CVE fixes done to the AppleDouble library back in 2022. The fix implemented the strict verification of field length, but no one seems to consider that the fields may be differing lengths on different platforms!

While the AppleDouble format can reserve extra bytes for a field, it still has a length specifier to tell a parser how large of a value to expect. Setting it to 4 or 8 are both valid.

andylemin:
@NJRoadfan I have reverted #1829 in the 2nd commit.

@rdmark However we cannot merge any PRs now until #3087 is merged as the new Performance Dashboard changes in that PR have broken the legacy Flamegraph runners.

andylemin:
@rdmark and @NJRoadfan The original commit with the fix has already gone into main (https://github.com/Netatalk/netatalk/pull/3075/changes/fb1d7093cde71688e50a76f8d100070d7c7c1e1b) under #3075 as I had locks-tests rebased from this (dragonfly-fixes) - some VM workflows (now with spectests) would not pass without it.

Sorry this became so messy! The fallout from building the full spectest matrix for all platforms, and the dashboard changes tied the fate of these together.

You can ignore "This branch had an error being deployed" - that is residue from the hack I had to put in place to solve the dashboard deploy race condition when we had both legacy and new performance workflows. I can now remove the aggressive publishing logic which brute forced the race (that has now been solved by merging #3075) in the follow up PR.

Ready for Merge 👍

rdmark (approved):
this PR has been reduced to a single cleanup fix, so shouldn't the title and description should be updated to reflect this?

looks good otherwise!

PR #3087, opened 2026-06-09
afpd: fix error handling in afp_exchangefiles and auth_load

andylemin:
After enabling debugoptimised (and AFP_ASSERT) on all CI workflows the macos workflow started asserting.
These fixes resolve this. Fix confirmed using #3075 

afp_exchangefiles, metadata-swap block (only reached after the file renames, so a wrong error here leaves the volume half-swapped):
- two bare `return -1` returned a non-AFP error code and skipped the rename unwind, leaking the open adoubles; set AFPERR_MISC and goto the unwind path (closing the source adouble we opened first)
- the three ad_copy_header() failure branches jumped to the unwind path without setting err, so a metadata-copy failure rolled back the swap but returned AFP_OK (silent success on a failed exchange); set AFPERR_MISC

auth_load: the uam-not-found log called stat() a second time to print a status, clobbering errno; print strerror(errno) from the original stat.

### DIFF
--- a/etc/afpd/auth.c
+++ b/etc/afpd/auth.c
@@ -1295,7 +1295,7 @@ int auth_load(AFPObj *obj, const char *path, const char *list)
                 LOG(log_error, logtype_afpd, "uam: %s load failure", p);
             }
         } else {
-            LOG(log_info, logtype_afpd, "uam: uam not found (status=%d)", stat(name, &st));
+            LOG(log_info, logtype_afpd, "uam: uam not found (%s)", strerror(errno));
         }
 
         p = strtok_r(NULL, ", ", &last);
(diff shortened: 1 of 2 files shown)


andylemin:
ready for review

rdmark:
Incidentally, the exchangefiles method was one that I implemented yesterday in the Netatalk Client, a key part of what macOS TextEdit uses when you edit a text file in place on an AFP volume.

andylemin:
@rdmark This is ready, but this and all PRs now depends on #3075

andylemin:
@rdmark rebased and ready for final review and merge

rdmark (approved):
nicely touched up error handling 👍

PR #3088, opened 2026-06-10
ci: bump dependencies for OpenBSD 7.9 build job

rdmark:
OpenBSD 7.9 is now the default distribution in the vmactions runner, so a number of port versions need to be adjusted accordingly


rdmark:
@andylemin I see now that you're baking the same change into your big spectest overhaul, we can discard this in favor of yours

PR #3090, opened 2026-06-12
initscripts: fix OpenRC service script for netatalk

DerVerruckteFuchs:
Include the full path for the PID file, not just the lockfile path. Only including the lockfile path on Alpine Linux results in OpenRC falsely reporting that the netatalk service has crashed after a few seconds.

I think this may be due to OpenRC trying to use the lockfile directory as the PID file, and failing to do so. This could be occuring during netatalk forking, or some other event/state where OpenRC needs to access the PID file.

### DIFF
--- a/distrib/initscripts/openrc.netatalk.in
+++ b/distrib/initscripts/openrc.netatalk.in
@@ -8,7 +8,7 @@ description_reload="Reload configuration"
 
 name=$RC_SVCNAME
 command="@sbindir@/${RC_SVCNAME}"
-pidfile=@lockfile_path@
+pidfile="@lockfile_path@/${RC_SVCNAME}"
 
 depend() {
 		need net


andylemin:
Hi @DerVerruckteFuchs Thanks for the PR. 

I traced the daemon's actual PID file path through the build system and source:

Build: meson.build:2141 sets PATH_NETATALK_LOCK to lockfile_path + '/netatalk' + lockfile_suffix. On Linux, lockfile_path defaults to /var/lock and lockfile_suffix is empty meson.build:2117-2131.

Daemon: netatalk.c:661 calls create_lockfile("netatalk", PATH_NETATALK_LOCK), which fopens the path and writes getpid() server_lock.c:138-161 — a genuine PID file.

So the daemon writes its PID to /var/lock/netatalk, and ${RC_SVCNAME} resolves to netatalk (the installed service name per meson.build:196, output: service).
The new pidfile now points exactly at the file the daemon creates 👍

The fix mirrors systemd, which already uses the correct form: systemd.netatalk.service.in:13 → PIDFile=@lockfile_path@/netatalk.

Looks like a well-targeted fix that aligns the OpenRC script with both the daemon's actual behavior and the existing systemd units.. Nice

andylemin (approved):

PR #3091, opened 2026-06-13
ci: Performance Dashboard condition should check for forks

rdmark:
we don't want to run the Performance Dashboard job when a PR is originating from a fork, because it won't have the proper permissions to do what it has to do


andylemin:
Thanks @rdmark I noticed this as well in the recent PR we received. You beat me to it ;)

andylemin (approved):

PR #3092, opened 2026-06-13
testsuite: generate test431 AppleDouble data programmatically

rdmark:
Build the legacy AppleDouble v2 sidecar directly in test431 and write it with normal file I/O. The generated 746-byte payload is identical to the former binary fixture, preserving the resource fork and Finder info conversion checks.


rdmark:
@andylemin ready for your review!

this eliminates the only binary blob file that we distribute and install with the software (when testsuite is enabled); arbitrary binary test fixtures is a security risk in my opinion

NJRoadfan:
Looking at this, its WAY easier to generate these in Python.

andylemin:
> Looking at this, its WAY easier to generate these in Python.

@NJRoadfan Hahaha, @rdmark just made me go through the same "pain" with #3093 for the performance dashboard 😂 The dashboards now look identical to the python variants (Perl is just...)
Of course I'm teasing because it made me improve a bunch of other stuff too

andylemin:
### Correctness ✅
Byte-for-byte verified. Compiled the PR's generation logic standalone and compared its output against main:test/testsuite/data/test431_data: identical, 746 bytes. The entry table, dates, TEXTTEST type/creator, PRIVDEV/INO/SYN/SV~ private data, FinderInfo flag byte (688=2), and resource-fork test\n payload all match the original fixture.

Nothing in the tree still uses _PATH_TESTDATA_DIR, test431_data, test-data, or subdir('data') (only a stale plans/*.md doc, unrelated).
Headers OK. fcntl.h/unistd.h/sys/stat.h are included; uint32_t, errno/EINTR, and memcpy/strerror are already used pervasively in this file, so types/symbols resolve.

### Quality improvements
Fixes a pre-existing snprintf bug. The old code compared snprintf(...) > sizeof(cmd) — signed int vs size_t (unsigned), which mishandles truncation and the off-by-one boundary. The new len < 0 || len >= (int)sizeof(cmd) is correct.
EINTR-safe write loop with partial-write accumulation — robust.

### Minor suggestions (non-blocking)
Repetition: the `{ int len = snprintf(...); if (len < 0 || len >= (int)sizeof(cmd)) {...} }` block appears ~4–5 times. A small static helper (e.g. build_path(cmd, sizeof, fmt, ...)) would shrink the function and centralize the truncation check.

andylemin (approved):

rdmark:
please forgive my irrational hatred for python ^^;

PR #3093, opened 2026-06-13
ci: split build/spectest workflows, dedup boilerplate, drop Python

andylemin:
Address PR review follow-ups on the workflow structure and toolchain:

- Split build.yml: it now holds only the pure Linux build + unit-test jobs. The eight non-Linux VM legs (DragonflyBSD, FreeBSD UFS/ZFS, NetBSD, OpenBSD, OmniOS, OpenIndiana, Solaris) that also run the AFP spectest move to a new spectest-vms.yml.

- Extract the per-leg spectest boilerplate into a shared POSIX-sh driver, .github/scripts/vm_spectest.sh. Each VM job now exports only its per-OS contract (CONFDIR, LOCKDIR, SERVICE_START/STOP, AFP_ADOUBLE) and calls the script.

- Replace the three copy-pasted gh-pages publish blocks in containers.yml with a reusable composite action, .github/actions/publish-gh-pages. Folder-scoped pruning and rebase-and-retry push are preserved, so the chart jobs still publish concurrently without trampling each other.

- Port the chart renderers from Python+matplotlib to Perl+gnuplot (plot_lantest.pl, plot_speedtest.pl), removing matplotlib from CI in favour of gnuplot-nox. Output matches the former charts; the render steps install gnuplot-nox and fonts-dejavu-core.

This is the fast-follow from #3075


andylemin:
Hi @rdmark Well I managed to get perl-based plotting working which actually looks good!
Does this workflow break-out and standardisation match what you were looking for?
This is now finished and ready for review

(just fixed all the "github-actions Bot added a commit that referenced this pull request" noise)

rdmark:
you made it happen with Perl, what a feat! 🤩

I will do the review now

.github/workflows/build.yml
```diff
@@ -1,5 +1,10 @@
 name: Builds
 
+# Pure build + unit test jobs on Linux distributions.
+# The non-Linux VM legs that also run the AFP spectest live in
+# spectest-vms.yml; container spectests in containers.yml; the macOS leg
+# in spectest-macos.yml.
```
rdmark, line 6: to this point, perhaps it is time to rename this workflow to `build-linux.yml`?

.github/workflows/spectest-vms.yml
```diff
@@ -0,0 +1,613 @@
+name: VM Spec Tests
```
rdmark, line 1: to more clearly express the practical function, do we want to call this something like `name: BSD, illumos, Solaris Spec Tests` and rename the file `spectests-bsd-illumos-solaris.yml`?

rdmark (approved):
about half of the lines in the Perl scripts are code comments, so at roughly 200 LOC each they're quite nice and compact scripts!

my comments are only concerning the big picture naming and structure, to start a discussion rather than expecting any changes in this PR; we can iterate moving forward

in addition to the inline comments, I would argue that `containers.yml` has also way outgrown the initial premise from years ago (build and sanity test the containers) -- I think the testsuite runs from containers.yml should be broken out at some point, supported by other GH Actions module

.github/workflows/spectest-vms.yml
andylemin: Agreed, but its too long, how about spectests-linux.yml and spectest-unix.yml?

rdmark: sure, that's a fair compromise

andylemin:
@rdmark I'm going to merge this and will follow up on the others later 👍

andylemin:
> about half of the lines in the Perl scripts are code comments, so at roughly 200 LOC each they're quite nice and compact scripts!
> 
> my comments are only concerning the big picture naming and structure, to start a discussion rather than expecting any changes in this PR; we can iterate moving forward
> 
> in addition to the inline comments, I would argue that `containers.yml` has also way outgrown the initial premise from years ago (build and sanity test the containers) -- I think the testsuite runs from containers.yml should be broken out at some point, supported by other GH Actions module

I'll clean it up. I left the comments in place as I'm not a Perl pro....

PR #3094, opened 2026-06-14
lantest: test improvements and optimisations, emit CSV aggregates, perf plotter labels

andylemin:
Fixes a misleading lantest byte range lock test and normalises other tests for comparability.

Lock test fix: The old "Lock/Unlock 10000 times" test locked then immediately unlocked one byte range on one file, repeatedly so the server only ever held one lock — hiding per-fork lock-tracking cost.

Split into:
Byte-range lock/unlock 2000 different ranges in one fork — acquires 2000 distinct locks then releases each.
NEW - Lock then unlock 2000 open forks — stresses the open-fork/refnum table, exposing any O(n) lock-tracking overhead.

Normalisation:
Every test now works on ~2000 units for comparable timings (however some tests still do more operations on those same files, so not directly comparable but easier to weight now), and hidden per-test iteration multipliers were removed (the harness already repeats runs). Cache-hits/validation/deep-traversal/dir-tree/mixed-ops all rescaled; deep-traversal is now pure path resolution.

Try to do and measure just one activity..

Shared-file pipeline:
File tests reuse one 2000-file set instead of each creating/deleting its own:
Create → Write → Fork-lock → Stat/open/read/close → Enumerate → Delete.
NEW - a "Write 1024 bytes to 2000 files" test; reworks the read test to reuse pipeline files.

Result: 14 tests (2 new), reordered; all [N AFP ops] labels recomputed (TOTAL_AFP_OPS = 56,586, verified); -f dependencies wired; man page + --help synced.

Supporting changes:
 - afp_lantest: print the aggregates block (total ops, total time, avg time per AFP op) in CSV mode too, under a '# Aggregate' marker so CI can parse it; tabular mode unchanged.
 - plot_lantest.pl: update SHORT_LABELS for the renamed/reordered 14 tests and reduce x-tick rotation to 12 deg so the longer labels fit.
 - containers.yml: add a 'Summarise lantest results' step that writes the aggregates and a per-test median table (slowest first) to the run summary.
 - Trim redundant comments from both plot scripts (keep only non-obvious rationale).

### DIFF
--- a/doc/manpages/man1/afp_lantest.1.md
+++ b/doc/manpages/man1/afp_lantest.1.md
@@ -56,7 +56,7 @@ features.
 : Server hostname or IP address (default: localhost)
 
 **-K**
-: Run cache-focused tests only (tests 9-12)
+: Run cache-focused tests only (tests 11-14)
 
 **-n** *iterations*
 : Number of test iterations to run (default: 2). For iterations > 5 outliers will be removed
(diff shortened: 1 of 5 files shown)


andylemin:
Ready for review

doc/manpages/man1/afp_lantest.1.md
```diff
@@ -91,45 +91,55 @@ Configure the UAM in netatalk's afp.conf:
 
 The following tests are available:
 
-**(1) Opening, stating and reading 512 bytes from 1000 files**
-: Tests basic file access patterns with many small operations
+Most tests operate on a shared set of 2000 files, normalised so their timings
```
rdmark: nitpick: good if we can standardize on on US English in user facing docs/strings ("normalized")

rdmark (approved):
the lantest is really getting buffed to a shine! it's progressed so vary far beyond the original premise of a "Helios Lantest for AFP"

doc/manpages/man1/afp_lantest.1.md
andylemin: Sorry. I can't help doing this 😂

andylemin:
> the lantest is really getting buffed to a shine! it's progressed so vary far beyond the original premise of a "Helios Lantest for AFP"

Haha, yes! Thanks @rdmark.

It's been a fair few late nights, but after adding Spectests to all the build VMs (adding a good selection of different platforms and filesystems), updating Lantest's lock test coverage, and visualising lantest and speedtest - I now finally feel confident to address all the locking issues.

Locking changes are not a topic to take lightly! Every change must be backed by robust evidence - which we now have.

PR #3095, opened 2026-06-16
macipgw: Send back a complete config packet

NJRoadfan:
Currently macipgw is not sending out a complete config packet with space for eight IP addresses. Fix that and tidy up error message handling while we are at it. This matches the behavior of the Shiva/Kinetics K-STAR routing software and Apple's IP Gateway software.

### DIFF
--- a/contrib/macipgw/macip.c
+++ b/contrib/macipgw/macip.c
@@ -55,6 +55,7 @@
 #define DDPTYPE_MACIP (22)
 #define MACIP_ASSIGN (1)
 #define MACIP_SERVER (3)
+#define MACIP_ERROR (-1)
 #define MACIP_PORT (72)
 
 #define MACIP_MAXMTU (586)
@@ -78,7 +79,7 @@
 struct macip_req_control {
     uint16_t mipr_version;
     uint16_t _mipr_pad1;
-    uint32_t mipr_function;
+    int32_t mipr_function;
 } __attribute__((__packed__));
 
 struct macip_req_data {
@@ -87,6 +88,10 @@ struct macip_req_data {
     uint32_t mipr_broadcast;
     uint32_t _mipr_pad2;
     uint32_t mipr_subnet;
+    uint32_t _mipr_pad3;
+    uint32_t _mipr_pad4;
+    uint32_t _mipr_pad5;
+    char mipr_error[22];
 } __attribute__((__packed__));
 
 struct macip_req {
@@ -131,6 +136,8 @@ static struct macip_data gMacip;
 static struct zones gZones;
 static outputfunc_t gOutput;
 
+static const char *error_noip = "No Address Available.";
+static const char *error_noop = "Unknown Operation.";
 
 static uint16_t cksum(char *buffer, int len)
 {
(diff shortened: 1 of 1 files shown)


contrib/macipgw/macip.c
```diff
@@ -415,35 +422,36 @@ static void config_input(ATP atp, struct sockaddr_at *faddr, char *packet,
 
     rq = (struct macip_req *) buffer;
     bzero(rq, sizeof(struct macip_req));
+    /* We always send these fields */
+    rq->miprc.mipr_version = htons(1U);
+    rq->miprd.mipr_nameserver =
+        htonl(gMacip.nameserver);
+    rq->miprd.mipr_broadcast = htonl(gMacip.broadcast);
+    rq->miprd.mipr_subnet = htonl(gMacip.mask);
+    /* Set packet length to size of macip_req minus error message buffer */
```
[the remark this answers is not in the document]
NJRoadfan: This is by design. We allocate an extra null terminator on the reply packet even if we aren't sending an error.

contrib/macipgw/macip.c
```diff
@@ -415,35 +422,36 @@ static void config_input(ATP atp, struct sockaddr_at *faddr, char *packet,
...
-            len = sizeof(struct macip_req);
             arp_set(ip, &sat);
 
             if (gDebug & DEBUG_MACIP) {
                 printf("assigned %s.\n", iptoa(ip));
             }
         } else {
-            rq->miprc.mipr_function = htonl(0);
-            len = sizeof(struct macip_req);
+            rq->miprc.mipr_function = htonl(MACIP_ERROR);
```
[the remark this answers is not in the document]
NJRoadfan: The buffer is sized to prevent overruns, we reserve a null terminator in line 432.

contrib/macipgw/macip.c
```diff
@@ -78,7 +79,7 @@
 struct macip_req_control {
     uint16_t mipr_version;
     uint16_t _mipr_pad1;
-    uint32_t mipr_function;
+    int32_t mipr_function;
```
andylemin, line 82: The type changed from uint32_t to int32_t. The request side at line 408 still reads this same field via `f = ntohl(rq->miprc.mipr_function)` into an `int f`, and the debug `printf("%d", f)` now interprets it as signed. The wire format is unchanged (still 4 bytes), so this is fine — just confirming the sign change was intentional and not only to make MACIP_ERROR (-1) compile or something?

contrib/macipgw/macip.c
```diff
@@ -131,6 +136,8 @@ static struct macip_data gMacip;
 static struct zones gZones;
 static outputfunc_t gOutput;
 
+const char *error_noip = "No Address Available.";
+const char *error_noop = "Unknown Operation.";
```
andylemin: Minor: these should both be `static const char *const` to keep internal linkage (matching the file's other file-scope helpers) and for const-correctness?

contrib/macipgw/macip.c
NJRoadfan: Its intentional. Per the spec, function is a signed value and field set to ERROR should return a value of FFFF. Previous code wasn't properly returning errors at all!

contrib/macipgw/macip.c
NJRoadfan: Its on the todo for later.

andylemin (approved):

rdmark (approved):

Issue #65, opened 2016-07-04
Invalid Open Flags When Flushing A File Before Deletion


Alex-Richman:
Hi,

It seems like AFPD is calling open() with invalid flags when flushing a file before deletion (and possibly other cases where flush() is called).

I noticed this while developing a fuse file system.  My open() implementation originally threw an EINVAL when none of the O_RDONLY, O_WRONLY or O_RDWR flags were set.  This caused AFP to fail to delete files.
The open(2) man page documents that applications are required to set one of these flags: "The argument flags **must** include one of the following access modes: O_RDONLY, O_WRONLY, or O_RDWR."

I don't see this behaviour with SMB or a few other protocols that I've tried.

Here's a log of the operations performed by AFPD:
// Open called with no RDONLY/WRONLY/RDRW flags?
[DEBU] [2016:07:04 09:50:43]     real_open(node: 0x2107180, path: '/test', fi: 0x7f17896aeca0)
Custom debug: open flags: 163840
  O_RDONLY: 0
  O_WRONLY: 0
  O_RDRW: 0
[DEBU] [2016:07:04 09:50:43]     <= 0 (Success)

[DEBU] [2016:07:04 09:50:43]     real_getattr(node: 0x2107180, path: '/.AppleDouble', stbuf: 0x7f1789eafbb0)
[DEBU] [2016:07:04 09:50:43]     <= 0 (Success)

[DEBU] [2016:07:04 09:50:43]     real_getattr(node: 0x2107180, path: '/.AppleDouble/test', stbuf: 0x7f1788eadbb0)
[DEBU] [2016:07:04 09:50:43]     <= -2 (No such file or directory)

// Flush called on the opened FD
[DEBU] [2016:07:04 09:50:43]     real_flush(node: 0x2107180, fi: 0x7f17896aeca0)
[DEBU] [2016:07:04 09:50:43]     <= 0 (Success)

// Close called on the opened FD
[DEBU] [2016:07:04 09:50:43]     real_release(node: 0x2107180, fi: 0x7f1789eafca0)
[DEBU] [2016:07:04 09:50:43]     <= 0 (Success)

I assume that most file system implementations simply ignore the invalid flags and open the FD anyway (indeed that's what I've changed my own to do) but it still seems like a good idea to fix here to avoid compatibility issues with other esoteric use cases.

Cheers,
- Alex.

slowfranklin:
Hi Alex,

[...]

PR #3104, opened 2026-06-16
adouble: give ad2openflags a fallback access mode; open exchange metadata RDWR

andylemin:
`ad2openflags()` is the single funnel translating `ADFLAGS_*` to the `O_*` flags for `open(2)`, feeding `ad_open_df`/`hf_v2`/`hf_ea`/`rf_ea`.  If a caller passed neither `ADFLAGS_RDWR` nor `ADFLAGS_RDONLY`, no access-mode bit was set and `open()` was invoked with `flags == 0`.
Linux defines `O_RDONLY == 0`, so that looks like a read-only open by accident; strict-POSIX backends (FreeBSD, FUSE, illumos) check `(flags & O_ACCMODE) != 0` and reject it with `EINVAL`.

This resolves GitHub #65.

The only access-mode-less callers in the tree were the two metadata opens in `afp_exchangefiles()`'s header-swap block, so `FPExchangeFiles` failed with `AFPERR_PARAM` on those backends.

Two fixes:

  * ad2openflags(): if neither `ADFLAGS_RDWR` nor `ADFLAGS_RDONLY` is set, default to read-only and log the offender.  The check is on `adflags`, not oflags: `O_RDONLY == 0` means a finished `oflags` can't distinguish "no mode" from a real read-only open, whereas `ADFLAGS_RDWR`/`RDONLY` are distinct bits.  This is a portability safety net for any future caller, not the exchange fix.

  * afp_exchangefiles(): open both metadata forks `ADFLAGS_HF` | `ADFLAGS_RDWR`. `RDWR` (not `RDONLY`) is required because this block rewrites the header and `ad_flush()` only persists when the fd carries `O_RDWR`.  `RDONLY` would also have been wrong-but-quiet: the v2 backend force-upgrades `HF` to `O_RDWR` only inside the `ADFLAGS_RDONLY` branch, but the EA backend does not, so on EA the swap would be silently dropped while `FPExchangeFiles` reported success.  `RDWR` is correct on both backends.

With both offenders fixed the new warning fires only on a genuine programming error (verified: zero occurrences across the full spectest run, all `FPExchangeFiles` cases green).

### DIFF
--- a/etc/afpd/file.c
+++ b/etc/afpd/file.c
@@ -2954,7 +2954,9 @@ int afp_exchangefiles(AFPObj *obj, char *ibuf, size_t ibuflen _U_,
         ad_init_offsets(&adtmp);
 
         if (!AD_META_OPEN(adsp)) {
-            if (ad_open(adsp, p, ADFLAGS_HF) != 0) {
+            /* RDWR: this block rewrites the metadata header (ad_flush only
+             * persists when the fd carries O_RDWR). */
+            if (ad_open(adsp, p, ADFLAGS_HF | ADFLAGS_RDWR) != 0) {
                 err = AFPERR_MISC;
                 goto err_temp_to_dest;
             }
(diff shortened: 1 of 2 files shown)


andylemin:
@rdmark this is ready for review

rdmark (approved):

PR #3106, opened 2026-06-17
ci(lantest): add % of total runtime to candle labels and full results table

andylemin:
Lantest candle labels now show each test's mean runtime and its relative share of the total (mean ms + % of total), so a test's proportion of one full pass is comparable across runs even when absolute times drift. The chart subtitle gains an 'Avg total runtime' figure (sum of per-test means).

The PR dashboard's 'Slowest operations (median)' section is replaced with a full per-test table of all tests by mean runtime (slowest first), and the inline aggregates now report avg total runtime alongside avg time per AFP op.


andylemin:
@rdmark Ready for review and merge

rdmark (approved):

PR #3108, opened 2026-06-18
afpd: close() leaked fork on afp_openfork() error; guard refcount underflow; add fault-injection unit tests

andylemin:
afp_openfork() opens the data/resource fork, then a second ad_open() for the AppleDouble metadata fork, then does getforkparams / WriteInhibit / fork_setmode.  On failure after the first ad_open() succeeded, control jumped to the openfork_err: label, which ran:
    
        openfork_err:
            of_dealloc(ofork);
            if (errno == EACCES) return ...;
            return ret;
    
of_dealloc() unhashes the ofork, drops the struct-adouble refcount and frees it, but never calls ad_close().  So the fd opened by the first ad_open() was leaked into the long-lived afpd child every time a later step reached the label: the metadata open failing with EROFS/EMFILE/ENFILE/EISDIR, getforkparams failing, or the WriteInhibit branch firing.
The lock side effects of that first open (share-mode F_SETLK from ad_open_df) were likewise never released.
    
 Two switch statements compounded it: the EROFS case lacked a break and fell through, so a read-only mount returned AFPERR_NFILE ("too many open files") instead of AFPERR_VLOCK, and the label could not derive the right code from errno because the cleanup close(2) clobbers it.
    
Scope: reachable on every read-only mount (EROFS) and under fd pressure (EMFILE/ENFILE); each occurrence is a permanent fd leak until the session ends.  Latent on the common read-write success path.
    
Fix.  Each error case sets ret explicitly (no fall-through), and every failure after the first ad_open() funnels through the single openfork_err: label, which closes whatever is actually open -- ad_close(DF|RF|HF|SETSHRMD) keyed on the AD_*_OPEN state, exactly once -- before of_dealloc().

The metadata-open, getforkparams, WriteInhibit and fork_setmode error paths all goto the label rather than running their own cleanup, so the fd and its share-mode locks are released in every failure combination.  fork_setmode's bespoke errno->AFPERR mapping (EAGAIN/EACCES/EINVAL -> AFPERR_DENYCONF, else AFPERR_PARAM) is done before the goto, while errno still holds the fault value, because the cleanup ad_close()/close(2) clobbers it.  SETSHRMD releases the data fork's share-mode locks; the open-state guard makes a failure before the first open close nothing.
    
Defense-in-depth: ad_close()'s four adf_refcount decrements (data, HF meta, AD_VERSION2 RF-closes-HF, EA reso) and of_dealloc()'s struct-level ad_refcount decrement are underflow-guarded.  A decrement that would go below zero -- a stray double ad_close()/of_dealloc() on a shared adouble -- is logged and trips AFP_ASSERT (active in non-NDEBUG builds) instead of silently corrupting the count and closing a sibling fork's fd. Behaviour-neutral on every correct path.
    
Testing:
Build-time fault-injection unit test framework (test/afpd/).  afpdtest links the real afpd objects plus libatalk, so it can drive ad_open/ad_close/of_alloc directly in one process against a scratch volume -- reaching faults the black-box spectest cannot.

Because libatalk is a shared library, a link-time --wrap on the executable cannot intercept its internal libc calls, so faultinject.c is built two ways from one source: compiled into afpdtest it provides the `fi` control block; built as libfaultinject.so (-DFAULTINJECT_PRELOAD) and LD_PRELOADed it interposes open/close/fcntl/fchdir at dynamic resolution, forwarding to the real symbol via dlsym(RTLD_NEXT) unless a failure is armed.  Under glibc with _FILE_OFFSET_BITS=64 the single open() definition is renamed by <fcntl.h> to open64() (the symbol LFS callers reference), so no separate open64() interposer is needed.

A test arms an fi armed/fail_after/errno counter narrowly around the call under test and resets after.  test.c logs to /dev/stderr rather than /dev/tty (no controlling terminal under meson test / CI / docker, where /dev/tty writes ENXIO and every LOG() was silently dropped), so a passing run is clean and a failure emits context; each utest_* returns 0 or a distinct nonzero code surfaced as "# got: N".
    
Interception is not reliable on every platform -- macOS's two-level namespace ignores a plain symbol preload -- so the self-test is a capability probe that returns TEST_SKIP when injection does not reach libatalk, and injection-dependent tests skip too, keeping coverage honest per platform.
    
Tests (subtests_lock.c), two kinds -- broad accounting-invariant tests that pin a ledger across a family of sequences, and targeted tests for nuance accounting cannot see:
    
 - utest_faultinject_selftest (framework): probes whether the preload intercepts open() here, then arms EMFILE on the next open and confirms ad_open() inside libatalk fails -- gates the injection mechanism itself; the errno check lives only in the probe because ad_open()'s error path clobbers errno.
 - utest_openfork_no_fd_leak (accounting): runs the openfork_err cleanup on an open data fork and asserts both ledgers agree the fd is gone (ad_data_fileno == -1 and fcntl(fd) -> EBADF), with a negative control so it cannot pass vacuously -- the regression net for this fix.
 - utest_shared_adouble_refcount_balance (accounting): walks a multi-reference open/close sequence asserting adf_refcount and fileno stay coupled (fd open iff refcount > 0; an early close must not close the shared fd).
 - utest_adclose_underflow_aborts (targeted): manufactures the fileno>=0 / refcount<=0 desync the guard defends and asserts ad_close() trips AFP_ASSERT -> SIGABRT, run in a fork()ed child so the harness survives; skips under NDEBUG where AFP_ASSERT is a no-op.

Make the unit test summary string a required argument of the TEST/TEST_int/TEST_expr/TEST_int_or_skip macros:
Unit tests are now self-describing.

### DIFF
--- a/test/afpd/faultinject.h
+++ b/test/afpd/faultinject.h
@@ -1,0 +1,52 @@
+/*
+  Copyright (c) 2026 andylemin
+
+  This program is free software; you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation; either version 2 of the License, or
+  (at your option) any later version.
+
+  This program is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+*/
+
+#ifndef FAULTINJECT_H
+#define FAULTINJECT_H
+
+/*!
+ * @file
+ * @brief Test-only fault-injection control block (see faultinject.c).
+ *
+ * Usage in a test (LD_PRELOAD interposition is global, so arm narrowly):
+ *   fault_inject_reset();
+ *   fi.open_armed = 1; fi.open_fail_after = 1; fi.open_errno = EMFILE;
+ *   ... run only the operation under test ...
+ *   fault_inject_reset();
+ *
+ * fail_after = N: the next N armed calls succeed, call N+1 fails once.
+ */
+
+struct fault_inject {
+    int open_armed;
+    int open_fail_after;    /*!< succeed this many armed opens, then fail once */
+    int open_errno;
+    int close_armed;
+    int close_fail_after;
+    int close_errno;
+    int fcntl_armed;
+    int fcntl_fail_after;
+    int fcntl_errno;
+    int fchdir_armed;
+    int fchdir_fail_after;
+    int fchdir_errno;
+};
+
+extern struct fault_inject fi;
+
+void fault_inject_reset(void);
+
+int faultinject_open_works(const char *scratch_path);
+
+#endif /* FAULTINJECT_H */
--- a/etc/afpd/fork.c
+++ b/etc/afpd/fork.c
@@ -626,18 +626,22 @@ int afp_openfork(AFPObj *obj _U_, char *ibuf, size_t ibuflen _U_, char *rbuf,
         fullpathname(s_path->u_name),
         (fork == OPENFORK_DATA) ? "data" : "reso",
         !(access & OPENACC_WR) ? "O_RDONLY" : "O_RDWR");
-    ret = AFPERR_NOOBJ;
 
     /* First ad_open(), opens data or resource fork */
     if (ad_open(ofork->of_ad, upath, adflags, 0666) < 0) {
+        /* Set ret per case (no fall-through): the cleanup label returns ret and
+         * must not consult errno, which the cleanup ad_close()/close(2) clobbers. */
         switch (errno) {
         case EROFS:
             ret = AFPERR_VLOCK;
+            goto openfork_err;
 
         case EACCES:
+            ret = (access & OPENACC_WR) ? AFPERR_LOCK : AFPERR_ACCESS;
             goto openfork_err;
 
         case ENOENT:
+            ret = AFPERR_NOOBJ;
             goto openfork_err;
 
         case EMFILE :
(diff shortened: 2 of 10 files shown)


andylemin:
augment review

andylemin:
@rdmark This is one of the most complex changes in the planned series. It has received extensive planning, multi-lens reviews, and many rounds of cross-cutting analysis, and is now complete - no known outstanding issues in this cleanup code remain. Includes detailed invariant unit-tests to validate accounting.

test/afpd/faultinject.h
```diff
@@ -0,0 +1,61 @@
...
+
+/*!
+ * @brief Is the LD_PRELOAD open() interposer active on this platform?
+ *
+ * Arms a one-shot open() failure and probes it against the given scratch path.
+ * Returns 1 if the injected failure fired (interception works), 0 otherwise
+ * (e.g. macOS two-level namespace, or the preload was not loaded).  Leaves `fi`
+ * reset.  Injection-dependent tests should gate on this and return TEST_SKIP
+ * when it is 0, so coverage stays honest per platform.
+ */
```
rdmark: our convention is to document functions in the implementation file and not the header file

rdmark (approved):
it looked daunting at first but most of the new lines were in test code :)

only one code documentation convention remark

test/afpd/faultinject.h
andylemin: Totally agree.. Thanks! 🙂

andylemin:
@rdmark Yes there is a lot of new test scaffolding. The actual changes are small, but they are critical path and result in material correctness changes - hence the test demands. This scaffolding opens up many fault injection capabilities and I prefer this approach for invariant testing over hoping to hit AFP_ASSERT

PR #3110, opened 2026-06-18
adouble: don't kill afpd on fchdir-restore failure in ad_metadataat/ad_openat

andylemin:
Two related robustness fixes. The first resolves two process-killing error paths in the AppleDouble layer; the second fixes the errno-clobbering root cause that made those (and many other) error paths fragile.

Fix 1 — ad_metadataat() / ad_openat() no longer terminate the process
Bug. Both functions snapshot the current directory as an fd, fchdir() into a caller-supplied target dir, do their work, then fchdir() back. _If filesystem changes in anyway, during work afpd crashes_.

Two defects:
- Process termination on restore failure. If the fchdir() back failed, ad_metadataat() called exit() and ad_openat() called AFP_PANIC() — tearing down the entire afpd child.
- Skip-restore on early error. A failure after fchdir(dirfd) succeeded took an error exit that closed the saved fd but skipped the fchdir() back, leaving the process in the target directory.

Scope & trigger:
- ad_openat() is reachable — called with a real dirfd from copyfile() and the deletefile() data-fork open, so its AFP_PANIC fires on a restore failure (saved-cwd fd invalidated by an unmount, EIO on a flaky mount, or EACCES on revoked permission). Rare, but whole-process blast radius.
- ad_metadataat() is latent — its only checkAttrib=1 caller passes dirfd == -1, so this is forward-looking hardening.
The skip-restore defect triggers on any error after a successful fchdir(dirfd) on a real-dirfd call.

On restore failure, anchor cwd to / (always valid — the daemon chdir("/")'d at startup), log, and return -1 instead of terminating; afpd re-resolves cwd from the CNID via movecwd() on its next path op. Restore cwd on every exit path (fixes the skip-restore defect). Return a meaningful errno so copyfile/deletefile map it to the right AFPERR: keep the operation's errno when the operation failed, surface the restore errno when only the restore failed. Snapshot opens now use O_DIRECTORY | O_CLOEXEC, and the entry open/fchdir checks are split so the log identifies which failed.

Fix 2 — make_log_entry() is now errno-transparent
Root cause. The logger calls write(), vasprintf() (and open() on first use), all of which set errno.
So LOG() on an error path could clobber the errno a caller was about to act on. Only ~23 of the tree's ~3116 LOG() sites worked around this with manual save/restore — the rest silently assumed errno survived, which it didn't.

The existing 23 manual handling sites for errno persistence due to LOG can now be removed after this.
NB; The most common source of this corruption-on-corruption is when the LOG line itself fails when it cannot log the desired error string/variable for any reason (common issue).

make_log_entry() now saves errno on entry and restores it on every exit, as a diagnostic routine should (like perror/syslog). This fixes the class everywhere and lets Fix 1 drop the save/restore that only guarded LOG() — the remaining saves in ad_open.c guard real syscalls (close()/chdir()), not the logger. This avoids various transient log errors from corrupting Netatalk errors, mostly defensive/correctness/code-cleanup fix.

### DIFF
--- a/libatalk/adouble/ad_open.c
+++ b/libatalk/adouble/ad_open.c
@@ -2468,25 +2468,58 @@ int ad_metadataat(int dirfd, const char *name, int flags, struct adouble *adp)
     int cwdfd = -1;
 
     if (dirfd != -1) {
-        if (((cwdfd = open(".", O_RDONLY)) == -1) || (fchdir(dirfd) != 0)) {
-            ret = -1;
-            goto exit;
+        if ((cwdfd = open(".", O_RDONLY | O_DIRECTORY | O_CLOEXEC)) == -1) {
+            LOG(log_error, logtype_ad,
+                "ad_metadataat: cannot snapshot cwd: %s", strerror(errno));
+            return -1;
+        }
+
+        if (fchdir(dirfd) != 0) {
+            /* Preserve fchdir's errno across close(). */
+            int saved = errno;
+            LOG(log_error, logtype_ad,
+                "ad_metadataat: fchdir(target) failed: %s", strerror(saved));
+            close(cwdfd);
+            errno = saved;
+            return -1;
         }
     }
 
     if (ad_metadata(name, flags, adp) < 0) {
         ret = -1;
-        goto exit;
+        /* fall through: restore cwd before returning */
     }
 
     if (dirfd != -1) {
+        /* Captured before the restore so its chdir()/close() can't clobber the
+         * operation's errno (which callers map to an AFPERR). */
+        int saved = errno;
+
         if (fchdir(cwdfd) != 0) {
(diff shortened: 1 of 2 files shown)


andylemin:
@rdmark ready for review

rdmark (approved):
the error handling looks much more sound now -- nice catch!

> The existing 23 manual handling sites for errno persistence due to LOG can now be removed after this.

do you plan a fast-follow to clean up these 23 manual handling sites?

andylemin:
@rdmark Yes, now that LOG does not clobber, and this situation is actually common in Netatalk - where an error log fails to log as the variable was not in the expected state (because there was an error in the first place!). Hence error-on-error or corruption-on-corruption.

The existing sites are now redundant, and I want to remove them - as the existing sites make the error handling opaque in itself ironically!

andylemin:
Requires follow up #3113

PR #3114, opened 2026-06-19
meson: define common afpd sources globally

rdmark:
rather than defining the afpd sources twice in the meson build scaffolding, we now have a global 'common' sources array which is used by both the afpd daemon and the afpd tests with a handful of local enrichments that are relevant to their respective contexts

as an added boon, we now build both the afpstats and ASP code into the test binary

### DIFF
--- a/etc/afpd/meson.build
+++ b/etc/afpd/meson.build
@@ -1,4 +1,4 @@
-afpd_sources = [
+afpd_common_sources = files(
     'ad_cache.c',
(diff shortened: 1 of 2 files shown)


rdmark:
@andylemin ready for your review; this should make it a bit easier to work with the tests

andylemin:
@rdmark this is great! LGTM

andylemin (approved):

PR #3123, opened 2026-06-24
nad: introduce StuffIt (.sit) endocing and decoding support

### LINKED ISSUE #3107: nad: support for encoding/decoding stuffit (sit) archives
StuffIt was the de-facto standard archiving and compression format for Macs throughout the 90s, adding support for it alongside MacBinary and BinHex would make for a near-complete Classic Mac file suite

there is a Rust implementation with a permissive license that we can learn from https://github.com/benletchford/stuffit-rs

rdmark:
this gives 'nad' the capability to compress to StuffIt with 'nad sit' and decompress StuffIt with 'nad unsit'

uses the stuffis-rs Rust library by Ben Letchford for encoding and decoding functionality

in the CI jobs, we build the Rust library on Alpine, Arch, Fedora, macOS, FreeBSD, and OpenBSD; the other platforms don't have recent enough Rust toolchains to build the library, or other issues with the dependencies

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -22,6 +22,11 @@ on:
 permissions:
   contents: read
 
+env:
+  STUFFIT_RS_COMMIT: 16336d7abff5c925e8081869bfa0dd6ee27f0658
+  LD_LIBRARY_PATH: /usr/local/lib
+  PKG_CONFIG_PATH: /usr/local/lib/pkgconfig
+
 jobs:
   build-alpine:
     name: Alpine Linux
@@ -40,6 +45,7 @@ jobs:
             bison \
             build-base \
             ca-certificates \
+            cargo \
             cracklib \
             cracklib-dev \
             cracklib-words \
@@ -51,6 +57,7 @@ jobs:
             dconf \
             flex \
             file-dev \
+            git \
             gcc \
             glib \
             iniparser-dev \
@@ -70,11 +77,19 @@ jobs:
             perl \
             pkgconfig \
             rpcsvc-proto-dev \
+            rust \
             sqlite-dev \
             talloc-dev \
             tinysparql-dev \
             valgrind \
             xapian-core-dev
+      - name: Install stuffit-ffi
+        run: |
+          git clone https://github.com/benletchford/stuffit-rs.git /tmp/stuffit-rs
+          cd /tmp/stuffit-rs
+          git checkout "$STUFFIT_RS_COMMIT"
+          cargo build --release --locked -p stuffit-ffi
+          make -C stuffit-ffi install-files PREFIX=/usr/local
       - name: Configure
         run: |
           meson setup build \
@@ -83,6 +98,7 @@ jobs:
             -Dwith-cups-pap-backend=true \
             -Dwith-eventloop=libev \
             -Dwith-statedir-path=/var/lib \
+            -Dwith-stuffit=true \
             -Dwith-tests=true \
             -Dwith-testsuite=true
       - name: Build
@@ -130,6 +146,7 @@ jobs:
             flex \
             file \
             gcc \
+            git \
             iniparser \
             libev \
             localsearch \
@@ -139,10 +156,18 @@ jobs:
             perl \
             pkgconfig \
             rpcsvc-proto \
+            rust \
             sqlite \
             talloc \
             tinysparql \
             xapian-core
+      - name: Install stuffit-ffi
+        run: |
+          git clone https://github.com/benletchford/stuffit-rs.git /tmp/stuffit-rs
+          cd /tmp/stuffit-rs
+          git checkout "$STUFFIT_RS_COMMIT"
+          cargo build --release --locked -p stuffit-ffi
+          make -C stuffit-ffi install-files PREFIX=/usr/local
       - name: Configure
         run: |
           meson setup build \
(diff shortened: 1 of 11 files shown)


rdmark:
this depends on the C ABI that I am proposing to the stuffit-rs project in https://github.com/benletchford/stuffit-rs/pull/4

rdmark:
@andylemin I’m well aware we still lack test coverage for this tool; promise to make that happen next :)

andylemin:
Hi @rdmark the more basic question?
What exactly does this do for us? How does it benefit users?

I think we need an example 🙂

I feel like I'm missing some big reveal for such a large dependancy? And how do we handle the platforms that don't have modern rust?

rdmark:
Fair question! This is a feature that I myself has wished for for years. The majority of very old archived Macintosh software are compressed in this StuffIt format. However bootstrapping an old Mac OS system can be challenging because you often don’t have a copy of StuffIt Expander handy by default, alternatively the version you have cannot handle the particular compression algorithm of the sit archives you want to decompress. Therefore, having this utility in `nad` can be a big boon and convenience when getting software onto an old Mac.

If nothing else, it will surely be well received by the vintage Mac community and a good promotional taking point. 

Agreed that the Rust dependency is heavy. However this Rust StuffIt library is really solid and capable, with support for multiple compression methods. 

When the StuffIt library is unavailable, we have graceful fallback to build nad without StuffIt support. It’s an optional feature all things considered. 

I’ve filed an upstream PR to improve the portability of the library.

andylemin:
Hi @rdmark 

Got it. I think I remember sit files actually now you've described it like this.

One last question. What is stopping a user installing stuffit themselves on the same server where they run Netatalk and achieving the same?

Ie, What is it about integrating with nad that makes things easier?

Should this be wrapped behind a --with-stuffit or --with-rust?

Or can we auto-detect a locally installed StuffIt dynamically? So a user can install it afterwards, and the integration appears?

Modern users who are build size sensitive and don't have classic machines (NAS builds etc) will likely need to ensure this is disabled?

I'm actually a fan of Rust, and I can see a lot more uses for it for the wider project anyway. So no concerns there.

andylemin (approved):

rdmark:
@andylemin the benefit of baking this into `nad` is that it gains the ability to read and write Netatalk style AppleDouble metadata, which is obviously mandatory if you want to extract a sit archive onto a Netatalk shared volume

there are three living open source CLI tools that I know of which are able to extract sit archives, but all of them output macOS style metadata sidecars only, which are not compatible with Netatalk

- [stuffit](https://github.com/benletchford/stuffit-rs) (the CLI tool that is built on the stuffit-rs library we also use)
- [sit](https://github.com/thecloudexpanse/sit)
- [unar](https://github.com/MacPaw/XADMaster)

I can see two alternative ways forward for better compatibility with the broader ecosystem:

1. convince / support these projects to implement Netatalk style AppleDouble sidecar support
2. implement native support for macOS style AppleDouble sidecars in Netatalk, and make it the default fallback for xattr long term

obviously, it would be preferable if you could use stuffit, sit, or unar to extract your StuffIt archive onto a Netatalk AFP volume and have it just work seamlessly :)

rdmark:
> Should this be wrapped behind a --with-stuffit or --with-rust?

do you mean in the build system or dynamically in `nad`?

> Or can we auto-detect a locally installed StuffIt dynamically? So a user can install it afterwards, and the integration appears?

instead of linking with libstuffit_ffi.so you mean? I think we run into to the same limitation that neither of the three CLI tools that can extract sit archives are able to output Netatalk compatible metadata by themselves

I chose the path of implementing this in `nad` since this allows us to leverage the adouble and vfs logic in libatalk

andylemin:
Hi @rdmark 

"Implement native support for macOS style AppleDouble sidecars in Netatalk, and make it the default fallback for xattr long term" - This piques my interest!
I admit I did not realise the Netatalk Adouble sidecar format was different to the macOS style metadata sidecars..

I think I understand the purpose and value, and the compromise as it stands 👍
Happy to merge what you have done here. But it makes me nervous building this by default?
We don't want to cause problems for the largest user base (modern macOS and NAS users)?

I think we need a couple of new wiki pages;
- A wiki page with build and tuning for high performance modern macOS users with lots of memory and CPU (full servers)
- A wiki page with build and tuning for high performance modern macOS users but with small memory and CPU (NAS boxes)
- A wiki page with build, tuning and command options for classic Mac users?

There is an increasibly stronger divergence between these, and one size does not fit all, and it would give you a place to show off nad properly for users who need it?

We should create an issue to track the implementation of macOS style metadata sidecars? 🙂

rdmark:
@andylemin no need to worry! `-Dwith-stuffit` is false by default, so unless users or downstream packagers deliberately opt in, this will not affect the size of the installation

how about we add a new section to the [Installation manual chapter](https://github.com/Netatalk/netatalk/blob/main/doc/manual/Installation.md) which outlines which packages to install for these three feature sets?

issue for macOS style metadata sidecars here: https://github.com/Netatalk/netatalk/issues/3140

andylemin:
@rdmark Perfect. Thanks.

Yes a section in the installation manual to confirm when these optional build extras are needed works.

The other cases are tuning. Like the samba page, we just need a set of pages with common config tuning examples.

The ZFS page needs an update.
The Samaba page needs an update.
High memory and low memory should be one page actually.
And a high performance page.
Classic Mac page.

There are lots of little kernel level things as well like increasing rlimit.

Anyway, we can think about preparing these as part of the next release which coincides with the next macOS release 👍

PR #3124, opened 2026-06-24
ci: create a Ubuntu build job with 32 bit toolchain

rdmark:
emulating a 32 bit build by forcing 32 bit toolchain on a 64 bit system

the current repos are missing 32 bit versions of libmariadb and libtracker-sparql-3.0 so skipping those features for now

despite the dtrace library being present, Ubuntu is missing a sys/sdt.h header which I'm chalking down as an Ubuntu bug and disabling it

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -376,6 +376,7 @@ jobs:
             libmagic-dev \
             libmariadb-dev \
             libpam0g-dev \
+            libsqlite3-dev \
             libtalloc-dev \
             libtirpc-dev \
             libtracker-sparql-3.0-dev \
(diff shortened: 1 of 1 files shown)


andylemin:
Hi @rdmark 

We also need to run `meson test` and the spectests?

There are some bugs I have come across recently which will get fixed soon, which are vulnerable on 32 bit systems.

The AFP/Netatalk locking architecture is an advanced high fidelity design.

Rather than only providing open for read/writing, and simple read locks and write locks, Netatalk maintains another 10 additional locking sharemodes/flags, and byte range locks (plus attributes such as deleteinhibit for AFP's 'Locked' status).

These provide extra locking states across the data file and resource fork content.

As all these additional locking flags are not supported by POSIX they have to be stored somewhere. So they are cleverly stored in the top addressable bytes of the data fork in the main file's fd.

With 64 bit systems this places them well beyond the end a file's real data/content, as you can place byte range locks on any address regardless of underlying file data.

Each lock mode bit is set with a single offset and length of 1, where offset is byterangemax -(1-10).

For 32 bit systems, there is a high chance that data will run to the maximum available file size (usually as a truncated file).

And the default locking mode is to set data content byte range locks (not sharemode lock bits) across the file data. However this means the file data lock and the sharemode lock bits overlap.
Ie, setting content locks tramples the special lock bits.

The required fix is to cap the effective maximum file data size to byterangemax-11, to reserve the lock bits headroom, avoiding content range and sharemode lock range collisions.

I will be adding tests for the locking contract soon. So we should add tests for 32bit here so they get executed once added?

andylemin:
Looking good, just spectests left to add? Not critical though as the locking test structure will be added as a unit test

andylemin (approved):

rdmark:
Cheers, I didn’t have time to annotate and clean this up before going to work this morning. 

Spectest is a valid follow up project. I think it would require either a special 32 bit runner container or the use of a vmactions runner. 

This is what’s currently not working in 32 bit on this Ubuntu runner:

- mariadb and libsparql don’t seem to have 32 bit packages (or they have an unexpected naming scheme)
- building with dtrace leads to compiler error with missing header
- building with Xapian leads to compiler error that our C++ wrapper is not 32 bit safe

all are worth looking closer at, I will file bug tickets later

rdmark:
the drace failure is because `sys/sdt.h` is missing with the 32 bit toolchain, I'm going to chalk this down as an Ubuntu bug and leave it be

the xapian failure was fixed with a CXX flag

PR #3145, opened 2026-06-29
deletefile(): POSIX-lock-safe, GET-only open-conflict check, and fix no-access/no-lock open blocking delete

andylemin:
`deletefile()`'s open-conflict check (decides whether a peer holds a lock/share-mode that must block the delete) accreted three layers that are no longer aligned, with five distinct defects: 1) stranded held locks on every platform, 2) over-blocking on `FPDelete`, 3) a dead check on rename/copy-overwrite, 4) invisible deny modes, and 5) a fail-open that unlinks when lock state can't be read.

This PR replaces the whole check with one `F_GETLK` (read-only) conflict read, through an existing held fd, or one safely-transient fd, and which reads lock state per-bit and is fail-closed, fixing all five).

## Root cause

POSIX advisory locks are owned by the process, not the fd: closing any fd to a file releases every lock on all fds the process holds on it (`fcntl(2)`; same on the BSDs, Solaris/illumos, AIX). `F_OFD_SETLK` avoids this but is absent on the BSDs/Solaris/AIX, so it is not an option.

In the kernel: an `F_SETLK` lock is keyed on the `(inode, process)` pair, **not** on the open file description or the fd number. The kernel keeps the lock list on the inode and tags each entry with the owning process (Linux `struct file_lock.fl_owner`/`fl_pid`; the BSDs key on `p_leader`; Solaris/illumos on the proc). The fd you used to *take* the lock is not recorded against it. So when *any* `close(2)` runs on *any* fd pointing at that inode, the kernel walks the inode's lock list and drops **every** entry owned by this process — there is no per-fd accounting to scope the release to the fd being closed. The man page states the consequence directly:

> *"If a process closes any file descriptor referring to a file, then all of the process's locks on that file are released, regardless of the file descriptor(s) on which the locks were obtained. This is bad: it means that a process can lose its locks on a file … because of a function call buried in a library."* — `fcntl(2)`, "Record locks"

That last sentence is exactly netatalk's situation: the "function call buried in a library" is `deletefile()` opening a transient second fd to probe, then closing it — silently discarding the share-mode lock the client's *open fork* still relies on, through a completely unrelated fd. `F_OFD_SETLK` keys the lock on the open file description instead, so a close of a different fd leaves it intact — but it does not exist on the BSDs/Solaris/AIX, so the fix cannot depend on it.

The old `deletefile()` trial-acquired a lock through a transient second fd at two sites — the `checkAttrib` NODELETE probe (`ad_metadataat(CHECK_OF)`) and the body `ad_openat` + `ad_tmplock`. Closing either transient fd dropped the session's own share-mode lock on the inode, so a just-opened-for-read file then read as busy. A trial-SET also answers the wrong question ("could I lock the whole fork?") rather than "does a peer hold a lock/share-mode the spec says blocks delete?".

## How the old detector got here (5 commits, ~12 years)

This is not anyone's mistake — it is what incremental, well-intentioned patches add up to when each is locally correct but no one revisits the whole. The trial-lock detector was assembled over a decade:

- **`31843674` — rufustfirefly, 2000 (initial revision).** `deletefile()` is born with the trial-acquire model: open the fork, `ad_tmplock()` it, infer "in use" from whether the lock is granted. Reasonable in a single-user world where the POSIX close-drops-locks trap rarely bit.
- **`3606b9e3` — jmarcus, 2002 ("trash will work in multi-user environments… read-write/read-only lock issues").** Extends the body probe to a whole-resource-fork `ADLOCK_FILELOCK` trial lock so an open file blocks deletion across sessions. This commit also *documented its own limits* — both surviving `FIXME`s are his: the "priority inversion" note, and *"doesn't work for RFORK open read only and fork open without deny mode"*. Defects 2 and 3 trace to the whole-fork trial lock introduced here.
- **`170ad4c6` — didg, 2003 (renamefile concurrent-access rework).** Reworks `renamefile()`/copy-overwrite and routes them through `deletefile(checkAttrib=0)`. Because that path never runs the attribute probe, its fd stays `O_RDONLY` and the trial `ADLOCK_WR` silently downgrades to a read lock — defect 3, the "rename slips through where delete is refused" half.
- **`96afda34` — didg, 2006 ("SFM compatible format for adouble").** Adds `check_attrib()` (the NODELETE / DOPEN / ROPEN attribute probe) and the `ad_openforks` band reader. That reader only ever probes the OPEN band bits (+0..+3), never `DENY_*` — defect 4 — and the CHECK_OF attribute open is what later forces the body fd to `O_RDWR` and produces the over-broad sweep of defect 2.
- **`0576bda9` — Frank Lahm, 2012 ("Refactor the locking").** Makes *"close locks on file close"* the accepted model (`of_closefork()` → `ad_unlock()` drops the fork's byte locks on every close), turning the close-drops-locks trap from edge case into routine — defect 1. The spectest recorded the flip: the old test asserting locks survive a sibling close (`test78`) was disabled as an *"unfixed `adf_unlock()` bug"* and replaced by one asserting the *inverse* (`test410`: *"in Netatalk 3 we now close locks on file close"*, `FPCloseFork` annotated *"this should drop the byterange lock!"*).

None of these is wrong in isolation; the defects above are emergent — the seams between the layers, surfacing once the 2012 refactor plus `F_OFD`-less platforms and EA backends made the close-drops-locks trap routine.

### Coverage matrix — what each layer could actually detect

Each conflict dimension a delete *should* weigh, against what the old detector saw on each of its two paths (`afp_delete`, i.e. `checkAttrib=1`, vs `rename`/`copy-overwrite`, i.e. `checkAttrib=0`), and what the new read does. This is the mental translation table: read down "new" and every cell is a deliberate ✓.

Legend: ✓ = detected and handled correctly · ~ = detected but **over-broad** (blocks more than it should) · ✗ = **blind** (never detected) · ✗-strand = actively *broke* state.

| conflict dimension | old `afp_delete` | old `rename`/`copy` | new code |
|---|---|---|---|
| NODELETE attribute (kFPDeleteInhibitBit) | ✓ (96afda34) | ✗ never checked | ✓ both paths |
| peer **data** content **write** lock | ✓ (2000) | ✓ (WR-probe survives the RD downgrade vs a peer WR) | ✓ |
| peer **data** content **read** lock | ✓ (RDWR body probe) | ✗ (RD-probe, RD-vs-RD doesn't conflict) | ✓ |
| peer **resource** content lock | ✓ (2000 → whole-rfork 2002) | ✗ (RD downgrade) | ✓ |
| share-mode **OPEN_RD/WR** (mere open) | ~ blocks (band sweep) | ✗ blind | ✓ OPEN_RD excluded from blocking, OPEN_WR still blocks (requires dirty fork accounting) |
| share-mode **DENY_RD/WR** | ~ blocks (sweep, not isolated) | ✗ blind | ✓ blocks (isolated bit) |
| **RSRC** share-mode mirrors (+2/3, +6/7) | ~ | ✗ | ✓ |
| **OPEN_NONE** marker (+8/9) | ~ blocks (**wrong** — claims nothing) | ✗ | ✓ excluded from blocking |
| own (same-process) held locks | ✗-strand | ✗-strand | ✓ never stranded |
| unreadable lock state | ✗ unlinks anyway | ✗ unlinks anyway | ✓ refuses (fail-closed) |

**What no old layer ever covered** (✗ across both old columns, now ✓): a peer's read-only content lock on the rename/copy path; the `DENY_*` band as a thing distinct from "anything is open"; same-process safety (every path stranded the caller's own locks); and fail-closed behaviour (every path could unlink on an unreadable probe).

**Where coverage is still deliberately not 100%** — the new read *sees* everything above, but two policy/cosmetic gaps remain, tracked in "Not in this PR":

- A **bare read-only open** (`OPEN_RD`, no deny, no byte lock) still blocks delete. The new code reads it correctly; *whether it should block* is an unresolved AFP/SMB-semantics question, so the policy mask conservatively keeps blocking it for now.
- The refusal reason is still a single `AFPERR_BUSY`, not split into `AFPERR_DENYCONF` vs `AFPERR_BUSY` — the read now *knows* which bit blocked, but the wire code isn't yet differentiated.

(Cross-process conflicts are *not* a gap: `F_GETLK` reports other processes' locks/band by design, so a peer afpd child's holds are seen regardless of the held-fd-reuse optimisation, which is purely same-process.)

## Lock layout (for the per-bit read)

Lock state is `fcntl` byte ranges partitioned by offset across two fds:

| fd | offset zone | contents |
|---|---|---|
| data fd | `[0, BYTELOCK_MAX]` | content byte-range locks (`FPByteRangeLock`) |
| data fd | `[AD_FILELOCK_BASE … +9]` (past EOF) | 10-entry share-mode band |
| resource-fork fd | `[0, BYTELOCK_MAX]` | resource content (only dimension not on the data fd) |

The share-mode band is ten fixed offsets past EOF, each a one-bit flag for an AFP open/deny/none state planted by `fork_setmode()`:

| offset (from base) | symbol | group |
|---|---|---|
| +0 / +1 | OPEN_WR / OPEN_RD | open access (data) |
| +2 / +3 | RSRC_OPEN_WR / RSRC_OPEN_RD | open access (resource mirror) |
| +4 / +5 | DENY_WR / DENY_RD | deny modes (data) |
| +6 / +7 | RSRC_DENY_WR / RSRC_DENY_RD | deny modes (resource mirror) |
| +8 / +9 | OPEN_NONE / RSRC_OPEN_NONE | "open, claiming nothing" markers |

Two POSIX properties the read relies on: band entries are stored `F_RDLCK`, so a probe must use `F_WRLCK` to detect them (a read lock does not conflict with a read lock); and `F_GETLK` never reports the caller's own locks (afpd is process-per-session, so a probe through a held fd reports only peers).

## Defects in the old detector (one function, five faults)

1. **Stranded locks (all platforms).** Both transient-fd probe sites drop the session's own locks on close.
2. **Over-blocks on `FPDelete`.** The NODELETE probe forces the data fd `O_RDWR` (CHECK_OF ⇒ SETSHRMD ⇒ `O_RDWR`); the body reuses it, so `ad_tmplock(ADLOCK_WR,0,0)` is a real `F_WRLCK` over `[0,∞)` that sweeps the band and conflicts with any entry, including no-claim markers (verified at maxdebug).
3. **Dead on rename/copy-overwrite (`checkAttrib=0`).** No CHECK_OF open, so the body fd is `O_RDONLY` and `ad_tmplock(ADLOCK_WR)` downgrades to `F_RDLCK` — blind to the band and to a peer read lock. (This is why rename slipped through where delete was refused.)
4. **Deny modes invisible.** The only band reader (`ad_openforks`) probes +0..+3 only, never `DENY_*`; `OPEN_NONE` (+8) and v2 resource-fork-only opens also unseen.
5. **Fail-open.** On `EACCES`/failed probe open the handle stayed `NULL`, the trial locks were skipped, and the code unlinked anyway.

Two of these were already flagged in-tree. The `deletefile()` resource-fork block on `main` carried two long-standing `FIXME`s, both from jmarcus's 2002 commit, and this PR's restructure **deletes the block that held them** — they are resolved, not relocated:

- **FIXME A — "priority inversion".** *"we have a pb here because we want to know if a file is open; there's a 'priority inversion' if you can't open the resource fork RW you can delete it if it's open because you can't get a write lock."* This is the trial-acquire model admitting it conflates *"can I lock it?"* with *"may I delete it?"*. The `F_GETLK` read separates the two (read state, then decide policy), so the inversion disappears.
- **FIXME B — "doesn't work for RFORK open read only and fork open without deny mode".** The exact statement of defect 3 (and part of 4): an RDONLY-downgraded probe sees neither a peer's read-only open nor the share-mode band. The explicit-`F_WRLCK` per-bit read sees both.

Worked example (the spectest case that flips RED→GREEN): a no-access open (`fork_setmode` plants only `OPEN_NONE`, an `F_RDLCK`, base+8 — spec: "None access allows no further access to the fork"); a cross-session `FPDelete`'s `F_WRLCK` over `[0,∞)` (defect 2) hits that marker and returns `AFPERR_BUSY`. One unbounded trial lock cannot exclude `OPEN_NONE` or distinguish it from `DENY_WR` — a per-bit read plus a policy decision is required.

## Fix

Invariant: this afpd child can only hold a lock on a file via an fd it keeps open in the owning ofork. So the conflict read is `F_GETLK`-only with two branches: reuse the held fd if the fork is open (no open/close → no strand), else open one transient fd (no lock held → nothing to strand). `deletefile()` becomes: NODELETE → one conflict read → policy → unlink, fail-closed (indeterminate read → `AFPERR_ACCESS` + log, never unlink).

New primitives:

- **`ad_testlock_range(ad, eid, off, len)`** (`libatalk/adouble/ad_lock.c`) — ranged `F_GETLK`; no `adf_lock[]` scan (no self-report through a held fd); explicit `F_WRLCK` (sees peer `F_RDLCK` band entries on any fd mode); zone-clamps a data-zone request by subtraction (`len > MAX - start`) so `len==0` never becomes "to infinity" into the band and never overflows `off_t`.
- **`ad_testlock_whole(ad, eid)`** — one unclamped whole-data-fd `F_GETLK` (content + band) as a negative-case fast path.
- **`of_get_locks(vol, dirfd, path, …)`** (`etc/afpd/ofork.c`) — single entry point; afpd-side because it needs `oforks[]` to find a held fd. Resolves held-or-transient fd per fork, reads data + resource content + each band bit, retries the open as root on `EACCES` (band is server bookkeeping, not a user permission; the unlink still runs as the user; `become_root`/`unbecome_root` paired, `errno` captured before restore). Returns tri-valued `OF_LOCKS_OK` / `OF_LOCKS_NOENT` / `OF_LOCKS_ERROR` and a positional band bitmap (bit *n* == base+*n*).

Platform-agnostic: no OFD, no build probe. One source conditional, `#if HAVE_EAFD && SOLARIS`, for the backend where the resource fork is an `O_XATTR` fd off the data fd.

`ad_open` NORF/NOHF error handling is also narrowed: a *missing* resource fork / header (`ENOENT`/`ENOATTR`) stays a non-error, but a hard open error (EIO, EACCES) now propagates instead of being masked to success — so `of_get_locks` fails closed on an indeterminate resource-fork read rather than reading it as unlocked.

### How each element maps to a defect / FIXME

Every fault above is closed by a specific design element — nothing is left open:

| # | Defect / FIXME | Closed by |
|---|---|---|
| 1 | Stranded held locks (all platforms) | `F_GETLK`-only read + held-fd-reuse / safely-transient branches — the open/close that strands a lock never happens |
| 2 | Over-blocks on `FPDelete` (whole-`[0,∞)` `F_WRLCK` sweeps the band) | per-bit `ad_testlock_range` + zone clamp (data probe can't reach the band) + `DELETE_BLOCKING_BAND_BITS` policy (no-claim markers excluded) |
| 3 / FIXME B | Dead on rename/copy-overwrite; RDONLY-downgraded probe | explicit `F_WRLCK` in `ad_testlock_range` — independent of fd mode, so it sees a peer's read open and the band on every caller |
| 4 | Deny modes / `OPEN_NONE` / rfork-only opens invisible | `of_get_locks` reads **every** band bit positionally (data + resource mirror), not just +0..+3 |
| 5 | Fail-open (unlink on unreadable state) | tri-valued return; `OF_LOCKS_ERROR` → `AFPERR_ACCESS` + log, never unlink; plus the `ad_open` NORF/NOHF narrowing so a hard read error is no longer masked to "unlocked" |
| FIXME A | "priority inversion" (can't-lock conflated with may-delete) | read state first (`F_GETLK`), then a separate policy decision — the two questions are no longer fused into one trial-SET |

The rewrite **removes** the `check_attrib()` + `ad_tmplock` block that carried both FIXMEs, so neither comment is relocated or left behind — `grep` for "priority inversion" / "RFORK open read only" in the tree now returns nothing.

## Behaviour changes
- `copyfile()`'s two error-path cleanup deletes now honour `deletefile()`'s return (no blind force-unlink of an uncertain-ownership destination; logs a refused removal).
- A no-access open no longer blocks delete: `DELETE_BLOCKING_BAND_BITS` excludes `OPEN_NONE`/`RSRC_OPEN_NONE` (one-line policy edit on the new per-bit read). Deny modes, open-for-write, and content byte-range locks still block.
- A plain open readonly no longer blocks delete: `DELETE_BLOCKING_BAND_BITS` excludes `OPEN_RD`/`RSRC_OPEN_RD` (one-line policy edit on the new per-bit read). Deny modes, open-for-write, and content byte-range locks still block.

Delete allowed with `OPEN_NONE`/`RSRC_OPEN_NONE`/`OPEN_RD`/`RSRC_OPEN_RD`; *This aligns with the AFP specification and with Samba.*
The AFP spec does not authoritatively require a bare read-access open to block a delete: its FPDelete result-code table (Reference, Table 16) lists only kFPObjectLocked (DeleteInhibit) as a held-claim refusal — kFPFileBusy is not even a listed FPDelete result — and the File Sharing Modes chapter builds all synchronization from deny modes, where a plain access mode ("None access allows no further access... except to close it") restricts nobody. Only an explicit deny mode is a claim against other users.
Samba matches this: source3/smbd/open.c share_conflict() gates a delete solely on the holder's FILE_SHARE_DELETE grant (the deny-type axis), never on whether the existing open was for read or write - *So both AFP and Samba users now get the same experience on the same share.*

Note, samba even allows delete for *open for write*. Both AFP and Samba docs focus on DENY modes. However I am stopping short of allowing delete on open for write, as this would require additional code changes to enable the delete sequence to check for DIRTY un-flushed pages and some other write pending semantics (which must always block delete even if user did not open for write with DENY mode. Deleting before forkflush has risk of corruption.

### Spectests
- test605 — A foreign (non-afpd) process holding a raw F_WRLCK share-mode band lock (OPEN_WR offset) makes a cross-session FPDelete correctly return AFPERR_BUSY.
- test608 — A "NONE"-access open (no Read/Write/DenyRead/DenyWrite → only an OPEN_NONE marker) does not block a cross-session FPDelete, which must succeed.
- test609 — Cross-session DeleteInhibit lifecycle: session 1 sets ATTRBIT_NODELETE → session 2 FPDelete is refused AFPERR_OLOCK; session 1 clears it → session 2 FPDelete succeeds.
- test610 — A "OPEN_RD"-access open (no Write/DenyRead/DenyWrite → only an OPEN_RD marker) does not block a cross-session FPDelete, which must succeed.
- AFP spectest: ea:sys 340 pass / 0 fail, ea:ad 341 pass / 0 fail.

### Unit tests
- New Unit Tests = 11
- afpd unit suite: 49/49, on both `ea=ad` (v2) and `ea=sys`. Covers: same-process close-drops-locks hazard control + the cure (held lock survives the read, via a forked peer); no self-report; `F_WRLCK` sees a peer `F_RDLCK`; zone clamp bounds `l_len`; ≥2-dimension whole-fd fast path; fail-closed `ERROR`, `NOENT`, and absent-resource-fork → `OK`; NODELETE on both backends. A capability probe skips fault-injection/`ea=sys` tests where unsupported.
- Fault-injection harness: also interposes `__open_2`/`__open64_2` so probe opens are intercepted under `_FORTIFY_SOURCE` (Ubuntu/glibc ≥ 2.42), where no-`O_CREAT` opens are rewritten away from `open`/`open64`. Verified on glibc and musl.

- Pre-existing `afp_copyfile` held-source resource-fork-handle corruption.
- Pre-existing 32-bit lock-range / band-overlap clamp.
- Splitting the delete refusal into `AFPERR_DENYCONF` vs `AFPERR_BUSY`.
- Investigate the 2012 close-drops-locks behaviour itself (`of_closefork()` → `ad_unlock()` in `0576bda9`): a single `FPCloseFork` still releases *every* fork's byte-range locks on the inode. This PR only stops the *delete probe* from triggering it; the general per-fork lock-release fix is unaddressed.
- Pending that fix, re-evaluate whether `test410`'s expectation should revert and `test78` be re-enabled — i.e. assert locks *survive* a sibling close again, rather than the suite encoding the bug as expected behaviour.

---

Every commit in this locking fix/audit series manipulates one or more layers of the stack below.
The diagram is the single source of truth for *what owns what* and *what the kernel actually tracks* — the lock/fork bugs are all divergences between the netatalk-side bookkeeping (left) and the kernel-side reality (right).

```mermaid
flowchart TD
    %% ── colour legend (by resource TYPE); rounded nodes via ("...") shape
    %%   wire/protocol = blue   netatalk tables = purple
    %%   netatalk structs = teal   lock bookkeeping = amber
    %%   kernel resources = red    on-disk = grey
    classDef wire   fill:#1e3a8a,stroke:#1e40af,color:#eff6ff,stroke-width:1px;
    classDef table  fill:#5b21b6,stroke:#6d28d9,color:#f5f3ff,stroke-width:1px;
    classDef struct fill:#0f766e,stroke:#0d9488,color:#f0fdfa,stroke-width:1px;
    classDef lock   fill:#b45309,stroke:#d97706,color:#fffbeb,stroke-width:1px;
    classDef kernel fill:#9f1239,stroke:#be123c,color:#fff1f2,stroke-width:1px;
    classDef disk   fill:#374151,stroke:#4b5563,color:#f9fafb,stroke-width:1px;

    subgraph WIRE["🌐 AFP wire protocol (client ⇄ afpd)"]
        OP("AFP op<br/><i>FPOpenFork · FPRead · FPByteRangeLock<br/>FPCloseFork · FPDelete · FPGetFileParms</i>")
        REFNUM("OForkRefNum (16-bit)<br/>1..0xffff, 0 reserved<br/><i>the only fork handle the client holds</i>")
    end

    subgraph AFPD["🟣 afpd child — one process per DSI session — netatalk bookkeeping"]
        direction TB

        subgraph TABLE["refnum table + inode index · etc/afpd/ofork.c"]
            OFORKS("oforks[] — struct ofork *[nforks]<br/>nforks = min(0xffff, getdtablesize()/2) · commit 0<br/>slot = refnum % nforks · refnum == slot · commit 12")
            HASH("ofork_table[64] — hash by (dev,inode)<br/><i>of_findname / of_findnameat / of_ad</i>")
            FREEQ("of_freeq[] — FIFO of free slots · commit 12<br/>O(1) alloc/free · refnum = slot")
        end

        OFORK("struct ofork — one per FPOpenFork<br/>of_refnum (uint16_t · = slot · = lock-owner id)<br/>key = {dev, inode}<br/>of_flags (RSRC / ACCWR / DIRTY / ERROR)")

        ADOUBLE("struct adouble — ONE per inode per child<br/><i>shared across sibling oforks via ad_ref</i><br/>ad_refcount (struct-level · freed at 0)<br/>ad_vers (AD_VERSION2 | AD_VERSION_EA)<br/>ad_open_forks (DOPEN/ROPEN) · data/meta/reso refcounts")

        subgraph FORKS["struct ad_fd members + backend indirections"]
            DF("ad_data_fork — struct ad_fd<br/>adf_fd (real OS fd | -1 | AD_SYMLINK)<br/>adf_refcount (fd-level)")
            RF("ad_resource_fork — struct ad_fd<br/>adf_fd · adf_refcount")
            MDP("ad_mdp → metadata fork<br/><i>(indirection)</i>")
            RFP("ad_rfp → resource fork<br/><i>(indirection)</i>")
        end

        LOCKARR("adf_lock[] — adf_lock_t per lock (0..N, grows on demand)<br/>struct flock (l_start · l_len · l_type)<br/>user = of_refnum (owner) · refcount* (shared reads)<br/>holds BOTH: share-mode band (AD_FILELOCK_*) + byte ranges")
    end

    subgraph KERNEL["🔴 OS kernel — the real resources"]
        direction TB
        FDLIMIT("RLIMIT_NOFILE (per process)<br/>raised toward RLIM_MAX = 133120 · commit 0<br/>~2 fds per fork on EA backends")
        OSFD("open file descriptions / fd table<br/><i>one real fd per opened fork file</i>")
        POSIXLOCK("POSIX advisory record locks · fcntl(F_SETLK)<br/><b>owned by the PROCESS</b><br/>⚠ POSIX Quirk: closing ANY fd to the inode drops<br/>ALL this process's locks on that inode")
        SHARE("Solaris/illumos F_SHARE reservations<br/><i>HAVE_FSHARE_T · OPTION_SHARE_RESERV</i>")
        INODE("inode on disk<br/><i>data fork + ._ sidecar (v2) or EA (ea)</i>")
    end

    %% ═══ cardinalities: 1 (exactly one) · 0..1 (optional) · 1..N · N..1 (shared)
    OP -->|"1"| REFNUM
    REFNUM -->|"1:1 · refnum == slot (commit 12)"| OFORKS
    OFORKS -->|"occupied slot → 1 ofork"| OFORK
    HASH -.->|"inode lookup · delete probe · of_ad"| OFORK
    FREEQ -.->|"free slots only"| OFORKS
    OFORK -->|"1 (mandatory) · N oforks : 1 adouble (ad_ref)"| ADOUBLE
    ADOUBLE ==>|"1 (data fork: always the lock/fd anchor)"| DF
    ADOUBLE -->|"0..1 (resource fork: OPTIONAL — ADFLAGS_NORF)"| RF
    ADOUBLE --> MDP
    ADOUBLE --> RFP

    %% backend-specific indirection wiring (which ad_fd the indirections point at)
    MDP -. "ea → ad_data_fork (HF shares data fd)" .-> DF
    MDP -. "v2 → ad_resource_fork (HF+RF share ._ fd)" .-> RF
    RFP -. "both backends → ad_resource_fork" .-> RF

    %% locks: one MANDATORY share-mode lock per open fork, plus 0..N byte-range
    DF ==>|"1 mandatory share-mode lock per fork<br/>(fork_setmode: AD_FILELOCK_OPEN_* — incl. OPEN_NONE)<br/>+ 0..N byte-range (FPByteRangeLock) · rfork share via rf2off"| LOCKARR
    RF -->|"0..N byte-range locks on the rfork's own fd"| LOCKARR

    %% netatalk bookkeeping mapped onto real kernel resources
    DF -->|"1:1 when open (adf_fd ≥ 0; else -1 / AD_SYMLINK)"| OSFD
    RF -->|"0..1 (only if rfork opened)"| OSFD
    OSFD -->|"counts against"| FDLIMIT
    OSFD -->|"1:1"| INODE
    LOCKARR -->|"each lock → set_lock(): F_SETLK · all platforms"| POSIXLOCK
    LOCKARR -. "0..N · Solaris + OPTION_SHARE_RESERV only" .-> SHARE
    POSIXLOCK -->|"keyed on inode + process"| INODE
    SHARE --> INODE

    %% the bug surface
    POSIXLOCK -. "⚠ POSIX Quirk drops these — commit 9B keeps the<br/>delete probe off transient fds (reuse a held fd,<br/>or probe only when none held)" .-> LOCKARR

    %% ── apply colours by type ──────────────────────────────────────────
    class OP,REFNUM wire;
    class OFORKS,HASH,FREEQ table;
    class OFORK,ADOUBLE,DF,RF,MDP,RFP struct;
    class LOCKARR lock;
    class FDLIMIT,OSFD,POSIXLOCK,SHARE kernel;
    class INODE disk;

    %% grey-white background for the outer/inner container boxes
    style WIRE   fill:#f1f5f9,stroke:#cbd5e1,color:#334155;
    style AFPD   fill:#f8fafc,stroke:#cbd5e1,color:#334155;
    style TABLE  fill:#eef2f6,stroke:#d1d9e0,color:#334155;
    style FORKS  fill:#eef2f6,stroke:#d1d9e0,color:#334155;
    style KERNEL fill:#f1f5f9,stroke:#cbd5e1,color:#334155;

    %% highlight the POSIX Quirk bug edge in red (last edge declared, index 23)
    linkStyle 23 stroke:#dc2626,stroke-width:2px;
```

### DIFF
--- a/test/afpd/test.sh
+++ b/test/afpd/test.sh
@@ -12,8 +12,21 @@ if [ $? -ne 0 ]; then
     exit 1
 fi
 
+AFPTESTVOLUMESYS=$(mktemp -d /tmp/AFPtestvolumeSys-XXXXXX)
+if [ $? -ne 0 ]; then
+    echo Error creating ea=sys AFP test volume >&2
+    exit 1
+fi
+
+AFPTESTCNIDSYS=$(mktemp -d /tmp/AFPtestCNIDSys-XXXXXX)
+if [ $? -ne 0 ]; then
+    echo Error creating ea=sys CNID test directory >&2
+    exit 1
+fi
+
 SIGNATURE=$(LC_ALL=C tr -dc 'A-Za-z0-9' < /dev/urandom | head -c 16)
 VOLUUID=$(uuidgen | tr 'a-z' 'A-Z')
+VOLUUIDSYS=$(uuidgen | tr 'a-z' 'A-Z')
 
 if [ -f test.conf ]; then
     echo "Removing stale configuration template test.conf ..."
@@ -31,11 +44,18 @@ cnid server = 127.0.0.1
 
 [test]
 cnid scheme = sqlite
-cnid server = localhost:4700
 ea = none
 path = $AFPTESTVOLUME
 vol dbpath = $AFPTESTCNID
 volume name = afpd_test
 volume uuid = $VOLUUID
+
+[testsys]
+cnid scheme = sqlite
+ea = sys
+path = $AFPTESTVOLUMESYS
+vol dbpath = $AFPTESTCNIDSYS
+volume name = afpd_test_sys
+volume uuid = $VOLUUIDSYS
 EOF
 echo [ok]
--- a/test/afpd/test_capabilities.c
+++ b/test/afpd/test_capabilities.c
@@ -1,0 +1,161 @@
+/*
+  Copyright (c) 2026 andylemin
+
+  This program is free software; you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation; either version 2 of the License, or
+  (at your option) any later version.
+
+  This program is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+*/
+
+/*!
+ * @file
+ * @brief Test capabilities: environment/build features a unit test may depend on.
+ *
+ * Some tests can only run where a feature is present (the LD_PRELOAD shim reaches
+ * libatalk; the filesystem supports user extended attributes for an ea=sys
+ * volume).  A test probes the capability it needs and returns TEST_SKIP when it is
+ * absent, so coverage stays honest per platform rather than failing where the
+ * feature is simply unavailable.  Each capability is probed once by actually
+ * exercising it (e.g. an xattr round-trip), not by trusting a higher-level
+ * operation that may succeed without the underlying support; the result is
+ * cached for the run.
+ */
+
+#ifdef HAVE_CONFIG_H
+#include "config.h"
+#endif /* HAVE_CONFIG_H */
+
+#include <errno.h>
+#include <fcntl.h>
+#include <stdio.h>
+#include <string.h>
+#include <sys/param.h>
+#include <unistd.h>
+
+#include <atalk/ea.h>
+#include <atalk/volume.h>
+
+#include "faultinject.h"
+#include "test_capabilities.h"
+
+/* Cached probe results: -1 = not yet probed, 0 = absent, 1 = present. */
+static int cap_cache[] = {
+    [TEST_CAP_FAULT_INJECT] = -1,
+    [TEST_CAP_EA_SYS]       = -1,
+};
+
+/*! @brief Short capability name for skip/log messages. */
+const char *test_capability_name(enum test_capability cap)
+{
+    switch (cap) {
+    case TEST_CAP_FAULT_INJECT:
+        return "fault-injection (LD_PRELOAD)";
+
+    case TEST_CAP_EA_SYS:
+        return "user extended attributes (ea=sys)";
+
+    default:
+        return "unknown";
+    }
+}
+
+/*!
+ * @brief Does the filesystem under vol->v_path support user extended attributes?
+ *
+ * Round-trips a real xattr on a scratch file: set, read back, remove.  An ea=sys
+ * volume stores AppleDouble metadata in xattrs, so without this the volume opens
+ * (openvol() does not write-probe) but every metadata write fails — which would
+ * surface as a test failure rather than an honest skip.  Probing here keeps the
+ * skip honest.
+ */
+static int probe_ea_sys(const struct vol *vol)
+{
+    char path[MAXPATHLEN + 1];
+    char buf[4];
+    int fd;
+    int ok = 0;
+
+    if (snprintf(path, sizeof(path), "%s/.cap_ea_probe", vol->v_path)
+            >= (int)sizeof(path)) {
+        return 0;
+    }
+
+    fd = open(path, O_RDWR | O_CREAT | O_TRUNC, 0600);
+
+    if (fd < 0) {
+        return 0;
+    }
+
+    close(fd);
+
+    /* set -> read back -> verify; AD_EA_META is netatalk's own metadata EA name,
+     * so the probe uses the exact namespace an ea=sys volume writes. */
+    if (sys_setxattr(path, AD_EA_META, "ok", 2, 0) == 0
+            && sys_getxattr(path, AD_EA_META, buf, sizeof(buf)) == 2) {
+        ok = 1;
+    }
+
+    (void)sys_removexattr(path, AD_EA_META);
+    (void)unlink(path);
+    return ok;
+}
+
+/*!
+ * @brief Is the LD_PRELOAD interposer active and reaching libc here?
+ *
+ * Arms a one-shot open() failure and checks it fired.  macOS's two-level
+ * namespace ignores a plain symbol preload, so the injection never takes effect
+ * there and fault-injection-dependent tests skip.
+ */
+static int probe_fault_inject(const struct vol *vol)
+{
+    char path[MAXPATHLEN + 1];
+
+    if (snprintf(path, sizeof(path), "%s/.cap_fi_probe", vol->v_path)
+            >= (int)sizeof(path)) {
+        return 0;
+    }
+
+    return faultinject_open_works(path);
+}
+
+/*!
+ * @brief Probe a capability; the result is cached for the run.
+ *
+ * @param cap  the capability to probe
+ * @param vol  supplies a filesystem path for capabilities probed against the
+ *             volume (TEST_CAP_EA_SYS, TEST_CAP_FAULT_INJECT use vol->v_path for
+ *             a scratch file); must not be NULL.
+ * @returns    1 if the capability is available, 0 otherwise.
+ */
+int test_capability(enum test_capability cap, const struct vol *vol)
+{
+    if (cap < 0 || cap >= (int)(sizeof(cap_cache) / sizeof(cap_cache[0]))) {
+        return 0;
+    }
+
+    if (cap_cache[cap] != -1) {
+        return cap_cache[cap];
+    }
+
+    switch (cap) {
+    case TEST_CAP_FAULT_INJECT:
+        cap_cache[cap] = probe_fault_inject(vol);
+        break;
+
+    case TEST_CAP_EA_SYS:
+        cap_cache[cap] = probe_ea_sys(vol);
+        break;
+
+    default:
+        cap_cache[cap] = 0;
+        break;
+    }
+
+    return cap_cache[cap];
+}
--- a/test/afpd/test_capabilities.h
+++ b/test/afpd/test_capabilities.h
@@ -1,0 +1,29 @@
+/*
+  Copyright (c) 2026 andylemin
+
+  This program is free software; you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation; either version 2 of the License, or
+  (at your option) any later version.
+
+  This program is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+*/
+
+#ifndef TEST_CAPABILITIES_H
+#define TEST_CAPABILITIES_H
+
+#include <atalk/volume.h>
+
+enum test_capability {
+    TEST_CAP_FAULT_INJECT,   /* LD_PRELOAD libc interposition reaches libatalk */
+    TEST_CAP_EA_SYS,         /* the test filesystem supports user xattrs (ea=sys) */
+};
+
+int test_capability(enum test_capability cap, const struct vol *vol);
+
+const char *test_capability_name(enum test_capability cap);
+
+#endif /* TEST_CAPABILITIES_H */
--- a/distrib/docker/testsuite_alp.Dockerfile
+++ b/distrib/docker/testsuite_alp.Dockerfile
@@ -94,6 +94,8 @@ RUN meson setup build \
     -Dwith-testsuite=true \
 &&  meson compile -C build
 
+RUN meson test -C build --print-errorlogs
+
 RUN meson install --destdir=/staging/ -C build
 
 FROM alpine:3.24.1@sha256:28bd5fe8b56d1bd048e5babf5b10710ebe0bae67db86916198a6eec434943f8b AS deploy
--- a/distrib/docker/testsuite_deb.Dockerfile
+++ b/distrib/docker/testsuite_deb.Dockerfile
@@ -107,6 +107,8 @@ RUN meson setup build \
     -Dwith-testsuite=true \
 &&  meson compile -C build
 
+RUN meson test -C build --print-errorlogs
+
 RUN meson install --destdir=/staging/ -C build
 
 FROM debian:13.5-slim@sha256:4e401d95de7083948053197a9c3913343cd06b706bf15eb6a0c3ccd26f436a0e AS deploy
--- a/etc/afpd/afp_dsi.c
+++ b/etc/afpd/afp_dsi.c
@@ -993,7 +993,7 @@ void afp_over_dsi(AFPObj *obj)
             function = (uint8_t) dsi->commands[0];
             /* AFP replay cache */
             rc_idx = dsi->clientID % REPLAYCACHE_SIZE;
-            LOG(log_debug, logtype_dsi, "DSI request ID: %u", dsi->clientID);
+            LOG(log_maxdebug, logtype_dsi, "DSI request ID: %u", dsi->clientID);
 
             if (replaycache[rc_idx].DSIreqID == dsi->clientID
                     && replaycache[rc_idx].AFPcommand == function) {
@@ -1007,14 +1007,14 @@ void afp_over_dsi(AFPObj *obj)
                 if (afp_switch[function]) {
                     dsi->datalen = DSI_DATASIZ;
                     dsi->flags |= DSI_RUNNING;
-                    LOG(log_debug, logtype_afpd, "<== Start AFP command: %s",
+                    LOG(log_maxdebug, logtype_afpd, "<== Start AFP command: %s",
                         AfpNum2name(function));
                     AFP_AFPFUNC_START(function, (char *)AfpNum2name(function));
                     err = (*afp_switch[function])(obj,
                                                   (char *)dsi->commands, dsi->cmdlen,
                                                   (char *)&dsi->data, &dsi->datalen);
                     AFP_AFPFUNC_DONE(function, (char *)AfpNum2name(function));
-                    LOG(log_debug, logtype_afpd, "==> Finished AFP command: %s -> %s",
+                    LOG(log_maxdebug, logtype_afpd, "==> Finished AFP command: %s -> %s",
                         AfpNum2name(function), AfpErr2name(err));
 
                     if (!idle_worker_is_active()) {
@@ -1054,14 +1054,14 @@ void afp_over_dsi(AFPObj *obj)
             if (afp_switch[function] != NULL) {
                 dsi->datalen = DSI_DATASIZ;
                 dsi->flags |= DSI_RUNNING;
-                LOG(log_debug, logtype_afpd, "<== Start AFP command: %s",
+                LOG(log_maxdebug, logtype_afpd, "<== Start AFP command: %s",
                     AfpNum2name(function));
                 AFP_AFPFUNC_START(function, (char *)AfpNum2name(function));
                 err = (*afp_switch[function])(obj,
                                               (char *)dsi->commands, dsi->cmdlen,
                                               (char *)&dsi->data, &dsi->datalen);
                 AFP_AFPFUNC_DONE(function, (char *)AfpNum2name(function));
-                LOG(log_debug, logtype_afpd, "==> Finished AFP command: %s -> %s",
+                LOG(log_maxdebug, logtype_afpd, "==> Finished AFP command: %s -> %s",
                     AfpNum2name(function), AfpErr2name(err));
                 dsi->flags &= ~DSI_RUNNING;
             } else {
--- a/etc/afpd/file.c
+++ b/etc/afpd/file.c
@@ -2012,7 +2012,18 @@ int copyfile(struct vol *s_vol,
         ad_close(adp, adflags);
 
         if (EEXIST != err) {
-            deletefile(d_vol, -1, dst, 0);
+            /* dst ownership is not certain here (the create failed for a
+             * non-EEXIST reason, or dst appeared via a race), so the conflict
+             * check governs the cleanup unlink: if it reports the object as
+             * held or unreadable, do not remove it — log and leave it. */
+            int dret = deletefile(d_vol, -1, dst, 0);
+
+            if (dret != AFP_OK && dret != AFPERR_NOOBJ) {
+                LOG(log_error, logtype_afpd,
+                    "copyfile('%s'): could not remove partial destination: %d",
+                    dst, dret);
+            }
+
             goto done;
         }
 
@@ -2072,7 +2083,17 @@ error:
     }
 
     if (err) {
-        deletefile(d_vol, -1, dst, 0);
+        /* dst is the server's own freshly-created file (the CREATE|EXCL open
+         * above succeeded, so nothing pre-existed).  Log if the conflict check
+         * refuses (e.g. a peer raced an open on it), so a stranded temp file is
+         * diagnosable rather than silent. */
+        int dret = deletefile(d_vol, -1, dst, 0);
+
+        if (dret != AFP_OK && dret != AFPERR_NOOBJ) {
+            LOG(log_error, logtype_afpd,
+                "copyfile('%s'): could not remove failed-copy destination: %d",
+                dst, dret);
+        }
     } else if (stat_result == 0) {
         /* set dest modification date to src date */
         struct utimbuf	ut;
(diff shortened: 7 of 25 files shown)


andylemin:
Hi @rdmark I have left my commits split out, to make it easier to follow (keep all the testing code separate from the actual change commit).
Don't merge yet, I want to run stress-testing using this PR. But please do start to read it and ask questions?

No regressions on the `Deleting 2000 files` test despite the new fidelity. I was worried the new resolution might slow things down. But you will see in the code I added a fast path for low resolution lookups, so it should be faster for the simple single lock checks, as well as being capable of actually testing full contract locks (which does trigger a lookup for every bit requested)..

andylemin:
augment review

andylemin:
@rdmark I have finished this now. I am still soak testing, but ready for review.
I have rewritten the summary to make it as clear as possible being a big change, and to demonstrate the amount of research I have done.

test/afpd/test.sh
```diff
@@ -37,5 +49,14 @@ path = $AFPTESTVOLUME
 vol dbpath = $AFPTESTCNID
 volume name = afpd_test
 volume uuid = $VOLUUID
+
+[testsys]
+cnid scheme = sqlite
+cnid server = localhost:4700
```
rdmark: minor thing, but `cnid server` is only used by the dbd backend, isn't it?

the preexisting `[test]` volume has the same superfluous line

test/afpd/test.sh
```diff
@@ -37,5 +49,14 @@ path = $AFPTESTVOLUME
...
+ea = sys
+path = $AFPTESTVOLUMESYS
+vol dbpath = $AFPTESTCNIDSYS
+volume name = afpd_test_sys
+volume uuid = $VOLUUID
```
rdmark: I think it would be preferable to generate a separate volume UUID for the second volume

test/afpd/test_capabilities.c
```diff
@@ -0,0 +1,146 @@
...
+/*!
+ * @brief Does the filesystem under vol->v_path support user extended attributes?
+ *
+ * Round-trips a real xattr on a scratch file: set, read back, remove.  An ea=sys
+ * volume stores AppleDouble metadata in xattrs, so without this the volume opens
+ * (openvol() does not write-probe) but every metadata write fails — which would
+ * surface as a test failure rather than an honest skip.  Probing here keeps the
+ * skip honest.
+ */
+static int probe_ea_sys(const struct vol *vol)
```
rdmark, line 76: we have another static function with the same purpose in libatalk/util/netatalk_conf.c -- `do_check_ea_support()`

while they are small functions and not much duplicated logic, perhaps an argument can be made that the application and test should call the same function, not least to avoid future drift?

not a mandatory fix for this PR, but something worth considering long term

test/afpd/test_capabilities.h
```diff
@@ -0,0 +1,52 @@
...
+
+/*!
+ * @brief Probe a capability; the result is cached for the run.
+ *
+ * @param cap  the capability to probe
+ * @param vol  supplies a filesystem path for capabilities probed against the
+ *             volume (TEST_CAP_EA_SYS, TEST_CAP_FAULT_INJECT use vol->v_path for
+ *             a scratch file); must not be NULL.
+ * @returns    1 if the capability is available, 0 otherwise.
+ */
```
rdmark: can we please put the doxygen code comments in the implementation file instead of the header? :)

rdmark (approved):
a handful of non-critical feedback comments for your consideration!

rdmark:
@andylemin if you want to go even deeper into code archaeology, I have stored all known historical tarballs here:

https://sourceforge.net/projects/netatalk/files/historical/

the oldest one is 1.3.3 which is the final stable code that Craig delivered, then you have 1.4b2 which is the final development version from the original author

I find it useful sometimes to diff these to the suite of 1.4b2+asun* tarballs, to identify the code refactored / written by Adrian

then there's about a one year gap between netatalk-1.4b2+asun2.1.3.tar.gz (final changelog item dated 1999-02-26) and the initial git revision, so we have a pretty solid picture of the evolution through these stages!

rdmark:
as a side note, the PR description deserves a blog post or tech notes article somewhere, a thrilling read, not to be lost to time in a throwaway GitHub PR ;)

test/afpd/test_capabilities.c
andylemin: Cool. I see it's static — reusing it means exporting it from libatalk (new declaration in a public header, symbol added to the library), which is an API change for a test convenience? So leaving this out for now 🫡

andylemin:
@rdmark Thank you, I'm pleased you enjoyed reading the work I put in. This is critical core code afterall..
I'm collecting PRs which I want to write up into a overall journey article as part of Spec 4.x launch materials. I don't have a blog to micro post these as I go.. Maybe I could use the wiki's (dedicated branch) in the meantime as a place to store them

All you feedback applied, thanks 🙂 Have you been able to test this yourself yet?

rdmark:
I have not tested in a live environment yet, but will start doing so once we merge to `main` :)

andylemin:
@rdmark Testing is looking good. The last change was to also exclude `OPEN_RD` / `RSRC_OPEN_RD` from the delete blocking mask. Added spectest to cover this, and Summary updated to clarify.
SMB and AFP (Samba/Netatalk) users now get the same experience.

Extend DELETE_BLOCKING_BAND_BITS to also exclude the read-only-open markers OPEN_RD / RSRC_OPEN_RD, alongside the already-excluded OPEN_NONE / RSRC_OPEN_NONE. A fork held open read-only with no deny mode makes no claim that should refuse a delete; only an explicit deny (DENY_*), a write open (OPEN_WR), DeleteInhibit, or a held content lock should block. OPEN_WR and all DENY_* bits keep blocking.

Samba source: source3/smbd/open.c share_conflict() gates a delete solely on the holder's FILE_SHARE_DELETE grant (the deny-type axis), never on whether the existing open was for read or write.

**Note, samba even allows delete for "open for write".** Both AFP and Samba docs focus on DENY modes. However I am stopping short of allowing delete on open for write, as this would require additional code changes to enable the delete sequence to check for DIRTY un-flushed pages and some other write pending fork semantics (which must always block delete even if user did not open for write with DENY mode. Deleting before forkflush completes has risk of corruption.

---

I have also lowered some logging levels from `debug` to `maxdebug`;

Per-request and per-I/O log lines (AFP command dispatch, DSI request/reply framing, socket buffer mechanics, and read/write op traces) were logged at debug, drowning out control-plane fork operations. Ie, the individual logs for data plane transfers were at debug level!
Lowered to `maxdebug` (where they belong), but left other Fork operations (open/flush/sync/close) at `debug`, to aid tracing of remaining locking audit items

andylemin:
@rdmark I am going to squash this lot now 🪨

andylemin:
@rdmark squashed and merged. I could not get it to fail...

And I took the liberty of adding OPEN_RD as non-blocking too following my Spec and Samba research. Please test.

This was a lot of work! The rest of the locking audit series should be a down hill from here.. (even though the best is yet to come - I can't wait to show you what I have ready to go).

And yes, you are right - this all deserves a blog style wiki page, just as the caching audit still does..

PR #3151, opened 2026-07-04
standardize the copyright statement style for user handle

rdmark:
use a consistent style for user handle identifiers that are not email addresses


rdmark:
@andylemin I changed my mind after seeing a bunch of examples in other projects: the @ prefix for your user handle doesn't look good here

this also normalizes the style of your copyright blurbs across the board, mainly for the selfish reason of getting more consistent output from the Debian copyright tool later :)

andylemin (approved):

PR #3152, opened 2026-07-06
dbd: clear AFPVOL_NOV2TOEACONV flag to allow AD to xattr conversion

### LINKED ISSUE #3150: dbd -c broken unless "convert appledouble = yes" is set
Attempting to convert a volume from AppleDouble v2 to EA using "dbd -c" fails. Command will output that it can't remove the .AppleDouble directories and will fail to convert, unless the volume is set up with "convert appledouble = yes". Issue exists with 4.5. Earlier versions always autoconverted v2 files on access and never exhibited the problem.

rdmark:
when afp.conf is set to 'convert appledouble = no' we enable the AFPVOL_NOV2TOEACONV flag, however the dbd tool should always be allowed to convert AD to xattr so clear the flag when running dbd

### DIFF
--- a/bin/dbd/cmd_dbd.c
+++ b/bin/dbd/cmd_dbd.c
@@ -308,6 +308,10 @@ int main(int argc, char **argv)
         exit(EXIT_FAILURE);
     }
 
+    if (flags & DBD_FLAGS_V2TOEA) {
+        vol->v_flags &= ~AFPVOL_NOV2TOEACONV;
+    }
+
     /* Sanity checks to ensure we can touch this volume */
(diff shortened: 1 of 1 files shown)


andylemin:
@rdmark this had quite a performance hit having it always check, so forcing on always for dbd is may be a performance regression.

I think we need to restrict this to just the dbd tooling rather than the whole dbd backend?

rdmark:
@andylemin this has nothing to do with the CNID backend of the same name: the code modified here is only ever executed by the `dbd` command

andylemin (approved):

PR #3154, opened 2026-07-06
afpd: convert FCE modify paths before notify scripts

rdmark:
FCE_FILE_MODIFY was building its event path from dir->d_fullpath plus of_name(), but of_name() is the AFP/Mac-side name. That produced a mixed-encoding path for notify scripts while create, delete, and move events used Unix-side paths.

Convert the open fork name with mtoupath() before composing the close-event path, and reuse that Unix-side name for cache/CNID lookups around the close. Also let pending modify events flush when only 'fce notify script' is configured and no UDP listener exists.

Fixes SourceForge [Bug No.604](https://sourceforge.net/p/netatalk/bugs/604/) reported by Leandro (lfluvisotto)

### DIFF
--- a/etc/afpd/fce_api.c
+++ b/etc/afpd/fce_api.c
@@ -720,7 +720,7 @@ static void check_saved_close_events(const AFPObj *obj)
  */
 void fce_pending_events(const AFPObj *obj)
 {
-    if (!udp_sockets) {
+    if (!udp_sockets && obj->fce_notify_script == NULL) {
         return;
     }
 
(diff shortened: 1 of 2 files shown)


rdmark:
sample test with MacRoman file name in OS9

example afp.conf settings -- note that the executable bit has to be set on fce_ev_script.sh

```ini
[Global]
fce version = 2
fce events = fmod,fdel,ddel,fcre,dcre,fmov,dmov,login,logout
fce holdfmod = 0
fce notify script = /usr/local/libexec/netatalk/fce_ev_script.sh
```

bug reproduced before fix

```
FCE Event: FCE_FILE_MODIFY, protocol: 2, ID: 14, pid: 60246-n , user: atalk1
-n , path: /opt/afpicon1/f?r? copyd.txt
```

with the fix

```
FCE Event: FCE_FILE_MODIFY, protocol: 2, ID: 3, pid: 83083-n , user: atalk1
-n , path: /opt/afpicon1/förå copyd.txt
```

andylemin (approved):

PR #3155, opened 2026-07-06
contrib: harden and install FCE notify script helper

rdmark:
Install the FCE notify script as a libexec helper when FCE support is enabled so downstream packages include it. Harden the helper to log via syslog by default while retaining optional file logging through NETATALK_FCE_LOG_FILE.

Set the executable bit on the FCE notify script since afpd invokes it with /bin/sh when events are emitted.

Document the helper with a new fce_ev_script(8) man page, update afp.conf and fce_listen references, and expand the FCE manual chapter with UDP and script capture examples.


rdmark:
augment review

andylemin (approved):

PR #3163, opened 2026-07-08
ci: use a stuffit-rs stable release tarball

rdmark:
transitions from checking out a particular git hash to downloading a stuffit-rs release tarball to build

at the same time we bump to stuffit-rs 0.1.8 which has a portability fix, allowing it to be built with older versions of rustc

therefore, we now build with StuffIt support in all jobs except for OpenIndiana, where the stuffit-rs Makefile is having portability issues

we also apply the new pattern for other software installed from tarball

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -23,7 +23,7 @@ permissions:
   contents: read
 
 env:
-  STUFFIT_RS_COMMIT: 16336d7abff5c925e8081869bfa0dd6ee27f0658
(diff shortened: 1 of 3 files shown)


rdmark:
this can now be built on all platforms except for OpenIndiana, where the stuffit-rs Makefile somehow is not working; I will look into make it more portable in the upstream project and revisit it here later

for the record, the requisite packages on OpenIndiana are:

- developer/cargo-c 
- developer/lang/rustc

andylemin (approved):

PR #3165, opened 2026-07-09
test: expand afp_logintest with UAM test suite

### LINKED ISSUE #3064: rewire testsuite for a secure auth layer
we want a testsuite that can use any client UAM, and to achieve that I can think of two approaches

- borrow the UAM implementations wholesale from netatalk_client (afpfs-ng fork)
- link with netatalk_client's libafpclient

the former is the quick win and least disruptive, but we end up with two copies of the client auth implementations; the latter is the Right Thing To Do from a software engineering perspective

this is about exploring the latter idea

rdmark:
leverage Netatalk Client's libafpclient shared library to build a test suite covering basic success and failure cases of each password based UAM

afp_logintest now runs a matrix that checks each UAM against libafpclient. For each selected libafpclient UAM, the test checks whether the UAM is known to libafpclient, whether the server advertises it when that can be discovered,
whether login succeeds with the supplied credentials, whether login fails with an intentionally wrong password for credentialed UAMs, and whether a tiny authenticated AFP smoke request succeeds after login.

the legacy tests in afp_logintest have been refactored, cleaned up, and renamed for this context; when libafpclient is not available, only the legacy tests will be built

expand the logintest jobs in the CI workflow to cover all supported AFP versions with appropriate UAMs for that version; however, removing the AFP 2.1 job because Netatalk does not in fact advertise AFP 2.1 support, which libafpclient checks for and throws an error for the mismatch

consistently use the 'ADD' containerfile keyword instead of curl

### DIFF
--- a/.github/workflows/containers.yml
+++ b/.github/workflows/containers.yml
@@ -1032,14 +1032,14 @@ jobs:
             -e AFP_USER="atalk1" \
             -e AFP_PASS="$CONT_PASSWD" \
             -e AFP_GROUP="afpusers" \
-            -e AFP_UAMS="uams_guest.so uams_clrtxt.so uams_dhx.so" \
+            -e AFP_UAMS="uams_guest.so uams_clrtxt.so uams_randnum.so uams_dhx.so uams_dhx2.so uams_srp.so" \
             -e TESTSUITE="login" \
             -e VERBOSE="1" \
             -e AFP_VERSION="7" \
             ${{ env.REGISTRY }}/netatalk/netatalk-testsuite:${{ github.sha }}
 
-  afp-logintest-afp21:
-    name: AFP login test AFP 2.1 - Alpine
+  afp-logintest-afp33:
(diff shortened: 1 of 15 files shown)


rdmark:
@andylemin this is ready for your review now! we only test the UAM happy path so far but I think it's a good start to get some coverage, we can expand it with tailored tests later: password length boundary, password changing, multi stage auth interruptions, etc.

I consider the current PR a proof-of-concept: after this I plan to tackle encrypted auth for the spectest. when I did the experiment a while ago, I realized that the spectest contains two classes of tests: those relying on strictly "normal" AFP operations, and those that cheat the system by using quirks of the libafptest library to create "weird" DSI or AFP requests. since libafpclient is an AFP library that plays by the rules, I think we need to create physical separation between straight laced and quirky spectests, to make it manageable in the long run.

thoughts?

andylemin (approved):

PR #3166, opened 2026-07-09
ci: bump to v0.2.0 of the stuffit-rs library

### LINKED ISSUE #3164: nad unsit function issues.
**Timestamps not being preserved**

Files extracted from Stuff-It archives are not retaining their timestamps. The timestamp from the archive should be applied at both the filesystem level and in Netatalk's AppleDouble metadata.

**Files with no/empty resource forks are having ._ sidecars created with `ad = ea` backend**

If a file does not have a resource fork, `nad unsit` is creating ._ files that consist of only an AppleDouble header with FinderInfo. This is not default or desirable behavior. Files lacking resource forks don't have ._ files associated with them on Netatalk shares. Suggest checking the length of the resource fork on a file. If it is zero (or the sidecar is 82 bytes), don't create the sidecar file.

rdmark:
the upstream project has implemented support for StuffIt time stamps, so bumping the library version for this purpose

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -23,7 +23,7 @@ permissions:
   contents: read
 
 env:
-  STUFFIT_RS_TARBALL_URL: https://github.com/benletchford/stuffit-rs/archive/refs/tags/v0.1.8.tar.gz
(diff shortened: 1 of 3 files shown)


rdmark:
superseded by https://github.com/Netatalk/netatalk/pull/3167

PR #3168, opened 2026-07-10
nad: avoid empty StuffIt resource sidecars

rdmark:
When unsit extracts an entry with no resource-fork data, do not open the resource fork in the AppleDouble backend and skip the resource-fork write. Opening the RF path is enough for the EA backend to flush an AppleDouble sidecar containing only header/FinderInfo bytes, which creates unwanted ._ files for ordinary data-only files.

Add an internal OPTION_NORF mode so StuffIt extraction can preserve data and metadata without creating zero-length resource fork storage.


NJRoadfan:
This appears to be working here.

andylemin (approved):

PR #3175, opened 2026-07-11
afpd: harden copyfile held-handle use; fix ad_write rlen-on-error; rfork-cache guard/log

andylemin:
Three small, independent correctness/robustness changes in the fork/adouble paths. Each is self-contained and could stand alone.

## `copyfile` — reuse a held handle, don't open/close onto it (hardening)

When the copy source is already open by the session, `of_ad()` returns the live `of->of_ad`, and `afp_copyfile()`/`copyfile()` then run `ad_open`/`ad_close` cycles **on that handle**. The refcount arithmetic happens to balance (verified on an `AD_VERSION2` volume — the shared count returns to zero cleanly, no `AFP_ASSERT`), so this does not crash today, but mutating a client-owned handle is fragile: a transient `ad_close` can drop the holder's process-owned POSIX advisory locks, and it couples copyfile to `ad_close`'s internal accounting.

Fix: reuse the held fds and never `ad_open`/`ad_close` `of->of_ad`, mirroring the rule `of_get_locks()` already follows. The destination fork-set is discovered from the on-disk resource fork via a private probe adouble (with the `HAVE_EAFD && SOLARIS` guard, where a private rfork open would strand the held locks). A held rfork-only source with no data fd is refused rather than opened transiently. Adds `FPCopyFile:test425`, the only test that exercises the held-source path.

## `ad_write` — don't fold a write error into `ad_rlen`

The `ADEID_RFORK` branch updated `ad_rlen` from `adf_pwrite()`'s return **before** checking it, so a failed rfork write (`ENOSPC`/`EDQUOT`/`EIO`) set `ad_rlen` to a length for bytes never written — which `ad_flush()` then persists. Now checks `cc < 0` first and leaves `ad_rlen` untouched on error. Backend-independent.

## `fork.c` — rfork-cache serve guard + log

`rfork_cache_serve_from_buf()` now rejects a NULL cache buffer before the `memcpy` (defense-in-depth against a broken invariant) and logs served-from-cache hits at `debug`, matching the AD-cache serve path.

## Testing

Full `testsuite_alp` container build; `ea:ad` spectest **345 passed, 0 failed** (`test425` green). Compiles clean, no new warnings.

### DIFF
--- a/etc/afpd/directory.c
+++ b/etc/afpd/directory.c
@@ -288,7 +288,7 @@ static int copydir(struct vol *vol, struct dir *ddir, int dirfd, char *src,
                     goto copydir_done;
                 }
             } else if (AFP_OK != (err = copyfile(vol, vol, ddir, dirfd, spath, dpath, NULL,
-                                                 NULL))) {
+                                                 NULL, 0))) {
                 goto copydir_done;
             } else {
                 /* keep the same time stamp. */
--- a/etc/afpd/file.c
+++ b/etc/afpd/file.c
@@ -1656,7 +1656,7 @@ int renamefile(struct vol *vol, struct dir *ddir, int sdir_fd, char *src,
             }
 
             if (AFP_OK != (rc = copyfile(vol, vol, ddir, sdir_fd, src, dst, newname,
-                                         NULL))) {
+                                         NULL, 0))) {
                 /* on error copyfile delete dest */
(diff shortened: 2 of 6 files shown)


andylemin:
@rdmark ready for review

rdmark (approved):
nice fixes!

PR #3176, opened 2026-07-11
testsuite: authenticate spectest T1/T2 with libafpclient

rdmark:
Replace afp_spectest’s single legacy-DSI connection model with explicit per-testset tier, backend, and precondition metadata. T1 and T2 now authenticate through libafpclient, while tests that depend on socket ownership, partial packets, malformed framing, or unusual session lifecycle remain on the legacy DSI backend as T3.

Add a libafpclient transport adapter behind the existing CONN/DSI test interface. The adapter establishes sessions with a selected UAM, dispatches the existing hand-built AFP payloads through libafpclient, queues replies for the legacy receive helpers, supports split DSI writes and large Read/ReadExt replies, and mirrors result codes and payloads into the buffers inspected by existing tests. Add a native libafpclient smoke path covering volume lookup, create/open, read/write, cleanup, and logout.

Teach the spectest registry to expose tier, backend, preconditions, and the reason for retaining legacy DSI in afp_spectest -l. Run contiguous backend groups with separate primary and secondary connections, allow -A to select libafpclient UAMs by short alias or protocol name, and keep T3 authentication fixed to Cleartxt Passwrd so encrypted-UAM runs can still transition to raw protocol coverage.

Split the partial-response deadlock cases out of FPRead into T3FPRead, rename the malformed-request and sleep/session suites to T3Error and T3FPzzz, and retain FPDisconnectOldSession as raw DSI. Remove obsolete FPGetUserInfo probes from no-access fixture setup so AFP 3.4 does not turn AFPERR_NOOP into a bogus directory ID.

Factor generic libafpclient initialization, discovery, connection, and teardown out of the login UAM tests and keep a logintest-specific wrapper around that shared support. Rename the cleartext selector to clrtxt and update the UAM documentation.

Update Meson dependencies, container entrypoints, documentation, and CI for the new backend model. Container test runners now build quoted argument vectors, accept AFP_SUBTEST and AFP_TESTSUITE_UAM, support very-verbose transport logging, fold readonly coverage into spectest, and distribute representative encrypted UAMs across the spectest jobs.


andylemin:
@rdmark Wow, interesting to see the impact the netatalk-client library has on performance. I would suggest merging #3177 first and rebasing this so we can see better where the performance hits are coming from the most?

This also suggests we need a way to tell afp_speedtest, afp_lantest to use the legacy/direct or new library connect methods?

We will probably need a flamegraph for the client library too..

rdmark:
@andylemin interesting observation -- picking the transport backend dynamically is trivial to accomplish, will make it happen when I get to that stage

I'm leaning towards pulling this back to Draft for a while and prepare a new netatalk-client release with a handful of fixes that should make this cleaner and more portable

andylemin:
I think its a great baseline to be able to compare between the two client methods, the performance dashboard's deltas are quite noticeable - 4 are 200-300% slower, whilst others are not that bad (maybe 50% worse). It shows nicely where the issues are. And the variance is very telling on the slowest ones..

This has been a really great way to test the client at scale rapidly! 🙂

andylemin:
@rdmark ive had an idea!

This should live in the Netatalk-client repo, not the Netatalk repo..

Then we would have both the pure Netatalk baselines in Netatalk CI, and the Netatalk-client + Netatalk baselines in Netatalk-client CI.

This would allow us to see what performance impact belongs to what code.. 

It means we would need replicate the whole Performance Dashboard stack in the Netatalk-client repo.

We would need to add some changes to Netatalk to make it selectable so you can select the Netatalk + Netatalk-client setup in the client repo.

To keep the numbers aligned as both evolve you would need to import the current Netatalk commit in the client repo instead of using a release, but that's fine as Netatalk is stable compared to the client?

What do you think?

rdmark:
the more I think of this idea the more I like it: a simple transport baseline that we control completely in this project, with an emphasis on the performance and correctness of Netatalk proper

this drawback is that we land back at the original functionality gap: what if we want to run these tests against a live Netatalk AFP server that is properly secured, or a late era Apple AFP server, neither which ought to have ClearTxt UAM enabled

two ideas:

1. implement DHX2 and SRP client UAMs in testsuite's libafptest (it's not THAT much code, but a little extra to maintain)
2. duplicate or move wholesale a subset of the spectest suite to Netatalk-client

in fact, the latter could be an opportunity to change the focus of the spectest: Netatalk's testsuite becomes a **Netatalk** testsuite without any Apple AFP server compatibility code (remove the `-m` "Mac" mode)

and Netatalk-client's testsuite takes over the role of an **AFP server** testsuite complete with "Mac" mode

thoughts?

rdmark:
pivoted to a the much less intrusive approach with a local client UAM module that can speak DHCAST128 and DHX2 only, in https://github.com/Netatalk/netatalk/pull/3247

PR #3179, opened 2026-07-12
introduce bstring compatibility header

rdmark:
there is a long-running portability wrinkle in the bstring library in that it declares a 'bgets' symbol in the public API which directly conflicts with the libgen API in Solaris/illumos

https://docs.oracle.com/cd/E86824_01/html/E54772/libgen-3lib.html#REFMAN3Flibgen-3lib

we have worked around this in our own bstring builds by compiling away that symbol on Solaris/illumos

however, we have now encountered a bstring binary package in the wild (SmartOS pkgsrc) which does not have this workaround

therefore, I propose this compatibility header that undefs away the symbol instead, since we do not use it anywhere in our codebase

this also includes a sha256 validation of the SmartOS bootstrap tarball in our OmniOS CI build pipeline

### DIFF
--- a/.github/workflows/spectest-vms.yml
+++ b/.github/workflows/spectest-vms.yml
@@ -28,6 +28,7 @@ env:
   CMARK_TARBALL_URL: https://github.com/commonmark/cmark/archive/refs/tags/0.31.2.tar.gz
   INIPARSER_TARBALL_URL: https://gitlab.com/iniparser/iniparser/-/archive/v4.2.6/iniparser-v4.2.6.tar.gz
   SMARTOS_BOOTSTRAP_URL: https://pkgsrc.smartos.org/packages/SmartOS/bootstrap/bootstrap-2025Q4-x86_64.tar.gz
+  SMARTOS_BOOTSTRAP_SHA256: 4842f72b610d44cf98560253b6bf16ff4056016e5fe2c6932bc195156ac75772
   STUFFIT_RS_TARBALL_URL: https://github.com/benletchford/stuffit-rs/archive/refs/tags/v0.2.0.tar.gz
 
 jobs:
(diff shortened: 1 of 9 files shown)


rdmark:
augment review

andylemin (approved):

rdmark:
bringing this conversation over from https://github.com/Netatalk/netatalk/pull/3180

> Hi @rdmark Personally I quite liked the way we were building our own. As it provides the greatest control: we control the build flags, where we don't for a packaged binary.

this is indeed convenient for us, but not sustainable for downstream distribution

downstream distributions either frown upon or outright disallow vendoring of dependent libraries!

plus there is something reassuring about delegating the packaging of 3rd party libraries to the dedicated community of packagers :)

> But I think we can do both: Keep meson.build:863-867 as-is (system-preferred, vendored fallback). Optionally keep the OmniOS pkgin libbstring — with the header universal, a system package is now safe. Convert all 28 `#include <bstrlib.h>` → `#include <atalk/bstrlib_compat.h>` (your branch does 7; extend to the remaining 21). Optionally add a guard rail so it can't regress: a tiny CI grep that fails if any file under `libatalk/`, `etc/`, `bin/`, `include/` includes `<bstrlib.h>` directly instead of the wrapper.

right, I chose to use the compat header only for source files where there were concrete conflicts with libgen.h through include chains... maybe that was a misguided choice on second thought, it's not transparent to future coders why this is

if we make the vendored bstring the default, then I want an option to opt in to the downstream binary package -- testing against the package is important for CI if nothing else!

> Then: User installs their own bstring → meson picks it up; compat header shields every include site regardless of how their package was built. ✅ User has none → vendored subproject builds (bgets compiled out). ✅ illumos + any bstring → no collision anywhere, not just at 7 files. ✅

you merged the PR as-is in the end, but happy to iterate on this approach moving forward

I also intend to reach out to the SmartOS folks and see if they can redo their package with the workaround

PR #3186, opened 2026-07-15
testsuite: keep write frames within DSI quantum in speedtest

rdmark:
Reserve the DSI and FPWriteExt request overhead when choosing the default remote I/O quantum. Sending a full server_quantum bytes of data made the complete DSIWrite frame exceed the negotiated limit, which the hardened DSI parser correctly rejects.


rdmark:
@andylemin after the DSI hardening, the speedtest started failing -- please validate my assumption and fix!

andylemin (approved):
Nicely spotted! Great fix..

PR #3188, opened 2026-07-16
[4.5] docs: make the manpage generation script portable

rdmark:
the pipefail flag isn't POSIX compatible, so replace it with temporary files to capture cmark failures

### DIFF
--- a/doc/manpages/make_man.sh
+++ b/doc/manpages/make_man.sh
@@ -1,7 +1,5 @@
 #!/bin/sh
 
-set -o pipefail
-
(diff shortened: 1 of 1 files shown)


rdmark:
@andylemin FYI, this came up last minute when I tried to deploy the website: the cloudflare deployment runner uses Ubuntu 24.04 where /bin/sh resolves to dash which does not support the pipefail flag (which seems to be a bash extension)

so I'm hotfixing this to unblock the release (it has turned out to be a pain to work around cmark bugs and quirks)

PR #3195, opened 2026-07-18
distrib: update webmin container to 2.652

rdmark:
rename env variables to clarify their purpose - validating the 'setup repos' shell script specifically


rdmark:
worth noting: the setup repos script seems to be more or less unchanging between releases, so the sha doesn't always have to be updated

andylemin (approved):

PR #3196, opened 2026-07-21
DSI Quantum optimisations, alignments and performance regression fixes

andylemin:
The DSI server quantum governs how much file data a client moves per FPWrite/FPWriteExt request, and — because clients derive their read chunk size from it — per read as well. The recent DSI frame-parsing hardening (CVE-2026-62319 / CVE-2026-62320, commit `cc0a9d16`) correctly rejects any frame whose *total* size exceeds the negotiated quantum. But the server was measuring the wrong thing: it bounded the whole DSI frame (header + request block + data) by the quantum, when the AFP spec defines the quantum as the **data** size only. A spec-legal write of exactly `quantum` data bytes therefore had a frame of `quantum + doff`, which the hardened parser tore down as oversized.

The tools papered over this by subtracting the frame overhead from every transfer (PR #3186, PR #3187), so clients never used the full negotiated quantum in either direction. That restored correctness but gave up throughput — a workaround for a server-side accounting bug, not a real protocol limit.

This PR fixes the accounting at the source, restores full-quantum transfers, aligns the advertised quantum to the connection's TCP MSS so no frame ends in a runt segment, and removes a redundant userspace copy on the write receive path. Along the way it closes an out-of-bounds read in the FPAddIcon write path. **All of the CVE hardening stays in place and is tightened.**

## Root cause

Per the AFP specification, the server request quantum is the size of the write data alone. `DSIOpenSession`, Table 1:

> DSI Option Server Request Quantum … the maximum size of data that the server is prepared to receive as part of a request, **not including the length of the DSI header and the AFP command**.

and `DSIWrite`:

> the amount of data to be written may be up to the size of the server request quantum.

The server instead enforced the quantum against the entire frame. In `dsi_stream_receive` the check `if (dsi_len > dsi->server_quantum)` ran *before* any WRITE/`doff` handling, so for a write frame `dsi_len` was `data + doff` and a full-quantum write (`data == quantum`) was rejected. Because clients size their read requests from the advertised quantum too, the client-side workaround (subtracting frame overhead) degraded **read** throughput as well as write.

Two further defects on the write path:

- **Redundant copy + latent truncation.** `dsi_writeinit` `memmove`d the already-buffered payload prefix out of the read-ahead buffer into the caller's buffer (a second userspace copy of up to the read-ahead size per write), and its `memmove(buf, start, MIN(buflen, bytes))` paired with an untruncated `bytes` return silently dropped data whenever the prefix exceeded the caller's buffer.
- **Client runt segments.** The test client sent each write frame as two syscalls — the header+request block via `dsi_stream_send`, then the payload via `dsi_stream_write` — so with `TCP_NODELAY` a runt segment preceded every payload.

## Model

The server quantum counts FPWrite/FPWriteExt **data** only; the 16-byte DSI header and the request block (12 bytes FPWrite, 20 bytes FPWriteExt) ride on top of it on the wire.

The session quantum is resolved **once per session, immediately after `accept()`** (before any buffer is allocated), from the configured value and the socket's measured MSS:

> the smallest 4 KiB multiple **at or above** the configured quantum whose FPWriteExt frame (data + 36 bytes) ends in a TCP segment at least 7/8 full — growing by at most 256 KiB, capped at 256 MiB (`DSI_SERVQUANT_MAX`). When no such multiple is found within the budget, the best-fill candidate wins; when no usable MSS is detected, the value is page-aligned only.

Growing (rather than rounding down) means the client never gets *less* than the administrator configured. Page alignment keeps the data a 4 KiB multiple; the 7/8-fill threshold is used because exact dual alignment is impossible when `gcd(4096, mss)` does not divide 36 (e.g. MSS 1448 or 8960). Buffers, inbound checks, and the advertised value all use the grown quantum, so **advertised == capacity** and every client frame fits the server's buffers exactly.

MSS is read from `TCP_INFO` (Linux `tcpi_snd_mss`) or `TCP_CONNECTION_INFO` (macOS `tcpi_maxseg`), falling back to `TCP_MAXSEG`; live values already account for active TCP options. Detection self-disables to page-alignment-only where no socket option is available.

## Fixes

**Server receive bound (`libatalk/dsi/dsi_stream.c`).** For `DSIFUNC_WRITE`, the data offset is validated against an exact `{12, 20}` whitelist, then `doff <= dsi_len`, then `dsi_len - doff > server_quantum` bounds the **data**. Non-write commands keep the strict `dsi_len > server_quantum` bound. This restores full-quantum transfers — a client may again send, and request, a complete `server_quantum` of data per frame.

**Zero-copy write handoff (`libatalk/dsi/dsi_write.c`).** `dsi_writeinit` now hands the buffered payload back by reference (`char **`), or drains it when passed `NULL`. The prefix is consumed in place — one copy, no truncation. `afp_addicon`'s error path drains and returns the AFP error instead of falling through and re-arming the transfer. Signature change ⇒ **soversion 20 → 21**.

**Single-writev client (`test/testsuite/afpclient.c`).** `dsi_stream_send3` sends header, request block, and payload in one `writev`, so no runt segment precedes the payload. `speedtest` selects FPWriteExt or legacy FPWrite automatically from the negotiated AFP version (FPWriteExt for AFP ≥ 3.0, matching macOS), and both `speedtest` and `lantest` size transfers from the full advertised quantum — no overhead subtraction, and a guard against a zero (unnegotiated) quantum.

**Config (`libatalk/util/netatalk_conf.c`).** `dsireadbuf` default 12 → 32, capped so `dsireadbuf × quantum` stays under 1 GiB per session (`NETATALK_READBUF_LIMIT`, re-clamped in the child after the quantum grows). Server-quantum values above the maximum now clamp to `DSI_SERVQUANT_MAX` instead of silently falling back to the 1 MiB default.

## Security

The CVE-2026-62319 / CVE-2026-62320 DSI frame-parsing hardening (`cc0a9d16`) **stays fully in place and is strengthened**:

- The write data offset is now validated against an exact `{12, 20}` whitelist — malformed offsets 13–19 are **rejected**, not parsed into a nonsense request block.
- The `doff <= dsi_len` bound and the ASC-3.7 zero-`doff` workaround are preserved.
- `dsi_tcp_open`'s first-frame whole-frame check is unchanged: a hostile full-quantum WRITE as the first frame is still killed there (conforming clients always send `DSIOpenSession` first).
- Nothing is clamped, truncated, or fallen back from — out-of-bounds input is **rejected**, in the server and in the tools.

This **supersedes** the frame-overhead-subtraction workarounds from PR #3186 / PR #3187 by fixing the server semantics they compensated for — restoring full-quantum throughput — while preserving their invariant that the client never emits a frame the hardened parser rejects.

### Additional vector closed — FPAddIcon out-of-bounds read

The old `dsi_writeinit` truncated the buffered prefix into the caller's buffer via `memmove(buf, start, MIN(buflen, bytes))` but returned the **untruncated** byte count. `afp_addicon` then wrote that count from a 64 KiB (`DSI_DATASIZ`) reply buffer — reading **past** the buffer and emitting stale heap bytes into the icon file whenever a buffered write prefix exceeded 64 KiB against the 1 MiB default quantum. The pointer handoff returns the exact buffered range, removing the over-read.

## Behaviour changes

- Full-quantum FPWrite/FPWriteExt is accepted again; transfers use the complete advertised quantum in both directions.
- The advertised quantum is MSS-aligned per session and logged at session open (e.g. `dsi_tcp_open: MSS 32768, session quantum 1077248`).
- `server quantum` maximum is 256 MiB (was 4 GiB); out-of-range configs clamp to the max (above) or the default (below the 32000 minimum).
- `dsireadbuf` default is 32 (max 1024), with the per-session product capped at 1 GiB.
- No on-wire or config change is required of existing deployments; the quantum is renegotiated on every `DSIOpenSession`.

### afpd unit tests (`test/afpd/test.c`)

| test | asserts |
|---|---|
| `utest_dsi_receive_accepts_full_quantum_write` (doff 12 & 20) | a full-quantum write in one record is accepted |
| `utest_dsi_receive_rejects_overquantum_write` (doff 12 & 20) | one data byte past the quantum is rejected |
| `utest_dsi_receive_rejects_overquantum_cmd` | non-write commands keep the strict whole-frame bound |
| `utest_dsi_receive_validates_doff` | `doff` outside `{12, 20}` rejected; `doff==0` maps to 12 |
| `utest_dsi_writeinit_no_duplicate_copy` | buffered payload handed off by reference, not copied |
| `utest_dsi_align_quantum` | growth, page alignment, fill threshold, 256 KiB / 256 MiB caps, and no uint32 overflow at the top of the range |
| `utest_readbuf_clamp` | `dsireadbuf × quantum` bounded, floored at 1 |

The two server bugs landed as RED tests one commit ahead of their fixes; those failing CI runs are preserved in this PR's checks history as the RED evidence, before the branch was squashed.

### Integration

- AFP spectest (single-container loopback): 341 pass / 0 fail; FPWriteExt full-quantum writes round-trip at the grown advertised size.
- `afp_speedtest` and `afp_lantest` complete against the built server, exercising both FPWriteExt (AFP ≥ 3.0) and legacy FPWrite (older AFP versions) via the negotiated version.
- The speedtest dashboard gains **avg mean** and **avg max** columns across the file-size sweep, so per-round-trip regressions on small transfers are visible alongside large-file peaks.

- Backport of the throughput regression fix to the stable 4.5 branch.
- Widening the legacy-mode 32-bit offset/size loop variables in `lantest` (pre-existing, out of scope).

---

Quantum/MSS scaling validation;
packet-size-compare


test/afpd/test.c
```diff
@@ -95,8 +94,7 @@ static int dsi_test_receive(uint8_t command, uint32_t code_or_doff,
     }
 
     dsi->socket = fds[1];
-    dsi->server_quantum = 16;
-    quantum = dsi->server_quantum;
+    dsi->server_quantum = quantum;
```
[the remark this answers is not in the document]
andylemin: Fixed in e9acab53: the harness quantum parameter is now `uint32_t`, matching `dsi->server_quantum`, so the parameter, the enforced limit, and the allocated buffer sizes cannot diverge.

test/afpd/test.c
```diff
@@ -139,13 +137,52 @@ static void dsi_test_cleanup(DSI *dsi)
...
+/* dsi_len = quantum + doff is spec-legal: the quantum bounds write data,
+ * not the whole frame.  Only the header drives accept/reject, so the
+ * harness sends just doff payload bytes; data is consumed by dsi_write. */
+static int utest_dsi_receive_accepts_full_quantum_write(uint32_t doff)
+{
+    DSI dsi;
+    uint8_t cmd[DSI_WROFF_FPWRITEEXT];
+    int ret;
+    memset(cmd, 0, sizeof(cmd));
+    ret = dsi_test_receive(DSIFUNC_WRITE, doff, 4096 + doff, cmd, doff,
```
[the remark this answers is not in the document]
andylemin: Correct reading, and it is intentional: this PR follows a RED-test-first commit sequence. This commit pins the spec-correct semantics (the server quantum bounds FPWrite/FPWriteExt *data*, not the whole frame - AFP spec, DSIOpenSession Table 1 and DSIWrite) and deliberately fails against the current code; the CI run on this commit is the recorded RED evidence. The receive-side fix (bounding `dsi_len - doff` for writes, with strict doff validation) lands in a subsequent commit on this PR, after which the suite goes green. The branch will be squashed before merge per the repo's rebase-only policy.

andylemin:
@rdmark Performance fully recovered, security patches maintained and one additional vector closed.
All VFS/DSI/NET layers now aligned..

This is not ready for merge - I am creating a new performance test rig which runs the speedtest and lantest tests;
- Run inside the existing spectest image under docker: afpd starts normally in the container's root namespace; the script creates a client namespace, joins them with a veth pair, and shapes both directions with tc netem.
- Network speed is configurable via env: RATE (e.g. 100mbit, 1gbit, 10gbit), DELAY (one-way; RTT is 2×), MTU (1500/9000), plus SIZE/ITER/MODE (FPWriteExt vs -E legacy vs both).
- Reports speedtest throughput per run, a mid-run ss -ti sample (connection MSS + segs_out), and the client namespace's total TCP OutSegs delta per run.

Ie, the purpose is to provide a static network test environment, rather than the CPU bound loopback based one the performance runners currently use. This will show performance on different real networks much more consistently, and will show the number of packets exchanged, exposing any runt and alignment issues.

We need to merge that PR before this one, so we have a before/after this PR for a true concrete comparison?

rdmark:
impressive work, this turned out to be a quite big changeset! do you think it's feasible to backport the performance regression to the stable 4.5 branch? I think it's worth putting out another bugfix release for this.

it makes sense to hold off on merging this while you introduce new testing capabilities to properly prove the improvement

andylemin:
> impressive work, this turned out to be a quite big changeset! do you think it's feasible to backport the performance regression to the stable 4.5 branch? I think it's worth putting out another bugfix release for this.

The key parts yes (byte accounting and the extra security fix). This includes several changes which do not need to be back-ported (such as the MSS alignment, which adjusts the quantum value to a figure which is divisible by the MSS, mitigating the runt packet at the end of every single DSI record).

> it makes sense to hold off on merging this while you introduce new testing capabilities to properly prove the improvement

Absolutely, This one needs to be grounded in evidence as there is a lot of hard theory work here.

andylemin:
**Performance BEFORE quantum security hardening;**
pr-3180

**Performance AFTER quantum security hardening (more secure but performance regression);**
pr-3193

All test files sizes bellow 8-16MiB have performance loss, as every test payload is a multiple of KiB/MiB, however the server DSI quantum became un-aligned during the security fixes, resulting in a runt DSI record for every single DSI record stream.
1MiB = 2 DSI Records
2MiB = 3 DSI Records
4MiB = 5 DSI Records
8MiB = 9 DSI Records (runt overhead (1/8) begins to diminish)

**Performance AFTER quantum security hardening and AFTER this fix**
pr-3196

After the fix, there are no runt DSI records at the end of each DSI stream, and there are no runt network packets at the end of every DSI record

andylemin:
@rdmark It looks bigger than it is due to the number of commits - mostly unittests due to the critical nature of the code. I added tests first (to ensure they failed) and then confirm they turned green after adding the respective fixes.

Once it is squashed it will be smaller and easier to review.

I have just modified the quantum alignment algorithm to scale up (instead of down) to the nearest value which is divisible by 4KiB and aligns the DSI record size with the connections negotiated MSS ensure no runt packets at the end of each record. So there is some churn which will vanish

andylemin:
augment review

test/afpd/test.c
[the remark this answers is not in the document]
andylemin: Addressed: the parameter is now `uint32_t quantum`, matching `dsi->server_quantum` — no truncation possible.

test/afpd/test.c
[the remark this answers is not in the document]
andylemin: Stale: the receive-side semantics are updated in this PR (commit "dsi: bound the server quantum against write data, not the whole frame"). For DSIFUNC_WRITE, dsi_stream_receive now checks `dsi_len - dsi_doff > server_quantum` after validating doff, so `dsi_len = quantum + doff` is accepted by design; non-write commands keep the strict `dsi_len > server_quantum` bound. The tests were pushed one commit ahead of the fix as intentional RED evidence.

etc/afpd/afp_dsi.c
```diff
@@ -1091,7 +1091,10 @@ void afp_over_dsi(AFPObj *obj)
          * we need to flush the rest of the packet if necessary. */
         default:
             LOG(log_info, logtype_afpd, "afp_dsi: spurious command %d", cmd);
-            dsi_writeinit(dsi, dsi->data, DSI_DATASIZ);
+            {
+                char *wbuf;
```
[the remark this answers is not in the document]
andylemin: Addressed: `dsi_writeinit` now accepts a NULL bufp for drain-only callers, and the flush sites in afp_dsi.c, fork.c, and desktop.c pass NULL — no unused locals.

test/testsuite/afpclient.c
```diff
@@ -1739,7 +1802,7 @@ unsigned int  AFPCreateFile(CONN *conn, uint16_t vol, char type, int did,
 }
 
 /* ------------------------------- */
-unsigned int AFPWriteHeader(DSI *dsi, uint16_t fork, int offset, int size,
+unsigned int AFPWriteHeader(DSI *dsi, uint16_t fork, off_t offset, size_t size,
```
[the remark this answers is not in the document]
andylemin: The truncation is intentional: legacy FPWrite/FPRead have 32-bit wire fields, and these helpers are only used for AFP < 3.0 or with the explicit `-E` legacy flag. The casts are now explicit (`htonl((uint32_t)offset)`) to state that. Tests needing 64-bit offsets use the FPWriteExt path, which serializes full off_t values.

andylemin:
@rdmark Ready for review.

Can be merged in any order as I have already tested using 3203 before and after this - all validated and ready to go

Interesting comparing the speedtests; "Performance BEFORE quantum security hardening", "Performance AFTER quantum security hardening, and the last Performance Dashboard 🙂

andylemin:
Closed accidentally. reopened

andylemin:
Hi @rdmark I'm keen to get this merged as I have some other changes on the go? 🙏

rdmark:
absolutely, is it okay if I take care of it after work today? (in about 4 hours)

andylemin:
> absolutely, is it okay if I take care of it after work today? (in about 4 hours)

Of course! Always.. No need to ask 🙂 Feel free to merge yourself. This has been extensively tested and is running in my live environment as well

doc/manpages/man5/afp.conf.5.md
```diff
@@ -438,22 +410,43 @@ which IP address to advertise, therefore the hostname is resolved to an
...
 
-max connections = *number* **(G)**
+zeroconf = *BOOLEAN* (default: *yes*) **(G)**
 
-> Sets the maximum number of clients that can simultaneously connect to
-the server (default is 200).
+> Whether to use automatic Zeroconf service
+registration if Avahi or mDNSResponder were compiled in.
+
+### Transfer Tuning
```
rdmark: can we make this a level 2 heading with `##`? roff man pages only have styles for h1 and h2.

libatalk/util/netatalk_conf.c
```diff
@@ -2967,11 +2996,19 @@ int afp_config_parse(AFPObj *AFPObj, char *processname)
     options->timeout        = getoption_int(config, INISEC_GLOBAL, "timeout",
                                             NULL, 4);
     options->dsireadbuf     = getoption_int(config, INISEC_GLOBAL, "dsireadbuf",
-                                            NULL, 12);
+                                            NULL, 32);
```
rdmark, line 2999: this new default needs to be reflected in the webmin module code as well, i.e. contrib/webmin_module/netatalk-lib.pl

test/testsuite/speedtest.c
```diff
@@ -1840,6 +1825,8 @@ void usage(char *av0)
     fprintf(stdout, "\t-w\tpassword\n");
     fprintf(stdout, "\t-c\toutput in CSV format (auto-enables statistics)\n");
     fprintf(stdout, "\t-L\tuse posix calls (default AFP calls)\n");
+    fprintf(stdout,
+            "\t-E\tuse legacy FPWrite instead of FPWriteExt (AFP >= 3.0)\n");
```
rdmark: can't we detect which method to use based on the AFP protocol version currently used? or do you have a reason for wanting to set it manually for testing?

andylemin: That's a very good point! I was using this to make testing easier, but I wonder if the server should even go as far as rejecting the wrong write mode based on the AFP Version?

libatalk/util/netatalk_conf.c
andylemin: I always forget

doc/manpages/man5/afp.conf.5.md
andylemin: Sure, good to know

andylemin: Done — changed to `## Transfer Tuning` in 3669c62c. (It was also the only h3 in the file; every other grouping under *Explanation of Global Parameters* is h2, so this matches the surrounding structure too.)

libatalk/util/netatalk_conf.c
andylemin: Good catch — updated `dsireadbuf` to 32 in `contrib/webmin_module/netatalk-lib.pl` (3669c62c). The other new/changed option, `server quantum`, is already carried by the webmin module (default 1048576), so no further webmin changes are needed.

test/testsuite/speedtest.c
andylemin: Agreed — there's no good reason to set it manually. Removed the `-E` flag entirely in 3669c62c; `speedtest` now picks the write method purely from the negotiated AFP version — FPWriteExt for AFP ≥ 3.0, legacy FPWrite below that:

```c
/* FPWriteExt was introduced in AFP 3.0; older versions use FPWrite. */
if (!Local && Version >= 30) {
    VFS.writeheader = FPWriteExtHeader;
    VFS.writefooter = FPWriteExtFooter;
}
```

To be clear about scope: this is a client-side (testing tool) change only. The **server still accepts both FPWrite and FPWriteExt regardless of the negotiated version** — we're not locking that down. The stale `-E` mentions in the commit message and PR description are updated too.

andylemin:
augment review

test/testsuite/lantest.c
```diff
@@ -1006,7 +1005,13 @@ void run_test(const int32_t dir)
...
     struct async_io_req air;
     static char temp[MAXPATHLEN];
-    numrw = rwsize / (dsi->server_quantum - FPWRITE_RQST_SIZE);
+
+    if (dsi->server_quantum == 0) {
+        fprintf(stderr, "no server quantum negotiated\n");
+        clean_exit(ERROR_NETWORK_PROTOCOL);
+    }
+
+    numrw = rwsize / dsi->server_quantum;
```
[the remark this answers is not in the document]
andylemin: Can never happen as tests params are managed

andylemin:
@rdmark Updated. Ready for review 🙏

rdmark (approved):
let's ship it!

PR #3209, opened 2026-07-25
meson: standardize boolean option naming

rdmark:
Centralize Boolean build options near the top of the root build file and name them according to intent, detected availability, selected implementation, build target, or installation action.

Update all Meson consumers, distinguish docs-only DBD documentation selection from the dependency-backed backend choice, and document the naming convention in CONTRIBUTING.md.


rdmark:
@andylemin the original trigger for this change was that the have_fce flag hadn't been declared in scope when using the `-Dwith-docs-only` build flag -- to avoid the inherent weakness in meson variable scopes I have centralized all feature flags near the top of the build file!

andylemin (approved):
Much more clear 🙂

PR #3220, opened 2026-07-28
distrib: update Webmin version to 2.653 in Dockerfile

rdmark:
bumping to the latest release version from the Webmin project


rdmark:
@andylemin heads-up, I filed https://github.com/webmin/webmin/issues/2803 with the upstream project to ask if they'd consider a longer retention period for historical deb packages

it is becoming a significant manual overhead to constantly keep up to date with the latest versions, so we may have to come up with a better scheme

andylemin (approved):

PR #3223, opened 2026-08-03
Samba interop: ea = samba defaults the coherency settings; strict boolean config parsing; afp read locks renamed strict locking; fail-closed volume config loader

andylemin:
Sharing a volume between AFP and SMB required four coordinated afp.conf settings, one of which (`afp read locks`) is misnamed, one of which (`ea` in `[Global]`) was silently ignored, and none of which failed loudly when misconfigured. This PR makes `ea = samba` the single switch: it selects the Samba-compatible metadata format (unchanged) **and defaults every related setting to the coherency-safe value** — explicit settings always win, with a logged warning when they weaken coherency. Underneath it, boolean config parsing becomes strict and diagnosed tree-wide, and the volume config loader stops parsing unlocked under lock contention.

Four commits, reviewable independently:

1. `libatalk:` one strict boolean validator for every boolean option
2. `afpd:` `ea` becomes `(G)/(V)` (the five string options that hand-rolled that fallback now share one helper); `ea = samba` defaults the Samba-coherency settings; `afp read locks` → `strict locking`
3. `libatalk:` fail closed when the afp.conf read lock cannot be taken; `load_volumes()` → `load_afp_conf_vols()`
4. `test:` end-to-end netatalk-client ⇄ samba-client interop CI leg, plus the container/CI variant renames

## Root cause

Three independent defects, one shared theme: config handling that silently does something other than what the operator wrote.

**Boolean parsing was first-letter matching.** Every boolean option funnels through iniparser's `iniparser_getboolean()`, which classifies by first character only: `y/Y/t/T/1` is true, `n/N/f/F/0` is false, and **everything else is silently treated as absent**. So `ture` parsed as true, `never` parsed as false, and `enabled`/`maybe` vanished without a log line. iniparser is a system dependency, so the fix sits on netatalk's side: all boolean reads already funnel through two wrappers, so one strict validator (`conf_parse_bool()`) rewired inside them upgrades every option with zero call-site churn.

**`ea` was volume-only and `afp read locks` gates writes too.** The `ea` parse never read `[Global]`, so every volume had to repeat it. And `OPTION_AFP_READ_LOCK` has gated **both** the per-read `F_RDLCK` and the per-write `F_WRLCK` since 2012 — the name describes half of what it does. It is also the option Samba interop hinges on; Samba's name for the same concept is `strict locking`.

**The volume loader's lock-retry guard was dead code.** `retries++` runs *before* the `if (!retries)` test, so the fail path is unreachable and after two failed lock attempts **the config is parsed with no lock held**. The `open()` return was also unchecked (fd −1 → every `read_lock` fails `EBADF` → same unlocked parse). A concurrent writer holding `F_WRLCK` let afpd parse a half-written config.

## Provenance

There is no bug to blame here — Samba interop was simply incomplete, and this completes the story the original commits started:

- **`904b58ca` (2011)** — the lock retry is born as a *countdown*, correctly fail-closed. **`374699dc` (2012)** — the conf rework flips it to count-up but keeps `if (!retries)` verbatim; the guard has been dead ever since, silently.
- **`8fa8b411` (2012-04)** — `afp read locks` is born gating the per-read lock; the name is accurate. **`f691d045` (2012-05)** — nine days later the write-path lock is gated behind the same flag; sensible, but the name is a misnomer from this commit on.
- **`2a1f4d31` (2015)** — `ea = samba` is born as a pure storage format (trailing NUL so `vfs_streams_xattr` reads netatalk's xattrs). The locking/caching settings safe concurrent access requires were left to documentation.
- iniparser's first-letter matching predates all of it — library behaviour netatalk inherited.

## What `ea = samba` now does

| coherency dimension | before | after, with only `ea = samba` |
|---|---|---|
| Apple metadata Samba can read/write | `ea = samba` per volume (`[Global]` ignored) | once in `[Global]`, inheritable, per-volume override |
| AFP I/O conflicts with Samba's locks | `afp read locks = yes` | **defaults on** (`strict locking`) |
| afpd notices Samba's file changes | `dircache validation freq = 1` | **defaults to 1** |
| cached rforks can't go stale behind Samba | not expressible | **volume excluded from rfork cache** unless budget set explicitly |
| samba volume on a no-EA filesystem | silent downgrade to AppleDouble v2 — metadata Samba cannot read | **volume refuses to load**, `log_error` names it |

**Defaults, not forced values.** An explicit setting always wins; each explicit setting that weakens coherency logs a verbose warning naming the risk, and each applied default logs a note. Explicitness is recorded at parse time in three new `struct afp_options` markers, robust to future compiled-default changes. `strict locking` is process-global while `ea` is per-volume, so any samba volume flips the process defaults (strict locking only *adds* spec-conforming locks on non-samba volumes; the rfork exclusion is per-volume).

**Boolean values, tree-wide:** everything documented (`yes/no`, `true/false`, `1/0`, any case) parses as before; the single-letter shorthands `y/n`/`t/f` stay accepted as whole words; `on/off`/`enabled/disabled` are newly accepted; anything else — including the old prefix accidents `ture`→true, `never`→false — is warned once at parse time (naming the option, the value, and the accepted spellings) and takes the option's default. Empty values stay silently unset, as generated configs emit them.

## Defect → fix

| # | defect | fix |
|---|---|---|
| 1 | booleans classified by first letter; garbled values silently dropped | `conf_parse_bool()` — whole-word, warn-and-ignore on invalid; rewired inside both wrappers (39 call sites, zero churn) |
| 2 | `[Global] ea` silently ignored | `vdgoption_str()`: volume → preset → `[Global]` → auto-detect; unknown values warn |
| 3 | `afp read locks` gates read **and** write locks | renamed `strict locking` (Samba's name); old key a deprecated alias with a warning, new key wins; tri-state read so an explicit `no` survives the samba default |
| 4 | four hand-aligned settings for Samba coherency | `apply_samba_defaults()` in `creatvol()`'s commit block — after the volume can no longer fail, so a failed volume never mutates process state |
| 5 | samba volume without EAs silently serves AppleDouble v2 | hard failure after `check_ea_support()`; probe warning now carries `errno` |
| 6 | rfork cache can serve bytes Samba changed | samba volumes excluded at the cache gate unless the budget is explicit |
| 7 | dircache freq copied to a static before volumes load | `should_validate_cache_entry()` reads the live option (1–100 clamp) |
| 8 | dead lock-retry guard → silent unlocked config parse | open+lock first, fail-closed on either; `errno` captured before `unbecome_root()` |
| 9 | failure mid-reload strands consumed mtime + `v_deleted` marks | the reorder is the fix: failing before either mutation leaves reload state untouched — next call retries |

## Behaviour changes

- A `[Global] ea =` value that previous releases silently ignored now takes effect (per-volume/preset still wins). The one config shape whose behaviour changes on upgrade; NEWS calls it out.
- Invalid boolean values warn and take the default; the undocumented prefix accidents no longer parse.
- An `ea = samba` volume whose filesystem cannot store EAs fails to load. Deliberately stricter than `ea = sys` (which downgrades): `sys` is a storage preference, `samba` is an interop contract.
- A transient `open()` failure on afp.conf previously fell through to an unlocked parse by path; it now fails closed and retries next call. afpd callers ignore the return, so degraded mode is "serve the previous volume list" — never an unlocked parse.
- Container/CI: `AFP_STRICT_LOCKING` is the canonical env var (`AFP_READ_LOCKS` deprecated fallback with a start-up note); the CI job is renamed accordingly. `OPTION_AFP_READ_LOCK` is removed without an alias — this cycle already changes the installed-header ABI (soversion 21→22, since 4.5.2 shipped 21), so out-of-tree consumers rebuild regardless.
- Webmin reads/writes the new key; a config from an older Webmin shows its `afp read locks` value and one save migrates the file.

## End to end Netatalk-Client to/from Samba-Client full chain testing

New CI job **`samba-client and netatalk-client interop (ea:samba) - Debian`** exercises the complete dual-protocol chain, both directions, from the client layer down:

```
Samba-Client (kernel CIFS mount + smbclient)
  — SMB — Samba (vfs_fruit + streams_xattr)
    — Disk (one shared directory, ea = samba format)
  — AFP — Netatalk (bare `ea = samba` volume, defaults only)
Netatalk-Client (FUSE mount, built from the latest github.com/Netatalk/netatalk-client release)
```

The leg validates at three points: the **Samba client mount** (filesystem operations trigger SMB actions), the **shared disk** (on-disk format and lock-band state), and the **netatalk-client mount** (filesystem operations trigger AFP actions). Checks are plain file operations across the two mounts, asserting changes through one client are **reflected** — or **blocked** — through the other. Every setting `ea = samba` defaults is exercised, each blocking check paired with its release-side control.

Three AFP-side checks cannot go through the FUSE mount and use a wire-level AFP client (`afparg`, same container, same protocol) instead — a netatalk-client capability gap, not a shortcut (verified against 0.9.5 source; tracked in Netatalk/netatalk-client#310):

- **byte-range locks**: the client registers no FUSE `.lock` handler, so a `fcntl()` lock on the AFP mount is satisfied locally and never reaches the server — itself a coherency finding for dual-exported shares;
- **deny-mode opens**: standard POSIX `open()` on linux has no deny concept and the client maps none for any OS including macOS clients;
- **DeleteInhibit**: no filesystem operation reaches the Finder attribute bits (these are Finder operations made against Metadata).

When netatalk-client#310's lock support lands, the strict-locking mirror checks can switch to plain mount-point locks, completing the full user-visible chain for that dimension.

| `ea = samba` default | client-chain proof |
|---|---|
| `strict locking = yes` | same equivalent operations both directions: a byte-range lock held through the CIFS mount blocks an AFP-mount write while held, and an AFP byte-range lock blocks a CIFS-mount write while held — each with its release-side control |
| deny modes (`fruit:locking = netatalk`) | same equivalent operations both directions: an AFP deny-mode open plants the on-disk band (verified in `/proc/locks`) and draws `NT_STATUS_SHARING_VIOLATION` for an SMB write until the fork closes; the SMB→AFP direction is a known Samba-side gap (below) |
| `dircache validation freq = 1` | names created/deleted through either mount are immediately listed/gone through the other |
| rfork cache exclusion | an EA already read once through either protocol re-reads fresh after the other overwrites it |
| AFP DeleteInhibit | set and cleared over AFP with an SMB-delete probe; the refusal is a known Samba-side gap (below) |
| `solaris share reservations` | skipped on Linux (F_SHARE is Solaris/illumos-only) |

Plus content and EA/ADS round-trips both directions, the on-disk trailing-NUL assertion, and a control proving the behaviour comes from the **defaults** (afp.conf carries only `ea = samba`; the afpd log shows the default applied). `afparg` gained scriptable `FPWrite`/`FPRead`/`FPSetEA`/`FPGetEA` one-shots plus `FPSetInhibit` and a timed-hold mode for `FPLockrw` (a signal-killed AFP session parks disconnected and keeps its share modes, so scripted holders must close cleanly).

## Testing

- **Unit (76/76 in the container; 10 new `utest_conf_*`):** the boolean table (28 cases); the `ea` fallback chain; strict-locking key/alias precedence and explicitness; samba defaults applied/explicit-wins/no-samba regression; the planned performance defaults (freq 100, rfork cache on) simulated and reverted by `ea = samba`; the EA hard-fail; defaults-not-leaked-on-failed-volume; the fail-closed loader under a forked `F_WRLCK` holder, with state-neutrality proven by the post-release reload.
- **Wire:** the AFP spectest passes in three configurations — strict locking enabled, stock defaults, and AppleDouble v2 metadata — alongside the client-chain interop job above. The full PR CI matrix is green, including the fuzzing build.

## New Discovered Issues

The client-chain leg runs the same equivalent operation through each protocol and compares outcomes. Three Samba-side gaps surfaced; each is asserted in the leg with the correct expected semantics and reported as `FAILED (Known)` without failing the job, so a Samba-side fix flips the line to PASS and flags the marker for removal. Tracked in #3229:

1. **Kernel CIFS client bypasses the share-mode check that smbclient trips.** With an AFP deny-read/write open held (band verified present on disk via `/proc/locks`), a conflicting write via smbclient is correctly refused `NT_STATUS_SHARING_VIOLATION` — but the identical write through a kernel CIFS mount (oplocks/leases disabled server-side, `actimeo=0,closetimeo=0` client-side) succeeds and lands on disk.
2. **SMB opens are invisible to AFP deny modes.** The symmetric direction: a read-write open held through the CIFS mount plants **no** share-mode band on disk, and a conflicting AFP deny-read/write open succeeds. `fruit_check_access()` in vfs_fruit reads netatalk's band and plants its own markers only for opens it evaluates with read access — in this configuration no band entry appeared for SMB opens at all, so AFP↔SMB deny-mode protection is currently one-directional (AFP→SMB via smbclient only).
3. **SMB deletes ignore AFP DeleteInhibit.** With `kFPDeleteInhibitBit` set over AFP (netatalk refuses cross-session AFP deletes with `AFPERR_OLOCK`), an SMB delete removes the file. The bit lives in netatalk's metadata EA, which vfs_fruit reads for FinderInfo but does not consult for delete operations.

- Unknown-*key* detection (`strict lokcing = yes` is silently stored and never read) — different mechanism, own change.
- `ea = sys ⇄ samba` on-disk migration tool (per-xattr NUL append/strip + volume marker) — home to be decided now that dbd is being deprecated.

---

Where each option's effective value now comes from, and where the samba defaults do — and do not — apply:

```mermaid
flowchart TD
    classDef conf   fill:#1e3a8a,stroke:#1e40af,color:#eff6ff,stroke-width:1px;
    classDef parse  fill:#5b21b6,stroke:#6d28d9,color:#f5f3ff,stroke-width:1px;
    classDef vol    fill:#0f766e,stroke:#0d9488,color:#f0fdfa,stroke-width:1px;
    classDef gate   fill:#b45309,stroke:#d97706,color:#fffbeb,stroke-width:1px;
    classDef fail   fill:#9f1239,stroke:#be123c,color:#fff1f2,stroke-width:1px;

    subgraph CONF["afp.conf (read under a mandatory read lock — fail-closed)"]
        VOLSEC("volume section")
        PRESET("vol preset")
        GLOBAL("[Global]")
    end

    subgraph PARSE["afp_config_parse() — process-global, once per process"]
        BOOL("conf_parse_bool()<br/><i>whole-word; invalid ⇒ warn + unset</i><br/>inside getoption_bool / vdgoption_bool")
        TRI("strict locking (canonical)<br/>afp read locks (deprecated alias, warns)<br/><i>tri-state: explicit no survives</i>")
        MARK("explicitness markers<br/>strict_locking_explicit<br/>dircache_validation_freq_explicit<br/>dircache_rfork_budget_explicit")
    end

    subgraph CV["creatvol() — per volume"]
        EA("ea resolve: volume → preset → [Global] → auto-detect<br/><i>vdgoption_str(); unknown value warns</i>")
        PROBE("check_ea_support() probe<br/><i>failure log now carries errno</i>")
        HARD{"ea = samba ∧<br/>no EA support?"}
        COMMIT("volume commit block<br/><i>no failures possible beyond here</i>")
        DEF("apply_samba_defaults()<br/><i>explicit wins ⇒ log_warning<br/>default applied ⇒ log_note</i>")
    end

    subgraph RT["session runtime"]
        IO("fork.c per-IO POSIX locks<br/>F_RDLCK per read · F_WRLCK per write")
        RFORK("Tier-2 rfork cache gate<br/><i>samba volume ⇒ excluded<br/>unless budget explicit</i>")
        DC("should_validate_cache_entry()<br/><i>live options read, 1–100 clamp</i>")
    end

    VOLSEC --> EA
    PRESET --> EA
    GLOBAL --> EA
    GLOBAL --> BOOL
    BOOL --> TRI
    TRI --> MARK
    EA --> PROBE
    PROBE --> HARD
    HARD -- "yes ⇒ EC_FAIL:<br/>volume not loaded,<br/>process state untouched" --> FAILN("log_error names volume + reason"):::fail
    HARD -- no --> COMMIT
    COMMIT --> DEF
    MARK -- "explicit ⇒ value survives,<br/>coherency warning" --> DEF
    DEF -- "defaults strict locking on" --> IO
    DEF -- "defaults freq = 1" --> DC
    DEF -. "per-volume exclusion<br/>(v_flags bit, not an option write)" .-> RFORK

    class VOLSEC,PRESET,GLOBAL conf;
    class BOOL,TRI,MARK parse;
    class EA,PROBE,COMMIT,DEF vol;
    class HARD,RFORK,DC,IO gate;
    style CONF fill:#f1f5f9,stroke:#cbd5e1,color:#334155;
    style PARSE fill:#f8fafc,stroke:#cbd5e1,color:#334155;
    style CV fill:#eef2f6,stroke:#d1d9e0,color:#334155;
    style RT fill:#f1f5f9,stroke:#cbd5e1,color:#334155;
```


contrib/webmin_module/edit_global_section.cgi
```diff
@@ -514,10 +514,18 @@ print &ui_tabs_end_tab('mode', 'network');
 print &ui_tabs_start_tab('mode', 'misc');
 print &ui_table_start($text{'edit_global_section_title_table'}, 'width="100%"', 2);
 
-@values = get_parameter_of_section($afpconfRef, $sectionRef, 'afp read locks', \%in);
+# Migration shim: a config written before the rename may carry only the
```
rdmark: when we renamed `dircachesize` to `dircache size` we didn't provide such a convenience for webmin module users (and a handful of earlier options that were renamed or consolidated) -- I'm on the fence whether we want to burden the webmin Perl code with a lot of fallback logic, although I agree it's very kind to the user :)

NEWS.md
```diff
@@ -4,6 +4,25 @@ Netatalk Changelog
...
+  setting for safe concurrent access (`strict locking = yes`,
+  `dircache validation freq = 1`, resource-fork cache off). Explicit settings
+  still win, with a logged warning when they weaken Samba coherency
+* FIX: libatalk: the afp.conf volume loader parsed the config file with no
+  read lock held after two failed lock attempts, and did not check the open()
+  of the config file; both now fail closed and leave the reload state
+  untouched. The function is renamed load_volumes() -> load_afp_conf_vols()
+  to say what it does
+* BREAKING: libatalk: struct afp_options gains three fields (installed header
+  globals.h); ABI layout change covered by this cycle's soversion 21 bump
```
rdmark: strictly speaking we bumped the soversion to 21 in the upcoming 4.5.2 release didn't we? if users go from 4.5.2 to 4.6.0 not expecting an ABI change they will get a surprise won't they?

rdmark:
please update the table of renamed/changed options in doc/manual/Upgrading.md as well -- `afp read lock` was added in v3.0 originally

rdmark:
not related to this changeset, but it would be great to have a metadata migration path between `ea = sys` and `ea = samba` (both directions)

if not fully automatic and transparent in afpd, it could for instance be an operational mode of the `dbd` database repair tool

contrib/webmin_module/edit_global_section.cgi
andylemin: This one is more important I think, as the default is off. Else we would have the fall back code in netatalk forever?
We didn't do it for dircachesize to dircache size etc, becuase the default does not matter so much and would not be a breaking change.

Both are still supported here also, but without the shim, when we remove the fall back, it would be breaking change? Ie, those who need it on, would have it reset to off, which may be bad for those with samba - which will be NAS boxes, which are the exact users of the webmin module?

NEWS.md
andylemin: Good catch — you're right, the 4.5 branch ships soversion 21 in 4.5.2 (the dsi_writeinit pointer-handoff change). Bumped main to 22 for this layout change and reworded the NEWS entry: "4.5.2 branch shipped soversion 21, so we skip to 22 on main".

include/atalk/globals.h
```diff
@@ -60,7 +60,7 @@
 #define OPTION_SERVERNOTIF   (1 << 2)
 #define OPTION_NOSENDFILE    (1 << 3)
 #define OPTION_VETOMSG       (1 << 4) /*!< whether to send an AFP message for veto file access */
-#define OPTION_AFP_READ_LOCK (1 << 5) /*!< whether to do AFP spec conforming read locks (default: no) */
+#define OPTION_STRICT_LOCKING (1 << 5) /*!< per-IO POSIX byte-range data locks on FPRead/FPWrite (default: no); formerly OPTION_AFP_READ_LOCK / "afp read locks" */
```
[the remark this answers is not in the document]
andylemin: Kept as a deprecated alias: `#define OPTION_AFP_READ_LOCK OPTION_STRICT_LOCKING` now sits next to the new define in globals.h, so out-of-tree code referencing the old name keeps compiling.

libatalk/util/netatalk_conf.c
```diff
@@ -2898,8 +3111,38 @@ int afp_config_parse(AFPObj *AFPObj, char *processname)
...
+        int strict_locking = getoption_bool(config, INISEC_GLOBAL,
+                                            "strict locking", NULL, -1);
+        int read_locks_alias = getoption_bool(config, INISEC_GLOBAL,
+                                              "afp read locks", NULL, -1);
+        const char *raw_alias = INIPARSER_GETSTR(config, INISEC_GLOBAL,
+                                "afp read locks", NULL);
+
+        /* Deprecation notice keys on the KEY being present (even with an
+         * invalid or empty value), not on it parsing. */
+        if (raw_alias && *raw_alias) {
```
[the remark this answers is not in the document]
andylemin: The comment was the bug, not the guard: an empty value (`afp read locks = ` with nothing after it) deliberately counts as unset everywhere in this file — generated configs commonly emit that shape — so no deprecation notice should fire for it. Reworded the comment to say exactly that (present with a non-empty value, even an invalid one).

libatalk/util/netatalk_conf.c
```diff
@@ -89,6 +89,15 @@ static int have_uservol = 0;
 static struct vol *Volumes = NULL;
 static uint16_t lastvid = 0;
 
+/* Test support: saturate the vid counter so the next creatvol() takes the
+ * vid-overflow failure path.  Unconditional, like
+ * dircache_reset_validation_counter() — test-only c_args cannot gate
+ * libatalk code. */
+void conf_testutil_set_lastvid(uint16_t vid)
```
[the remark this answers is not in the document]
andylemin: Partly intentional, partly unavoidable: the symbol must live in libatalk (it mutates the file-static vid counter), but its declaration was moved out of the installed netatalk_conf.h into the test-local subtests_conf.h during pre-PR review, so it's not part of the advertised API. It follows the same pattern as dircache_reset_validation_counter() — a production symbol that exists for test determinism. libatalk doesn't use symbol visibility annotations, so making it truly private would need an export map — out of scope here.

andylemin:
Upgrading.md updated — `afp read locks` → `strict locking` row added to the v3→v4 renamed-options table (change in 4.6.0, old name noted as a deprecated synonym that logs a warning).

On the `ea = sys` ⇄ `ea = samba` migration path: agreed, and a `dbd` operational mode looks like the right home — the transform is mechanical per xattr (append or strip the trailing NUL that `vfs_streams_xattr` expects) and dbd already walks every file on a volume with the CNID machinery loaded. Both manual sections this PR touches currently document the copy-via-AFP-client workaround as the migration path; happy to spec the dbd mode as a follow-up issue.

andylemin:
augment review

libatalk/util/netatalk_conf.c
```diff
@@ -1472,7 +1637,12 @@ static struct vol *creatvol(AFPObj *obj,
             name, volume->v_uuid);
     }
 
-    if ((val = getoption_str(obj->iniconfig, section, "ea", preset, NULL))) {
+    /* Resolve ea from: volume section -> preset -> [Global] -> built-in
+     * default (AFPVOL_EA_AUTO / AD_VERSION, already set at volume init).
+     * An empty value ("ea = ") counts as unset. */
+    val = vdgoption_str(obj->iniconfig, section, "ea", preset, NULL);
```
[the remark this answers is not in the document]
andylemin: The behaviour is deliberate, but the comment did misdescribe it — fixed. An explicitly emptied volume key (`ea = `) opts OUT of preset/[Global] inheritance and keeps auto-detect, rather than silently inheriting a global value the operator blanked. Generated configs that emit `ea = ` therefore get exactly what they got before this PR (auto-detect), never a surprise inherited backend. The comment now states the opt-out and its rationale. Covered by utest_conf_ea_fallback's absent-key case; the empty-key shape keeps today's behaviour byte-for-byte.

distrib/docker/samba_interop_test.sh
```diff
@@ -0,0 +1,477 @@
...
+    samba smbclient cifs-utils attr python3 fuse3 procps \
+    git ca-certificates meson ninja-build pkg-config build-essential \
+    libgcrypt20-dev libreadline-dev libfuse3-dev > /dev/null
+
+# --------------------------------------------------------------------------
+# Netatalk-Client: pull from GitHub and build (FUSE enabled)
+# --------------------------------------------------------------------------
+
+echo "*** Building netatalk-client from $NETATALK_CLIENT_REPO"
+git clone --quiet --depth 1 "$NETATALK_CLIENT_REPO" /tmp/netatalk-client
```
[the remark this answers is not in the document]
andylemin: Intended, and now called out in the script: the leg is an integration test between the two projects' heads, so a client-side regression should surface here rather than stay invisible until release. NETATALK_CLIENT_REPO is the pinning knob (any git URL, e.g. a fork or a tag mirror) if reproducibility is ever needed for a specific run.

contrib/webmin_module/edit_global_section.cgi
andylemin: What do you think @rdmark ? I've leaving it in for now for the above reasons

andylemin:
> Upgrading.md updated — `afp read locks` → `strict locking` row added to the v3→v4 renamed-options table (change in 4.6.0, old name noted as a deprecated synonym that logs a warning).
> 
> On the `ea = sys` ⇄ `ea = samba` migration path: agreed, and a `dbd` operational mode looks like the right home — the transform is mechanical per xattr (append or strip the trailing NUL that `vfs_streams_xattr` expects) and dbd already walks every file on a volume with the CNID machinery loaded. Both manual sections this PR touches currently document the copy-via-AFP-client workaround as the migration path; happy to spec the dbd mode as a follow-up issue.

@rdmark Actually dbd is being retired. All the CNID realted operations should be re-implemented in `nad` or something?

andylemin:
@rdmark Checkout the "Containers / samba-client and netatalk-client interop (ea:samba) - Debian" CI?
You might be pleased to see the Netatalk-client results. There are some issues to investigate (suppressing their return codes for now), but impressive results for Netatalk-Client

rdmark:
> @rdmark Actually dbd is being retired. All the CNID realted operations should be re-implemented in `nad` or something?

@andylemin we are indeed planning to retire etc/cnid_dbd however bin/dbd is completely decoupled from the former and operates on any CNID backend

andylemin:
> > @rdmark Actually dbd is being retired. All the CNID realted operations should be re-implemented in `nad` or something?
> 
> 
> 
> @andylemin we are indeed planning to retire etc/cnid_dbd however bin/dbd is completely decoupled from the former and operates on any CNID backend

Great, sounds like a 'nad cnid' namespace opportunity then 🙂

andylemin:
augment review

andylemin:
Issue #3229 raised for follow ups with Samba

include/atalk/globals.h
```diff
@@ -60,7 +60,7 @@
 #define OPTION_SERVERNOTIF   (1 << 2)
 #define OPTION_NOSENDFILE    (1 << 3)
 #define OPTION_VETOMSG       (1 << 4) /*!< whether to send an AFP message for veto file access */
-#define OPTION_AFP_READ_LOCK (1 << 5) /*!< whether to do AFP spec conforming read locks (default: no) */
+#define OPTION_STRICT_LOCKING (1 << 5) /*!< per-IO POSIX byte-range data locks on FPRead/FPWrite (default: no) */
```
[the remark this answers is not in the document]
andylemin: ABI has been updated

distrib/docker/env_setup_netatalk.sh
```diff
@@ -532,11 +540,11 @@ if [ "$TESTSUITE" = "spectest" ] && [ -z "$AFP_REMOTE" ]; then
     TEST_FLAGS="$TEST_FLAGS -c $NETATALK_SHARE_DIR"
 fi
 
-# When the server enables 'afp read locks', tell afp_spectest (-L) so the
+# When the server enables 'strict locking', tell afp_spectest (-L) so the
 # byte-range read-lock conflict tests run instead of skipping (T_LOCKING).
 # -L is an afp_spectest-only flag, so gate it on the spectest suite to avoid
 # passing an unknown option to the other test runners.
-if [ "$TESTSUITE" = "spectest" ] && [ "$AFP_READ_LOCKS" = "yes" ]; then
+if [ "$TESTSUITE" = "spectest" ] && [ "$AFP_STRICT_LOCKING" = "yes" ]; then
```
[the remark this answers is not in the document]
andylemin: non-issue for scripted use of this file. the official values are yes/no, other values are convenience only.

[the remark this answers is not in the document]
andylemin: non-issue for scripted use of this file. the official values are yes/no, other values are convenience only.

test/afpd/peer_lock.c
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    /* parent */
+    close(up[1]);
+    close(down[0]);
+    p->to_child = down[1];
+    p->from_child = up[0];
+
+    /* block until the peer reports the lock is held */
+    if (read(p->from_child, &b, 1) != 1) {
```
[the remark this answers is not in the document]
andylemin: Fixed — the failed-handshake path now reaps the child (waitpid) alongside closing both pipe ends, so no zombie survives a peer that dies before reporting.

libatalk/util/netatalk_conf.c
```diff
@@ -1065,31 +1141,149 @@ static uint32_t getoption_uint32_strict(INIPARSER_DICTIONARY *conf,
...
+        LOG(log_note, logtype_afpd,
+            "creatvol(\"%s\"): ea = samba: defaulting 'strict locking' to yes "
+            "(POSIX read+write data locks, visible to Samba)", volname);
+        options->flags |= OPTION_STRICT_LOCKING;
+    }
+
+    if (options->dircache_validation_freq_explicit) {
+        if (options->dircache_validation_freq != 1) {
+            LOG(log_warning, logtype_afpd,
+                "creatvol(\"%s\"): ea = samba, however 'dircache validation "
```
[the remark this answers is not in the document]
andylemin: Fixed — the warning now fires only for in-range values (2–100). Out-of-range values already fall back to the validate-always fail-safe at dircache_init(), so warning about them would misstate the runtime behaviour.

etc/afpd/dircache.c
```diff
@@ -1100,17 +1100,28 @@ static unsigned long queue_count;
...
+     * plumbing.  Same 1-100 validity range as
+     * dircache_set_validation_params(); out-of-range config values fall
+     * back to the static (which the setter kept at the fail-safe). */
+    unsigned int freq = (AFPobj
+                         && AFPobj->options.dircache_validation_freq > 0
+                         && AFPobj->options.dircache_validation_freq <= 100)
+                        ? (unsigned int)AFPobj->options.dircache_validation_freq
+                        : dircache_validation_freq;
 
     /* Validate every Nth access to detect external changes */
```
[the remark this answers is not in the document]
andylemin: Fixed — the unreachable freq == 0 branch is removed; a comment states the invariant (options read clamped to 1–100, setter rejects 0 for the static).

libatalk/util/netatalk_conf.c
```diff
@@ -2327,13 +2514,62 @@ int load_volumes(AFPObj *obj, lv_flags_t flags)
         ret = getpwuid_r(obj->uid, &pwent, pwbuf, bufsize, &pwresult);
 
         if (pwresult == NULL) {
-            LOG(log_error, logtype_afpd, "load_volumes: getpwuid_r: %s", strerror(errno));
+            LOG(log_error, logtype_afpd, "load_afp_conf_vols: getpwuid_r: %s",
```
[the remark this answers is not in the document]
andylemin: Fixed — the log now reports strerror(ret) (getpwuid_r's actual error), with a distinct "user not found" message for the ret == 0, no-entry case.

contrib/webmin_module/edit_global_section.cgi
```diff
@@ -514,10 +514,18 @@ print &ui_tabs_end_tab('mode', 'network');
...
+# deprecated 'afp read locks' key.  Show its value as the 'strict locking'
+# selection; save_global_section.cgi writes the new key and removes the
+# deprecated one.
+if (   !exists $$sectionRef{parameters}{'strict locking'}
```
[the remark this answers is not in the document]
andylemin: Fixed — the shim now also treats an empty-valued 'strict locking' key as absent, matching the server's parse where the alias wins in that case too.

distrib/docker/env_setup_netatalk.sh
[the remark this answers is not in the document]
andylemin: Fixed — the -L gate now matches every truthy spelling the server's boolean parser accepts (yes/y/true/t/on/enabled/1, case-insensitive), so the spectest flag tracks what the generated config actually enables.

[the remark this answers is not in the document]
andylemin: Valid gap, deliberately out of scope for the env-var gate: -L keys on the explicit env var, not on ea = samba's implicit default. A samba-EA spectest leg (which would need -L for full coverage) is future work; today no CI leg sets AFP_EA=samba, so no coverage is lost. Noted for the leg that adds it.

include/atalk/globals.h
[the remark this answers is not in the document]
andylemin: Intentional — discussed and decided earlier in this PR: this cycle already changes the installed-header ABI (struct afp_options layout, soversion 21→22), so out-of-tree consumers must rebuild against the new headers regardless; a clean break now beats carrying a deprecated macro. The NEWS BREAKING entry records it.

andylemin:
@rdmark This has been cleaned up a lot now and is looking much better.
Also improved the webmin code to be more modular and reusable.
Ready for final review

distrib/docker/env_setup_netatalk.sh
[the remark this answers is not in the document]
andylemin: On reflection, reverted to the plain `yes` test: the CI workflows and compose files all pass `AFP_STRICT_LOCKING=yes`, so the extra spellings were dead flexibility in a container-API gate. The env var contract stays simple; only afp.conf itself accepts the full boolean vocabulary.

rdmark:
> > > @rdmark Actually dbd is being retired. All the CNID realted operations should be re-implemented in `nad` or something?
> > 
> > 
> > @andylemin we are indeed planning to retire etc/cnid_dbd however bin/dbd is completely decoupled from the former and operates on any CNID backend
> 
> Great, sounds like a 'nad cnid' namespace opportunity then 🙂

That’s an interesting idea! I can see how it could work on a technical level, dbd is a quite thin application. 

My original plan has been to keep dbd around as is, perhaps invent a new meaning for the acronym (how about “database doctor”?)

My argument for keeping nad and dbd separate is that one operates on files on the (virtual) file system while the other operates on a database. They’re semantically divided this way.

andylemin:
That's a reasonable idea

distrib/docker/samba_interop_test.sh
```diff
@@ -0,0 +1,677 @@
...
+# --------------------------------------------------------------------------
+# Netatalk-Client: pull from GitHub and build (FUSE enabled)
+# --------------------------------------------------------------------------
+
+echo "*** Building netatalk-client from $NETATALK_CLIENT_REPO"
+# Deliberately tracks the client repo's default branch: this leg is an
+# integration test between the two projects' heads, so a client-side
+# regression should surface here.  Pin via NETATALK_CLIENT_REPO if a
+# reproducible build is ever needed.
+git clone --quiet --depth 1 "$NETATALK_CLIENT_REPO" /tmp/netatalk-client
```
rdmark: can't it be argued that testing bleeding edge netatalk-client should more appropriately be done over at the netatalk-client project?

it might end up somewhat challenging to troubleshoot failures in this test since there may be bugs on either side

not a blocker for this PR but I would like to move towards a pinned netatalk-client version to test against here sooner rather than later!

doc/manpages/man5/afp.conf.5.md
```diff
@@ -36,9 +36,13 @@ Any line ending in a **\\** (backslash) is continued on the next line in the
...
-string (no quotes needed) or a boolean, which may be given as yes/no,
-1/0 or true/false. Case is not significant in boolean values, but is
-preserved in string values. Some options such as **file perm** take
+string (no quotes needed) or a boolean. Booleans are preferably given as
+yes/no — the backwards-compatible form accepted by every Netatalk
+release — though true/false, on/off, enabled/disabled, 1/0 and the
+single-letter shorthands y/n and t/f are also accepted. Case is not
+significant in boolean values, but is preserved in string values. Any
+other boolean value is invalid: it is ignored with a logged warning and
+the option's default applies. Some options such as **file perm** take
```
rdmark: rather than fully documenting the iniparser library's logic how about we list up the most common options and then defer to the library's own docs?

this seems to be it: https://iniparser.gitlab.io/iniparser/iniparser_8h.html#afa30800883e2853074c038467f5fde2b

interestingly, this doesn't cover the "on/off" variant

contrib/webmin_module/netatalk-lib.pl
```diff
@@ -104,10 +104,14 @@
 
 my @yesNoSelectOptions = ('yes', 'no');
 
+# Renamed afp.conf options: new name => deprecated alias still accepted by
+# the server.  Drives the edit-page display fallback and the save-time
+# cleanup below; future renames only need a new entry here.
+our %netatalkDeprecatedAliases = ('strict locking' => 'afp read locks',);
```
rdmark: not required for this PR, but I want to throw `dircache size => dircachesize` and maybe even `server name => zeroconf name` into the mix later!

doc/manpages/man5/afp.conf.5.md
```diff
@@ -1225,11 +1224,14 @@ hosts deny = *IP host address/IP netmask bits* [ ... ] **(V)**
...
 > Example: hosts deny = 192.168.100/24 10.1.1.1 2001:db8::1428:57ab
 
-ea = *sys* | *samba* | *ad* | *none* (default: auto detect) **(V)**
+ea = *sys* | *samba* | *ad* | *none* (default: auto detect) **(G)**/**(V)**
 
 > Specify how Extended Attributes and Mac OS Resource Forks are
 stored.
 >
+> When set in the **[Global]** section it applies to all volumes; a volume
+section (or selected **vol preset**) may override it.
```
rdmark: isn't this how all G/V options behave? I suggest we move this paragraph up under the `The [Global] section` section of this man page

doc/manpages/man5/afp.conf.5.md
```diff
@@ -1245,6 +1247,15 @@ For read-only volumes, set this option explicitly.
...
 order to be compatible with Samba's vfs_streams_xattr.
 >
+> > Selecting **samba** also changes the server defaults to the values safe
+concurrent Samba access requires: **strict locking** defaults to *yes*,
+**dircache validation freq** defaults to *1*, and the volume is excluded
+from the resource-fork data cache. Explicit settings always win; every
+explicit setting that weakens Samba coherency logs a verbose warning
+describing the risk. It
+requires filesystem Extended Attribute support; a volume whose filesystem
+cannot store them fails to load.
```
rdmark, line 1254: I'm not sure how I feel about altering the semantics of `ea = samba` although I do see the argument that all three absolutely should be used together for a stable Samba interop setup

how do you feel about introducing a brand new boolean called `samba interop` or similar which activates these three options in one go? is arguably more discoverable and self-explanatory for users.

rdmark (commented):
it's great to see test coverage for the netatalk_conf.c code -- that module is such a spaghetti mess, surely many bugs lurking there especially around parsing invalid or boundary inputs

do you think we need to update the afparg.1 man page at all? I wasn't sure if those code changes alters the behavior of afparg notably

distrib/docker/samba_interop_test.sh
andylemin: Agree! Is the last committed version complete enough to do that now?

doc/manpages/man5/afp.conf.5.md
andylemin: This is not the iniparser anymore. I had to replace it with our own boolean parser which is better than iniparser bolean parser (this was not a strict requirement, but we a repeated nag whilst trying to clean up and unify things in the config parsing in general - for example logging). But I agree this is too long (and missing on/off) for the afp.conf.

contrib/webmin_module/netatalk-lib.pl
andylemin: Done! 🙂

doc/manpages/man5/afp.conf.5.md
andylemin: Fair challenge but I feel quite strongly about this for a few reasons;
If `ea` is set incorrectly, it can corrupt EAs for Samba users. All the other samba related options are locking and performance things. Adding `samba interop` adds _yet another_ option which users have to understand.
Searching for samba in the docs should have one hit, and `ea = samba` already provides this and is the critical know already - so updating the defaults for the other optional settings to align with this critical one, which is clearly named, is graceful and minimal user overhead.
It does not clobber their other settings if they want to override, and it gives log messages when the user has updated defaults.
This is simple and clear for users, transparently upgrade safe and graceful, and does not add to the existing excessive options count.

distrib/docker/samba_interop_test.sh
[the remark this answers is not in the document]
andylemin: Updated: the leg now builds the latest netatalk-client *release tag* (resolved via `git ls-remote --tags`, currently 0.9.5) instead of the default-branch snapshot, so runs are reproducible per release. `NETATALK_CLIENT_REF` overrides with any tag or branch for pinning or for testing an unreleased client.

contrib/webmin_module/edit_global_section.cgi
andylemin: Resolving now we have others using it too 🙂

doc/manpages/man5/afp.conf.5.md
andylemin: Cleaned up and reduced

doc/manpages/man5/afp.conf.5.md
andylemin: Yes, was redundant. Fixed

andylemin:
> it's great to see test coverage for the netatalk_conf.c code -- that module is such a spaghetti mess, surely many bugs lurking there especially around parsing invalid or boundary inputs
> 
> do you think we need to update the afparg.1 man page at all? I wasn't sure if those code changes alters the behavior of afparg notably

Yes it really is! Which is why I had to draw the diagram in the summary to get my head around what was going on!
I resolved a whole bunch of issues in netatalk_conf.c, which is what triggered the parser replacement.

Re; afparg.1 man page - good catch. Resolved.

andylemin:
@rdmark ready for next round - I knew this was going to be a big one as it touches so many things..

> netatalk_conf.c code -- that module is such a spaghetti mess

I have a task list of fixes we can make to improve it (a few key changes should improve it a lot).

Also notice the commit includes changes for the website and replaces the old Interoperability-with-Samba page with the content in the main [Netatalk Configuration](https://netatalk.io/manual/en/Configuration) page

distrib/docker/samba_interop_test.sh
rdmark: I believe v0.9.5 is stable enough to serve as baseline for testing, yes!

doc/manpages/man5/afp.conf.5.md
rdmark: aha, this sounds like an opportunity to file a feature request upstream?

would we go so far as to make our own iniparser routines if we're not confident with the upstream library?

doc/manpages/man5/afp.conf.5.md
rdmark: I see the argument!

my only objection is that "ea" doesn't really stand for (only) "extended attributes" anymore in this configuration, but it can be considered a nitpick at this point

rdmark (approved):
LGTM 👌

PR #3232, opened 2026-08-08
ci: runner-normalized Adj Δ% column for the lantest dashboard

andylemin:
Adds a runner-normalized **Adj Δ%** column to the lantest dashboard table, separating genuine code effects from CI runner speed variance.

## The problem

Every lantest op test in a CI run moves together with the runner it landed on: a fast host shifts all deltas down 1–3%, a slow one shifts them all up. A single run's `Cur Avg Δ%` therefore mixes two signals — the runner's speed relative to the historical fleet, and the PR's actual code effect. PR #3231 showed this clearly: every test drifted −1–3% (runner) while ServerCopy moved −18.5% (code).

## The method

This is the same trick as **median-of-ratios normalization** in genomics (DESeq2's size factors) and **market-relative returns** in finance: when most members of a population move with a common factor, the median isolates that factor because it ignores the minority that genuinely changed.

Walk-through, per dashboard render:

1. For each history-backed lantest table metric *i*, compute its ratio against its own historical average:
   `rᵢ = currentᵢ / hist_avgᵢ` (equivalently `1 + δᵢ`, where δᵢ is the raw delta)
2. Take the **run baseline** `m = median(rᵢ)` across those metrics — the runner's speed factor for this run. The median, not the mean, so the handful of tests the PR genuinely changed can't skew it.
3. Re-base each metric against the baseline, by **division**:
   `Adj Δᵢ = rᵢ / m − 1 = (1 + δᵢ) / (1 + m′) − 1` (where `m′ = m − 1` is the baseline expressed as a delta)
   Division rather than subtraction because runner speed is multiplicative — a runner that is 10% slower makes a 500 ms test 550 ms and a 100 ms test 110 ms; subtracting a flat 10 points would over-correct slow tests and under-correct fast ones.
4. Metrics with `|Adj Δ| ≥ 5%` render **bold**: standouts from the pack — code effects, not runner properties.
5. The baseline itself is printed under the table (`Run baseline: median op-test delta +10.0%`), so a strongly shifted run is visible at a glance.

Worked example (the scenario from #3231, simulated in testing — slow runner +10%, ServerCopy genuinely improved 18%):

| Metric | Current (ms) | Cur Avg Δ% | Adj Δ% | Hist avg |
|--------|--------------|------------|--------|----------|
| Creating 2000 files | 330 | +10.0% | +0.0% | 300 |
| Copying 2000 files server-side | 631 | −9.9% | **−18.1%** | 700 |
| Stat 2000 files | 275 | +10.0% | +0.0% | 250 |

The raw column shows the confounded −9.9%; the adjusted column recovers the true −18.1%.

## Scope decisions

- **`Cur Avg Δ%` stays.** Median-centering detects *differential* signals only — a change that shifts every test equally is indistinguishable from runner noise and gets normalized away by construction. Uniform regressions and broad improvements remain visible in the raw column, in the printed run baseline, and in the long-term `Avg total runtime` trend already stored in the YAML history.
- **Lantest only.** Speedtest metrics are not commensurable (peak/avg-mean/avg-max of four different operations), and runner IO variance does not track the CPU-bound op tests, so a speedtest median would be meaningless.
- **The 100MB streaming tests are excluded** (`--adjust-exclude one_large_file`) for the same reason: Write/Read 100MB are IO-bound and don't move with the CPU-bound op-test baseline. They contribute nothing to the median and show `—` in the adjusted column.
- **First run / no history**: the column renders `—` throughout and no baseline footer is printed.

## Mechanics

`perf_history.pl` gains `--adjust NAME` (per-section opt-in) and `--adjust-exclude REGEX`; both dashboard call sites (main merge and the push-retry regeneration) pass `--adjust lantest --adjust-exclude one_large_file`, so a retried publish regenerates identically.

## Testing

- Slow-runner simulation (all +10%, one −18% code improvement): baseline +10.0%, unaffected tests +0.0% adjusted, improvement recovered at −18.1% and bolded
- First-run empty history and excluded-metric rendering verified
- perltidy clean; both call sites verified; core-Perl only (no new dependencies)

Follow-up once history accumulates: replace the fixed 5% standout threshold with per-test historical stddev bands.

### DIFF
--- a/.github/scripts/perf_history.pl
+++ b/.github/scripts/perf_history.pl
@@ -41,11 +41,18 @@
 # Usage:
 #     perf_history.pl --history FILE --pr N --timestamp ISO8601 \
 #         --section name=metrics_file [--section name=metrics_file ...] \
+#         [--adjust name ...] [--adjust-exclude REGEX] \
 #         [--github-output FILE]
 #
 # For every --section the script writes a "<name>_table" and "<name>_inline"
 # output (empty string when there is no data) to --github-output (defaults
 # to $GITHUB_OUTPUT), and rewrites --history in place.
+#
+# Sections named by --adjust gain an "Adj %" column: each metric's
+# current/avg ratio divided by the section's runner baseline (the median
+# ratio across its table metrics, minus keys matching --adjust-exclude).
+# This cancels run-to-run runner-speed variance so genuine code effects
+# stand out; standouts beyond 5% render bold.
 
 use strict;
 use warnings;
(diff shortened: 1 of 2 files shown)


.github/workflows/containers.yml
```diff
@@ -1635,6 +1637,8 @@ jobs:
               --timestamp "$(date -u +%Y-%m-%dT%H:%M:%SZ)" \
               --section lantest=../lantest-metrics.txt \
               --section speedtest=../speedtest-metrics.txt \
+              --adjust lantest \
```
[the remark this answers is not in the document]
andylemin: Good catch. Fixed: the retry now writes its regenerated fragments to the step's outputs, and the comment template prefers those when present — so the posted tables always match the history that was actually published.

andylemin:
This is an interesting test.
This particular CI run was significantly slower than the past 2 recorded CI runs (Cur Avg Δ% and Adj Δ% are currently based on an average of just 2 PRs), and shows 30% slower (severe CI noise). 

Should be interesting to see how these numbers stabilise over time as more CI runs are collected in the history.

.github/scripts/perf_history.pl
```diff
@@ -206,23 +241,49 @@ sub render_section {
...
     my $table_md = '';
     if (@table_rows) {
+        my $adj_hdr = $adjust ? "Adj \x{394}% | " : '';
+        my $adj_sep = $adjust ? "-------|"        : '';
         $table_md =
-            "| Metric | Current ($unit_hdr) | Cur Avg \x{394}% | Hist avg "
+            "| Metric | Current ($unit_hdr) | Cur Avg \x{394}% | ${adj_hdr}Hist avg "
           . "| Hist min | Hist max |\n"
-          . "|--------|------------------|------------|----------"
+          . "|--------|------------------|------------|${adj_sep}----------"
```
rdmark, line 272: should there really be a `${adj_sep}` inserted in this table frame?

rdmark: nm, I understand the logic now

rdmark (approved):

PR #3245, opened 2026-08-12
doc: substitute manpage markdown titles for website

rdmark:
when publishing on the web, it's better to have a unique title for each rendered man page rather than the default 'Name' everywhere


rdmark:
@andylemin when trying the new website search it was obvious that each manual page needs a unique page title!

andylemin (approved):

PR #3247, opened 2026-08-13
testsuite: add DHX and DHX2 auth support to AFP test runners

rdmark:
Implement the DHCAST128 and DHX2 login exchanges in libafptest and expose them through the test runner -A option. This allows testing against secured AFP servers including Mac OS X personal file sharing where ClearTxt UAM is not available.

The insecure and legacy ClearTxt auth can be selected with `-A clrtxt` which is also the default fallback when no UAM is given.

Refactor the testsuite container entrypoint to allow for subtest execution, notably for the spectest, controlled by AFP_SUBTESTS container runtime option. Rework the entry point for the Readonly and Speedtest test suite using this structure, while also adding the FPZzzzz (sleep) test suite to CI.

Introduces a VERY_VERBOSE container runtime option for the testsuite which passes -V to the test runner for testsuite debug logging.

Fix a bug in the spectest code that failed to tokenize comma separated subtests passed to the test runner with the -f parameter.

### DIFF
--- a/distrib/docker/entrypoint_netatalk.sh
+++ b/distrib/docker/entrypoint_netatalk.sh
@@ -46,30 +46,58 @@ else
     echo ""
     # Temporarily disable exit-on-error to ensure we can dump logs even if test fails
     set +e
-    set -x
     TEST_EXIT_CODE=0
     case "$TESTSUITE" in
-        spectest)
-            afp_spectest $TEST_FLAGS -"$AFP_VERSION" -h "$AFP_HOST" -p "$AFP_PORT" -u "$AFP_USER" -d "$AFP_USER2" -w "$AFP_PASS" -s "$SHARE_NAME" -S "$SHARE_NAME2"
-            TEST_EXIT_CODE=$?
-            ;;
-        readonly)
-            echo "testfile uno" > /mnt/afpshare/first.txt
-            echo "testfile dos" > /mnt/afpshare/second.txt
-            mkdir /mnt/afpshare/third
-            afp_spectest $TEST_FLAGS -"$AFP_VERSION" -h "$AFP_HOST" -p "$AFP_PORT" -u "$AFP_USER" -w "$AFP_PASS" -s "$SHARE_NAME" -f Readonly_test
+        spec)
+            if [ "$AFP_SUBTESTS" = "Readonly" ]; then
+                echo "testfile uno" > /mnt/afpshare/first.txt
+                echo "testfile dos" > /mnt/afpshare/second.txt
+                mkdir /mnt/afpshare/third
+            fi
+            set -- -"$AFP_VERSION" -h "$AFP_HOST" -p "$AFP_PORT" \
+                -u "$AFP_USER" -w "$AFP_PASS" -s "$SHARE_NAME"
+            if [ -n "$AFP_USER2" ]; then
+                set -- "$@" -d "$AFP_USER2"
+            fi
+            if [ -n "$SHARE_NAME2" ]; then
+                set -- "$@" -S "$SHARE_NAME2"
+            fi
+            if [ -n "$AFP_TESTSUITE_UAM" ]; then
+                set -- "$@" -A "$AFP_TESTSUITE_UAM"
+            fi
+            if [ -n "$AFP_SUBTESTS" ]; then
+                set -- "$@" -f "$AFP_SUBTESTS"
+            fi
+            afp_spectest $TEST_FLAGS "$@"
             TEST_EXIT_CODE=$?
             ;;
         login)
-            afp_logintest $TEST_FLAGS -"$AFP_VERSION" -h "$AFP_HOST" -p "$AFP_PORT" -u "$AFP_USER" -w "$AFP_PASS"
+            set -- -"$AFP_VERSION" -h "$AFP_HOST" -p "$AFP_PORT" \
+                -u "$AFP_USER" -w "$AFP_PASS"
+            afp_logintest $TEST_FLAGS "$@"
             TEST_EXIT_CODE=$?
             ;;
         lan)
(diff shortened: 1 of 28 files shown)


rdmark:
augment review

rdmark:
@andylemin I am quite pleased with this approach: DHX and DHX2 implemented as a single tight module (since they share most of the logic) together with the legacy ClearTxt and Guest authentication enables testing against virtually any AFP server, I think.

note also the refactoring of the test container entrypoint scripts which I think is cleaner and easier to tweak in the future

rdmark:
@andylemin rebased and ready for your review!

distrib/docker/entrypoint_netatalk.sh
```diff
@@ -46,38 +46,65 @@ else
     echo ""
     # Temporarily disable exit-on-error to ensure we can dump logs even if test fails
     set +e
-    set -x
     TEST_EXIT_CODE=0
     case "$TESTSUITE" in
-        spectest)
-            afp_spectest $TEST_FLAGS -"$AFP_VERSION" -h "$AFP_HOST" -p "$AFP_PORT" -u "$AFP_USER" -d "$AFP_USER2" -w "$AFP_PASS" -s "$SHARE_NAME" -S "$SHARE_NAME2"
+        spec | spectest)
```
andylemin: Nice QoL addition, these short aliases. One thing I noticed: env_setup_netatalk.sh still matches only the long forms — `[ "$TESTSUITE" = "spectest" ]` gates the `-c $NETATALK_SHARE_DIR` and `-L` flags, and `[ "$TESTSUITE" = "lan" ]` gates the IO-monitoring setup. So `TESTSUITE=spec` (the form CONTAINERS.md now lists) would run spectest without `-c`, quietly skipping the tier-2 testsets. Worth normalizing `TESTSUITE` once in env_setup so the aliases behave the same everywhere?

test/testsuite/afptest_uam.c
```diff
@@ -0,0 +1,745 @@
...
+    ctxerror = gcry_cipher_setkey(ctx, K_hash, hash_len);
+    free(K_hash);
+    K_hash = NULL;
+
+    if (gcry_err_code(ctxerror) != GPG_ERR_NO_ERROR) {
+        goto dhx2_fail;
+    }
+
+    /* Set the initialization vector for client->server transfer. */
+    ctxerror = gcry_cipher_setiv(ctx, dhx_c2siv, sizeof(dhx_s2civ));
```
andylemin: Tiny one: this passes `dhx_c2siv` but takes `sizeof(dhx_s2civ)` (same again at the client-to-server setiv further down, before the final FPLoginCont). Both IVs are 8 bytes so no functional difference — just wondering if `sizeof(dhx_c2siv)` was intended, to save the next reader a double-take?

test/testsuite/afptest_uam.c
```diff
@@ -0,0 +1,745 @@
...
+    /* Compare our incremented nonce to the server's incremented copy
+     * of our original nonce value; if they don't match, something
+     * terrible has happened. */
+    if (memcmp(nonce_binary, d, 16) != 0) {
+        goto dhx2_fail;
+    }
+
+    d += sizeof(nonce_binary);
+    /* Pull the server's nonce value into an gcry_mpi_t. */
+    gcry_mpi_scan(&nonce, GCRYMPI_FMT_USG, d, sizeof(nonce_binary), NULL);
```
andylemin, line 626: Since `gcry_mpi_scan()` allocates a fresh MPI into its out-pointer, the placeholders from `gcry_mpi_new(0)` (p, g, Mb, nonce) end up orphaned when scanned over — and `nonce` gets scanned a second time here. Negligible for a one-shot login, and I can see the pattern comes straight from afpfs-ng; the reconnect-heavy tests do log in a few times per process though. Do you think it's worth releasing before re-scanning, or better to stay close to the original?

test/testsuite/FPZzzzz.c
```diff
@@ -18,6 +20,20 @@ static void pipe_handler(int signum)
     sigp = 1;
 }
 
+/* Match the login choice used for the primary spectest connection. */
+static unsigned int test_login(CONN *conn)
```
andylemin: This helper ends up in three testset files — FPDisconnectOldSession.c has an identical copy, and FPRead.c's differs only in deliberately keeping the plain FPopenLogin fallback. Would a shared helper in afphelper.c (with the FPRead variant kept or documented) be worthwhile, so the login-selection logic lives in one place?

andylemin:
Hi @rdmark sorry for the slow reply. I'm on holiday at the moment. This is a nice improvement on #3176.

distrib/docker/entrypoint_netatalk.sh
rdmark: good catch, then what do you say we keep only the short forms after all, to avoid having to remember to check both forms elsewhere in the scripts? on second thought supporting both may lead to bugs like this in the future.

test/testsuite/afptest_uam.c
rdmark: good catch, definitely a typo

test/testsuite/afptest_uam.c
rdmark: agreed that this is not an optimal pattern, fixing

distrib/docker/entrypoint_netatalk.sh
andylemin: I think we should be strict and have single forms which are consistent everywhere - as you say, fewer bugs and visible positive failures

test/testsuite/FPZzzzz.c
rdmark: yeah, let's make this a shared helper function

distrib/docker/entrypoint_netatalk.sh
rdmark: agreed, fixing!

rdmark:
@andylemin enjoy your vacation! appreciate you leaning in ;)

rdmark:
@andylemin ready for re-review, when you can!

andylemin (approved):

PR #3250, opened 2026-08-16
testsuite: introduce Junit XML test reporting for spectest

### LINKED ISSUE #3221: testsuite: output test results in industry standard formats
introduce optional support for having our testsuite test runner library outputting test results in an industry standard format that can be parsed by other software like GitHub actions test reporting actions

a common one is JUnit XML

rdmark:
the -j argument takes a file path to the xml files where to save Junit compatible XML test results

have each spectest CI job post its test report to PR workflow run summary pages, and do the same for the preexisting TAP unit test reports

### DIFF
--- a/.github/scripts/vm_spectest.sh
+++ b/.github/scripts/vm_spectest.sh
@@ -62,6 +62,8 @@ echo "==== Running AFP spec test (AFP 3.4) ===="
 set +e
 # TEST_FLAGS holds multiple flags (e.g. "-c /path" or "-a"); must split.
 # shellcheck disable=SC2086
+JUNIT_REPORT_PATH="$PWD/afp-spectest.junit.xml"
+rm -f "$JUNIT_REPORT_PATH"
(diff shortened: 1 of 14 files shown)


rdmark:
@andylemin you can see an example of the new junit test reporting on workflow Summary pages such as:

https://github.com/Netatalk/netatalk/actions/runs/31955910503?pr=3250

I think this gives you a better overview of pass rates, a more human friendly list of tests with test status, and overall test duration (as opposed to job duration)

and more broadly, junit test reports can be consumed by downstream users including distribution packagers

andylemin:
@rdmark This is interesting, I've not come across this testing pattern with GitHub before.
Will definitely have a play with this to see how we can improve our testing data.
I admit my recent works, whilst informative, have been bespoke and not portable (and I'm sure I've been reinventing some wheels).

andylemin (approved):

PR #3257, opened 2026-08-18
afpd: ignore permission options without UNIX privileges

### LINKED ISSUE #95: set  file  perm    directory perm  not  take effect
[s1]   
 file perm = 0222
    directory perm = 0444
    hosts allow =
    rolist = ""
    rwlist = "u1"
    valid users = "u1"

3.1.11

rdmark:
Make all runtime uses of the volume umask, file perm, and directory perm options go through accessors that return zero unless the volume enables UNIX privileges. This preserves the configured values while preventing these permission modifications from affecting volumes configured with unix priv = no.

In the afp.conf man page, document the bitwise-OR behavior and their interaction with unix priv.

Add test coverage for for the effective configuration values.

Thanks to @ypfjyhh1523 for the report.

### DIFF
--- a/doc/manpages/man5/afp.conf.5.md
+++ b/doc/manpages/man5/afp.conf.5.md
@@ -1294,9 +1294,15 @@ this writing).
 
 file perm = *mode* **(V)**; directory perm = *mode* **(V)**
 
-> Add(or) with the client requested permissions: **file perm** is for files
-only, **directory perm** is for directories only. Don't use with
-"**unix priv = no**".
+> Add (bitwise OR) this mode to permissions requested by the client:
+**file perm** applies to files only, and **directory perm** applies to
+directories only. These options can grant permission bits, but cannot remove
+bits requested by the client or set a fixed mode. For example, a client
+requesting mode `0644` with `file perm = 0222` results in mode `0666`.
+>
+> The modes are applied when the client changes permissions; they are not
+initial creation modes. See **umask** to mask permission bits. These options
+are ignored with "**unix priv = no**".
 >
 > **Example**: Volume for a collaborative workgroup
 
@@ -1305,7 +1311,8 @@ only, **directory perm** is for directories only. Don't use with
 
 umask = *mode* **(V)**
 
-> Set perm mask. Don't use with "**unix priv = no**".
+> Set a permission mask, which removes bits from modes to which it is applied.
+This option is ignored with "**unix priv = no**".
 
 preexec = *command* **(V)**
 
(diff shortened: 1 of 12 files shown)


ypfjyhh1523:
这是来自QQ邮箱的假期自动回复邮件。 您好，我最近正在休假中，无法亲自回复您的邮件。我将在假期结束后，尽快给您回复。

doc/manpages/man5/afp.conf.5.md
```diff
@@ -1294,9 +1294,15 @@ this writing).
...
-"**unix priv = no**".
+> Add (bitwise OR) this mode to permissions requested by the client:
+**file perm** applies to files only, and **directory perm** applies to
+directories only. These options can grant permission bits, but cannot remove
+bits requested by the client or set a fixed mode. For example, a client
+requesting mode `0644` with `file perm = 0222` results in mode `0666`.
+>
+> The modes are applied when the client changes permissions; they are not
+initial creation modes. See **umask** to mask permission bits. Don't use these
+options with "**unix priv = no**".
```
andylemin: This is unclear? What happens if you do use them with `unix priv = no`?

Better to just say they are ignored with `unix priv = no`

andylemin (approved):

doc/manpages/man5/afp.conf.5.md
rdmark: good point, in fact this made me think: the current logic is a bit awkward in that it still attempts to apply Unix perm modes when 'unix priv = 0' -- I think we should make it so that the configure parser ignores these settings in this case

andylemin: I'm not too keen on putting business logic into the config parser when possible (assuming that's what you meant?). We need to simplify the config parser?

I think it's more consistent and robust to keep the logic skip with the main code as it's a direct dependancy, rather than having conditionals in the parser hold down other values.

I know the recent samba changes did the opposite of this, however the justification for adding the samba alignment logic in the config parser is because each setting was not "dependant" on another - it was just an optimised config pattern.

What do you think? I would think a simple `if` would suffice (even if a few are required) as it's not hot path, and makes the dependency clear at the point of implementation?

rdmark: right, thanks for calling that out, I fell back to bad habits

here's a fresh approach: a suite of macros that conditionally look up the priv/umask settings only when AFPVOL_UNIX_PRIV is flagged

the config parser logic is untouched -- quite clean solution if I may say so myself :)

andylemin: Nice! 🙂

PR #3258, opened 2026-08-18
doc: overhaul the ACL chapter in the manual with support matrix

### LINKED ISSUE #239: ACLs "Allow" Only take precedence over POSIX permissions for Files BUT Not Folders - ACLs "Deny" DO take precedence over POSIX permissions
According to the Netatalk documentation, when LDAP is setup and a directory user has his permissions read by the netatalk server, the Folder's or File's ACLs take precedence over POSIX/Unix permissions.

Running Netatalk under MacOS Big Sur, if a directory group or directory user encounters a file or a folder on the AFP share with an ACL with a "deny" set against that group or user, but a POSIX/Unix permission of 777, the file will be invisible to the user or group, which is the expected behavior. In this instance ACLs DO take precedence over POSIX permissions.

However, if a directory group or directory user encounters a folder on the AFP share with an ACL with an "Allow" Read/Write set for that group or user, but a POSIX/Unix permission of 770, the folder will be invisible to the user or group, which is not the expected behavior. ACLs Don't take precedence over POSIX permissions.

ACLs Precedence fails for Folders, but works partially for files:
Additionally,  if a directory group or directory user encounters a FOLDER on the AFP share with an ACL with an "Allow" Read/Write set for that group or user, but a POSIX/Unix permission of 774, the folder will not allow any new files to be created inside the folder, however existing files can be deleted and renamed AND new folders can be created--which is not the expected behavior. In regards to files with the ACL set for "Allow" Read/Write for that group or user and a POSIX/Unix permission of 770, the ACL take precedence and the file is readable and changeable by the network user, yet the preview for the file fails in MacOS Finder.

- - - - - - 
afpd 3.1.14 - Apple Filing Protocol (AFP) daemon of Netatalk

This program is free software; you can redistribute it and/or modify it under
the terms of the GNU General Public License as published by the Free Software
Foundation; either version 2 of the License, or (at your option) any later
version. Please see the file COPYING for further information and details.

afpd has been compiled with support for these features:

          AFP versions:	2.2 3.0 3.1 3.2 3.3 3.4 
         CNID backends:	dbd last tdb mysql 
      Zeroconf support:	mDNSResponder
   Admin group support:	Yes
    Valid shell checks:	Yes
            EA support:	ad | sys
           ACL support:	Yes
          LDAP support:	Yes
         D-Bus support:	Yes
     Spotlight support:	Yes

              afp.conf:	/usr/local/etc/afp.conf
           extmap.conf:	/usr/local/etc/extmap.conf
       state directory:	/usr/local/var/netatalk/
    afp_signature.conf:	/usr/local/var/netatalk/afp_signature.conf
      afp_voluuid.conf:	/usr/local/var/netatalk/afp_voluuid.conf
       UAM search path:	/usr/local/lib/netatalk/
  Server messages path:	/usr/local/var/netatalk/msg/

rdmark:
Rather than a general discussion on ACLs, rewrite the ACL chapter in the manual with a concrete discussion about Netatalk's interpretation and implementation across supported platforms.

Document the supported ACL backends, including a statement that native Darwin ACLs are unsupported, and describe Netatalk's AFP and POSIX ACL mappings.

### DIFF
--- a/doc/manual/ACL.md
+++ b/doc/manual/ACL.md
@@ -1,116 +1,103 @@
 # ACL Support
 
-ACL support for AFP is implemented for ZFS ACLs on Solaris and derived platforms,
-and for POSIX 1e ACLs on Linux.
+Netatalk maps ACLs on a supported server filesystem to AFP ACLs and effective
+AFP permissions. It does not provide one ACL implementation for every server
+platform or make ACLs override the host filesystem's access checks.
 
-## Configuration
+## Support matrix
 
-For basic operation, no additional configuration is needed.
-Netatalk reads ACLs on the fly and calculates effective permissions,
-which are then sent to the AFP client via the so-called UARights permission bits.
-On a Mac, the Finder uses these bits to adjust permissions in its windows.
-For example, a folder whose UNIX mode is read-only
-but whose ACL grants the user write access will display the effective read-write permission.
-Without this permission mapping,
-the Finder would display a read-only icon and the user would not be able to write to the folder.
+| Server platform and ACL backend | Support | AFP behavior |
+| --- | --- | --- |
+| Solaris/illumos with NFSv4/ZFS ACLs | Supported | Maps NFSv4 ACEs to and from the AFP/macOS ACL format. |
+| FreeBSD with `libsunacl`/ZFS ACLs | Supported | Uses the NFSv4/ZFS ACL implementation. |
+| Linux with POSIX 1e ACLs | Supported | Maps the less expressive POSIX ACL model to AFP; see [POSIX ACLs](#posix-acls). |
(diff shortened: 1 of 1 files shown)


andylemin (approved):
This is great!

I've been meaning to do something like this since I fixed part of the ACL implementation during the cache refactor (dircache was previously not storing ACLs properly). I only tested with ZFS ACLs.

Some clarity on the UID/GID to UUID requirements might be required. This currently reads like LDAP/AD is mandatory for ACLs.

rdmark:
@andylemin two paragraphs on LDAP-less UUID mapping added to the commit!

andylemin (approved):

andylemin:
Thanks @rdmark

PR #3261, opened 2026-08-21
webmin: restore use of ui_checked_columns_row elements

rdmark:
a historical bug in Webmin v2.111 triggered us to refactor ui_checked_columns_row elements to ui_columns_row with manually generated checkbox rows

the bug has resolved so restoring the more sophisticated structure

cf. https://github.com/webmin/webmin/issues/2164

also, update Webmin version to 2.660 in Dockerfile


rdmark:
for posterity, the workaround was applied May 2024 before the webmin module code moved to netatalk proper:

https://github.com/Netatalk/netatalk-webmin/commit/898fa279558455f85e35778ad3692578998cb3a8

andylemin (approved):

PR #3264, opened 2026-08-24
afpd: a stock config gets the documented 64K dircache

andylemin:
## Root cause

Two constants both mean "the dircache size default", and they drifted apart:

- `DEFAULT_DIRCACHE_SIZE = 65536` in `etc/afpd/dircache.h` — what `dircache_init()` applies, and what the man page, the docker image, and the webmin module all document.
- `DEFAULT_MAX_DIRCACHE_SIZE = 8192` in `include/atalk/globals.h` — what an unset `dircache size` actually resolves to in `afp_config_parse()`.

Because 8192 is above the 1024 minimum, `dircache_init()` accepted it verbatim — so **every stock configuration has been running an 8× smaller directory cache than documented**. Explicit `dircache size` settings were never affected.

## How the drift got here

| when | commit | what happened |
|---|---|---|
| 2012-02-17 | 6655a7d0 | volume loading moved to libatalk, bringing the config-side default constant (8192) with it — the documented default at the time |
| later | 9aa80804 | the dircache work raised the documented default to 65536 (man page and `dircache.h`), but the config-side constant in `globals.h` was not part of that change |

Two constants meaning "the default" diverged, and nothing tied them together.

## Defect → fix

| defect | fix |
|---|---|
| stock config resolves 8192 while all documentation says 65536; nothing prevents the two constants drifting apart again | collapsed into a **single** `DEFAULT_DIRCACHE_SIZE = 65536` in `globals.h`, which `dircache.h` already includes — one definition, nothing to drift |
| size-bounds behaviour was untestable (`dircache_maxsize` is a static set inside `dircache_init()`, which also builds live cache state) | resolution extracted into a pure `dircache_resolve_size()` helper; `dircache_init()` calls it |
| two stale comments describe the resolution as `max(DEFAULT_DIRCACHE_SIZE, min(size, MAX_DIRCACHE_SIZE))` — never the actual algorithm — and reference a nonexistent `DIRCACHE_FREE` | comments now describe the real resolution (unset/below-min → default, in-range → round up to power of two, oversize → clamp); `DIRCACHE_FREE_QUANTUM` |
| webmin labels the option "(bytes)" though the unit is entries, and its form cap (2097152) exceeds `MAX_DIRCACHE_SIZE` (1048576) | label and cap corrected |
| the shutdown statistics lines print the configured cache maximum unit-less (`config_max:`), inviting the same bytes-vs-entries misreading | renamed to `config_max_entries:` in both the ARC and LRU lines; lantest only matches the line prefix, so nothing parses the old label; `doc/developer/dircache.md` field list and examples updated |
| `doc/developer/dircache.md` still described the dircache as "by default just 8192 entries" — the very stale default this PR fixes | reworded to the corrected 65536 default |

## Behaviour changes

- Stock install: 8K → 64K-entry dircache (roughly ~1.5 MB → ~12 MB under LRU at full occupancy).
- Explicit `dircache size` values: unchanged. Bounds unchanged: minimum 1024 (deliberately kept — the spectest CI legs run `dircache size = 1024` to exercise cache exhaustion and eviction), maximum 1048576, in-range values still round up to the next power of two.
- The man page bounds line was already correct and is untouched.

## Size resolution

| requested `dircache size` | `dircache_resolve_size()` result |
|---|---|
| unset (≤ 0) | 65536 |
| 1 – 1023 | 65536, with a warning |
| 1024 – 1048576 | rounded up to the next power of two |
| > 1048576 | clamped to 1048576, with a warning |

## Testing

- New unit test `utest_conf_dircache_resolve_size` in `test/afpd/subtests_conf.c`: bounds matrix over the pure helper — `-1 → 65536`, `512 → 65536` (warn), `1024 → 1024`, `100000 → 131072`, `2000000 → 1048576` (clamp warn)
- `meson test` (Alpine container, CI build flags): 3/3 suites, 88 afpd subtests passed
- AFP spectest, ea=sys and ea=ad legs (single container, CI env including `dircache size = 1024`): both exit 0, zero failures

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,9 @@ Netatalk Changelog
 Changes in 4.6.0
 ----------------
 
+* FIX: afpd: a stock configuration now gets the documented 64K-entry
+  directory cache; a stale compiled default silently capped it at 8K
+  entries. Explicit `dircache size` settings are unaffected
 * UPD: afpd: `ea` is now a (G)/(V) option; a value in [Global] applies to all
   volumes and can be overridden per volume. A `[Global] ea =` setting that
   previous releases silently ignored now takes effect
--- a/contrib/webmin_module/edit_global_section.cgi
+++ b/contrib/webmin_module/edit_global_section.cgi
@@ -537,7 +537,7 @@ print &ui_table_row(
 @values = get_parameter_of_section($afpconfRef, $sectionRef, 'dircache size', \%in);
 print &ui_table_row(
                     $text{'edit_global_section_dircachesize'},
-                    "<input name='p_dircache size' type='number' min='0' max='2097152' value='"
+                    "<input name='p_dircache size' type='number' min='0' max='1048576' value='"
                     . $values[0] . "'>" . " "
                     . ($values[2] ? html_escape($values[2]) . ": " . html_escape($values[1]) : '') . "\n"
 );
--- a/contrib/webmin_module/lang/en
+++ b/contrib/webmin_module/lang/en
@@ -230,7 +230,7 @@ edit_global_section_cnid_mysql_db=CNID MySQL database name
 edit_global_section_cnid_scheme=CNID scheme
 edit_global_section_cnid_server=CNID server IP address and port
 edit_global_section_dbus_daemon=D-Bus daemon directory
-edit_global_section_dircachesize=Directory cache size (bytes)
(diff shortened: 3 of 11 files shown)


andylemin:
augment review

andylemin:
@rdmark ready for review

rdmark (approved):

PR #3265, opened 2026-08-24
afpd: default dircache mode to ARC

andylemin:
## What changes

A configuration that says nothing about `dircache mode` now gets **ARC** instead of LRU. Explicit `dircache mode = lru` remains the opt-out and behaves exactly as before; explicit `arc` is unchanged. An invalid value now falls back to ARC (previously LRU), with the warning message updated to match.

ARC has shipped opt-in for a full release cycle and delivers 10-50% better hit ratios than LRU (2× on sequential scans). With the cache-correctness work merged, the conservative-default question flips: the safe choice is to give every server the well-tested fast path and let memory-constrained systems opt down.

## Memory: what ARC actually costs

LRU and ARC use the **same amount of memory for the same number of live entries**. ARC *can* use up to 2× (~24 MB vs ~12 MB at the 64K default) because instead of dropping an evicted entry outright it may keep it as a **ghost** — signalled as expired, but retained — so a ghost that is referenced again is promoted back to a full entry without rebuilding it from disk. Ghosts that stay cold are truly dropped.

The man page entry previously overstated this as "ARC uses approximately 2× memory (~100% overhead)"; it now describes the ghost mechanics and frames 2× as the ceiling, not the steady state.

## Defect → fix

| change | where |
|---|---|
| missing-key default `lru` → `arc`; invalid-value fallback `lru` → `arc` (warning text updated) | `libatalk/util/netatalk_conf.c` |
| `(default: *lru*)` → `(default: *arc*)`; body rewritten with the ghost/promotion mechanics; recommendation reworded | `doc/manpages/man5/afp.conf.5.md` |
| webmin default `lru` → `arc` | `contrib/webmin_module/netatalk-lib.pl` |
| generated afp.conf hardcoded `dircache mode = ${AFP_DIRCACHE_MODE:-lru}`, which would have shadowed the new compiled default in every container; now emits the key only when `AFP_DIRCACHE_MODE` is set (same idiom as the `strict locking` line in the same heredoc) | `distrib/docker/env_setup_netatalk.sh` |
| NEWS entry | `NEWS.md` |

## Behaviour changes

- Stock install: LRU → ARC replacement policy; dircache memory ceiling rises from ~12 MB to ~24 MB at the default size (ghosts only, as they accrue).
- Docker image with `AFP_DIRCACHE_MODE` unset: previously wrote an explicit `lru` into afp.conf; now inherits the compiled default (arc). CI legs set the variable explicitly and are unaffected.
- Explicit settings everywhere: unchanged.

## Testing

- AFP spectest, ea=sys and ea=ad legs (single container, CI env): both exit 0, zero failures — both legs now exercise the ARC default continuously.


andylemin:
augment review

andylemin:
@rdmark ready for review

rdmark (approved):
sounds reasonable to make ARC the default now 👍

andylemin:
@dependabot rebase

rdmark:
> @dependabot rebase

I do this sometimes too on my own PRs :-D

PR #3266, opened 2026-08-24
afpd: enable the resource-fork data cache by default

andylemin:
## What changes

A configuration that says nothing about `dircache rfork budget` now gets **32768 KB (32 MB)** of resource-fork data cache instead of 0 (disabled). `dircache rfork budget = 0` remains the explicit opt-out; the hard caps (`RFORK_BUDGET_MAX_KB`, `RFORK_ENTRY_MAX_KB`) are unchanged.

What is eligible for caching is bounded per fork by the existing **`dircache rfork maxsize`** option, whose default — **1024 KB (1 MB) per resource fork** — is unchanged by this PR: only resource forks up to that size are ever cached; larger forks bypass the cache entirely regardless of remaining budget. So the default cache holds the small custom-icon/FinderInfo resource forks that dominate FPGetFileDirParams/FPEnumerate traffic — thousands of metadata only rforks, or a few dozen folder Icons (within the 32 MB budget) — while streaming-sized forks never enter it.

The default 32 MB budget is deliberately conservative: meaningful for metadata-heavy browsing, irrelevant on any modern server. With the cache-correctness hardening and the `ea = samba` carve-out both merged, the cache is safe to run everywhere it is coherent.
NB; For extensive Icons use, increase `dircache rfork budget` up to 10GB (10485760).

## `ea = samba` interaction — verified, no change needed

- Samba volumes stay excluded from the rfork cache: the Tier-2 gate in `fork.c` skips them unless the budget was set *explicitly*, and a compiled default keeps `dircache_rfork_budget_explicit` false.
- `apply_samba_defaults()` logs the "excluded from the rfork cache by default" note on samba volumes — the designed behaviour, previously vacuous with a default budget of 0.
- The ASP session path continues to zero the budget before `dircache_init()`; AppleTalk children stay cache-free.

## Docker: emit-only-when-set, and why that is load-bearing

The generated afp.conf hardcoded `dircache rfork budget = ${AFP_DIRCACHE_RFORK_BUDGET:-0}`, which would have shadowed the new compiled default in every container. It now emits the key only when `AFP_DIRCACHE_RFORK_BUDGET` is set (the same idiom as the `strict locking` line in the same heredoc).

Simply bumping the fallback value would have been wrong, not just untidy: a hardcoded value in the generated config makes the option **explicit** (`conf_key_present()`), and the fork.c gate honours an explicit budget *even on `ea = samba` volumes* — containers would have served cached resource-fork data that Samba can mutate behind afpd's back. Unset must mean unset for the samba carve-out to work.

## Defect → fix

| change | where |
|---|---|
| default budget 0 → 32768 | `libatalk/util/netatalk_conf.c` |
| `(default: *0*)` → `(default: *32768*)`; "set to 0 to disable" | `doc/manpages/man5/afp.conf.5.md` |
| webmin default `0` → `32768` | `contrib/webmin_module/netatalk-lib.pl` |
| emit `dircache rfork budget` only when the env var is set | `distrib/docker/env_setup_netatalk.sh` |
| NEWS entry | `NEWS.md` |
| manual pass: `Dircache.md` records the arc and rfork-budget defaults and states ARC's memory cost as a ceiling (up to 2x, from retained ghosts); `Upgrading.md`'s current `cnid scheme` default is sqlite | `doc/manual/Dircache.md`, `doc/manual/Upgrading.md` |

`dircache rfork maxsize` (per-fork cap, default 1024 KB) is intentionally untouched at every layer: config default, man page, webmin, and the docker entrypoint's `${AFP_DIRCACHE_RFORK_MAXSIZE:-1024}` fallback, which matches the compiled default.

## Behaviour changes

- Stock install: up to 32 MB of resource-fork cache (as entries accrue), holding only forks ≤ 1 MB each, on top of the dircache itself.
- Docker image with `AFP_DIRCACHE_RFORK_BUDGET` unset: previously wrote an explicit `0`; now inherits the compiled default *without* setting the explicit flag, so samba volumes in containers keep their exclusion. CI legs set the variable explicitly and are unaffected.
- Explicit settings everywhere: unchanged, including negative values still clamping to 0 and the 32-bit overflow clamp.

## Testing

- Existing `ea = samba` tests pass unchanged — their negative log assertions only cover paths where `apply_samba_defaults()` never runs, verified before landing.
- AFP spectest: ea=sys leg (no budget env — exercises exactly the new default path in a real server) and ea=ad leg (explicit budget) both exit 0, zero failures.


andylemin:
augment review

andylemin:
@rdmark ready for review

doc/manpages/man5/afp.conf.5.md
```diff
@@ -574,10 +574,10 @@ Example: 64K cache uses ~24 MB (ARC) vs ~12 MB (LRU).
 > **Recommendation**: Use **arc** for servers with 8GB+ RAM. Use **lru** for
 constrained systems. Increase **dircache size** (max entries) for best results.
 
-dircache rfork budget = *number* (default: *0*) **(G)**
+dircache rfork budget = *number* (default: *32768*) **(G)**
```
rdmark, line 580: how about we mention here in the documentation that the rfork dircache is ignored with `ea = samba` ? I wouldn't have realized it if you hadn't mentioned it!

doc/manual/Dircache.md
```diff
@@ -58,11 +58,11 @@ always update dircache entries immediately regardless of this setting.
 If Netatalk is the only process accessing the volume,
 you can safely set a value of 100 for maximum performance.
 
-dircache mode = *lru* | *arc* (default: *lru*) **(G)**
+dircache mode = *lru* | *arc* (default: *arc*) **(G)**
 
 Cache replacement algorithm. Netatalk supports two cache eviction policies:
 
-- **LRU (Least Recently Used)** — the legacy algorithm and the default.
+- **LRU (Least Recently Used)** — the legacy algorithm.
```
rdmark, line 65: nitpick: the cache algorithm updates are arguably a better fit for https://github.com/Netatalk/netatalk/pull/3265

rdmark (approved):
this is more of a broad point not directly related to the PR contents, but is it fair to say that "resource forks" in this context actually means any EA metadata, or is it literally applied only to Classic Mac OS resource forks? I'm thinking we could shore up documentation to explain to users when to apply this setting.

doc/manpages/man5/afp.conf.5.md
andylemin: Ok, fair - I will add "ignored when `ea = samba` is set" next to each of the settings it impacts 👍

andylemin:
> this is more of a broad point not directly related to the PR contents, but is it fair to say that "resource forks" in this context actually means any EA metadata, or is it literally applied only to Classic Mac OS resource forks? I'm thinking we could shore up documentation to explain to users when to apply this setting.

Yes, it is all resource forks including AppleDouble metadata and Icns for example. Not Extended Attributes per-se, just Resource Forks which are stored as Extended Attributes. Good idea, will do

doc/manual/Dircache.md
andylemin: Yes, some of the docs got blended a little bit otherwise it would have become difficult to rebase and merge the PRs

PR #3269, opened 2026-08-24
netatalk: start cnid_metad only when a volume uses the dbd scheme

### LINKED ISSUE #3268: netatalk: only start cnid_metad when a volume uses the dbd CNID scheme
The netatalk master starts `cnid_metad` whenever the dbd backend is compiled in (`CNID_BACKEND_DBD` gate in `etc/netatalk/netatalk.c`), regardless of whether any configured volume actually has `cnid scheme = dbd`. All three backends are compiled by default and the default scheme is sqlite (in-process), so a stock install runs an idle `cnid_metad`/`cnid_dbd` daemon that nothing talks to.

The startup decision should be runtime-conditional: start `cnid_metad` only when at least one volume resolves to the dbd scheme (and keep the compile gate for builds without dbd). Config reload would need to re-evaluate, since a reload can add or remove dbd volumes.

Noticed while cleaning up the sqlite-default residuals in #3267.

andylemin:
Fixes #3268.

## Problem

The netatalk master started `cnid_metad` whenever the dbd backend was merely *compiled in* (`CNID_BACKEND_DBD`), regardless of whether any volume has `cnid scheme = dbd`. All three backends compile by default and the default scheme is sqlite (in-process), so a stock install ran an idle `cnid_metad`/`cnid_dbd` pair that nothing talks to.

## Design

The master already has the machinery: it loads volumes before starting services and re-loads them on SIGHUP, and its timer re-launches any service whose pid sentinel is `NETATALK_SRV_NEEDED`. The change makes the needed/optional decision from the configuration in one small exported check plus three call sites:

- **`conf_cnid_scheme_in_use()`** (netatalk_conf.c): scans the loaded volume list, then the `[Homes]` section — homes volumes are instantiated per-user at login, so outside AFP sessions they never appear in the master's volume list; the section's scheme is resolved exactly as `creatvol()` resolves it (section → preset → global → compiled default). It lives in netatalk_conf.c because those resolution helpers are static there.
- **Startup**: skip `cnid_metad` when no volume resolves to dbd.
- **Timer restart guard**: the reaper marks a dead `cnid_metad` as `NETATALK_SRV_ERROR`, which equals `NETATALK_SRV_NEEDED` — without a guard the timer would resurrect a daemon a reload just made unnecessary; the guard demotes instead. This one check makes stop transitions converge through the existing machinery with no new state.
- **SIGHUP**: after the volume reload, stop a running daemon no dbd volume needs (SIGTERM → reap → guarded demotion) or mark it needed when one now does (next timer tick starts it). Both transitions are logged.

Robustness: if the master's volume load fails, an empty list is indistinguishable from "no volumes", so a `volumes_loaded` flag makes every decision fail conservative — the daemon runs, exactly as today. Builds without the dbd backend are unchanged: the pid sentinel can never become needed there (the SIGHUP re-evaluation is compile-gated), so the master never attempts to start a binary that isn't installed.

## Reload semantics

Start/stop on SIGHUP applies to **added or removed** volumes. Changing `cnid scheme` on an *existing* volume takes effect at restart only — `creatvol()` does not change options for volumes once loaded, which is netatalk's long-standing reload behaviour for all volume options; this PR matches it rather than changing it.

## Testing

All cases run against live containers built from this branch.

| build | configuration | expected | result |
|---|---|---|---|
| all backends (default) | `cnid scheme = sqlite` volumes | no `cnid_metad` | ✓ 0 processes |
| all backends | `cnid scheme = dbd` volumes | daemon running | ✓ |
| all backends | only share is `[Homes]`, global `cnid scheme = dbd` | daemon running | ✓ — the homes resolution path |
| all backends | SIGHUP: add a new dbd volume | started | ✓ "Starting 'cnid_metad': a volume now uses the dbd CNID scheme", process appears |
| all backends | SIGHUP: remove that volume | stopped, not restarted | ✓ "Stopping 'cnid_metad': …", cnid_metad logs "shutting down on SIGTERM", process gone, no restart — the timer guard held |
| all backends | `kill -9` a needed `cnid_metad` | restarted by the master | ✓ existing behaviour preserved |
| all backends | AFP spectest ea=sys and ea=ad legs | green | ✓ exit 0, zero failures |
| all backends | `meson test` | green | ✓ zero failures |
| **no dbd** (`-Dwith-cnid-backends=sqlite,mysql`) | build + unit suite | green; `cnid_metad` not installed | ✓ |
| no dbd | sqlite volume, live master | master + afpd run, no metad | ✓ |
| no dbd | volume **explicitly** `cnid scheme = dbd` | inert: no metad, no start-error loop | ✓ — the volume fails at client open as on any backend-less build; `cnid_open()` has no backend fallback (verified) |
| **dbd only** (`-Dwith-cnid-backends=dbd`) | stock config, **no scheme set anywhere** | daemon running | ✓ — the compiled default is dbd here, exercising the check's default-resolution end to end |
| dbd only | volume explicitly `sqlite` (unavailable backend) | no metad | ✓ — masks nothing; the volume fails at open, pre-existing |

(The dbd-only build cannot run the unit suite: `with-tests` uses the sqlite backend because the cnid_metad / cnid_dbd daemon combo needs root privileges to run, so dbd was deliberately left out of the afpd tests.)

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,10 @@ Netatalk Changelog
 Changes in 4.6.0
 ----------------
 
+* UPD: netatalk: cnid_metad is now started only when a volume uses the
+  dbd CNID scheme, and started or stopped on config reload as volumes
+  change; previously it always ran when the dbd backend was compiled in,
+  GitHub #3269
 * UPD: afpd: the resource-fork data cache is now enabled by default
   (`dircache rfork budget = 32768`, 32 MB). `ea = samba` volumes remain
   excluded (Samba mutates resource forks behind afpd's back). Set
--- a/etc/netatalk/netatalk.c
+++ b/etc/netatalk/netatalk.c
@@ -75,6 +75,9 @@ static pid_t cnid_metad_pid = NETATALK_SRV_NEEDED;
 static pid_t cnid_metad_pid = NETATALK_SRV_OPTIONAL;
 #endif
 static pid_t dbus_pid = NETATALK_SRV_OPTIONAL;
+/* whether the volume list loaded; without it the cnid_metad
+ * decisions fail conservative (run the daemon) */
+static bool volumes_loaded;
 static uint afpd_restarts, cnid_metad_restarts, dbus_restarts _U_;
 #ifdef WITH_LIBEV
 static struct ev_loop *loop;
@@ -386,7 +389,10 @@ static void sighup_impl(void)
 {
     LOG(log_note, logtype_afpd,
         "Received SIGHUP, sending all processes signal to reload config");
-    load_afp_conf_vols(&obj, LV_ALL);
+
+    if (load_afp_conf_vols(&obj, LV_ALL) == 0) {
+        volumes_loaded = true;
+    }
 
     if (!(obj.options.flags & OPTION_NOZEROCONF)) {
         zeroconf_deregister();
@@ -395,6 +401,23 @@ static void sighup_impl(void)
     }
 
     kill_childs(SIGHUP, &afpd_pid, &cnid_metad_pid, NULL);
+#ifdef CNID_BACKEND_DBD
+
+    if (volumes_loaded) {
+        bool dbd_in_use = conf_cnid_scheme_in_use(&obj, "dbd");
+
+        if (service_running(cnid_metad_pid) && !dbd_in_use) {
+            LOG(log_note, logtype_afpd,
+                "Stopping 'cnid_metad': no volume uses the dbd CNID scheme");
+            kill_childs(SIGTERM, &cnid_metad_pid, NULL);
+        } else if (!service_running(cnid_metad_pid) && dbd_in_use) {
+            LOG(log_note, logtype_afpd,
+                "Starting 'cnid_metad': a volume now uses the dbd CNID scheme");
+            cnid_metad_pid = NETATALK_SRV_NEEDED;
+        }
+    }
+
+#endif
 }
 
 /*! SIGCHLD implementation, returns true if all services have exited during shutdown */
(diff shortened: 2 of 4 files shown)


andylemin:
@rdmark ready for review

NEWS.md
```diff
@@ -4,6 +4,9 @@ Netatalk Changelog
 Changes in 4.6.0
 ----------------
 
+* UPD: netatalk: cnid_metad is now started only when a volume uses the
+  dbd CNID scheme, and started or stopped on config reload as volumes
+  change; previously it always ran when the dbd backend was compiled in
```
rdmark: nitpick, but I'd love to get back to referencing issue or PR tickets for each changelog item: GitHub https://github.com/Netatalk/netatalk/pull/3269

rdmark (approved):
this is a nice improvement! one more step towards completely decoupling from the old dbd structure

rdmark:
> (The dbd-only build cannot run the unit suite: meson's with-tests requires the sqlite backend — a build-system constraint, not a gap.)

not strictly speaking a build system constraint, rather that root privileges are required to run the cnid_metad / cnid_dbd daemon combo so I decided not to use it for the afpd tests

we could easily add support for the mysql backend in the afpd tests if it's valuable

NEWS.md
andylemin: Resolved across all of them, thanks for the call out - fell back to a bad habit. Agree that all updates should have traceable provenance 👍

andylemin: Done — this entry now ends with `, GitHub #3269`, and the whole 4.6.0 section has been brought back to the historical convention (every entry references its PR).

andylemin:
@rdmark feedback resolved

PR #3271, opened 2026-08-25
Spotlight cnid search: honor every "word" of a multi-word Finder search query, quoted phrases, folder scope, and the results limit

andylemin:
The cnid Spotlight (Finder search) backend served only a fraction of what Finder asks for. Four defects:

1. **Multi-word searches matched only the first word.** Finder splits `two words` into one predicate pair per word joined with `||`; extraction returned after the first supported predicate, so the OR the client asked for was dropped.
2. **Quoted phrase searches returned nothing.** `"two words"` arrives as `*=="\"two words\"*"cdw`; the closing-quote scan used a plain `strchr(…, '"')`, stopped at the `\"` delimiter, and extracted a lone `\` — below `SL_CNID_MIN_TERMLEN`, so zero results.
3. **Folder-scoped searches returned the whole volume.** Finder sends the folder in `kMDScopeArray` and the RPC layer parsed it, but this backend never consulted it.
4. **The result limit was ignored** by this backend, and the option it came from (`sparql results limit`) meant something different in each of the other two.

## Fix

**Extraction.** `sl_cnid_extract_terms()` collects up to 8 terms from every supported filename predicate (`kMDItemFSName`, `kMDItemDisplayName`, `_kMDItemFileName`, then `*==` only if no named attribute matched), requiring the `=` to directly follow the key so a key appearing inside another predicate's value cannot bind elsewhere. `sl_cnid_quoted_value()` scans to the first *unescaped* quote, strips wildcards, drops phrase delimiters and unescapes. One search runs per term and the results are united, dropping CNIDs matched by more than one term. Phrase adjacency needs no new machinery: the CNID search is a filename substring match, so `two words` only matches adjacent occurrences.

**Scope, enforced in the database** (for the supported sqlite and mysql schemes; see the dbd note below). Post-filtering an unscoped search cannot be correct: the candidate buffer caps at 10000 volume-wide matches, so on a large volume it fills with out-of-scope entries before an in-scope one is seen. The schema stores `(Id, Name, Did)` — parent links, no paths — so the scope path is resolved once per query to its directory CNID with `cnid_for_path()`, and the new `cnid_find_scoped()` restricts matching to that subtree:

- **sqlite / mysql** — a recursive CTE seeded with the scope DID, joined scope-set-first (`CROSS JOIN` / `STRAIGHT_JOIN`) so both the subtree walk and the name match drive off the `(Did, Name)` index: cost follows the subtree, not the volume. A server without recursive CTEs (MySQL < 8.0 / MariaDB < 10.2), or one refusing the recursion depth at execute time, falls back to the unscoped query.
- **dbd** — *not a target of this work: dbd is deprecated since the CNID default moved to sqlite.* It gets a minimal implementation only, so a dbd volume is not left worse off: the daemon checks each name-index match's ancestor chain against the scope, with a per-request verdict cache, and the scope rides in the previously unused `rqst.did` field of the SEARCH op so the wire format is unchanged and version skew degrades to unscoped behaviour in both directions. It gets none of the optimisation the SQL schemes do — BDB has no subtree index, so the scan is unchanged — and its error handling is deliberately left incomplete (see below). The supported schemes are sqlite and mysql.

Scope paths are canonicalised at parse time: trailing slashes stripped, and the components *below* the volume root converted from the client's UTF8-MAC form per component — the volume path prefix is left untouched, since it already holds volume-charset bytes that this volume's casefold and precomposition rules must not rewrite. A scope that is not inside the volume, or that escapes it via `..`, falls back to the volume root rather than reaching `cnid_for_path()`, whose contract requires a path inside the volume; an over-long scope is an error, not a silent truncation to an ancestor. A path-prefix check on resolved results covers what the database-side scope cannot: daemon version skew, the mysql fallback, and renames racing the query.

**The result limit — option renamed `sparql results limit` → `spotlight results limit`.** (It was never SPARQL-specific; the old name remains a deprecated alias that logs a warning, and the webmin module registers the rename so an existing key is shown and cleaned up). It now defaults to 10000 and means the same thing everywhere — `0` removes the limit for every backend, where previously `0` meant unlimited for localsearch, 10000 for xapian, and nothing at all for cnid, which could not exceed 10000 results by construction. A nonzero value is clamped to `[100, 16000000]`; negative or unparsable values fall back to the default. The end-of-life *dbd* scheme is excluded and keeps a fixed 10000 cap. The cnid candidate buffer is now sized from the limit, and for an unlimited search it starts at 10000 and grows until the database reports no further matches. Directory path resolution is memoized across results, cutting `cnid_resolve()` round-trips for every multi-result search. Zero-result replies pack CNIDs and filemeta as header-only containers, which the unmarshaller rejected as malformed — accepted now, with the empty CNID element added to its container like the empty filemeta beside it, plus guards on the three RPC handlers that read the first CNID of a client-supplied array without checking it holds one, and an empty array is treated as no filter rather than one that matches nothing.

`cnid_find()` keeps its public signature and behaviour; the `cnid_db` backend function-pointer for find gains the scope parameter (installed header — noted BREAKING in NEWS).

## Coverage

| Finder query (wire form) | before | after |
|---|---|---|
| single word | ✓ | ✓ |
| single predicate containing a space | ✓ | ✓ |
| quoted phrase — `*=="\"two words\"*"cdw` | ✗ 0 results | ✓ substring `two words` |
| multi-word, `‖`-joined per-word predicates | ✗ first word only | ✓ union of all words |
| name matching several words | n/a | ✓ counted once |
| under-length term among valid ones | ✗ aborted extraction | ✓ skipped, rest honored |
| key appearing inside another predicate's value | ✗ could bind a foreign value | ✓ ignored |
| scoped to a folder (`kMDScopeArray`) | ✗ whole volume | ✓ subtree only, complete to the candidate cap |
| scope with a trailing slash / non-ASCII name | ✗ 0 results | ✓ resolves |
| scope outside the volume or via `..` | n/a | ✓ clamped to the volume root |
| result limit | ✗ ignored by cnid; different meaning per backend | ✓ one option, same meaning everywhere |

Content queries (`kMDItemTextContent` alone) still return zero results gracefully — the CNID database indexes names only, by design.

## Behaviour changes

- All filename predicates are searched and united; a name matching several words appears once. With more than one term the buffer is sorted for dedup, so result order changes — Spotlight result sets are unordered; single-term queries keep the previous order.
- An under-length term no longer aborts extraction. At most 8 terms per query (one per typed word in practice), logged when capped.
- Folder-scoped searches return only in-scope results, complete up to the same 10000-candidate cap unscoped searches always had.
- **Option renamed:** `sparql results limit` → `spotlight results limit`. The old name still works and logs a deprecation warning, so existing configurations keep running; update them at your convenience.
- `spotlight results limit` now defaults to **10000** and caps every backend identically; `0` is unlimited. Deployments that relied on the old unlimited-by-default behaviour should set `0` explicitly.
- Docs updated: NEWS, the manual's Search page, `afp.conf(5)` and the developer indexing notes, referring to the feature as *Spotlight (Finder search)* so users know macOS searches network volumes through Finder search and not the menu-bar widget.

## Testing

**test632** — seven Finder search query shapes against a three-file fixture: plain-space term, quoted phrase, per-word OR (a fixture file matching both terms proves dedup), the universal Finder form with `kMDItemTextContent` predicates that must be ignored, a key embedded in another predicate's value, and short-term-alone / short-term-then-valid.

**test633** — six Finder search scope phases: unscoped baseline, scoped to a subdirectory, trailing-slash scope, a scope named in the client's decomposed form against a precomposed on-disk name, a `..`-escaping scope, and a scope naming another volume (plus one merely sharing a path prefix). The scope is sent through a new `FPSpotlightOpenQueryScoped()` helper that adds `kMDScopeArray` exactly as Finder does, deriving the server-side path from `FPSpotlightOpen`'s reply.

Both were written first and recorded failing against the unmodified backend, then green after. Verified locally: full spectest **290/290**; `FPSpotlightRPC` **8/8 on all three CNID schemes** (dbd, sqlite, mysql with embedded MariaDB); 8/8 with a decomposed non-ASCII volume path; ASAN clean over the extractor across the escape, phrase, short-term and foreign-key inputs; clang `-Werror`, doxygen-strict and markdownlint gates clean.

**Unit tests** cover the limit option end to end: unset, an explicit value, an explicit `0`, a value below the minimum and one above the maximum, a negative and an unparsable value, an empty canonical key falling through to the alias, and canonical-beats-alias precedence, plus the deprecation warning.

Standalone harnesses additionally cover extraction and union/dedupe, the per-term buffer reservation (a broad first term must not starve later words — verified failing before the fix and passing after), the scope boundary check (`/srv/afp2` vs `/srv/afp`), the limit clamp, and the path memo. The scoped SQL was validated against the real cnid sqlite schema with `EXPLAIN QUERY PLAN` confirming both the recursion and the outer join use the `(Did, Name)` covering index.

The limit was also verified behaviourally rather than by unit test alone: a limit of 20 is raised to the 100 minimum and truncates a 250-file corpus to exactly 100, an unlimited search grows the candidate buffer (`10000 → 80000 → …`) until the database stops reporting matches, and the dbd cap fires only on dbd volumes.

## Not in this PR

- Quoted-phrase support in the localsearch and xapian backends (both still fail to parse phrase queries).
- Date-range filters (`InRange(…)`) as a post-hoc filter from the per-result `stat()`.
- A uniform case/diacritic contract: `cd` insensitivity currently follows each CNID scheme's matching (SQLite `LIKE` is ASCII-only; MySQL follows the table collation).
- An `afp_spotlight` CLI over the same testsuite client library, usable as a live smoke test.
- Hardening the deprecated `cnid_dbd` scheme: a BDB engine error during the ancestor walk is still treated as out-of-scope rather than failing the search, and its scoped search cannot reach subtree cost. dbd is not a target of this work.
- Pre-existing in the xapian backend: the limit is applied to the candidate window *before* scope and permission filtering, so a scoped search can under-deliver while reporting no truncation.
- Pre-existing in `sl_rpc_fetchAttributeNamesForOIDArray()`: the `EC_CLEANUP` path dereferences the reply array whose allocation failure triggered the jump (out-of-memory only).

---

```mermaid
%%{init: {'flowchart': {'curve': 'basis'}}}%%
flowchart LR
    classDef wire    fill:#1e3a8a,stroke:#1e40af,color:#eff6ff,stroke-width:1px;
    classDef extract fill:#5b21b6,stroke:#6d28d9,color:#f5f3ff,stroke-width:1px;
    classDef db      fill:#0f766e,stroke:#0d9488,color:#f0fdfa,stroke-width:1px;
    classDef post    fill:#b45309,stroke:#d97706,color:#fffbeb,stroke-width:1px;

    Q("Finder query + kMDScopeArray<br/><i>one predicate pair per word, ‖-joined;<br/>phrases in escaped quotes</i>"):::wire
    X("extract terms<br/>named attrs → *== fallback<br/>'=' must follow the key<br/>dedupe, max 8"):::extract
    S("scope → CNID<br/>canonicalise, convert below<br/>the volume root, clamp escapes<br/>cnid_for_path(), once"):::extract
    F("cnid_find_scoped() per term<br/>sqlite/mysql: recursive CTE on (Did, Name)<br/>dbd: ancestor check in the daemon"):::db
    U("unite + dedupe<br/><i>a name matching several<br/>words counts once</i>"):::post
    L("'spotlight results limit' cap<br/><i>0 = unlimited</i>"):::post
    R("CNID → path<br/>memoized ancestor walk<br/>scope guard + stat + access(R_OK)<br/>20 results per reply page"):::post

    Q --> X --> F
    Q --> S --> F
    F --> U --> L --> R
```


andylemin:
Working well so far. Found some minor edge cases which need fixing first

andylemin:
Fixed some security issues where users could do `../othervol/word` to search other vols, and some oddities when using NFD encoded vol names. Working great and is very usable now the path filter is working correctly

andylemin:
@rdmark This is a joy to use! It is fast, works intuitively, and works by default without user effort.

DBD still has some limitation (is missing the new features), and we will not fix.. SQLite and MySQL work beautifully.
I think we need to say in the news as a headline that DBD is officially depreciated and is no longer receiving fixes or integration with new features, and we strongly recommend removing `cnid scheme = dbd` from all configurations. DBD will be removed entirely in upcoming releases.

Also updated the docs to better clarify; _macOS searches network volumes through Finder search only, not the Spotlight menu-bar widget, which is reserved for local volumes — macOS client behaviour, not a Netatalk limitation._

Ready for review

doc/developer/indexing.md
```diff
@@ -78,12 +78,12 @@ flowchart TD
 The Meson option `with-spotlight-backends` controls which backends are
 built. Supported values are `cnid`, `localsearch`, and `xapian`.
 
-The global `sparql results limit` option is shared by LocalSearch and
-Xapian despite its historical name. A value of `0` leaves LocalSearch
-SPARQL queries unlimited. Xapian applies a 10000 candidate-result safety
-cap in that case because the backend materializes candidate paths before
-paging validated results back to the client. The CNID backend has its own
-fixed CNID search reply cap.
+The global `sparql results limit` option is shared by all backends
```
rdmark: might be worth renaming this option now! we can add one of those compatibility fallbacks.

README.md
```diff
@@ -35,8 +35,8 @@ so Netatalk can act as a seamless bridge between new and old Macs, and 3rd party
 
 Compared to common file transfer protocols like NFS and FTP, Netatalk delivers a Mac-like user experience,
 with seamless integration of Mac filesystem metadata, such as the resource forks that were common before Mac OS X.
-Modern macOS features such as Zeroconf (Bonjour) service discovery, Time Machine backups, and Spotlight indexed search
-are also supported.
+Modern macOS features such as Zeroconf (Bonjour) service discovery, Time Machine backups,
+and Spotlight (Finder search) indexed search are also supported.
```
rdmark: nitpick: remove "indexed search", it's redundant now

rdmark (approved):
massive quality of life improvement!

doc/developer/indexing.md
andylemin: Done. Now `spotlight results limit` 🙂

README.md
andylemin: Agreed, resolved

andylemin:
> massive quality of life improvement!

It is especially good if you are using modern file managers like "Forklift file manager". Finder is quite clunky, as each time you use Finder search, you have to click on the folder you are in (it always defaults to searching the local Mac).

File managers like Forklift and Pathfinder allow you to set Spotlight and "current folder" as the default. So you just work as normal, and when you search it automatically uses your current path context and Netatalk CNID, without extra clicks. It feels like a drop-in upgrade now.

PR #3277, opened 2026-08-27
netatalk: add unprivileged single-user mode

### LINKED ISSUE #1914: Allow netatalk to run as non-root
In this day and age, it's getting common for people to want to run services as regular users in a sandboxed environment, without the need for root access to the host system.

In the context of netatalk, you would want to for instance to spin up a quick AFP server for sharing the contents of your home directory to other Macs.

Right now, there's no logic or validation that stops one from launching the netatalk controller daemon and its child daemons as a regular user. In fact, it appears to mostly work. However, we run into (quite insidious) errors when trying to actually mount f.e. the home directory from a client.

I suspect that there's something about the logic for creation of and writing to the CNID database that assumes the controlling daemons run as root.

rdmark:
Allow a non-root user to run the controller with a safe, explicitly configured single-user AFP service. Restrict the mode to a user-owned SRP verifier and static SQLite volumes with explicit UUIDs and signatures, pass the mode through to afpd, and authenticate only the invoking account.

Require a private absolute PID path, reject privilege-dependent features, and disable configuration reloads. Keep the SQLite CNID database, WAL, and shared-memory files private to the serving user.

Add non-root SRP verifier initialization and robust password-input error handling in afppasswd, update the manuals, and cover command-line parsing.

### DIFF
--- a/bin/afppasswd/afppasswd.c
+++ b/bin/afppasswd/afppasswd.c
@@ -54,6 +54,7 @@
 #define OPT_ADDUSER (1 << 3)
 #define OPT_NOCRACK (1 << 4)
 #define OPT_RANDNUM (1 << 5)
+#define OPT_INITIAL (1 << 6)
 
 #define PASSWD_ILLEGAL '*'
 
@@ -112,6 +113,17 @@ static const unsigned char srp_N_bytes[SRP_NBYTES] = {
 
 static char buf[MAXPATHLEN + 1];
 static const unsigned char hextable[] = "0123456789ABCDEF";
+
+static char *get_password(const char *prompt)
+{
+    char *password = getpass(prompt);
+
+    if (password == NULL) {
+        fprintf(stderr, "afppasswd: unable to read password.\n");
+    }
+
+    return password;
+}
 
 static int randnum_make_keypath(const char *path, char *keypath,
                                 size_t keypath_size)
@@ -497,7 +509,8 @@ static int update_srp_passwd(const char *path, const char *name, int flags,
         if (p && name_len == (size_t)(p - line) && strncmp(line, name, name_len) == 0) {
             p++;
 
-            if (!(flags & OPT_ISROOT) && (*p == PASSWD_ILLEGAL)) {
+            if (!(flags & OPT_ISROOT) && !(flags & OPT_INITIAL)
+                    && (*p == PASSWD_ILLEGAL)) {
                 fprintf(stderr, "Your password is disabled. Please see your administrator.\n");
                 err = -1;
                 goto done;
@@ -553,11 +566,15 @@ static int update_srp_passwd(const char *path, const char *name, int flags,
 found_entry:
 
     /* Verify old password for non-root users */
-    if ((flags & OPT_ISROOT) == 0) {
+    if ((flags & OPT_ISROOT) == 0 && !(flags & OPT_INITIAL)) {
         /* For SRP, we can't verify old password from verifier alone.
          * We would need to recompute the verifier from the old password
          * and compare. */
-        passwd = getpass("Enter OLD AFP password: ");
+        if ((passwd = get_password("Enter OLD AFP password: ")) == NULL) {
+            err = -1;
+            goto done;
+        }
+
         /* Parse existing salt from file */
         unsigned char old_salt[SRP_SALT_LEN];
 
@@ -612,7 +629,11 @@ found_entry:
 
     /* Get new password */
     if (pass_len < 1) {
-        passwd = getpass("Enter NEW AFP password: ");
+        if ((passwd = get_password("Enter NEW AFP password: ")) == NULL) {
+            err = -1;
+            goto done;
+        }
+
         size_t passwd_len = strnlen(passwd, SRP_PASSWDLEN + 1);
 
         if (passwd_len > SRP_PASSWDLEN) {
@@ -641,7 +662,10 @@ found_entry:
 #endif
 
     if (pass_len < 1) {
-        passwd = getpass("Enter NEW AFP password again: ");
+        if ((passwd = get_password("Enter NEW AFP password again: ")) == NULL) {
+            err = -1;
+            goto done;
+        }
 
         if (strcmp(passwd, password) != 0) {
             fprintf(stderr, "afppasswd: passwords don't match!\n");
@@ -743,6 +767,51 @@ static int create_srp_file(const char *path, uid_t minuid)
     return err;
 }
 
+/* Create a private verifier file containing only the calling user. This is
+ * the bootstrap path for a rootless, single-user AFP server. */
+static int create_srp_file_for_user(const char *path, const char *name)
+{
+    int fd;
+    size_t namelen;
+    int n;
+    namelen = strnlen(name, sizeof(buf));
+
+    if (namelen == sizeof(buf) || namelen + SRP_FORMAT_LEN > sizeof(buf) - 1) {
+        fprintf(stderr, "afppasswd: username is too long.\n");
+        return -1;
+    }
+
+    if ((fd = open(path, O_CREAT | O_TRUNC | O_WRONLY, 0600)) < 0) {
+        fprintf(stderr, "afppasswd: can't create %s: %s\n", path, strerror(errno));
+        return -1;
+    }
+
+    if (fchmod(fd, 0600) < 0) {
+        fprintf(stderr, "afppasswd: can't set permissions on %s: %s\n", path,
+                strerror(errno));
+        close(fd);
+        return -1;
+    }
+
+    n = snprintf(buf, sizeof(buf), "%s:", name);
+    memset(buf + n, PASSWD_ILLEGAL, SRP_HEX_SALT_LEN);
+    n += SRP_HEX_SALT_LEN;
+    buf[n++] = ':';
+    memset(buf + n, PASSWD_ILLEGAL, SRP_HEX_V_LEN);
+    n += SRP_HEX_V_LEN;
+    buf[n++] = '\n';
+
+    if (write(fd, buf, n) != n) {
+        fprintf(stderr, "afppasswd: problem writing to %s: %s\n", path,
+                strerror(errno));
+        close(fd);
+        return -1;
+    }
+
+    close(fd);
+    return 0;
+}
+
 /* -------------------- RandNum (legacy) functions -------------------- */
 
 /* this matches the code in uam_randnum.c */
@@ -805,7 +874,10 @@ found_entry:
 
     /* need to verify against old password */
     if ((flags & OPT_ISROOT) == 0) {
-        passwd = getpass("Enter OLD AFP password: ");
+        if ((passwd = get_password("Enter OLD AFP password: ")) == NULL) {
+            err = -1;
+            goto update_done;
+        }
 
         if (convert_passwd(p, NULL, keyfd) < 0) {
             err = -1;
@@ -821,7 +893,11 @@ found_entry:
 
     /* new password */
     if (pass_len < 1) {
-        passwd = getpass("Enter NEW AFP password: ");
+        if ((passwd = get_password("Enter NEW AFP password: ")) == NULL) {
+            err = -1;
+            goto update_done;
+        }
+
         size_t passwd_len = strnlen(passwd, PASSWDLEN + 1);
 
         if (passwd_len > PASSWDLEN) {
@@ -864,7 +940,10 @@ found_entry:
 #endif /* USE_CRACKLIB */
 
     if (pass_len < 1) {
-        passwd = getpass("Enter NEW AFP password again: ");
+        if ((passwd = get_password("Enter NEW AFP password again: ")) == NULL) {
+            err = -1;
+            goto update_done;
+        }
     }
 
     if (strcmp(passwd, password) == 0 || pass_len > 0) {
@@ -956,7 +1035,7 @@ static void print_usage(void)
             "Usage (root): afppasswd [-cfr] [-a username] [-u minuid] [-p path] [-w string]\n");
 #endif
     fprintf(stderr,
-            "Usage (user): afppasswd [-r]\n");
+            "Usage (user): afppasswd [-cfr] [-p path] [-w string]\n");
     fprintf(stderr, "  -a user   add or update the named user\n");
     fprintf(stderr,
             "  -c        create and initialize password file or specific user\n");
@@ -1063,8 +1142,34 @@ int main(int argc, char **argv)
 
     if (flags & OPT_CREATE) {
         if ((flags & OPT_ISROOT) == 0) {
-            fprintf(stderr, "afppasswd: only root can create the RandNum password file.\n");
-            return -1;
+            const struct passwd *pwd;
+
+            if (flags & OPT_RANDNUM) {
+                fprintf(stderr,
+                        "afppasswd: only root can create a legacy RandNum password file.\n");
+                return -1;
+            }
+
+            if (flags & OPT_ADDUSER) {
+                fprintf(stderr, "afppasswd: only root can add another user.\n");
+                return -1;
+            }
+
+            if (!i && ((flags & OPT_FORCE) == 0)) {
+                fprintf(stderr, "afppasswd: SRP verifier file already exists.\n");
+                return -1;
+            }
+
+            if ((pwd = getpwuid(uid)) == NULL) {
+                fprintf(stderr, "afppasswd: can't get password entry.\n");
+                return -1;
+            }
+
+            if (create_srp_file_for_user(path, pwd->pw_name) != 0) {
+                return -1;
+            }
+
+            return update_srp_passwd(path, pwd->pw_name, flags | OPT_INITIAL, pass);
         }
 
         if (!i && ((flags & OPT_FORCE) == 0)) {
--- a/etc/afpd/afp_options.c
+++ b/etc/afpd/afp_options.c
@@ -182,7 +182,7 @@ static void show_paths(void)
  */
 static void show_usage(void)
 {
-    fprintf(stderr, "Usage:\tafpd [-d] [-F configfile]\n");
+    fprintf(stderr, "Usage:\tafpd [-du] [-F configfile]\n");
     fprintf(stderr, "\tafpd -h|-v|-V\n");
 }
 
@@ -191,10 +191,14 @@ void afp_options_parse_cmdline(AFPObj *obj, int ac, char **av)
     int c, err = 0;
     optind = 1;
 
-    while (EOF != (c = getopt(ac, av, "dF:vVh"))) {
+    while (EOF != (c = getopt(ac, av, "duF:vVh"))) {
         switch (c) {
         case 'd':
             obj->cmdlineflags |= OPTION_DEBUG;
+            break;
+
+        case 'u':
+            obj->cmdlineflags |= OPTION_UNPRIVILEGED;
             break;
 
         case 'F':
--- a/etc/netatalk/netatalk.c
+++ b/etc/netatalk/netatalk.c
@@ -17,11 +17,13 @@
 
 #include <errno.h>
 #include <fcntl.h>
+#include <getopt.h>
 #include <poll.h>
 #include <signal.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include <ctype.h>
 #include <sys/param.h>
 #include <sys/stat.h>
 #include <sys/resource.h>
@@ -29,6 +31,7 @@
 #include <sys/time.h>
 #include <sys/uio.h>
 #include <sys/wait.h>
+#include <unistd.h>
 
 #include <bstrlib.h>
 
@@ -63,6 +66,7 @@
 
 /* forward declarations */
 static pid_t run_process(const char *path, ...);
+static pid_t run_afpd(void);
 static void kill_childs(int sig, ...);
 static void netatalk_exit(int ret);
 
@@ -90,6 +94,10 @@ struct event *sigterm_ev, *sigquit_ev, *sigchld_ev, *sighup_ev, *timer_ev;
 static int in_shutdown;
 static const char *dbus_path _U_;
 
+/* The lock path is normally compiled in for a system service. In rootless
+ * mode it is supplied by the caller and must live in private user state. */
+static const char *lockfile_path = PATH_NETATALK_LOCK;
+
 /******************************************************************
  * Misc stuff
  ******************************************************************/
@@ -101,6 +109,214 @@ static bool service_running(pid_t pid)
     }
 
     return false;
+}
+
+static bool srp_is_the_only_uam(const char *uamlist)
+{
+    const unsigned char *p = (const unsigned char *)uamlist;
+    size_t count = 0;
+
+    if (p == NULL) {
+        return false;
+    }
+
+    while (*p != '\0') {
+        const unsigned char *start;
+        size_t len;
+
+        while (*p != '\0' && (isspace(*p) || *p == ',')) {
+            p++;
+        }
+
+        if (*p == '\0') {
+            break;
+        }
+
+        start = p;
+
+        while (*p != '\0' && !isspace(*p) && *p != ',') {
+            p++;
+        }
+
+        len = (size_t)(p - start);
+
+        if (len != strlen("uams_srp.so") || strncmp((const char *)start,
+                "uams_srp.so", len) != 0) {
+            return false;
+        }
+
+        count++;
+    }
+
+    return count == 1;
+}
+
+static bool dbpath_parent_is_writable(const char *path)
+{
+    char parent[MAXPATHLEN];
+    char *slash;
+    size_t len;
+
+    if (path == NULL || strlcpy(parent, path, sizeof(parent)) >= sizeof(parent)) {
+        return false;
+    }
+
+    len = strlen(parent);
+
+    while (len > 1 && parent[len - 1] == '/') {
+        parent[--len] = '\0';
+    }
+
+    slash = strrchr(parent, '/');
+
+    if (slash == NULL) {
+        strlcpy(parent, ".", sizeof(parent));
+    } else if (slash == parent) {
+        parent[1] = '\0';
+    } else {
+        *slash = '\0';
+    }
+
+    return access(parent, W_OK | X_OK) == 0;
+}
+
+static bool pidfile_path_is_private(const char *path)
+{
+    char parent[MAXPATHLEN];
+    char *slash;
+    struct stat st;
+    size_t len;
+
+    if (path == NULL || strlcpy(parent, path, sizeof(parent)) >= sizeof(parent)) {
+        return false;
+    }
+
+    len = strlen(parent);
+
+    while (len > 1 && parent[len - 1] == '/') {
+        parent[--len] = '\0';
+    }
+
+    slash = strrchr(parent, '/');
+
+    if (slash == NULL) {
+        strlcpy(parent, ".", sizeof(parent));
+    } else if (slash == parent) {
+        parent[1] = '\0';
+    } else {
+        *slash = '\0';
+    }
+
+    if (stat(parent, &st) != 0 || !S_ISDIR(st.st_mode) || st.st_uid != getuid()
+            || (st.st_mode & 0077) != 0) {
+        return false;
+    }
+
+    if (lstat(path, &st) == 0) {
+        return S_ISREG(st.st_mode) && st.st_uid == getuid();
+    }
+
+    return errno == ENOENT;
+}
+
+static int validate_unprivileged_config(void)
+{
+    struct stat st;
+    uid_t uid = getuid();
+
+    if (obj.options.flags & OPTION_DDP) {
+        fprintf(stderr, "netatalk: --unprivileged does not support AppleTalk.\n");
+        return -1;
+    }
+
+    if (obj.options.flags & OPTION_AFPSTATS) {
+        fprintf(stderr, "netatalk: --unprivileged does not support afpstats.\n");
+        return -1;
+    }
+
+    if (obj.options.force_user || obj.options.force_group
+            || obj.options.admingid != 0) {
+        fprintf(stderr,
+                "netatalk: --unprivileged does not support admin or forced identities.\n");
+        return -1;
+    }
+
+    if (obj.options.signatureopt == NULL || obj.options.signatureopt[0] == '\0') {
+        fprintf(stderr,
+                "netatalk: --unprivileged requires an explicit [Global] signature.\n");
+        return -1;
+    }
+
+    if (!srp_is_the_only_uam(obj.options.uamlist)) {
+        fprintf(stderr,
+                "netatalk: --unprivileged requires 'uam list = uams_srp.so'.\n");
+        return -1;
+    }
+
+    if (stat(obj.options.configfile, &st) != 0 || st.st_uid != uid
+            || (st.st_mode & 0022) != 0) {
+        fprintf(stderr,
+                "netatalk: --unprivileged requires a configuration file owned by the calling user and not writable by group or others.\n");
+        return -1;
+    }
+
+    if (stat(obj.options.srppasswdfile, &st) != 0 || !S_ISREG(st.st_mode)
+            || st.st_uid != uid || (st.st_mode & 0077) != 0) {
+        fprintf(stderr,
+                "netatalk: --unprivileged requires a mode-0600 SRP verifier file owned by the calling user.\n");
+        return -1;
+    }
+
+    if (INIPARSER_GETSTR(obj.iniconfig, INISEC_HOMES, "basedir regex",
+                         NULL) != NULL) {
+        fprintf(stderr, "netatalk: --unprivileged does not support [Homes] volumes.\n");
+        return -1;
+    }
+
+    if (!volumes_loaded || getvolumes() == NULL) {
+        fprintf(stderr,
+                "netatalk: --unprivileged requires at least one static volume.\n");
+        return -1;
+    }
+
+    for (const struct vol *vol = getvolumes(); vol != NULL; vol = vol->v_next) {
+        if (vol->v_cnidscheme == NULL || strcasecmp(vol->v_cnidscheme, "sqlite") != 0) {
+            fprintf(stderr, "netatalk: --unprivileged volume '%s' must use sqlite CNID.\n",
+                    vol->v_localname);
+            return -1;
+        }
+
+        if (vol->v_uuid == NULL || vol->v_uuid[0] == '\0') {
+            fprintf(stderr,
+                    "netatalk: --unprivileged volume '%s' requires an explicit volume uuid.\n",
+                    vol->v_localname);
+            return -1;
+        }
+
+        if (vol->v_flags & AFPVOL_SPOTLIGHT) {
+            fprintf(stderr,
+                    "netatalk: --unprivileged does not support Spotlight on volume '%s'.\n",
+                    vol->v_localname);
+            return -1;
+        }
+
+        if (access(vol->v_path, R_OK | X_OK) != 0
+                || (!(vol->v_flags & AFPVOL_RO) && access(vol->v_path, W_OK | X_OK) != 0)) {
+            fprintf(stderr,
+                    "netatalk: --unprivileged cannot access volume '%s' as the calling user.\n",
+                    vol->v_localname);
+            return -1;
+        }
+
+        if (!dbpath_parent_is_writable(vol->v_dbpath)) {
+            fprintf(stderr,
+                    "netatalk: --unprivileged requires a writable parent directory for vol dbpath of volume '%s'.\n",
+                    vol->v_localname);
+            return -1;
+        }
+    }
+
+    return 0;
 }
 
 #ifdef SPOTLIGHT_BACKEND_LOCALSEARCH
@@ -387,6 +603,12 @@ static void sigquit_impl(void)
 /*! SIGHUP implementation */
 static void sighup_impl(void)
 {
+    if (obj.cmdlineflags & OPTION_UNPRIVILEGED) {
+        LOG(log_note, logtype_afpd,
+            "Ignoring SIGHUP: configuration reload is disabled in unprivileged mode");
+        return;
+    }
+
     LOG(log_note, logtype_afpd,
         "Received SIGHUP, sending all processes signal to reload config");
 
@@ -471,8 +693,7 @@ static void timer_impl(void)
         afpd_restarts++;
         LOG(log_note, logtype_afpd, "Restarting 'afpd' (restarts: %u)", afpd_restarts);
 
-        if ((afpd_pid = run_process(_PATH_AFPD, "-d", "-F", obj.options.configfile,
-                                    NULL)) == -1) {
+        if ((afpd_pid = run_afpd()) == -1) {
             LOG(log_error, logtype_default, "Error starting 'afpd'");
         }
     }
@@ -633,7 +854,7 @@ static void kill_childs(int sig, ...)
 /*! this get called when error conditions are met that require us to exit gracefully */
 static void netatalk_exit(int ret)
 {
-    server_unlock(PATH_NETATALK_LOCK);
+    server_unlock(lockfile_path);
     exit(ret);
 }
 
@@ -669,6 +890,15 @@ static pid_t run_process(const char *path, ...)
     }
 
     return pid;
+}
+
+static pid_t run_afpd(void)
+{
+    if (obj.cmdlineflags & OPTION_UNPRIVILEGED) {
+        return run_process(_PATH_AFPD, "-d", "-u", "-F", obj.options.configfile, NULL);
+    }
+
+    return run_process(_PATH_AFPD, "-d", "-F", obj.options.configfile, NULL);
 }
 
 static void show_netatalk_version(void)
@@ -713,7 +943,8 @@ static void show_netatalk_paths(void)
 
 static void usage(void)
 {
-    printf("usage: netatalk [-F configfile] \n");
+    printf("usage: netatalk [-d] [-F configfile] \n");
+    printf("       netatalk -u -P pidfile [-d] [-F configfile] \n");
     printf("       netatalk -d \n");
     printf("       netatalk -v|-V \n");
 }
@@ -737,7 +968,16 @@ static bool any_volume_uses_localsearch(void)
 
 int main(int argc, char **argv)
 {
-    int c, ret, debug = 0;
+    static const struct option long_options[] = {
+        { "debug",        no_argument,       NULL, 'd' },
+        { "config",       required_argument, NULL, 'F' },
+        { "pidfile",      required_argument, NULL, 'P' },
+        { "unprivileged", no_argument,       NULL, 'u' },
+        { "version",      no_argument,       NULL, 'v' },
+        { NULL,            0,                 NULL,  0  }
+    };
+    int c, ret, debug = 0, unprivileged = 0;
+    const char *pidfile = NULL;
     sigset_t blocksigs;
 #ifndef WITH_LIBEV
     struct timeval tv;
@@ -745,7 +985,7 @@ int main(int argc, char **argv)
     /* Log SIGBUS/SIGSEGV SBT */
     fault_setup(NULL);
 
-    while ((c = getopt(argc, argv, ":dF:vV")) != -1) {
+    while ((c = getopt_long(argc, argv, ":dF:P:uvV", long_options, NULL)) != -1) {
         switch (c) {
         case 'd':
             debug = 1;
@@ -753,6 +993,15 @@ int main(int argc, char **argv)
 
         case 'F':
             obj.cmdlineconfigfile = strdup(optarg);
+            break;
+
+        case 'P':
+            pidfile = optarg;
+            break;
+
+        case 'u':
+            unprivileged = 1;
+            obj.cmdlineflags |= OPTION_UNPRIVILEGED;
             break;
 
         case 'v':       /* version */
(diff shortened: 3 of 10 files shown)


rdmark:
@andylemin something that's been on my mind for a few years, now finally possible! this involved quite a lot of scaffolding so hardly a plug'n'play thing for beginners to run locally, but perhaps it could be the foundation of a variant of the containerized netatalk that runs non-privileged?

bin/afppasswd/afppasswd.c
```diff
@@ -743,6 +745,51 @@ static int create_srp_file(const char *path, uid_t minuid)
...
+    size_t namelen;
+    int n;
+    namelen = strnlen(name, sizeof(buf));
+
+    if (namelen == sizeof(buf) || namelen + SRP_FORMAT_LEN > sizeof(buf) - 1) {
+        fprintf(stderr, "afppasswd: username is too long.\n");
+        return -1;
+    }
+
+    if ((fd = open(path, O_CREAT | O_TRUNC | O_WRONLY, 0600)) < 0) {
```
andylemin, line 784: `afppasswd` is installed `rwsr-xr-x` (`bin/afppasswd/meson.build:28`) and nothing in the program drops euid, so this `open()` runs as root against whatever `-p` supplies — no `O_EXCL`, no `O_NOFOLLOW`, no path restriction.

How do we feel about `afppasswd -c -f -w x -p /etc/shadow` from an unprivileged account? It truncates the target and `fchmod`s it `0600`. Non-root `-c` was refused outright before, so this is the first path that gets there.

Would dropping to the real uid before creating, or requiring the path to resolve under the caller's own home plus `O_EXCL|O_NOFOLLOW`, be enough to close it?

bin/afppasswd/afppasswd.c
```diff
@@ -1063,8 +1110,34 @@ int main(int argc, char **argv)
...
+                        "afppasswd: only root can create a legacy RandNum password file.\n");
+                return -1;
+            }
+
+            if (flags & OPT_ADDUSER) {
+                fprintf(stderr, "afppasswd: only root can add another user.\n");
+                return -1;
+            }
+
+            if (!i && ((flags & OPT_FORCE) == 0)) {
```
andylemin, line 1158: With `-p` omitted, `path` is still `_PATH_AFPDSRPPWFILE`, so a plain `afppasswd -c -f` from a regular user lands on the shared system verifier and `create_srp_file_for_user()` truncates it — everyone else's entries included.

Should non-root `-c` require an explicit `-p`, and refuse a target it doesn't already own?

bin/afppasswd/afppasswd.c
```diff
@@ -553,7 +555,7 @@ static int update_srp_passwd(const char *path, const char *name, int flags,
 found_entry:
 
     /* Verify old password for non-root users */
-    if ((flags & OPT_ISROOT) == 0) {
+    if ((flags & OPT_ISROOT) == 0 && !(flags & OPT_INITIAL)) {
```
andylemin, line 569: `OPT_INITIAL` is only ever set immediately after `create_srp_file_for_user()`, so skipping the old-password proof is self-consistent in isolation. But the file it just wrote over may have held a real verifier, so combined with the default `-p` above, does this let someone set a new AFP credential without knowing the current one?

Would deriving this from "we created this file in this run and it was genuinely new" be tighter than a caller-supplied flag?

bin/afppasswd/afppasswd.c
```diff
@@ -497,7 +498,8 @@ static int update_srp_passwd(const char *path, const char *name, int flags,
         if (p && name_len == (size_t)(p - line) && strncmp(line, name, name_len) == 0) {
             p++;
 
-            if (!(flags & OPT_ISROOT) && (*p == PASSWD_ILLEGAL)) {
+            if (!(flags & OPT_ISROOT) && !(flags & OPT_INITIAL)
```
andylemin, line 512: Same thought on the `*` marker: `OPT_INITIAL` skips it, and the truncate upstream means an admin-disabled entry is gone by the time we reach here anyway. Is re-enabling a disabled account meant to be reachable without root?

bin/afppasswd/afppasswd.c
```diff
@@ -1063,8 +1110,34 @@ int main(int argc, char **argv)
...
+            if ((pwd = getpwuid(uid)) == NULL) {
+                fprintf(stderr, "afppasswd: can't get password entry.\n");
+                return -1;
+            }
+
+            if (create_srp_file_for_user(path, pwd->pw_name) != 0) {
+                return -1;
+            }
+
+            return update_srp_passwd(path, pwd->pw_name, flags | OPT_INITIAL, pass);
```
andylemin, line 1172: By this point `create_srp_file_for_user()` has already truncated the file and written the `*` placeholder, so a password mismatch, a cracklib rejection, or Ctrl-C leaves the previous verifier destroyed and the account disabled.

Recovery looks awkward too: `-c` without `-f` is refused above, and a plain update takes the non-`OPT_INITIAL` branch and prints "Your password is disabled. Please see your administrator." — in a mode whose premise is that there's no administrator.

Could we collect and validate the password first, then create and write once?

andylemin (commented):
Thanks for this — the restriction set is well chosen and `validate_unprivileged_config()` is thorough. A few things I'd like to talk through, mostly around the setuid-root `afppasswd` path and the config-parse/`daemonize()` reordering. Comments inline.

Two I think need resolving before this lands:

- the new non-root `afppasswd -c` path reaching `open(O_CREAT|O_TRUNC)` with euid 0 on a caller-supplied `-p` target
- `afp_config_parse()` moving ahead of `daemonize()`, which silently drops file logging for the default root service

One more that didn't fall on a changed line, so I couldn't anchor it: `update_srp_passwd()` dereferences `getpass()` without a NULL check (lines 562, 617, 646), which segfaults when there's no controlling terminal. That's pre-existing, but the manual now shows a script-shaped `afppasswd -c -p <path>`, so it's newly easy to reach — worth a cheap guard while we're here?

bin/afppasswd/afppasswd.c
```diff
@@ -743,6 +745,51 @@ static int create_srp_file(const char *path, uid_t minuid)
...
+/* Create a private verifier file containing only the calling user. This is
+ * the bootstrap path for a rootless, single-user AFP server. */
+static int create_srp_file_for_user(const char *path, const char *name)
+{
+    int fd;
+    size_t namelen;
+    int n;
+    namelen = strnlen(name, sizeof(buf));
+
+    if (namelen == sizeof(buf) || namelen + SRP_FORMAT_LEN > sizeof(buf) - 1) {
```
andylemin, line 779: This bounds against `buf` (`MAXPATHLEN + 1`), which admits roughly a 605-char name and a 719-byte record, while `update_srp_passwd()`'s own reader (`line[...]`, 675 bytes) and the parser in `uams_srp.c` use smaller line buffers — so a long name would be split across two `fgets()` reads.

Should this bound on `USERNAME_MAX_LEN` instead, to match what the readers can actually consume?

etc/netatalk/netatalk.c
```diff
@@ -769,26 +1012,59 @@ int main(int argc, char **argv)
...
+    } else if (getuid() != 0 || geteuid() != 0) {
+        fprintf(stderr,
+                "netatalk: must run as root; use --unprivileged (-u) for a single-user AFP server.\n");
+        exit(EXIT_FAILURE);
+    } else if (pidfile != NULL) {
+        fprintf(stderr, "netatalk: -P is only valid together with --unprivileged.\n");
+        exit(EXIT_FAILURE);
+    }
+
+    if (afp_config_parse(&obj, "netatalk") != 0) {
```
andylemin, line 1056: `afp_config_parse()` calls `setuplog()`, which opens the configured `log file` and caches the fd. `daemonize()` just below then `closeall(0)`s it, so the first `LOG()` gets `EBADF` and `make_log_entry()` sets that logtype's fd to -1 permanently — the controller goes silent for the rest of its life, including restart and shutdown diagnostics. `sighup_impl()` only reloads volumes, so there's no recovery path, and `afpd` still parses after daemonizing (`etc/afpd/main.c:338` then `345`).

The move is clearly needed so validation can see the parsed config. Would re-calling `setuplog()` after `daemonize()`, the way `etc/afpd/afp_dsi.c:961` does, be the cheapest way to keep both?

etc/netatalk/netatalk.c
```diff
@@ -103,6 +111,214 @@ static bool service_running(pid_t pid)
...
+    }
+
+    if (lstat(path, &st) == 0) {
+        return S_ISREG(st.st_mode) && st.st_uid == getuid();
+    }
+
+    return errno == ENOENT;
+}
+
+static int validate_unprivileged_config(void)
```
andylemin, line 222: This is a nice thorough gate. One gap: nothing checks `afp port`, which defaults to `548`. The `[Global]` block in the manual doesn't override it, so validation passes and then `afpd` can't bind a privileged port as a non-root process, exits, and `timer_impl()` re-forks it every second — `NETATALK_SRV_ERROR` is `#define`d equal to `NETATALK_SRV_NEEDED`, and there's no restart cap.

The bind failure is logged at `log_info`, below the default `note`, so there's no visible clue either. Worth rejecting a privileged port here?

etc/netatalk/netatalk.c
```diff
@@ -103,6 +111,214 @@ static bool service_running(pid_t pid)
...
+    }
+
+    if (!volumes_loaded || getvolumes() == NULL) {
+        fprintf(stderr,
+                "netatalk: --unprivileged requires at least one static volume.\n");
+        return -1;
+    }
+
+    for (const struct vol *vol = getvolumes(); vol != NULL; vol = vol->v_next) {
+        if (vol->v_cnidscheme == NULL || strcasecmp(vol->v_cnidscheme, "sqlite") != 0) {
```
andylemin, line 283: Requiring sqlite makes sense for this mode, but `cnid_sqlite_open()` `chmod`s the database to `0666` and, once the directory already exists, `fchmod`s it to `01777` (`libatalk/cnid/sqlite/cnid_sqlite.c:1332` and `1366`). `become_root()` is a no-op when `getuid() != 0`, so both succeed as the invoking user on files they own.

Does that undercut the private-state premise the rest of this function enforces? Any local account could then read every path name in the share out of the CNID db, and write to it.

etc/netatalk/netatalk.c
```diff
@@ -103,6 +111,214 @@ static bool service_running(pid_t pid)
...
+    }
+
+    if (stat(obj.options.configfile, &st) != 0 || st.st_uid != uid
+            || (st.st_mode & 0022) != 0) {
+        fprintf(stderr,
+                "netatalk: --unprivileged requires a configuration file owned by the calling user and not writable by group or others.\n");
+        return -1;
+    }
+
+    if (stat(obj.options.srppasswdfile, &st) != 0 || !S_ISREG(st.st_mode)
```
andylemin, line 263: This is the check the documented bootstrap trips over. Because `afppasswd` is setuid root, `afppasswd -c -p "$HOME/.config/netatalk/afppasswd.srp"` creates the file owned by root, so `st.st_uid != uid` fires here — and `afpd` running as the user couldn't read a root-owned `0600` file even without the check.

Should the bootstrap command in the manual and this check be reconciled, so the documented path works end to end?

etc/netatalk/netatalk.c
```diff
@@ -103,6 +111,214 @@ static bool service_running(pid_t pid)
...
+        return -1;
+    }
+
+    if (!srp_is_the_only_uam(obj.options.uamlist)) {
+        fprintf(stderr,
+                "netatalk: --unprivileged requires 'uam list = uams_srp.so'.\n");
+        return -1;
+    }
+
+    if (stat(obj.options.configfile, &st) != 0 || st.st_uid != uid
```
andylemin, line 256: Should the containing directory be checked too? A config sitting in a world-writable directory passes this test, and since `afpd` re-parses `-F` independently and enforces only the uid comparison in `login()`, anything substituted between here and `run_afpd()` gets served without any of these constraints applied.

etc/netatalk/netatalk.c
```diff
@@ -769,26 +1012,59 @@ int main(int argc, char **argv)
...
+        exit(EXIT_FAILURE);
+    }
+
+    if (afp_config_parse(&obj, "netatalk") != 0) {
+        exit(EXITERR_CONF);
+    }
+
+    volumes_loaded = (load_afp_conf_vols(&obj, LV_ALL) == 0);
+
+    if (unprivileged && validate_unprivileged_config() != 0) {
```
andylemin, line 1062: Worth calling this again on reload? `sighup_impl()` re-runs `load_afp_conf_vols()` with no re-validation, so a `SIGHUP` can bring in a `dbd` volume — which then sets `cnid_metad_pid = NETATALK_SRV_NEEDED` and has `timer_impl()` spawn the root-designed `cnid_metad` as the unprivileged user, respawning once a second.

etc/netatalk/netatalk.c
```diff
@@ -769,26 +1012,59 @@ int main(int argc, char **argv)
...
+
+        if (pidfile == NULL || pidfile[0] == '\0') {
+            fprintf(stderr,
+                    "netatalk: --unprivileged requires -P with a PID file in private user state.\n");
+            exit(EXIT_FAILURE);
+        }
+
+        lockfile_path = pidfile;
+
+        if (!pidfile_path_is_private(lockfile_path)) {
```
andylemin, line 1042: Should `-P` be required to be absolute? A relative path validates here and in `check_lockfile()` against the invoking cwd, but `daemonize()` does `chdir("/")` before `create_lockfile()`, so the daemon exits `EXITERR_SYS` trying to write `/netatalk.pid` — after the parent has already `_exit(0)`, so the caller sees success and the diagnostic goes to a `/dev/null`'d stderr.

With `-d` there's no `chdir`, so the same command works, which would hide it during testing.

etc/netatalk/netatalk.c
```diff
@@ -769,26 +1012,59 @@ int main(int argc, char **argv)
...
+    } else if (pidfile != NULL) {
+        fprintf(stderr, "netatalk: -P is only valid together with --unprivileged.\n");
+        exit(EXIT_FAILURE);
+    }
+
+    if (afp_config_parse(&obj, "netatalk") != 0) {
+        exit(EXITERR_CONF);
+    }
+
+    volumes_loaded = (load_afp_conf_vols(&obj, LV_ALL) == 0);
```
andylemin, line 1060: One side effect of parsing before `check_lockfile()`: a second `netatalk` invocation now loads all volumes first, so `get_vol_uuid()` can generate and append a UUID to `afp_voluuid.conf` and every volume path gets EA-probed, before we notice another instance is already running. Previously a duplicate start exited before touching anything.

Could the lockfile check move back ahead of this?

etc/netatalk/netatalk.c
```diff
@@ -103,6 +111,214 @@ static bool service_running(pid_t pid)
...
+            return false;
+        }
+
+        count++;
+    }
+
+    return count == 1;
+}
+
+static bool dbpath_parent_is_writable(const char *path)
```
andylemin, line 154: The trailing-slash strip and `strrchr` parent derivation here is identical to `pidfile_path_is_private()` just below — the two differ only in what they check afterwards. Since both feed security decisions, would a single shared helper be safer than keeping two copies in sync? I couldn't find a dirname-style primitive in `libatalk` to reuse, so a small static one here might be the way.

etc/afpd/afp_options.c
```diff
@@ -182,7 +182,7 @@ static void show_paths(void)
  */
 static void show_usage(void)
 {
-    fprintf(stderr, "Usage:\tafpd [-d] [-F configfile]\n");
+    fprintf(stderr, "Usage:\tafpd [-du] [-F configfile]\n");
```
andylemin, line 185: `doc/manpages/man8/afpd.8.md` still lists only `-d`, `-v`, `-V`, `-h` and `-F` — should `-u` be documented there to match this? (`netatalk.8.md` covers `-u` and `-P` nicely.)

test/afpd/test.c
```diff
@@ -492,6 +492,15 @@ int main(int argc, char *argv[])
                      "config loader fails closed under lock contention, state-neutral");
     TEST_int_or_skip(utest_conf_dircache_resolve_size(), 0,
                      "dircache_resolve_size: default, minimum, round-up, clamp");
+    {
+        AFPObj unprivileged_obj = { 0 };
+        char *unprivileged_args[] = { "afpd", "-u" };
+        TEST(afp_options_parse_cmdline(&unprivileged_obj, 2, unprivileged_args),
+             "parse afpd rootless command-line option");
+        TEST_expr(reti = 0,
```
andylemin, line 546: This asserts the flag bit gets set, which would stay green even if the `pw_uid != getuid()` check in `auth.c` were removed — and that's the only runtime enforcement point for the whole mode.

Worth adding a case asserting a uid mismatch is rejected and a match accepted, plus something over `validate_unprivileged_config()`? It carries most of the security contract and currently has no coverage.

andylemin (changes requested):

rdmark:
@andylemin just wanted to leave a note to say that I have this on my radar -- a lot of great feedback here that I want to work through properly

do you think it's okay to push this into the 4.6.0 release when all the feedback is addressed, or better to postpone?

andylemin:
> @andylemin just wanted to leave a note to say that I have this on my radar -- a lot of great feedback here that I want to work through properly
> 
> do you think it's okay to push this into the 4.6.0 release when all the feedback is addressed, or better to postpone?

We are getting close to the release date now, but at the same time, this is an "it either works or it doesn't change", so I don't mind.

rdmark:
@andylemin I basically concur with all feedback given :)

can you please take a closer look at the SQLite hacks to make the database files only owner accessible in unprivileged mode?

andylemin:
@rdmark rather than calling this unprivileged mode, I think we should call this single user mode.

Unprivileged makes this sound "less than", while it is actually "more than" as no one but the user can access anything, not even the CNID at rest.

One could argue this is a perfect design for home directories, where each user/home has its own single user process, and the CNID is kept separately under the homes path in a hidden location or something.

The SQLite in-process library-based design is perfect for single user mode, as it allows for many concurrent single user databases (being part of afpd they never collide). And you only need MySQL for multiuser mode (keeping the advanced machinery only where needed).

I see lots of value and potential in having both "private"/"single user" shares and standard multi user shares coexisting on the same server.

The more I think about it, the more elegant it appears;

---

Thought some more; Whilst not your original separate daemon/port intention I know, but the piece we might be missing is a per-volume `private = yes` option: the parent still federates every connection on 548 and the session child already runs as the authenticated user, but a "private volume" could just tells cnid_sqlite_open() to use a per-user database (vol dbpath already expands $u/$h) at 0700/0600 instead of the shared world-writable one — which means private and shared volumes can coexist on the same server with no new process model.

This class of change (separate DB for private vols) would also help to solve the Xapian/LocalSearch privacy issues (Xapian/LocalSearch would need separate daemons, but private SQLite CNID becomes a trivial solve for home directories)..

bin/afppasswd/afppasswd.c
andylemin: Thanks for the `get_password()` wrapper — that's the `getpass()` NULL path cleanly closed across all seven call sites.

This one still stands, though. `create_srp_file_for_user()` opens the caller-supplied `-p` path with `O_CREAT | O_TRUNC` while euid is still 0: the binary installs `rwsr-xr-x` (`bin/afppasswd/meson.build:28`) and the only credential call in the file is `getuid()` at line 1064, so nothing drops privilege first. `afppasswd -c -f -p /etc/shadow` from any account truncates it as root.

Cheapest shape is probably to drop to the real uid before touching anything a caller named — the non-root create path needs no root at all, since it only writes a file the user owns. Alternatively refuse `-p` outright when `geteuid() != getuid()`. Happy to look at either.

bin/afppasswd/afppasswd.c
andylemin: Still open as well, and it compounds the setuid issue above: with `-p` omitted, `path` is set to `_PATH_AFPDSRPPWFILE` (line ~1128) before the new non-root branch runs, so a plain `afppasswd -c -f` from a regular account calls `create_srp_file_for_user()` on the **shared system verifier** and replaces it with one entry for the calling user.

Requiring an explicit `-p` in the non-root create path would close it, or defaulting that path to something under `$HOME` when the caller is not root. Combined with the privilege drop, the blast radius goes away entirely.

etc/netatalk/netatalk.c
andylemin: This is the reordering rather than anything specific to the new mode, which is why I think it is worth fixing before merge. On main the order is `check_lockfile()` (line 772) then `daemonize()` (776) then `afp_config_parse()` (787); here the parse moved to 1056 and `daemonize()` runs at 1070.

`afp_config_parse()` calls `setuplog()` (`netatalk_conf.c:3196`), which opens the configured log file and caches the fd. `daemonize()` then calls `closeall(0)` (`libatalk/util/unix.c:153`), closing it. The first `LOG()` write fails and `make_log_entry()` closes and sets `type_configs[logtype].fd = -1` permanently (`logger.c:549-555`) — there is no reopen path — so file logging is lost for the whole run, including for the default root service.

Two ways out: call `setuplog()` again after `daemonize()`, or keep `daemonize()` ahead of the parse and run `validate_unprivileged_config()` after it.

etc/netatalk/netatalk.c
andylemin: Had a proper look at this, as asked. The afpd side is right: directory `0700` with the ownership check, db/WAL/SHM `0600`, and fail-closed when the chmod fails — that is exactly the posture the mode needs.

The one residual is that `unprivileged` is derived from `vol->v_obj->options.flags & OPTION_UNPRIVILEGED`, i.e. from the current process's command line. Any other tool the same user runs against that volume takes the `else` branch and chmods the database, WAL and SHM back to `0666` (and would create the directory `01777`); running `nad` once is enough to undo the privacy.

Keying the decision on the state already on disk rather than on the invocation would make it stick — e.g. if the CNID directory is ours and already `0700`, keep `0600` regardless of how this process was started.

etc/netatalk/netatalk.c
andylemin: This is the interlock between the two halves of the PR, and I think it is the one a first-time user hits soonest. The documented bootstrap is `afppasswd -c -p "$HOME/.config/netatalk/afppasswd.srp"`, run as the ordinary user — but because `afppasswd` is setuid root, the file is created owned by **root**, and this check then rejects it on `st.st_uid != uid`.

Following the manual end to end therefore cannot currently produce a verifier that passes validation. The privilege drop discussed on the `afppasswd` thread fixes this one at the same time.

etc/netatalk/netatalk.c
andylemin: Still no `afp port` check, and I think the effect is worse than I first realised because the documented `[Global]` in `doc/manual/Configuration.md` does not set one either.

So the manual's own config leaves afpd trying to bind 548 as a regular user. `dsi_tcp_init()` logs the failure at `log_info` (`libatalk/dsi/dsi_tcp.c:505`), which is below the default `note` level, afpd exits, and because `NETATALK_SRV_ERROR == NETATALK_SRV_NEEDED` the parent's one-second timer (`ev_timer_init(&timer_ev, timer_cb, 1.0, 1.0)`) restarts it indefinitely. The operator sees a climbing "Restarting 'afpd'" counter with no stated reason.

Validating `afp port` >= 1024 in the mode and adding it to the example config would make the documented path work first time.

bin/afppasswd/afppasswd.c
andylemin: Still create-then-prompt: `create_srp_file_for_user()` writes the placeholder record before `update_srp_passwd()` asks for the password, so a typo at the confirmation prompt or a Ctrl-C leaves a disabled verifier behind — and on an existing file, the previous contents are already gone.

Collecting and validating the password first and writing once would fix it. If you would rather keep the two functions separate, writing to a temp file in the same directory and `rename()`ing on success gives the same atomicity.

etc/netatalk/netatalk.c
andylemin: Still only the config file itself is stat'd, so a config in a world-writable directory passes, and its directory owner can swap the file between this validation and afpd's own re-parse.

`pidfile_path_is_private()` right above already derives and checks the parent directory — reusing it for the config path would close this and remove the asymmetry between the two checks.

etc/netatalk/netatalk.c
andylemin: Order is still parse (1056) then `load_afp_conf_vols()` (1060) then `check_lockfile()` (1066), so a second `netatalk` invocation loads every volume — appending UUIDs to `afp_voluuid.conf`, running the EA probes — before it discovers the lock and exits.

Moving `check_lockfile()` above the volume load keeps the duplicate-instance case free of side effects.

bin/afppasswd/afppasswd.c
andylemin: The trailing-slash strip and `strrchr` parent derivation is still duplicated between the two helpers. Since both are security checks, a single helper taking the path plus a "needs write" flag would keep them from drifting apart later.

bin/afppasswd/afppasswd.c
andylemin: `OPT_INITIAL` itself reads fine now that it is only ever set immediately after the create — thanks for tightening that.

The residual is inherited from the default-path thread rather than from this flag: because a non-root `-c -f` can still land on the shared verifier, `OPT_INITIAL` then skips the old-password proof on an entry the caller never owned. Fixing the path and privilege side closes this without any change here.

bin/afppasswd/afppasswd.c
andylemin: Same relationship as the old-password thread: the `*` skip is reasonable for a genuinely fresh file, and only becomes a re-enable-without-root path while a non-root `-c` can still target the shared verifier. No change needed here once that is fixed.

etc/netatalk/netatalk.c
andylemin: The trailing-slash strip and `strrchr` parent derivation is still duplicated between the two helpers. Since both are security checks, a single helper taking the path plus a "needs write" flag would keep them from drifting apart later.

etc/afpd/afp_options.c
andylemin: `doc/manpages/man8/afpd.8.md` still documents only `-d`, `-v`, `-V`, `-h` and `-F`, so `-u` remains accepted and advertised in the usage string but absent from the man page. The `netatalk.8.md` wording you added would carry over nicely.

test/afpd/test.c
andylemin: The new case is a good start — parsing `-u` and asserting the flag lands is worth having.

It does still pass if the behaviour behind the flag is removed: deleting the `pw_uid != getuid()` check in `auth.c` keeps this green, because nothing here exercises the refusal. A case that logs in with a non-matching uid and expects rejection would pin the part that matters.

andylemin:
Went back through the threads against `5444e7ce` — thanks for working through this, and sorry for the reopens: the intent is only to keep the remaining items visible, not to relitigate anything you have already done.

Genuinely closed, and I have left those resolved:

- the `getpass()` NULL path, via the new `get_password()` wrapper — all seven call sites converted, each failure taking an existing cleanup label
- `-P` must be absolute
- re-validation on `SIGHUP` — you solved this more strongly than I suggested by refusing reloads outright in the mode, which also closes the `dbd`-via-reload path that would have respawned `cnid_metad`

The SQLite work you asked me to look at is right for the afpd path; the only gap is that it keys on the process's own command line, so another tool run by the same user undoes it. Details on that thread.

Of what is left, two are the ones I would treat as merge blockers, both unchanged so far: the setuid-root `O_CREAT | O_TRUNC` on a caller-supplied `-p`, and the `setuplog()`-before-`daemonize()` reordering, which costs file logging for the default root service as well as the new mode. The rest are smaller, and two of them (the old-password proof and the `*` marker) fall out for free once the `afppasswd` path and privilege drop are sorted.

Worth flagging one thing that is not in any thread: the manual's own example cannot start a working server yet, because the `[Global]` block sets no `afp port` and the documented `afppasswd` bootstrap produces a root-owned verifier that `validate_unprivileged_config()` then rejects. Both are covered on their threads.

Happy to re-review as soon as you want another pass, and equally happy to take any of the mechanical ones off your hands if that helps get it in.

rdmark:
@andylemin I see your point about the naming -- "unprivileged" is a description of the mechanics rather than the utility for the end user. if we lean into the CNID database isolation in the documentation, I might favor `netatalk --private` as the name of the argument, since it clearly signals the value and outcome. what do you think?

and yes, if you are up for it I'm happy for you to take over this PR and take care of the finishing touches! I'm feeling under the weather this week so at a limited mental capacity.

if you like you can cherry-pick the current commit into a fresh feature branch.

andylemin:
@rdmark Hope you feel better soon. Take care

andylemin:
@rdmark Happy to take this over.

Let me have a think about how to approach it.
I understand the use-case as you have built it so far.

But I need to find the correct compromise so it works with the standard multi-user architecture (where the parent process federates users).

It needs to work in such a way that the standard model gains the same benefits; individual users can access both private vols/CNID databases _and_ shared vols/CNID databases.

I believe this is the more impactful and common use-case which can work by default for home shares, but needs the child afpd process to maintain multiple CNIDs.

SQLite fits the use case beautifully as its biggest limitation is multiple user contention. And its biggest strength is being in-process library.

So an enterprise setup would have a child proc with a MySQL connection for the master CNID holding all the shared vols, and then as many private in-proc SQLite objects as needed.
Still only a single socket to manage for the master, adding zero new admin overhead.

So requires some form of CNID backend array.
I'm leaning towards keeping the required master CNID exactly as-is, and having a separate array of optional references to only private SQLite objects.

This keeps the existing code structurally the same (where the master CNID can be different backends), and the new private vol array (which can only be SQLite). These private backends will need to be stored inline within the same private vols for both scalability, security and compartmentalisation.

andylemin:
@rdmark playing devils advocate; 

Given the above described functionality which offers a containerised-like vol access (with the private CNID inside the target vol), what is the remaining benefit of this PR as it stands?

The remaining feature it would offer would be a single-user parent. I think that would only benefit very specific use cases or locked down systems where an operator has no admin access (so would need manual start etc)?

What do you think?

andylemin:
This one is definitely for after the upcoming release

rdmark:
@andylemin the original use case I had in mind was a non-sysadmin user who have a regular account and want to share their home directory casually

it could be someone who installs Netatalk on macOS via Homebrew, for instance, where the default assumption/expectation is that an installed app runs as a non-root user

I've also toyed with the idea of a macOS GUI app that emulates old-school "personal file sharing" by launching Netatalk -- which obviously would run as a non-privileged user!

another use case would be for testing: maybe the spectest could by default launch its own netatalk process with a temporary file system

andylemin:
Working on this now when I can

andylemin (approved):

andylemin:
Approved by accident, not ready for merge. updates underway

andylemin:
Will refactor the plan for this after #3307 is merged as that changes some things for this..

bin/afppasswd/afppasswd.c
andylemin: Closed: afppasswd permanently drops an inherited setuid root to the real uid before it touches anything (main()), and the non-root create path opens only the caller's own mode-0700 directory, with O_EXCL|O_NOFOLLOW on the uid-named file, refusing a store that is not the caller's. afppasswd_test.sh covers the create and the refusals.

bin/afppasswd/afppasswd.c
andylemin: Closed: a non-root -c requires an explicit -p and refuses a store the caller does not own, so the shared system verifier stays root's alone. afppasswd_test.sh leg "non-root create without -p refused by policy".

bin/afppasswd/afppasswd.c
andylemin: Closed with the two threads above: update_srp_passwd() no longer has a bootstrap case, and create_private_srp_verifier() writes one complete record for a file the caller owns, so there is no first-run skip of the old-password proof on someone else's entry.

bin/afppasswd/afppasswd.c
andylemin: Closed the same way: the bootstrap never writes a placeholder record, and a non-root run cannot reach the shared verifier, so re-enabling a disabled account stays root's.

bin/afppasswd/afppasswd.c
andylemin: Closed: collect_new_password() runs first; the directory and the uid-named file are created only after the password is collected and confirmed, and the record is written once under the write lock through srp_write_record(). An aborted prompt leaves an existing verifier intact, with or without -f (afppasswd_test.sh legs 3 and 4).

bin/afppasswd/afppasswd.c
andylemin: Closed: the writer validates the name with srp_valid_username() against SRP_USERNAME_MAX_LEN and every reader (uams_srp.c and both afppasswd readers) sizes its line as SRP_USERNAME_MAX_LEN + SRP_FORMAT_LEN + 1, so a record cannot straddle two fgets() reads. The duplicated parent derivation is one helper now, path_parent().

etc/netatalk/netatalk.c
andylemin: Closed: daemonize() itself closes the syslog connection before closeall() and reopens every configured log once the standard descriptors are /dev/null again, so any daemon that logs before it daemonizes keeps its log; a mkstemp-form log file is reopened by its generated name. logreopen_test.sh covers the file, syslog and mkstemp cases, and start_root_netatalk in test/integration/lib.sh asserts the startup line on every daemonized root start.

etc/netatalk/netatalk.c
andylemin: Closed as a warning rather than a refusal: an afp port below 1024 prints a warning naming the fix, because macOS and Linux (ip_unprivileged_port_start) can both grant the bind. The manual's example sets afp port = 5548 and the CI leg runs on it. singleuser_test.sh leg "a privileged afp port warns without rejecting the config".

etc/netatalk/netatalk.c
andylemin: Closed: cnid_sqlite_open() keeps state owner-only when the single-user flag is set or when the non-root opener already holds the directory as its own mode-0700 directory, so the serving user's nad and dbd keep what the server keeps; a directory that does not exist yet is created shared and the server tightens it at its next open; root and a foreign-owned directory follow the shared rules. Legs: nad keeps owner-only state; nad before the first start creates shared state the server then tightens.

etc/netatalk/netatalk.c
andylemin: Closed: with the setuid privilege dropped, afppasswd -c -p creates the directory and the uid-named file as the calling user, and this check (a 0700 directory owned by the caller holding the caller's regular file with one link, owner-readable, no group or other bits, the UAM's own rule) passes for the documented bootstrap end to end. singleuser_test.sh runs that bootstrap and logs in over SRP.

etc/netatalk/netatalk.c
andylemin: Closed: the configuration file is judged by the same helper as the PID file, dir_is_private(): owned by the caller, no group or other write bits, inside a mode-0700 directory the caller owns. Legs "config file not private" and "config directory not private".

etc/netatalk/netatalk.c
andylemin: Closed: check_lockfile() runs before afp_config_parse(), so a second start against a running instance exits before any parsing or volume load, and a single-user volume without volume uuid is skipped at load instead of having one generated, so afp_voluuid.conf is never written. Legs: a second start is refused by the lock before its config is parsed; no load side effect on afp_voluuid.conf.

etc/netatalk/netatalk.c
andylemin: Closed: one path_parent() over dirname(3) derives the parent for both, owned_private_dir() is the shared "0700 directory owned by the caller" test, and dir_is_private() is the general private-path check the PID file and the configuration file both use.

etc/afpd/afp_options.c
andylemin: Closed: afpd.8 documents -u with the wording from netatalk.8.

doc/manpages/man1/afp_spectest.1.md
```diff
@@ -89,6 +89,10 @@ run the byte-range read-lock conflict tests instead of skipping them
 **-m**
 : Run tests in AppleShare (Mac) AFP server compatibility mode
 
+**-N**
+: Server under test runs without root privileges (netatalk --single-user);
+skip the tests that rely on root reading a file whose mode denies its owner
+
```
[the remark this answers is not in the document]
andylemin: Done in df8513dc: the synopsis lists [-1234567aCEiLlmNVv], matching usage().

libatalk/cnid/sqlite/cnid_sqlite.c
```diff
@@ -1844,23 +1856,56 @@ struct _cnid_db *cnid_sqlite_open(struct cnid_open_args *args)
...
+                    EC_FAIL;
+                }
+
+                if ((st.st_mode & 0777) != 0700 && fchmod(dirfd, 0700) != 0) {
+                    LOG(log_error, logtype_cnid,
+                        "Can't make CNID DB directory '%s' owner-only: %s",
+                        dirpath, strerror(errno));
+                    close(dirfd);
+                    EC_FAIL;
+                }
```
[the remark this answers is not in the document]
andylemin: Good catch, thank you. Fixed in df8513dc at startup: validate_singleuser_config() refuses 'vol dbnest = yes' ("does not support 'vol dbnest': the CNID directory would be the volume itself, which the mode keeps owner-only"), Configuration.md lists it among the unsupported options, and singleuser_test.sh has a "vol dbnest" leg.

libatalk/util/logger.c
```diff
@@ -351,6 +379,52 @@ static void log_setup(const char *filename, enum loglevels loglevel,
...
+    for (int i = 0; i < logtype_end_of_list_marker; i++) {
+        if (!type_configs[i].set || type_configs[i].filename == NULL) {
+            continue;
+        }
+
+        if (type_configs[i].fd != -1) {
+            close(type_configs[i].fd);
+        }
+
+        type_configs[i].fd = log_open_file(type_configs[i].filename);
```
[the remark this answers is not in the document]
andylemin: Right, thank you. Fixed in df8513dc: daemonize() now calls log_close_all() before closeall(), which closes the syslog connection and every log file and sets each fd to -1 while the numbers are still ours, and log_reopen() opens only types whose fd is -1, so it never closes a number another type was just given.

test/integration/singleuser_test.sh
```diff
@@ -0,0 +1,442 @@
...
+    }
+owner_afparg -s backup -f FPEnumerateExt > /dev/null \
+    || {
+        stop_netatalk $USER1
+        fail "the volume with usable state was not served"
+    }
+stop_netatalk $USER1
+grep -q "not owned by the server user" $LOG1 \
+    || fail "the backend's ownership diagnostic is missing from the log"
+chown -R $USER1 $CNID1
```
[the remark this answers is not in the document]
andylemin: Done in df8513dc: the desktop-database leg asserts $CNID1 is still 700 owned by the serving user once .AppleDesktop exists.

test/testsuite/afptest_srp.c
```diff
@@ -0,0 +1,292 @@
...
+/*!
+ * @brief M1 = SHA1((H(N) XOR H(g)) | H(user) | salt | A | B | K)
+ *
+ * A and B are the stripped (leading zeros removed) forms, matching Tom Wu's
+ * reference SRP-6a derivation and uams_srp.c.
+ *
+ * @param[out] m1_out  SRP_SHA1_LEN bytes
+ *
+ * @returns 0 on success, -1 on failure
+ */
```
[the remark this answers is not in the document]
andylemin: Done in 288c2a04: every parameter of sha1_multi(), srp_derive_x(), srp_compute_u(), srp_compute_proofs() and srp_compute_server_proof() is documented; the Doxygen job passes on df8513dc.

etc/netatalk/netatalk.c
```diff
@@ -103,6 +114,298 @@ static bool service_running(pid_t pid)
...
+        return false;
+    }
+
+    while (len > 1 && dir[len - 1] == '/') {
+        dir[--len] = '\0';
+    }
+
+    if (lstat(dir, &st) == 0) {
+        return S_ISDIR(st.st_mode) && access(dir, W_OK | X_OK) == 0;
+    }
```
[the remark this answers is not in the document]
andylemin: This one is deliberate: an existing CNID directory owned by another account is refused at the first mount by cnid_sqlite_open(), and only that volume fails, so a server with one bad volume still serves the others; the startup clause checks what the mode can fix at startup (writable, not a symbolic link, or creatable). The PR body lists it under "where coverage is deliberately not 100%", and singleuser_test.sh covers it ("foreign-owned CNID state refuses the mount of that volume only").

andylemin:
Hi @rdmark the five open sonarqube issues are all rule S7688, "Use `[[` instead of `[` for conditional tests", on `test/integration/lib.sh`. We can't fix these becuase:

- The integration scripts are `#!/bin/sh` and run under the Alpine testsuite image's `/bin/sh`, which is busybox ash. `[[ ]]` is a bash and ksh extension; POSIX sh leaves it undefined (shellcheck reports it as SC3010, "In POSIX sh, [[ ]] is undefined").
- busybox ash happens to accept `[[`, but Debian's `/bin/sh` is dash, which rejects it. A follow-up for this PR is to ship `lib.sh` into the Debian testsuite image as well, so the scripts must stay portable across the two.
- There is no bash in the Alpine testsuite image, so switching the shebang is not an option without growing the image.
- No other `sh` script in the repository uses `[[`; every `[` in these scripts is a plain string or file test with quoted operands, which is the case the rule is guarding against.

The other 44 findings from the first scan were fixed (redundant casts, an implicit size_t to int conversion, repeated string literals, positional parameters assigned to named variables, explicit `return` at the end of every function).

andylemin:
This has proven quite a tricky PR to complete! I would appreciate your analysis on these commits to ensure this is what you were originally planning before we squash

libatalk/util/logger.c
```diff
@@ -351,6 +379,62 @@ static void log_setup(const char *filename, enum loglevels loglevel,
...
+ */
+void log_reopen(void)
+{
+    for (int i = 0; i < logtype_end_of_list_marker; i++) {
+        if (!type_configs[i].set || type_configs[i].filename == NULL
+                || type_configs[i].fd != -1) {
+            continue;
+        }
+
+        type_configs[i].fd = log_open_file(type_configs[i].filename);
```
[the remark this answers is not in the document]
andylemin: Right, thank you. Fixed in 0b54f882: log_setup() stores the name through log_absolute_name() (the name as given when it starts with /, otherwise getcwd() joined with it), for the open() and the mkstemp() branches alike, so log_reopen() after daemonize()'s chdir("/") opens the file in the launch directory. logreopen_test.sh leg 5 starts the controller from /var/log with 'log file = afpd-rel.log' and asserts the startup line reaches /var/log/afpd-rel.log; it was RED before (the line landed in /afpd-rel.log).

test/integration/singleuser_test.sh
```diff
@@ -0,0 +1,476 @@
...
+pass "the serving user's nad keeps the CNID state owner-only"
+
+# --- state the user's nad creates before the server's first start is the
+# shared kind any non-root tool creates, and the server's first open makes it
+# owner-only
+rm -rf $CNID1
+su $USER1 -c "nad -F $CONF/afp.conf ls $SHARE" > /dev/null \
+    || fail "nad could not create the CNID state as its user"
+[ "$(stat -c %a $CNID1)" != 700 ] \
+    || fail "nad created owner-only state for a server that has not started"
```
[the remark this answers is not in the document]
andylemin: Done in 0b54f882: the leg asserts [ -d "$CNID1" ] before it judges the mode, so an absent directory fails with its own message instead of passing on stat's empty output.

test/integration/singleuser_test.sh
```diff
@@ -0,0 +1,484 @@
...
+# --- a connection that opens a DSI session and drops before logging in must
+# end its afpd child rather than wait out 'disconnect time'; the bytes are a
+# DSIOpenSession request. The child's own last line is the observable, and only
+# the listening afpd may remain: an exited child the parent has not reaped yet
+# is a zombie, not a survivor, so state Z is not counted.
+only_afpd_parent() {
+    [ "$(ps -o pid,stat,user,comm \
+        | awk -v u="$USER1" 'NR > 1 && $4 == "afpd" && $3 == u && $2 !~ /^Z/' \
+        | wc -l)" -le 1 ]
+}
```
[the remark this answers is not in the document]
andylemin: Good catch, thank you. Fixed in 0b54f882: only_afpd_parent() takes the pids from pgrep -u and reads each state from /proc/<pid>/stat, so busybox ps's eight-character user column is no longer involved; a zombie is still not counted as a survivor, and the leg also asserts the child's own "Disconnected session terminating" log line.

libatalk/util/logger.c
```diff
@@ -304,13 +348,19 @@ static void log_setup(const char *filename, enum loglevels loglevel,
...
     } else {
-        become_root();
-        type_configs[logtype].fd = open(filename,
-                                        O_CREAT | O_WRONLY | O_APPEND,
-                                        S_IRUSR | S_IWUSR | S_IRGRP | S_IROTH);
-        unbecome_root();
+        type_configs[logtype].fd = log_open_file(filename);
+
+        if (type_configs[logtype].fd != -1) {
+            type_configs[logtype].filename = log_absolute_name(filename);
```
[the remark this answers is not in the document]
andylemin: Agreed, thank you. Fixed in cefc67aa: when log_absolute_name() returns NULL after the open (or mkstemp), log_setup() closes the descriptor and takes the existing open-failure path, so the type is left unset rather than configured with a descriptor that daemonize() would make stale. fd 1 for /dev/tty is exempt, as before.

test/integration/singleuser_test.sh
```diff
@@ -0,0 +1,492 @@
...
+
+    [ "$oap_live" -le 1 ]
+}
+wait_for 10 only_afpd_parent \
+    || fail "a child of an earlier login is still running"
+{
+    printf '\000\004\000\001\000\000\000\000\000\000\000\000\000\000\000\000'
+    sleep 1
+} | timeout 3 nc 127.0.0.1 5548 > /dev/null 2>&1 || true
+wait_for 10 grep -q "Disconnected session terminating" $LOG1 \
```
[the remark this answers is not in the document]
andylemin: Done in cefc67aa: the leg records the log's line count before the DSIOpenSession drop and greps only the lines written after it, the pattern controller_started uses; $LOG1 is quoted as well.

andylemin:
@rdmark Finally got there.. Over to you.

This will need a lot of scrutiny and fresh eyes, as this capability has a lot of implications across the codebase - but it does provide a much tighter security surface by dropping root entirely.. It is a notable feature for single-user use cases.

I also wonder if we should make the handshake more opaque so port probing cannot identify the protocol. As it cannot work on the standard port this could be worthwhile.

Issue #2640, opened 2026-01-17
move CNID repair logic out of CNID backends


rdmark:
all stable CNID backends (dbd, myql, sqlite) have complex logic for handling inconsistent / illegal CNID states, and repairing the underlying database when needed

the most complex logic originates in `etc/cnid_dbd/dbd_lookup.c` (`dbd_lookup()`) and described as the "CNID salvation spec" in a code comment at the top of this file

mysql and sqlite backends attempts to emulate this logic in their own lookup functions

we have identified a few issues with this design:

- it's awkward / unexpected for database repair to be done in a lookup function, adding to the cognitive complexity of the function
- each CNID backend has their own flavor of this database repair logic, leading to hard-to-detect and troubleshoot edge case bugs
- a better pattern would be for the CNID backend primitives to be strictly abstractions to database operations, and the business logic residing higher up (libatalk/cnid or etc/afpd for instance)

PR #3278, opened 2026-08-28
cnid: classify contended vs corrupt vs broken sqlite backends end to end; stop contention and disk-full ending AFP sessions

andylemin:
`get_id()` ends the AFP session on `CNID_ERR_DB` — the right answer for a CNID backend that has gone away, and the wrong one for a lock that will free. The sqlite backend (now the default) reported every failure as `CNID_ERR_DB`, so two sessions writing to one volume — or a full disk — disconnected clients mid-operation, and a contended *read* came back "not found", which `afp_resolveid` turns into a permanent `AFPERR_NOID`.

This PR gives the backend a real error contract (`CNID_ERR_BUSY` for what clears, `CNID_ERR_CORRUPT` for unusable data, `EROFS` for a read-only database), applies it to every path — write, read, bind, cleanup — and keeps the classification alive from the sqlite result code to the AFP reply. `CNID_ERR_DB` is no longer emitted by this backend at all: it means the backend process is gone, which a local database file cannot be. Along the way it fixes the adjacent defects it found (silent 64-bit truncation that could delete the wrong file, a NULL dereference, a byte-order bug in id validation, a failed COMMIT stranding the volume's write lock), and makes the write paths materially cheaper: the common-case update drops from six SQL statements to one, delete from three to one, and every value is bound on cached prepared statements instead of being pasted into SQL text ([details below](#per-operation-sql-cost)).

The test harness also stops destroying the evidence when something does go wrong mid-run: a dying server child now yields a named test failure, a finished JUnit report, and an afpd log dump in the CI job, where it previously killed the whole suite by SIGPIPE and left nothing but an exit code that looked like runner trouble ([details](#harness-observability)).

## Root cause

Three layers each made a locally reasonable choice; the composition is wrong:

1. **afpd's policy** (`get_id()`): `CNID_ERR_DB` ⇒ backend gone ⇒ `exit(EXITERR_SYS)`. Correct for the dbd backend, where it means the `cnid_dbd` companion process died and nothing can be catalogued until it returns.
2. **The backend's classification**: the mysql backend mapped every SQL failure to `CNID_ERR_DB` — defensible for a remote server. The sqlite backend was built on that shape, but an embedded WAL engine answers `SQLITE_BUSY` routinely and by design; there is no remote server to lose.
3. **errno as the carrier**: the classification travels in `errno`, but every cleanup path (`sqlite3_reset()`, `ROLLBACK`) enters the VFS and can overwrite it — and the not-found sentinel `CNID_DBD_RES_NOTFOUND` shares its value (1) with `EPERM`, so wherever raw errno gets switched on, "row absent" reads as a permission error.

With sqlite the default, that composition turned ordinary concurrency into disconnects.

## How the error contract got here

Each commit is locally correct; the defects are the seams between them:

- **`c08d8903` — didg, 2009.** `valide()` is born: reject junk ids from a confused DB. It compares the network-order id against the host-order `CNID_START` — invisible on big-endian, and volumes rarely reached id 0x01000000.
- **`f254fd61` — Ralph Boehme, 2013.** The MySQL backend, and with it "any SQL failure → `CNID_ERR_DB`". For a remote server, mostly true.
- **`3e7f6ed1` — Daniel Markstedt, 2023.** The tdb fallback is removed and `CNID_ERR_DB` becomes deliberately session-fatal — the right call for a dead `cnid_dbd` with no fallback left to mask it.
- **`68efab70` — Daniel Markstedt, 2025.** The SQLite backend arrives, built on the mysql implementation — inheriting the one-class error shape into an engine where `SQLITE_BUSY` is a routine answer, not a failure.
- **sqlite becomes the default backend, 2025–2026.** The composition is now every new install's default path.

## The defects, before and after

| failure | before | after |
|---|---|---|
| peer transaction holds the write lock | write path: **session disconnected**; read path: permanent `AFPERR_NOID` | `CNID_ERR_BUSY` → `AFPERR_MISC`, the client retries |
| disk full / I/O error / OOM | **session disconnected** | `CNID_ERR_BUSY` → `AFPERR_MISC`; the session survives while the condition clears |
| read-only database | **session disconnected** | `EROFS` → `AFPERR_VLOCK`, the answer for a locked volume — not an invitation to retry |
| unrecognised sqlite result (`SQLITE_ERROR`, `SCHEMA`, `CANTOPEN`, `LOCKED`, `MISUSE`) | **session disconnected** via `CNID_ERR_DB` | `CNID_ERR_CORRUPT` → the one operation fails |
| corrupt database / not-a-db file | disconnected, with no hint what to do | `CNID_ERR_CORRUPT` → `AFPERR_MISC`, rebuild advice in the log |
| 64-bit column that does not fit a CNID (`get`, `resolve`, lookup's duplicate cleanup) | silently truncated — the wrong file addressed, or **deleted** | range-checked through `cnid_sqlite_rowid_ok()` → `CNID_ERR_CORRUPT` |
| lookup's `Did` column outside 32 bits | unchecked: a truncated `Did` equals the queried did, so the identity comparison passes on a row `resolve` calls corrupt | range-checked with `Id` |
| `find` returning a narrowed id | an unrelated live file's CNID entered an `FPCatSearch` result, and the client opens that file | out-of-range row skipped and reported; the sound matches still return |
| `FPExchangeFiles`, `FPMoveAndRename` on a backend that could not answer | read as "no CNID recorded": the on-disk rename completed with the id swap skipped, silently breaking both files' identity | classified → `AFPERR_MISC`, nothing renamed |
| `movecwd()` returning to a caller that switches on raw errno | its self-heal's CNID calls left a `CNID_ERR_*` value there, so ~20 call sites hit their `default:` arm | the chdir errno is restored before every failure return, and `dirlookup_strict`'s `AFPERR_MISC` is not overwritten |
| `dir_modify()` on a curdir whose CNID lookup failed | fell through and re-indexed curdir under DID 0 (`CNID_INVALID`), tripping the dircache DID assert | the DID re-index runs only for a valid CNID |
| `get_id()`'s `CNID_ERR_CLOSE` arm | left `afp_errno` untouched, so callers replied with the previous call's value | sets `afp_errno` like every other arm |
| NULL statement handle after a failed re-prepare in `wipe` | `sqlite3_step(NULL)` — undefined | `cnid_sqlite_stmt_ready()` fails it closed as `CNID_ERR_CLOSE`; `wipe` finalizes the rest |
| parameter guards | evaluated after the dereference they guard, so the `CNID_ERR_PARAM` arms were unreachable | validated first; `resolve` also checks `id` |
| empty `volume uuid =` | iniparser returns `""`, not NULL, bypassing UUID generation: every such volume shared one CNID table named `""` and collided on the `volumes` primary key | routed to `get_vol_uuid()`, which generates via `randombytes()`; `cnid_sqlite_open()` refuses an empty UUID |
| NULL `Name` column in `resolve` | **afpd segfault** (`strlcpy` on NULL; the db file is world-writable) | guarded → `CNID_ERR_CORRUPT` |
| `getstamp`: missing `volumes` row; parameter failure | session-fatal; returned 0, which callers read as success | `CNID_ERR_CORRUPT`; returns −1 |
| bind failure | logged, then returned with stale errno | classified like every other failure |
| valid CNID ≥ 0x01000000 on little-endian (~16.7M creates) | falsely rejected as corruption by `valide()` and `getmetadata` — and a rebuild reproduces the rejection | `ntohl()` compares; the genuinely reserved ids 1–16 are now actually rejected |
| not-found leaking as `EPERM` | the stale-cache self-heal answered `AFPERR_ACCESS` for a directory another session had deleted (spectest `Error:test174`) | not-found is carried as `CNID_ERR_NOTFOUND`, out of the errno range, and `movecwd()` latches the chdir errno before its self-heal runs |
| `find` after a handled constraint | the connection's sticky `sqlite3_errcode()` made a successful search look failed | reads the step return |
| failed COMMIT | ownership released anyway — the write lock stranded for the connection's life, blocking every sibling session | ownership kept until COMMIT succeeds; the cleanup rollback closes what a failed COMMIT leaves open |
| stale volume-entry cleanup; `wipe` quoting | a blanket DELETE off an incomplete scan could orphan CNID tables for good; `-DSQLITE_DQS=0` builds could never wipe/rebuild | per-UUID removal — converges past the collection cap, cannot orphan; single-quoted literals |
| stale `VolUUID` read back out of the world-writable db | interpolated into the `sqlite_sequence` and `volumes` deletes | bound as a parameter; a UUID that is not 32 hex digits names no table this code created, so only its row goes and the scan converges |

## The fix, in brief

- **The contract.** BUSY = nothing is wrong with the backend (contention, or an environmental condition that clears); CORRUPT = unusable data, rebuild advised; `EROFS` = a read-only database, which `afp_deleteid()` already answers with `AFPERR_VLOCK`. `SQLITE_PROTOCOL` (WAL's bounded handshake giving up) is BUSY; `SQLITE_LOCKED` deliberately is not — the busy handler never waits on it, so a retry finds the same state. `CNID_ERR_DB` is left unused here: it means the backend process is gone, and a local file has no such state, so an unrecognised code fails the one operation as CORRUPT.
- **Classified everywhere.** `cnid_sqlite_set_errno()` runs at every failure site: steps, execs, and binds.
- **The classification survives its own cleanup.** `cnid_sqlite_stmt_reset()` and an errno-preserving rollback wrap every cleanup path, and `cnid_sqlite_add` proceeds to its insert only on a genuine not-found — any other failure cannot be fixed by inserting.
- **Not-found moved out of the errno range.** `CNID_ERR_NOTFOUND` replaces the errno use of the dbd wire constant `CNID_DBD_RES_NOTFOUND` (== `EPERM`), whose value leaked into raw errno switches downstream.
- **Consumers act on the difference, through one predicate.** `CNID_ERRNO_IS_BACKEND_FAILURE()` in `include/atalk/cnid.h` replaces the hand-copied classification triplets. `get_id()`, `getmetadata()`, `dirlookup()`, `afp_resolveid()`, `afp_deleteid()`, `afp_exchangefiles()`, `moveandrename()` and `enumerate()`'s v2→EA conversion answer BUSY/CORRUPT with `AFPERR_MISC` instead of a permanent error, and honour each other's classification instead of flattening it.
- **errno is primed in the wrapper, not at each call site.** `libatalk/cnid/cnid.c` clears it before dispatching to any backend, so one reset covers every backend and every caller — dbd and mysql say nothing on not-found, so a stale value would otherwise be read as the verdict.
- **One prepared-statement skeleton.** The eleven per-volume statements go through `init_prepared_stmt_one()` rather than eleven copies of the same finalize/asprintf/prepare block.

## Per-operation SQL cost

Beyond classification, the rewrite cuts what each CNID operation costs the engine — fewer statements, no SQL parsing on the hot paths, and no value interpolated into statement text:

| operation | before | after |
|---|---|---|
| `update` (common case) | **6 statements**: exec `BEGIN`, three `DELETE`s built with `asprintf` and parsed per call, prepared `INSERT`, exec `COMMIT` — delete-then-reinsert on *every* update, inside an **unbounded** retry loop | **1 statement**: a cached prepared `UPDATE` under autocommit; the reinsert runs only if the row vanished |
| `update` (proven UNIQUE collision) | the same six, again and again | one bounded second attempt: two cached prepared `DELETE`s + the write, inside one owned transaction |
| `delete` | exec `BEGIN` + prepared `DELETE` + exec `COMMIT` (3 statements, 2 parsed per call) | **1** prepared `DELETE` — a single-statement autocommit is already atomic |
| `add` | insert (re)attempted after *any* lookup failure, loop unbounded | insert only on a genuine not-found; cycle bounded by `CNID_SQLITE_ADD_ATTEMPTS` |
| parameter passing on the paths above | ids, dev/ino — and the **filename** — interpolated into SQL text: one parse per call, and a name containing a double quote broke the statement outright | all values bound as parameters on per-volume cached prepared statements (eleven now; `update` and the two collision `DELETE`s are newly cached) |
| session close | WAL carried over, growing across sessions | `SQLITE_CHECKPOINT_TRUNCATE` folds the WAL back — with the busy timeout zeroed first so it skips rather than stalls behind readers, retrying at the next close |

## Behaviour changes

- Ordinary contention and disk-full fail the operation; they do not end the session. No sqlite result code ends the session any more.
- Transient backend trouble is retryable (`AFPERR_MISC`) — never a permanent `AFPERR_NOID`/`AFPERR_NOOBJ`/`AFPERR_PARAM` for an object that exists.
- Corruption is reported as corruption, with rebuild advice — and can no longer address the wrong file, delete through a truncated id, or crash afpd.
- Little-endian volumes past ~16.7M CNIDs resolve correctly.
- A failed COMMIT no longer blocks the volume for the life of the connection, and `-DSQLITE_DQS=0` builds can wipe/rebuild the database.
- A read-only database answers `AFPERR_VLOCK` (volume locked) rather than inviting a retry that cannot succeed.
- `FPExchangeFiles` and `FPMoveAndRename` fail rather than completing a rename whose CNID swap was skipped.
- A volume with an empty `volume uuid` gets a generated one instead of sharing a CNID table named `""`. Where the UUID file cannot be written, its CNID database is refused instead — a louder failure than before for that state.
- Filenames containing a double quote no longer break the update path (they were interpolated into the SQL text; all values are bound now).

### Unit tests
- afpd unit suite: **0 fail** (Alpine container, meson `-Dwith-tests=true`), including nine CNID tests:
  - **the headline behaviour**: a peer sqlite connection holds `BEGIN EXCLUSIVE`; `cnid_add()` must return `CNID_ERR_BUSY` (not session-fatal `CNID_ERR_DB`), and the very same add must succeed once the lock is released;
  - a not-found answer classifies as `CNID_ERR_NOTFOUND`, out of the syscall errno range — not the `EPERM`-valued dbd wire sentinel;
  - 64-bit rows injected through a peer connection classify as `CNID_ERR_CORRUPT` in `cnid_get`/`cnid_resolve` instead of truncating to a live file's id;
  - the duplicate-row cleanup never deletes through a truncated rowid (a live row whose id equals the narrowed value survives);
  - a search never returns a narrowed id: a row with `Id` = 2^32 + 4242 must not put CNID 4242 into an `FPCatSearch` result (fails without the `find` range check);
  - the wrapper errno contract, starting from a poisoned `errno = EPERM` (the `CNID_DBD_RES_NOTFOUND` collision);
  - the CNID error codes stay mutually distinct and clear of the system errno range;
  - `valide()` byte order and the `cnid_resolve()` `".."` rejection, via fake backends (the former fails without the `ntohl()` fix on little-endian).
- A probe row is only planted when the AUTOINCREMENT high-water mark it raises can be read back and restored, so a failed read cannot leave the volume minting out-of-range rowids for the rest of the suite. sqlite-specific tests compile out and skip on builds without the sqlite backend.
- Valgrind (`meson test --wrapper='valgrind --leak-check=full --error-exitcode=1'`): 0 errors, no leaks.
- The harness generates its volume UUIDs from `/dev/urandom`, so the suite no longer needs `uuidgen` (absent from the project's own container images, where it silently produced empty volume UUIDs).

### Spectests (sqlite backend, single-container loopback)
- ea = sys leg: **288 passed / 94 skipped / 0 failed**
- ea = ad leg (`AFP_ADOUBLE=1`): **288 passed / 94 skipped / 0 failed**
- ARC stress leg (`AFP_DIRCACHESIZE=1024`, `AFP_DIRCACHE_MODE=arc`, `AFP_DIRCACHE_VALIDATION_FREQ=100`): **288 passed / 94 skipped / 0 failed**, including `Error:test174`, the cross-session deleted-directory case that exposed the `EPERM`-valued not-found leak

### Harness observability

A server child dying mid-run used to kill `afp_spectest` with SIGPIPE before the summary or the JUnit report were written — the job then looked like a CI-runner failure with no artifacts (three distinct one-off leg failures on this PR shared exactly that shape, each passing on rerun). Now:

- `afp_spectest` ignores SIGPIPE, so a write to a dead session fails the test in flight with EPIPE and the run completes with a summary and a report;
- the `FPDisconnectOldSession` tests restore the process-wide ignore rather than `SIG_DFL`;
- the container entrypoint dumps `/var/log/afpd.log` on any non-zero exit, so a dying child ships its own post-mortem.

Validated by killing afpd 35s into a run: the suite recorded 22 failures, wrote the JUnit report, dumped the server log, and exited 1 — where it previously died at exit 141 with nothing.

- **Everything reset-shaped lands in the exhaustion-recovery follow-up branch**, and is entirely absent here: `CNID_ERR_RESET` (defined there, where it is first raised and announced to every session), the dead depletion guard in `cnid_sqlite_add` (`(uint32_t) lastid > UINT32_MAX` cannot fire; that branch rewrites the block with the correct 64-bit compare and the recovery announcement), the hint-channel rework, and two mechanical cleanups (`get_id()`'s switch tails, the triplicated `sqlite_sequence` seeding SQL) whose sites collide with its hunks.
- **The mysql backend** gets this contract in its own follow-up branch — its resolve/not-found errno, `getstamp`'s 0-on-error, and the `CNID_ERR_NOTFOUND` adoption land there.
- **`catsearch`, `mangle` and `spotlight`** still ignore errno on their resolve paths; all degrade gracefully (mangled-name fallback, skipped search entry) rather than answer wrongly and permanently.
- **`spotlight.c`'s `memcpy(uuid.sl_uuid, v->v_uuid, 16)`** is guarded only against a NULL `v_uuid`, so a short one is a 16-byte read off a smaller buffer. Generation now keeps it 36 bytes, but the read wants a length check of its own.
- **The dbd backend is untouched**: its `CNID_ERR_DB` genuinely means the companion process died, its exhaustion signal `CNID_ERR_MAX` still falls to `get_id()`'s `default:` arm, and it never wipes, so it degrades safely.

---

The diagram is the contract: which sqlite results become which classification, what preserves it in transit, and what each consumer does with it. Every defect above is a divergence between two adjacent columns.

```mermaid
flowchart LR
    classDef rc     fill:#374151,stroke:#4b5563,color:#f9fafb,stroke-width:1px;
    classDef cls    fill:#0f766e,stroke:#0d9488,color:#f0fdfa,stroke-width:1px;
    classDef err    fill:#b45309,stroke:#d97706,color:#fffbeb,stroke-width:1px;
    classDef cons   fill:#5b21b6,stroke:#6d28d9,color:#f5f3ff,stroke-width:1px;
    classDef ok     fill:#1e3a8a,stroke:#1e40af,color:#eff6ff,stroke-width:1px;
    classDef fatal  fill:#9f1239,stroke:#be123c,color:#fff1f2,stroke-width:1px;

    subgraph SQLITE["sqlite result codes"]
        BUSYRC("SQLITE_BUSY<br/>SQLITE_PROTOCOL<br/><i>lock will free</i>"):::rc
        ENVRC("SQLITE_FULL · IOERR<br/>NOMEM · READONLY<br/><i>environment, clears</i>"):::rc
        CORRC("SQLITE_CORRUPT<br/>SQLITE_NOTADB"):::rc
        ROWRC("row guards (cnid_sqlite_rowid_ok):<br/>Id/Did outside 32-bit<br/>NULL Name · no stamp row"):::rc
        RORC("SQLITE_READONLY<br/><i>read-only database</i>"):::rc
        DEFRC("everything else<br/><i>incl. SQLITE_LOCKED —<br/>busy handler never waits on it</i>"):::rc
        DONERC("SQLITE_DONE<br/><i>no row</i>"):::rc
    end

    SET("cnid_sqlite_set_errno()<br/>at every step, exec and bind"):::cls

    subgraph CARRY["classification in transit"]
        PRESERVE("cnid_sqlite_stmt_reset()<br/>errno-preserving rollback<br/><i>cleanups cannot clobber errno</i>"):::cls
        VALIDE("wrapper valide(): ntohl(id) < CNID_START<br/>→ CNID_ERR_CORRUPT"):::cls
    end

    EBUSY("CNID_ERR_BUSY"):::err
    ECORR("CNID_ERR_CORRUPT"):::err
    EROFSN("EROFS"):::err
    ENF("CNID_ERR_NOTFOUND<br/><i>out of the errno range,<br/>unlike the dbd sentinel<br/>== EPERM</i>"):::err

    subgraph AFPD["afpd consumers"]
        GETID("get_id()<br/><i>cnid_add, write path</i>"):::cons
        RESOLVE("afp_resolveid()<br/>afp_deleteid()<br/>dirlookup()"):::cons
    end

    MISC("AFPERR_MISC<br/><i>operation fails,<br/>client retries,<br/>session survives</i>"):::ok
    VLOCK("AFPERR_VLOCK<br/><i>volume locked</i>"):::ok
    NOID("AFPERR_NOID / NOOBJ<br/><i>only for a genuine<br/>not-found</i>"):::ok
    EXIT("exit(EXITERR_SYS)<br/><i>dbd only: no sqlite<br/>result reaches this</i>"):::fatal

    BUSYRC --> SET
    ENVRC --> SET
    CORRC --> SET
    ROWRC --> SET
    RORC --> SET
    DEFRC --> SET
    DONERC --> ENF
    SET --> EBUSY
    SET --> ECORR
    SET --> EROFSN
    EBUSY --> PRESERVE
    ECORR --> PRESERVE
    EROFSN --> PRESERVE
    ENF --> PRESERVE
    PRESERVE --> VALIDE
    VALIDE --> GETID
    VALIDE --> RESOLVE
    GETID -->|"BUSY · CORRUPT"| MISC
    RESOLVE -->|"BUSY · CORRUPT"| MISC
    RESOLVE -->|"EROFS"| VLOCK
    RESOLVE -->|"NOTFOUND"| NOID
```

### DIFF
--- a/distrib/docker/entrypoint_netatalk.sh
+++ b/distrib/docker/entrypoint_netatalk.sh
@@ -114,8 +114,9 @@ else
     set -e
 
     echo "==== TESTS (${TESTSUITE}) COMPLETED (exit code: $TEST_EXIT_CODE) ===="
-    # Display Netatalk's server logs if SERVER_LOGS environment variable is set
-    if [ -n "$SERVER_LOGS" ]; then
+    # Display Netatalk's server logs if SERVER_LOGS is set, and always on a
+    # failed run: a dying afpd child takes its diagnosis with it otherwise
+    if [ -n "$SERVER_LOGS" ] || [ "$TEST_EXIT_CODE" -ne 0 ]; then
         if [ -f /var/log/afpd.log ]; then
             echo "/var/log/afpd.log log lines: $(wc -l /var/log/afpd.log | awk '{print $1}')"
             echo "==== AFPD LOG CONTENT ===="
--- a/etc/afpd/directory.c
+++ b/etc/afpd/directory.c
@@ -657,7 +657,17 @@ static struct dir *dirlookup_internal(const struct vol *vol, cnid_t did,
      * Outputs: upath = basename only ("mydir") + cnid modified to parent DID
      * So we need to build the rest of the fullpath */
     if (upath == NULL) {
-        afp_errno = AFPERR_NOOBJ;
+        /* AFPERR_NOOBJ is permanent; a backend that could not answer has not
+         * said the directory is gone */
+        if (CNID_ERRNO_IS_BACKEND_FAILURE()) {
+            LOG(log_warning, logtype_afpd,
+                DIRLOOKUP_LOG_FMT ": CNID backend error, errno=0x%x",
+                ntohl(did), CNID_ERRNO());
+            afp_errno = AFPERR_MISC;
+        } else {
+            afp_errno = AFPERR_NOOBJ;
+        }
+
         err = 1;
         goto exit;
     }
@@ -1225,29 +1235,32 @@ int dir_modify(const struct vol *vol, struct dir *dir,
             AFP_CNID_DONE();
 
             if (new_cnid == CNID_INVALID) {
+                const char *why = CNID_ERRNO_IS_BACKEND_FAILURE()
+                                  ? "CNID backend could not answer" : "no CNID for new inode";
+
                 if (dir == curdir) {
                     /* Do not remove curdir — callers rely on it being valid.
-                     * AD cache already invalidated (AD_RLEN_UNKNOWN above) */
+                     * AD cache already invalidated (AD_RLEN_UNKNOWN above).
+                     * The DID re-index below runs only for a valid CNID, as
+                     * CNID_INVALID is DID 0. */
                     LOG(log_error, logtype_afpd,
-                        "dir_modify: inode change on curdir! — no CNID for "
-                        "new inode, keeping stale entry (did:%u, ino:%llu->%llu)",
-                        ntohl(dir->d_did),
+                        "dir_modify: inode change on curdir! — %s, "
+                        "keeping stale entry (did:%u, ino:%llu->%llu)",
+                        why, ntohl(dir->d_did),
                         (unsigned long long)dir->dcache_ino,
                         (unsigned long long)args->st->st_ino);
                 } else {
                     /* No CNID. Remove stale entry. */
                     LOG(log_debug, logtype_afpd,
-                        "dir_modify: inode change — no CNID for new inode, "
+                        "dir_modify: inode change — %s, "
                         "removing stale entry did:%u (ino:%llu->%llu)",
-                        ntohl(dir->d_did),
+                        why, ntohl(dir->d_did),
                         (unsigned long long)dir->dcache_ino,
                         (unsigned long long)args->st->st_ino);
                     dir_remove(vol, dir, 0);
                     return 0;
                 }
-            }
-
-            if (new_cnid != dir->d_did) {
+            } else if (new_cnid != dir->d_did) {
                 LOG(log_debug, logtype_afpd,
                     "dir_modify: inode change — CNID refreshed "
                     "did:%u->%u (ino:%llu->%llu)",
@@ -2072,6 +2085,8 @@ int movecwd(const struct vol *vol, struct dir *dir)
         LOG(log_error, logtype_afpd, "movecwd: NULL parameter (vol:%p, dir:%p)",
             (void*)vol, (void*)dir);
         afp_errno = AFPERR_PARAM;
+        /* Callers read errno right after this returns */
+        errno = EINVAL;
         return -1;
     }
 
@@ -2090,6 +2105,12 @@ int movecwd(const struct vol *vol, struct dir *dir)
         ntohl(dir->d_did), dir->d_fullpath ? cfrombstr(dir->d_fullpath) : "(null)");
 
     if ((ret = ochdir_vol(cfrombstr(dir->d_fullpath), vol)) != 0) {
+        /* The CNID lookups and chdirs in the self-heal below overwrite errno,
+         * while the reply must reflect why this chdir failed */
+        int chdir_errno = errno;
+        /* Set when the self-heal's lookup failed because the CNID backend
+         * could not answer, rather than because the directory is gone */
+        int backend_failure = 0;
         /* Failed to change directory */
         LOG(log_debug, logtype_afpd, "movecwd(\"%s\"): %s",
             dir->d_fullpath ? cfrombstr(dir->d_fullpath) : "(null)", strerror(errno));
@@ -2121,12 +2142,17 @@ int movecwd(const struct vol *vol, struct dir *dir)
                     return 0;
                 }
 
+                chdir_errno = errno;
                 LOG(log_info, logtype_afpd,
                     "movecwd: CNID retry chdir failed: %s", strerror(errno));
             } else {
+                /* dirlookup_strict sets AFPERR_MISC when the backend could not
+                 * answer; the errno switch below must not overrule it */
+                backend_failure = (afp_errno == AFPERR_MISC);
                 LOG(log_debug, logtype_afpd,
-                    "movecwd: dirlookup_strict returned NULL for did:%u (does not exist)",
-                    ntohl(saved_did));
+                    "movecwd: dirlookup_strict returned NULL for did:%u (%s)",
+                    ntohl(saved_did),
+                    backend_failure ? "CNID backend could not answer" : "does not exist");
             }
 
             /* Special movecwd requirement: sync process CWD with curdir pointer.
@@ -2149,19 +2175,25 @@ int movecwd(const struct vol *vol, struct dir *dir)
             }
 
             curdir = vol->v_root;
+            errno = chdir_errno;
             return -1;
         }
 
-        switch (errno) {
-        case EACCES:
-        case EPERM:
-            afp_errno = AFPERR_ACCESS;
-            break;
+        if (!backend_failure) {
+            switch (chdir_errno) {
+            case EACCES:
+            case EPERM:
+                afp_errno = AFPERR_ACCESS;
+                break;
 
-        default:
-            afp_errno = AFPERR_NOOBJ;
+            default:
+                afp_errno = AFPERR_NOOBJ;
+            }
         }
 
+        /* Callers switch on errno after this returns and have no arm for the
+         * CNID_ERR_* values the self-heal above may have left there */
+        errno = chdir_errno;
         return -1;
     }
 
--- a/etc/afpd/enumerate.c
+++ b/etc/afpd/enumerate.c
@@ -455,6 +455,14 @@ static int enumerate(AFPObj *obj _U_, char *ibuf, size_t ibuflen _U_,
                         LOG(log_error, logtype_afpd, "enumerate: error updating CNID of \"%s\"",
                             fullpathname(convname));
                     }
+                } else if (CNID_ERRNO_IS_BACKEND_FAILURE()) {
+                    /* ad_convert has already renamed the file, so the CNID row
+                     * names a path that is gone and the next lookup mints a
+                     * fresh CNID for it */
+                    LOG(log_error, logtype_afpd,
+                        "enumerate: CNID backend error (errno=0x%x), the CNID of "
+                        "converted \"%s\" still names the pre-conversion file",
+                        CNID_ERRNO(), fullpathname(convname));
                 }
             }
         }
--- a/etc/afpd/file.c
+++ b/etc/afpd/file.c
@@ -373,7 +373,26 @@ uint32_t get_id(struct vol *vol,
         if (dbcnid == CNID_INVALID) {
             switch (errno) {
             case CNID_ERR_CLOSE: /* the db is closed */
-                break;
+                /* Callers reply with afp_errno after a CNID_INVALID result */
+                LOG(log_error, logtype_afpd,
+                    "get_id: CNID database is not open for \"%s\"", upath);
+                afp_errno = AFPERR_MISC;
+                goto exit;
+
+            case CNID_ERR_BUSY:
+                /* Contention, not a broken backend: retryable, unlike the
+                 * CNID_ERR_DB arm below */
+                LOG(log_warning, logtype_afpd,
+                    "get_id: CNID backend busy for \"%s\"", upath);
+                afp_errno = AFPERR_MISC;
+                goto exit;
+
+            case CNID_ERR_CORRUPT:
+                LOG(log_error, logtype_afpd,
+                    "get_id: CNID backend returned unusable data for \"%s\"; "
+                    "the CNID database for this volume should be rebuilt", upath);
+                afp_errno = AFPERR_MISC;
+                goto exit;
 
             case CNID_ERR_DB:
                 LOG(log_error, logtype_afpd,
@@ -551,6 +570,15 @@ int getmetadata(const AFPObj *obj,
                         AFP_CNID_START("cnid_get");
                         id = cnid_get(vol->v_cdb, dir->d_did, upath, len);
                         AFP_CNID_DONE();
+
+                        /* The AFPERR_NOOBJ below is permanent; a backend that
+                         * could not answer has not said the object is gone */
+                        if (id == CNID_INVALID && CNID_ERRNO_IS_BACKEND_FAILURE()) {
+                            LOG(log_warning, logtype_afpd,
+                                "getmetadata(\"%s\"): CNID backend error, errno=0x%x",
+                                upath, CNID_ERRNO());
+                            return AFPERR_MISC;
+                        }
                     }
 
                     if (id == CNID_INVALID || ntohl(id) < CNID_START) {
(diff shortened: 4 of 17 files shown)


andylemin:
Hi @rdmark As you can probably assert from the changes, these bug fixes are pretty important.

It has also resulted in some nice latency reductions for some AFP operations due to the more efficient SQL routines.
The improvements to; Creating files, Creating dirs, Copying files (ADD operations), and deleting files (DELETE operations) align well with the SQL changes. The Adj Δ% results for these are consistent across all test runs (albeit slow or fast CI runner).

The biggest gains would have been for rename and move (6 statements down to 1), but we don't have any lantest tests for these! (need to resolve this gap now I've noticed it).

andylemin:
@rdmark ready for review

rdmark (approved):
the new CNID tests will make me sleep better at night -- in the future we should consider expanding the fuzzer to hammer the CNID backend with all sorts of weird data!

this constitutes almost half of a rewrite of the sqlite backend I think, impressive work. have you tested larger number of current users? I'm curious if it would be more resilient to concurrent database writes by hundreds of afpd accessors for instance.

it reminds me of the discussion that led up to: https://github.com/Netatalk/netatalk/issues/2640  -- we should revisit this one once we retire the dbd backend!

rdmark:
@andylemin while on the topic, do you agree with my decision to remove the in-memory tdb CNID fallback upon backend failure? I concluded that it was a band-aid solution that made the file server behave less deterministically in a way that wasn't transparent/obvious to the user what actions they should take. the argument for the in-memory fallback, as I recall it, was to allow users to "rescue" files off of the AFP volume before corruption hit.

andylemin:
> do you agree with my decision to remove the in-memory tdb

@rdmark Yes, 100% 🙂 I see no reason to keep this, and there are more risks of bugs when we are trying to maintain many backends.. The same can be done much more reliably with SQLite and tmpfs..

I am testing the heck out of this now actually, hence the extra pushes this evening.. And reviewing the SQLite code itself to ensure that the detailed error contracts being defined here actually align with SQLite internals. Please continue to review, I not quite ready to merge yet - I need to prove some things to myself first.

andylemin:
Discovered that the prepare statement has the same contention issues in the SQLite backend as the actual step operations, so adding the same error handling there seems to have improved the SQLite flakiness under contention 🤞

andylemin:
PS; agreed on #2640, the database truncation on CNID ID exhaustion PR I have lined up is a step towards this

PR #3281, opened 2026-08-31
cnid: reach exhaustion recovery, announce it, and classify mysql errors

andylemin:
A volume's CNID space is 32 bits. When it runs out the sqlite backend is meant to empty the table and start again, but the guard that detects exhaustion reads `if ((uint32_t) lastid > UINT32_MAX)` — a comparison the cast makes always false. The recovery has never run. Past the ceiling `htonl((uint32_t) lastid)` truncates instead, so rowid 2³²+17 becomes CNID 17 and two files answer to one ID.

Reaching the reset is only half of it. Emptying the table invalidates every CNID held by **every session on the volume**: each one now names nothing, or — once the table refills from 17 — a different file. So this PR does two things. It makes the recovery reachable, and it makes the recovery tell (reset) every affected session, because recycling IDs silently under live clients is worse than the exhaustion it fixes.

Both SQL backends are in scope. The unreachable guard is sqlite's, but mysql does the same wipe-and-reseed, so once a reset became a reportable condition mysql had to answer for it too — which meant first making its `errno` mean something.

## Root cause

Three layers, each locally reasonable:

1. **The comparison.** `lastid` is a `uint64_t` from `sqlite3_last_insert_rowid()`, and `(uint32_t) lastid` cannot exceed `UINT32_MAX`. The branch is unreachable, and the compiler is right not to warn: the cast is well defined, just always false.
2. **The recovery assumed one writer.** Emptying and reseeding is correct for a single process. With one afpd child per session sharing the table it is a global event with purely local effect — the child that wipes carries on, and its siblings never learn.
3. **Nothing above the backend could hear it.** The backend's only channel to afpd is `errno`, and no code for a table reset existed, so even a correct guard had nothing to report with.

mysql failed the same way from the other side: one classification for everything, `CNID_ERR_DB`, which afpd answers with `exit(EXITERR_SYS)`. InnoDB reports a deadlock immediately rather than waiting, and sessions inserting into one volume's table hit deadlocks and lock-wait timeouts routinely — so a condition that deserved a retry ended the session instead.

## How the guard got here

- **`f254fd61` — Ralph Boehme, 2013.** The MySQL backend introduces depletion: a `Depleted` column, a table wipe, an `AUTO_INCREMENT` reseed. Sound for a backend whose one writer is a remote server. `CNID_ERR_DB` was the only classification available, and the codes that tell a lock from a crash were never consulted.
- **`68efab70` — Daniel Markstedt, 2025.** The SQLite backend inherits that shape, truncating comparison included. Because the branch never fires, no test and no deployment ever exercised it.
- **sqlite becomes the default (2025–2026).** The unreachable recovery is now on every install's default path, and the process model has changed underneath it: one connection per session, all on one file.

## What each caller did with exhaustion, before and after

Five call paths reach `cnid_add()`. Because the guard never fired, **none** of them had ever met this condition.

| caller | before | after |
|---|---|---|
| `get_id()` (create/lookup, the main path) | recycled CNID returned as success | `CNID_ERR_RESET` → announce → session ends |
| `reenumerate_loop()` (directory resync) | kept walking, numbering entries out of the reseeded table | stops, returns −1 |
| `afp_resolveid()` | retried `cnid_resolve()` on the pre-wipe CNID, answering **AFP_OK with another file's metadata** | honours the walk's −1, returns `AFPERR_NOID` |
| `crit_check()` (FPCatSearch) | took the reset for "no match" and packed reseeded CNIDs into the reply | ends the walk, `AFPERR_MISC` |
| Spotlight (`cnid`, `localsearch`, `xapian`) | `cnid_for_path()` allocates, so a search could trip the wipe; one reported no results, two skipped the row and returned pre-wipe rows as a *resumable* success | all three end the query |
| `dbd` scanvol | skipped the entry and **kept scanning**, renumbering every remaining file against the emptied table and rewriting their AppleDouble IDs, then exiting 0 | stops, and fails its exit status so the rerun it advises is not optional |
| `nad`, `megatron` (no session to end) | generic "Error resolving CNID"; the wipe reported nowhere | every call site names the reset behind the failure |

## The notification

Siblings hear about it over the existing dircache-hint pipe. A cache-mend hint may be dropped; this one ends the receiving session, so it cannot be. It survives a full buffer, where a best-effort hint yields its slot, but it is metered against its own per-child budget so a session's cache-mend chatter cannot consume it — and it leaves with the call that buffered it, because the batching trigger, an idle poll or a full buffer, can be starved indefinitely on a busy master.

Beyond that it rides the ordinary batch writer: no retry ladder, no separate delivery path. A 22-byte message into a 64 KB pipe fails only for a child already wedged and about to be reaped, and a once-in-years event does not justify machinery in the master's single-threaded loop. The child keeps one `poll`-then-retry on its own IPC write, the one place where a lost write costs *every* sibling the notification.

Two details keep that safe under a burst. The dedupe records a volume's tag when the hint is relayed, before delivery is known, so a batch that fails with a reset in it releases the tag again — a sibling whose pipe stayed full has no other way to hear, and re-broadcasting is harmless because anyone who already heard is ending their session anyway. And the waits are budgeted per second rather than per flush, since every reset forces a flush and a per-flush budget would multiply the stall.

**The hint cannot be keyed on `v_vid`**, a per-process counter assigned in each child's own volume-load order: with `valid users` filters or `[Homes]`, the same vid names different volumes in different children, so acting on it would end a session whose CNIDs are fine and leave the affected one stale. `cnid_volume_tag()` is an FNV-1a of `v_uuid`, which every child that opens the volume agrees on.

**No AFP attention is sent.** The protocol guarantees an ID is never reused for another object for the volume's lifetime, so "the ID space was recycled" has no representation. The nearest candidate, `AFPATTN_VOLCHANGED`, only means "call FPGetVolParms" — it invalidates no cached ID and costs a round trip to say nothing useful. The disconnect is the signal, and the die path already sends `AFPATTN_SHUTDOWN`.

## The mysql error contract

The table below lists the mysql defects one by one. Two are worth drawing out, because this change is what made them dangerous.

A stale errno: both INSERT chokepoints failed without setting `errno` at all, so the caller read whatever an earlier operation had left there. That was merely wrong before. With `CNID_ERR_RESET` now in the vocabulary, a leftover one turns a single deadlock into a disconnect of every session on the volume.

An undetected wipe: `cnid_mysql_execute()` returned `mysql_query()`'s non-zero while its two `EC_NEG1()` call sites tested for `-1`, so a failed multi-statement wipe read as success — announcing a reset of a table that was never emptied, then doing it again on the next insert.

## Defect → fix

| # | Defect | Closed by |
|---|---|---|
| 1 | depletion guard unreachable; rowids past 2³² truncate into live CNIDs | 64-bit compare, `lastid >= (uint64_t) UINT32_MAX` |
| 2 | recovery invisible to siblings | `CACHE_HINT_VOLUME_RESET` over the hint pipe, keyed on `cnid_volume_tag()` |
| 3 | `CNID_ERR_RESET` had no producer or consumer | raised by both backends; handled in `get_id()`, both Spotlight paths, and the hint receiver |
| 4 | five walks would serve reseeded CNIDs in a reply sent before the session ends | each stops its own walk (table above) |
| 5 | a wipe from `dbd`/`nad` announced nothing | `dbd` stops and fails its exit status; every `nad`/`megatron` call site names the reset |
| 6 | `dbd -f` discarded the AppleDouble hints it was restoring | both backends' `wipe()` clear the latched depleted flag |
| 7 | a planted hint could bring the wipe within reach of ordinary file creation | 2²⁸ reserve on both backends, gating *insertion* only, so legacy rows keep their identity |
| 8 | a stale hint let a later direct lookup rewrite an unrelated row's identity | both backends clear the hint at the end of the add that carried it |
| 9 | no warning before the cliff | one-shot near-depletion warning naming `dbd` as the remedy |
| 10 | the condition travels in errno, which the IPC send and the backends' cleanup clobbered — and mysql's new insert gate turns a clobbered errno into a total allocation stop | every path on both backends saves and restores it |
| 11 | a reset to a busy sibling was dropped outright, and a failed delivery still counted as broadcast, so the next duplicate was deduped away | the master waits once for pipe space; a failed batch releases the tag for re-broadcast |
| 12 | resets were exempt from the attack guard, and the wait budget was per flush while every reset forces its own | a dedicated per-child reset budget, and one wait budget shared across flushes each second |
| 13 | mysql reseeded and continued, returning an id from the emptied space | raises `CNID_ERR_RESET` like sqlite |
| 14 | mysql read lock waits, deadlocks and contended reads as a dead backend or a missing id — the latter a permanent `AFPERR_NOID` | `cnid_mysql_set_errno()` classifies them; `_lookup`, `_get` and `_resolve` classify their statement, `store_result`, `bind_result` and `fetch` failures |
| 15 | mysql's two INSERT chokepoints set no errno, so the caller read a stale one — including a stale `CNID_ERR_RESET` | both classify before failing |
| 16 | mysql carried not-found as `CNID_DBD_RES_NOTFOUND`, numerically `EPERM` | `CNID_ERR_NOTFOUND`, and the post-lookup insert gate tests for it exactly |
| 17 | a later statement of a multi-statement wipe could fail unnoticed: `mysql_next_result()` signals it with a value a `== 0` drain loop reads as end-of-batch | one helper that separates error from end-of-batch, checked before the reset is announced and before the latch is cleared; it also owns the result pointer, so the double-free path is gone |
| 18 | the exec wrapper's failure return could not satisfy its two `EC_NEG1()` call sites | normalised to `-1` |
| 19 | mysql's add and update retry loops were unbounded, and `CR_SERVER_LOST` recovery re-armed its own | all three bounded |
| 20 | `cnid_mysql_getstamp()` returned `CNID_INVALID` (0) on parameter failure, which its int-returning callers read as success | returns `-1` |

## Behaviour changes

- A volume that reaches the ceiling now recovers instead of handing out colliding IDs, and every session on it is disconnected so clients reconnect against the rebuilt table.
- A search, an `FPCatSearch` or a directory resync that trips the wipe stops, rather than completing with recycled IDs in a successful reply.
- `dbd` stops at a reset and fails its exit status, where it used to carry on and leave the volume half renumbered while exiting 0. `nad` and `megatron` report a wipe they performed in silence. Note that `dbd` now exits non-zero for any scanvol failure, not only a reset.
- On mysql, lock contention no longer disconnects the client: a deadlock or lock-wait timeout is a retryable `AFPERR_MISC` instead of `exit(EXITERR_SYS)`, and a contended read is no longer a permanent "not found".
- `dir_modify()` evicts a dircache entry whose inode changed when the backend cannot answer, so the next access re-stats. Keeping it would serve the pre-change stat indefinitely, since enumerate's cache-hit path rebuilds `struct stat` from the cached fields and never re-stats. `curdir` cannot be evicted, so it keeps its old inode instead, leaving the trigger that brings us back here.

## Testing

**Unit tests.** afpd suite: 3/3 meson tests, 102 subtests, 0 failures, no compiler warnings, built with the full CNID backend set including mysql.

`utest_cnid_add_depletion_resets` covers the headline path. A peer sqlite connection moves the volume's `AUTOINCREMENT` mark to the ceiling, so a real `cnid_add()` empties the table and has to answer `CNID_ERR_RESET` rather than a recycled id. It then restores the volume — `cnid_wipe()` to clear the depleted mark on disk and in the connection, then the saved sequence — and asserts the mark is clear, because left set it would quietly discard every AppleDouble hint in later tests and later runs. Proven RED/GREEN: remove the wipe and the test reports the volume still depleted.

`utest_cnid_volume_tag_identity` pins the tag's cross-process property: the same UUID in two volume structs with different vids must agree, a one-character UUID change must not alias, and a volume with no UUID yields the reserved 0.

**Spectests** (sqlite, single-container loopback): 292 passed / 0 failed with ea = sys, and the same with ea = ad (`AFP_ADOUBLE=1`).

## Scope

Verified by review rather than by test:

- The relay-side guards — the reset's rate budget, the dedupe and its release on a failed delivery, the per-second wait budget — are parent-side and static in `ipc_relay_cache_hint()`/`hint_flush_pending()`, and no harness drives the master's IPC loop.
- Nothing exercises the mysql changes: the depletion test drives a peer *sqlite* connection, and there is no mysql equivalent nor anything that induces a deadlock. A mysql leg for that test, and a lock-contention test pinning `CNID_ERR_BUSY`, are the natural follow-ups.

Left open deliberately:

- **A reset during a primary reconnect is swallowed.** `afp_dsi_die()` returns without exiting while `DSI_RECONINPROG` is set, so the `raise(SIGTERM)` this design depends on can leave that session alive holding stale CNIDs. #3285 carries the `process_deferred_signals()` fix with the fuller teardown that path needs; duplicating it here would collide.
- **A reset ends sessions parked in `DSI_DISCONNECTED`**, destroying the reconnect window and the user's open forks — on a session serving CNIDs to nobody. Telling "end this session" from "refuse its reclaim" is a decision beyond this change.
- **A reset tripped by `dbd` or `nad` reaches no running session.** Those tools have no IPC channel to the master, and an afpd connection reads the `Depleted` column only at open. Closing it needs a generation check on the backend, or a signal path from the tools.
- **The hint's identity range.** Outside the insertable band a hint still acts as lookup's identity token, where `hint == retid` selects update-in-place. Narrowing identity to `[CNID_START, UINT32_MAX]` is the likely fix; the legacy-rows rationale justifies only that range, not the full 32 bits.
- **The volume tag is 32-bit**, and the master's dedupe is keyed on it alone, so two volumes colliding under FNV-1a would cross-disconnect. The hint payload is a fixed 8-byte packed struct with a `_Static_assert`, so widening it means changing the shared wire format.
- **`cnid_volume_tag()` is a fourth open-coded FNV-1a** (`libatalk/acl/cache.c`, `sl_xapian.c`, `dircache.c`), and `vol->v_stamp` — which comes *from* the backend — may be a better identity source than the optional `v_uuid`.

---

The diagram is the mechanism: who detects exhaustion, how the one process that wipes reaches every other, and where each walk stops so no reply carries a recycled ID.

```mermaid
flowchart LR
    classDef be   fill:#0f766e,stroke:#0d9488,color:#f0fdfa,stroke-width:1px;
    classDef walk fill:#5b21b6,stroke:#6d28d9,color:#f5f3ff,stroke-width:1px;
    classDef ipc  fill:#b45309,stroke:#d97706,color:#fffbeb,stroke-width:1px;
    classDef end1 fill:#9f1239,stroke:#be123c,color:#fff1f2,stroke-width:1px;
    classDef tool fill:#374151,stroke:#4b5563,color:#f9fafb,stroke-width:1px;

    subgraph CHILD["afpd child that reaches the ceiling"]
        direction TB
        ADD("cnid_sqlite_add() · cnid_mysql_add()<br/>lastid at the 32-bit ceiling<br/><i>empty table, reseed to CNID_START-1</i>"):::be
        ERR("errno = CNID_ERR_RESET<br/><i>distinct from BUSY · CORRUPT · NOTFOUND · DB</i>"):::be
        WALKS("the five walks that mint CNIDs<br/>get_id · reenumerate_loop<br/>afp_resolveid · crit_check · Spotlight×3<br/><i>each stops its own walk</i>"):::walk
        RESET("cnid_volume_reset()<br/><i>tag = FNV-1a(v_uuid)</i>"):::ipc
        DIE1("deferred die<br/><i>between requests</i>"):::end1
    end

    subgraph MASTER["netatalk master"]
        direction TB
        RELAY("ipc_relay_cache_hint()<br/><i>own rate budget, deduped per volume,<br/>survives a full buffer,<br/>flushes with this call</i>"):::ipc
        SORT("hint_flush_pending()<br/><i>reset sorts ahead of the cache-mend batch,<br/>waits budgeted per second,<br/>undelivered reset is re-broadcastable</i>"):::ipc
    end

    subgraph SIBS["sibling sessions"]
        direction TB
        RECV("process_cache_hints()<br/><i>resolve volume by tag, not vid</i>"):::ipc
        DIE2("deferred die<br/><i>stop draining hints</i>"):::end1
    end

    subgraph TOOLS["dbd · nad — no session to end"]
        REPORT("end the scan, fail the exit status<br/><i>a rebuild's gap is visible</i>"):::tool
    end

    ADD --> ERR
    ERR --> WALKS
    WALKS -->|"one announces"| RESET
    ERR -.->|"tool context"| REPORT
    RESET --> RELAY
    RESET --> DIE1
    RELAY --> SORT
    SORT -->|"one write per sibling"| RECV
    RECV --> DIE2
```

### DIFF
--- a/bin/dbd/cmd_dbd.c
+++ b/bin/dbd/cmd_dbd.c
@@ -325,6 +325,7 @@ int main(int argc, char **argv)
 
     if (cmd_dbd_scanvol(vol, flags) < 0) {
         dbd_log(LOGSTD, "Error repairing database.");
+        EC_STATUS(-1);
     }
 
 EC_CLEANUP:
--- a/bin/dbd/cmd_dbd_scanvol.c
+++ b/bin/dbd/cmd_dbd_scanvol.c
@@ -690,6 +690,16 @@ static cnid_t check_cnid(const char *name, cnid_t did, struct stat *st,
     /* Get CNID from database */
     if ((db_cnid = cnid_add(vol->v_cdb, st, did, (char *)name, strlen(name),
                             ad_cnid)) == CNID_INVALID) {
+        if (CNID_ERRNO() == CNID_ERR_RESET) {
+            dbd_log(LOGSTD,
+                    "CNID space exhausted at \"%s/%s\": the table was emptied "
+                    "and this scan is incomplete. Rerun to renumber the volume.",
+                    cwdbuf, name);
+            /* Walking on would renumber every remaining file against the
+             * emptied table, leaving the volume half old, half new ids */
+            longjmp(jmp, 2);
+        }
+
         return CNID_INVALID;
     }
 
@@ -978,8 +988,15 @@ int cmd_dbd_scanvol(struct vol *vol_in, dbd_flags_t flags)
      */
     cnid_getstamp(vol->v_cdb, stamp, sizeof(stamp));
 
-    if (setjmp(jmp) != 0) {
+    switch (setjmp(jmp)) {
+    case 1:
         EC_EXIT_STATUS(0); /* Got signal, jump from dbd_readdir */
+
+    case 2:
+        EC_EXIT_STATUS(-1); /* CNID table reset, jump from check_cnid */
+
+    default:
+        break;
     }
 
     strlcpy(cwdbuf, vol->v_path, sizeof(cwdbuf));
--- a/bin/nad/megatron.c
+++ b/bin/nad/megatron.c
@@ -263,6 +263,7 @@ static int update_created_file_cnid(const struct nad_volume *volume,
     cnid = cnid_for_path(vol->v_cdb, vol->v_path, path, &did);
 
     if (cnid == CNID_INVALID) {
+        nad_report_cnid_reset(vol->v_path);
         fprintf(stderr, "Error resolving CNID for %s\n", path);
         errno = EIO;
         return -1;
--- a/bin/nad/nad.h
+++ b/bin/nad/nad.h
@@ -81,6 +81,7 @@ extern void nad_not_inside_volume(FILE *out, const char *path, int force_hint);
 extern void closevol(afpvol_t *vol);
 extern cnid_t cnid_for_paths_parent(const afpvol_t *vol, const char *path,
                                     cnid_t *did);
+extern void nad_report_cnid_reset(const char *volpath);
 extern char *utompath(const struct vol *, const char *);
 extern int convert_dots_encoding(const afpvol_t *svol, const afpvol_t *dvol,
                                  char *path);
--- a/bin/nad/nad_adouble.c
+++ b/bin/nad/nad_adouble.c
@@ -747,6 +747,7 @@ static int nad_update_cnid(void)
                          &did);
 
     if (cnid == CNID_INVALID) {
+        nad_report_cnid_reset(nad.vol->v_path);
         fprintf(stderr, "Error resolving CNID for %s\n", nad.adpath[DATA]);
         errno = EIO;
         return -1;
--- a/bin/nad/nad_cp.c
+++ b/bin/nad/nad_cp.c
@@ -521,6 +521,7 @@ static int copy(const char *path,
 
             if ((did = cnid_for_path(dvolume.vol->v_cdb, dvolume.vol->v_path, to.p_path,
                                      &pdid)) == CNID_INVALID) {
+                nad_report_cnid_reset(dvolume.vol->v_path);
                 NAD_INFO("Error resolving CNID for %s", to.p_path);
                 badcp = rval = 1;
                 return -1;
@@ -598,6 +599,7 @@ static int copy(const char *path,
 
             if ((cnid = cnid_for_path(dvolume.vol->v_cdb, dvolume.vol->v_path, to.p_path,
                                       &did)) == CNID_INVALID) {
+                nad_report_cnid_reset(dvolume.vol->v_path);
                 NAD_INFO("Error resolving CNID for %s", to.p_path);
                 badcp = rval = 1;
                 return -1;
--- a/bin/nad/nad_mkdir.c
+++ b/bin/nad/nad_mkdir.c
@@ -84,6 +84,7 @@ static int mkdir_with_cnid(const char *path, mode_t mode, afpvol_t *vol)
 
     if ((did = cnid_for_path(vol->vol->v_cdb, vol->vol->v_path, path,
                              &pdid)) == CNID_INVALID) {
+        nad_report_cnid_reset(vol->vol->v_path);
         NAD_INFO("Error resolving CNID for %s", path);
         return -1;
     }
--- a/bin/nad/nad_mv.c
+++ b/bin/nad/nad_mv.c
@@ -302,6 +302,7 @@ static int do_move(const char *from, const char *to)
     if (!mustcopy) {
         if ((cnid = cnid_for_path(svolume.vol->v_cdb, svolume.vol->v_path, from,
                                   &did)) == CNID_INVALID) {
+            nad_report_cnid_reset(svolume.vol->v_path);
             NAD_INFO("Couldn't resolve CNID for %s", from);
             return -1;
         }
--- a/bin/nad/nad_rm.c
+++ b/bin/nad/nad_rm.c
@@ -196,6 +196,7 @@ static int rm(const char *path,
 
             if ((cnid = cnid_for_path(volume.vol->v_cdb, volume.vol->v_path, path,
                                       &did)) == CNID_INVALID) {
+                nad_report_cnid_reset(volume.vol->v_path);
                 NAD_INFO("Error resolving CNID for %s", path);
                 return -1;
             }
@@ -235,6 +236,7 @@ static int rm(const char *path,
             /* Get CNID of Parent and add new childir to CNID database */
             if ((did = cnid_for_path(volume.vol->v_cdb, volume.vol->v_path, path,
                                      &pdid)) == CNID_INVALID) {
+                nad_report_cnid_reset(volume.vol->v_path);
                 NAD_INFO("Error resolving CNID for %s", path);
                 return -1;
             }
@@ -286,6 +288,7 @@ static int rm(const char *path,
 
             if ((cnid = cnid_for_path(volume.vol->v_cdb, volume.vol->v_path, path,
                                       &did)) == CNID_INVALID) {
+                nad_report_cnid_reset(volume.vol->v_path);
                 NAD_INFO("Error resolving CNID for %s", path);
                 return -1;
             }
--- a/bin/nad/nad_rmdir.c
+++ b/bin/nad/nad_rmdir.c
@@ -113,6 +113,7 @@ static int rmdir_with_cnid(const char *path, afpvol_t *vol)
         /* Resolve CNID before any modifications */
         if ((did = cnid_for_path(vol->vol->v_cdb, vol->vol->v_path, path,
                                  &pdid)) == CNID_INVALID) {
+            nad_report_cnid_reset(vol->vol->v_path);
             NAD_INFO("Error resolving CNID for %s", path);
             return -1;
         }
--- a/bin/nad/nad_util.c
+++ b/bin/nad/nad_util.c
@@ -298,6 +298,24 @@ int convert_dots_encoding(const afpvol_t *svol, const afpvol_t *dvol,
 }
 
 /*!
+ * @brief Report a CNID table reset behind a failed resolve
+ *
+ * A reset empties the volume's CNID table as a side effect of the failing
+ * call, which the caller's generic resolve error would hide.
+ *
+ * @param[in] volpath  path of the volume the failing call ran against
+ */
+void nad_report_cnid_reset(const char *volpath)
+{
+    if (CNID_ERRNO() == CNID_ERR_RESET) {
+        fprintf(stderr,
+                "CNID space exhausted for volume \"%s\": the table was "
+                "emptied, so every session on it now holds stale CNIDs\n",
+                volpath);
+    }
+}
+
+/*!
  * @brief Resolves CNID of a given paths parent directory
  *
  * path might be:
@@ -352,6 +370,7 @@ cnid_t cnid_for_paths_parent(const afpvol_t *vol,
                              cfrombstr(l->entry[i]),
                              blength(l->entry[i]),
                              0)) == CNID_INVALID) {
+            nad_report_cnid_reset(vol->vol->v_path);
             EC_FAIL;
         }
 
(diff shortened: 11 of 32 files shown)


andylemin:
@rdmark FYI, I improved logging of fault and exception handling, and discovered that the occasional CI run failures are actually real Netatalk faults.

It does not occur on ARM, only Intel, and only 1 in 5 runs - Hence the hundreds of pushes to trigger CI runs after adding logging.

It's an occasional free of an already unlinked object. Trying to find the root cause rather than just patching it.

Has been around for some time, but only able to trace it now following other improvements.

Will continue until I find it.

andylemin:
@rdmark Ready for review

andylemin:
@rdmark Also polished and story complete for both SQLite and MySQL now. I was originally dealing with MySQL separately, but doing them together in one large PR made it easier to rationalise the common elements.

rdmark (approved):
I think the second paragraph in doc/manual/CNID.md needs an overhaul after this!

andylemin:
Good catch — that paragraph described the pre-PR behaviour, where depletion meant the volume simply stopped accepting new files.

Rewritten in `9fc9d726`. It now splits the depletion facts from what each backend does about it, because the answer genuinely differs:

- *dbd* stops allocating (`CNID_ERR_MAX`), so the old "no new files can be written" text is still right — but only for that backend.
- *sqlite* and *mysql* empty the table and restart numbering, which is the spec violation worth being explicit about: an ID issued before the reset names nothing, or names a different file once the table refills. The paragraph now says every session on the volume is disconnected for that reason, that Mac aliases and other client-side references to the old IDs break, and that a **dbd** rebuild is the better way to renumber deliberately.

The vague "some backends may attempt to reuse available IDs" sentence is gone, since it is now specific and true of both SQL backends.

One asymmetry I documented rather than papered over: the near-depletion warning exists only in *sqlite*, so the last sentence names that backend explicitly. Mirroring it in *mysql* is about ten lines and I am happy to add it if you would rather the two read the same — I left it out to avoid adding code to an approved PR.

I left `doc/po/CNID.md.ja.po` untouched, since po4a regenerates the catalogues and earlier doc commits followed the same pattern.

andylemin:
Agreed on both — mysql now logs the same warning, so the doc no longer has to single out *sqlite*.

The warning fires at the same threshold as sqlite (`CNID_MYSQL_MAX_HINT`, i.e. `UINT32_MAX` minus the 2²⁸ hint reserve), latches a `CNID_MYSQL_FLAG_NEAR_DEPLETION` flag so a scan warns once rather than per file, and `cnid_mysql_wipe()` clears that flag alongside the depleted one — otherwise a rebuilt volume would never warn again on that connection.

Wording updated as suggested:

> Both backends log a warning as a volume approaches the ceiling,
> so it can optionally be renumbered with the **dbd** utility beforehand,
> during a maintenance window.

`fb0b53ee`. Clean build, no format-string warnings, 3/3 meson tests and 102 subtests passing.

andylemin:
@rdmark PS working on this reminded me just how confusing it is for the `dbd` command to be the tool for all backends, not just the DBD backend 😕 I think we should reconsider renaming this, or importing it into the nad namespace.
I know you suggested we change the abbreviation, but for the past 20+ years it has meant DBD and people are not going to notice a documentation update. But they will notice if it vanishes and they see the single management command has everything they need. Just a thought, not urgent, but it was painful and even writing the docs was awkward with the mixed meaning.

I added `The dbd command-line utility can verify, repair, and rebuild the CNID database of any backend, not just dbd.`

PR #3285, opened 2026-09-02
dircache: remove entries by stored node; fix duplicate-key creation, curdir loss, missing volume purge, and dead asserts

andylemin:
The dircache indexes every entry twice — on `(vid, did)` and on `(vid, pdid, uname)` — plus an LRU/ARC queue. Every removal used to find its nodes *by key* and free whatever the lookup returned. That is safe only while each key has one holder, and nothing enforced it: the assert that rejects a second node under an existing key had been compiled out of every build since the table was imported in 2005.

This PR makes removal work from nodes the entry stores itself, then closes each path that could mint a duplicate in the first place, and re-arms the asserts that police the whole thing. Arming them immediately caught a further defect, unrelated to duplicates, in ARC's insert ordering.

## When it actually fired

One spectest case, ever: `Dircache:test506`, *"stat subdir in poisoned path"*, from the suite written to attack this cache. It is the only case that vacates a `(vid, pdid, uname)` key in one session and re-occupies it with a different CNID from another:

1. Session A creates `t506 dir` / `t506 subdir1` / `t506 subdir2`, nested, so all three land in A's cache.
2. Session B, a second afpd child on the same volume, moves and renames `subdir1` away.
3. Session B re-creates a *new* directory under the old name `subdir1`, same parent. That is the poison — a different CNID now holds a key A still maps to the old one.
4. Session A resolves `subdir2` via `FPGetFileDirParams`, walking its cached path through the stale `subdir1`.

Both key dimensions collide at once, and `curdir` sits on the entry being retired, which is what turns the expunge into a re-publish. The panic named the child of that poisoned parent: `did:19 pdid:18 name:"t506 subdir2"`.

**Why only one test, and only one CI round in three to five.** Three samples have to line up, none under the test's control:

- Validation is sampled. At the jobs' `AFP_DIRCACHE_VALIDATION_FREQ=100`, one lookup in a hundred stat-validates, so whether the stale entry is noticed at all — and the expunge runs — depends on where the test lands in that cycle.
- Which of the two entries dies is a bucket-order accident: `hash_lookup()` returns whichever sits first in the chain.
- The orphan has to be touched again. The panicking entry was a B1 ghost (`flags:0xc0`, `arc_list:3`), so it had to age out of T1/T2 first, which depends on pressure from every earlier test at `AFP_DIRCACHESIZE=1024`.

Miss one and the round is green — which is why 16 consecutive local runs stayed clean while CI failed about one round in three.

**The duplicates were not rare.** A trap at the point of duplicate *birth* fired in ~23 legs of a single run. Only the combination above turned one into a crash, which is why this read as a flaky test rather than a systemic key collision.

## Root cause

The crash was an entry absent from both tables whose own history showed no removal had run against it. From the instrumented build used to trace it (this wording is from the hunt; the fix rewrote that reporting):

```
dircache_remove: not in didname index: did:19 pdid:18 name:"t506 subdir2"
flags:0xc0 arc_list:3 qidx:null linked(dircache):no linked(didname):no
```

With two key-equal entries, removing B frees A's node: A stays reachable with its node gone, B stays linked while its caller frees it.

The duplicates were born in the code meant to prevent them. `dircache_add()` expunged a key-holder with `dir_remove()`, whose curdir recovery re-enters the cache:

```
dircache_add(X) for did:19
 └─ expunge finds duplicate D (did:19), which is curdir
     └─ dir_remove(D) → removes D, curdir = NULL
         └─ recovery: dirlookup(19) → miss → dircache_add(D')   ← re-publishes did:19
 └─ ...inserts X for did:19  →  two live entries, one key
```

It compounds, because a single-probe expunge removes at most one holder per index: three entries for `did:19` were observed, serials 384/385/386.

**The hash itself was never at fault.** Both tables pair their hash function with a real comparator, and every survivor's stored hash matched a freshly computed one. The entries were genuinely key-equal, not misrouted.

## How the cache got here (6 commits, ~21 years)

- **`e8bbed26` — didg, 2005.** kazlib imported with `#define NDEBUG` baked into `hash.c`, killing all 25 asserts — including `hash_insert()`'s uniqueness check — in every build type from day one.
- **`6ee7ecd6` — franklahm, 2010.** The dircache rewrite: two keyed indexes, an LRU queue, and removal by `hash_lookup()` from birth. Sound while every key has one holder, which nothing enforced.
- **`1a8fee3f` — franklahm, 2010.** "Pay attention to curdir on cache eviction" — first of many scattered curdir special-cases in removal paths.
- **`4e1a09e4` — andylemin, 2025.** Validation hardening gives `dir_remove()` its `dirlookup()` → parent → volume-root curdir recovery. Valuable on its own, but it makes `dir_remove()` re-entrant into the cache.
- **`d6d88ea8` — andylemin, 2026.** ARC mode: ghosts are full `struct dir`s, so ghost trims become extra removal paths, and MRU ordering cannot protect a sole node (both MRU and LRU).
- **`a2fa3170` — andylemin, 2026.** `dir_modify()`'s re-key removes the entry's own old didname key and inserts the new one, without checking whether a stale entry holds the destination. The second duplicate factory, on the rename path.

## Coverage matrix

✓ handled · ~ detected, but by aborting the session · ✗ silent · ✗-strand actively breaks other state.

| hazard | old code (asserts dead) | old code, asserts armed | new code |
|---|---|---|---|
| duplicate born by `dircache_add()` expunge recovery | ✗ silent | ~ abort at the re-insert | ✓ expunge retires without recovery |
| duplicate born by `dir_modify()` re-key onto a held name | ✗ silent | ~ abort in `hash_insert()` | ✓ stale holder retired first |
| key-equal duplicate present at removal | ✗-strand (wrong node freed) | ✗-strand (uniqueness gone by then) | ✓ removal is by stored node |
| stored node naming another entry | ✗ frees what a table points at | ✗ same | ✓ unreachable by construction; fatal if it ever occurs |
| insert while the cache sits at capacity | ✗ silent overfill past `maxcount` | ~ abort in `hash_insert()` | ✓ room made before the inserts |
| `curdir` taken by an expunge, new entry is a file | ✗ `curdir == NULL`, later deref crashes | ✗ same | ✓ volume root, or rootParent before it exists |
| abandoned re-key after an on-disk rename | ✗ stale identity stays cached | ✗ same | ✓ entry removed outright |
| entries of a closed volume | ✗ met as live on vid reuse | ✗ same | ✓ purged at `closevol()`, nothing exempt |
| sole ghost trimmed mid-promotion | ✗ freed under the promotion | ✗ same | ✓ pinned, selection skips it |
| `fullpath` on the `dirlookup()` error path | ✗ latent double free (unreachable while adds cannot fail) | ✗ same | ✓ ownership released at `dir_new()` |
| O(table) `hash_verify()` per mutation | (dead) | ✗ every cache op O(cache) in `plain`/`debugoptimized` | ✓ confined to `debug` |

## Index layout

| structure | key | entry's back-pointer |
|---|---|---|
| `dircache` hash | `(d_vid, d_did)` | `d_index_node` *(new)* |
| `index_didname` hash | `(d_vid, d_pdid, d_u_name)` | `d_didname_node` *(new)* |
| LRU queue / ARC T1·T2·B1·B2 | — (position) | `qidx_node` *(existing)* |

Both hashes use the `struct dir` as key and store the computed `hkey` on the node, and the comparator reads the entry's current fields. Removal-by-key therefore had two ways to go wrong that back-pointers cannot: a key-equal duplicate resolves to the wrong node, and a key mutated after insert resolves to nothing. `hash_delete()` navigates by the stored `hkey` and never rehashes, and kazlib nodes never move — grow and shrink reallocate only the chain-head array — so a stored node pointer stays valid for the entry's whole indexed life.

## Defects

1. **Wrong-node unlink.** With two key-equal entries, removal frees the node the key resolves to, not the entry's own.
2. **The expunge was a duplicate factory.** `dircache_add()` → `dir_remove()` → curdir recovery → `dircache_add()` re-published the key being cleared.
3. **The re-key was a second factory.** `dircache_reindex_didname()` inserted the destination key blind; a stale holder made two nodes under one key, or an abort once the uniqueness assert is armed.
4. **Partial removal.** The queue side was drained before the hash lookups could fail, so a miss left the entry half-unlinked with `queue_count != hash_nodecount` — the state the count asserts panic on.
5. **curdir holes.** The post-expunge re-point skipped files, leaving `curdir == NULL` with dozens of unguarded dereferences downstream.
6. **No volume teardown.** `closevol()` left the closing volume's entries cached, and the persistent vid met them as live on reopen.

Arming the asserts exposed a seventh, independent of all that: **ARC inserted into a full cache.** A ghost holds a hash slot just as a cached entry does, so the table sits at its declared 2c capacity once the working set exceeds the cache size. Case IV made room *after* `dircache_add()` had inserted, so `hash_insert()` got a full table and its capacity assert killed the session child. No corruption or race needed — just a volume walked wider than twice the cache — in every build that does not define `NDEBUG`, `plain` included.

## Fix

The invariant: **an entry's index membership is recorded on the entry, and removal consumes that record.** A missing stored node means "not in that index, skip the leg". Nothing else is survivable, because freeing memory a table still points at is the one unrecoverable move.

- **`hash_alloc_insert_node()`** (`hash.c`) — `hash_alloc_insert()` that returns the inserted node; the old entry point wraps it.
- **`d_index_node` / `d_didname_node`** (`directory.h`) — the stored nodes, set on insert and cleared on removal.
- **`dircache_remove()`** — deletes by stored node, skipping indexes the entry is not in. `dir_free()`'s refuse-to-free guard now checks both stored nodes too.
- **`dircache_expunge_duplicate()`** — retires a key-holder with a plain full-index removal, no `dirlookup()` recovery, handing curdir back for the caller to re-point at the replacement entry or the volume root. `dircache_reindex_didname()` now calls it for a stale destination holder.
- **`arc_case_iv_make_room()` / `arc_case_iv_insert()`** — Case IV splits, so `dircache_add()` evicts before its inserts and queues after. LRU mode always did this; ARC was the outlier.
- **`dircache_purge_vol()`** — from `closevol()`, removes each of the volume's entries as the scan reaches it, deferred free. Nothing is exempt: `v_root` is freed next, so a kept-back entry would outlive its volume under a reusable vid. A per-vid count kept at the insert and delete funnels answers the common case, nothing cached, with one array read instead of a table scan.
- **`dircache_remove_children()`** — same remove-as-you-scan shape, dropping a capacity-doubling array, its OOM path and a second loop. It cannot use `dir_remove()`, whose curdir recovery would insert mid-scan; it re-resolves a removed curdir once the scan is over, a rename having kept the CNID that resolves it.
- **`dircache_flush_deferred_for_vol()`** — scans with `hash_scan_next()` rather than walking chain heads, so all three table walks share one shape. The two resumable per-chain scans keep chain-level access, which is their point.
- **One pin on a promoting ghost, not two** — `arc_case_ii/iii` also pushed the ghost to MRU before `arc_replace()`. The skip pointer works from any position, so that splice was dead work that made the tested mechanism look optional.
- **`arc_ghost_trim_candidate(q, skip)`** — trim selection takes the pinned entry as an argument and steps over it; a sole pinned node yields no candidate.
- **`queue_remove(node)`** (`libatalk/util/queue.c`) — unlinks a node from whichever queue holds it, replacing `dequeue(node->prev)`, which worked only by exploiting the sentinel layout.
- **`HASH_ASSERT_VERIFY`** (`hash.c`) — `#define NDEBUG` removed here and in `bin/nad/ftw.c`. The cheap O(chain) asserts go to `b_ndebug`, live in `debug`/`debugoptimized`/`plain`; the O(table) `hash_verify()` walks are confined to `debug`, so no CI or distro build pays O(cache) per operation.

### Detected corruption is fatal

`dircache_remove()` reports a stored node that does not name its entry — the entry's identity and flags, which index disagreed, what that node carries — dumps the cache and panics. It does not try to continue. That matches how the cache already treats a broken invariant (`dir_free()` on an indexed entry, `hash_insert()`'s uniqueness assert, the ARC migration failures), and it is honest about reachability: every node is created with its entry as its data and freed only through that entry's own removal, so once keyed deletion is gone, nothing can produce a mismatched node. A tripwire is the right shape for that; recovery code behind it would be untestable by construction.

Declining the removal instead — leaving the entry published — was tried and is worse. A declined entry can never be retired, so it is served stale for the rest of the session, `dir_remove()` reports success without unpublishing, `dir_modify()` strands an entry in no index and on no queue, and the ARC requeue paths break their own size invariants. Five near-identical requeue blocks went with it.

### Riding along

Found while instrumenting the hunt:

- Worker threads blocked every signal, including the synchronous fault signals. The kernel cannot queue those to a thread that blocks them and forces the default action, so a fault in the idle worker died silently. The two handled fault signals stay deliverable, each thread installs an alternate stack so stack exhaustion is reportable, `fault.c` unwinds with libunwind where available, and the handler's re-entry counter is atomic now two threads can reach it.
- libunwind's include directories reach the target that actually compiles `fault.c`. Only the shared library carried the dependency, so any platform whose `libunwind.h` sits outside the default include path failed to build.
- A session whose reconnect handoff failed carried on after being told to die. It now terminates, running `afp_dsi_close()` and keeping `afp_dsi_die()`'s exit mapping — the failed handoff leaves this child owning the client socket, so volume close, CNID close and logout still have to run.
- The `Icon\r` cleanup queues the entry it removes for deallocation instead of unlinking it from every index and leaking it.

### Defect → fix

| # | Defect | Closed by |
|---|---|---|
| 1 | wrong-node unlink | stored nodes: removal cannot resolve to a bystander |
| 2 | expunge re-published the key | `dircache_expunge_duplicate()`, no recovery re-entry |
| 3 | re-key published a second node | `dircache_reindex_didname()` retires the stale holder first |
| 4 | partial removal, skewed counts | no lookup left to fail mid-way; a mismatched node is fatal, not partial |
| 5 | curdir NULL | curdir lands on the replacement, the volume root, or rootParent |
| 6 | entries outliving their volume | `dircache_purge_vol()` in `closevol()` |
| 7 | insert into a full cache | Case IV's eviction half runs before the inserts |

## Behaviour changes

- `struct dir` grows two pointers (16 bytes per entry) so removal is by node.
- `dircache_remove()` stays `void`; a stored node naming another entry is fatal rather than silently freeing what a table points at.
- `dircache_add()` keeps its contract: returns 0, or the process exits on a hash insert failure.
- Asserts live in `debug`/`debugoptimized`/`plain`, compiled out in `release`; `hash_verify()` only in `debug`.
- The three performance-dashboard jobs build **release**, so their numbers and the flamegraph measure shipped code rather than assertions. The flamegraph keeps `-g`, which `release` omits and perf needs. The perf image builds only on the events whose jobs consume it.
- Comments call the 2c index the dircache. "Directory" is the ARC paper's word for it and reads here as a filesystem directory.

## Testing

Unit tests, each proven RED against the defect it closes:

- `test007` — adds over both keys with `curdir` on the entry being retired. RED with `dir_remove()` restored in the expunge.
- `test008` — renames onto a key a stale entry holds. RED without the reindex expunge: `SIGABRT` on the uniqueness assert.
- `test009` — a file takes over curdir's `(vid, did)` key. RED with the old file-gated re-point: returns `curdir == NULL`.
- `test010` — saturates the cache to 2c and keeps inserting. RED with the eviction half after the inserts: `Assertion failed: hash->nodecount < hash->maxcount`, `SIGABRT`. It promotes rather than looks up, because the planted names do not exist on disk and a validating lookup would stat through the parent and expunge.
- `dircache_test_ghost_trim_selection` — five selection cases including the sole-pinned-ghost regression.

The suite runs twice, since both cache modes ship: **105 subtests per leg, 0 failures, in ARC and in LRU.** The harness initialises the dircache after `AFPobj` is published so it honours the configured mode, and a second meson leg forces LRU with `AFPDTEST_DIRCACHE_MODE=lru` — `index_queue` and `dircache_evict()` are reachable in that mode only. Before this, every unit test silently ran LRU and the shipped default was never exercised.

Suites and measurements:

- Full spectest on the final tree: **292 passed, 0 failed** — ARC, `AFP_DIRCACHE_VALIDATION_FREQ=100`, sqlite CNID, the configuration the original failure lived in. `Dircache:test500`–`test509` all pass, with no panic, assert or fault in the run.
- Assert cost measured, not assumed: spectest 65s → 64s with all 25 kazlib asserts live. Stripping verified in the binary: 2 assert strings in the debug image, 0 in release.
- `closevol` leak, measured on the purge as first written: 89 of 115 volume closes left entries behind → 0 of 115.
- The original failure reproduced about one CI round in three. Two 10-round reroll campaigns ran all green on the key-uniqueness work (20 consecutive rounds); everything layered on top is covered by the RED/GREEN unit tests, the spectest above, and the CI matrix.

## Not in this PR

Three candidates were weighed and left out, each costing more than it returns:

- **`arc_promoting_ghost` stays.** Removing the pin needs one of two trades. Detaching the ghost's node before `arc_replace()` means re-enqueueing into T2 after — a `malloc`/`free` per ghost hit where the node move is currently allocation-free, plus a new enqueue-failure panic. Moving it into T2 *before* `arc_replace()` lets `arc_find_victim()` pick it, since that takes the LRU and at `t2_size == 1` the LRU is the entry about to be returned as a hit; suppressing that needs a skip inside `arc_find_victim`, which is the same pin relocated. The explicit pin is now the only mechanism, and the unit test drives it.
- **`vid_entry_count[]` stays.** It is BSS, so demand-zero: a child dirties about a page per vid in use, not 256 KiB. Moving the counter to `struct vol` needs `getvolbyvid()` inside `dircache_remove()`, because five call sites pass `vol == NULL` — three ghost trims, the Case IV victim, `dircache_evict()` — adding a volume-list walk to the eviction path to reclaim pages nothing touches.
- **The `containers.yml` matrix fold stays.** The two testsuite-image jobs differ only by suffix and `BUILDTYPE`, but folding them is CI refactoring unrelated to cache correctness, validated only by pushing it, and a mistake there breaks the CI that validates this PR.

---

The crash class lived in the red edge — resolving an entry by key at removal time — which the stored back-pointers replace.

```mermaid
flowchart TD
    classDef wire   fill:#1e3a8a,stroke:#1e40af,color:#eff6ff,stroke-width:1px;
    classDef api    fill:#5b21b6,stroke:#6d28d9,color:#f5f3ff,stroke-width:1px;
    classDef struct fill:#0f766e,stroke:#0d9488,color:#f0fdfa,stroke-width:1px;
    classDef index  fill:#b45309,stroke:#d97706,color:#fffbeb,stroke-width:1px;
    classDef global fill:#9f1239,stroke:#be123c,color:#fff1f2,stroke-width:1px;

    subgraph WIRE["🌐 AFP ops that touch the cache"]
        OPS("FPEnumerate · FPCreateDir/File · FPMoveAndRename<br/>FPDelete · FPCloseVol · every dirlookup")
    end

    subgraph API["🟣 dircache API — etc/afpd/dircache.c"]
        ADD("dircache_add()<br/>expunge both keys → make room → insert both → queue")
        REM("dircache_remove()<br/>delete by stored node, skip absent indexes<br/><i>node naming another entry = panic</i>")
        REK("dircache_reindex_didname()<br/>retire stale holder → re-insert")
        PUR("dircache_purge_vol()<br/>removes each entry as the scan reaches it")
        SRCH("dircache_search_by_did / _by_name<br/>keyed lookups (unchanged — reads only)")
    end

    subgraph ENTRY["🟢 struct dir — one per cached object"]
        DIR("d_vid · d_did · d_pdid · d_u_name<br/>d_flags (DIRF_INDEXED · DIRF_ISFILE · DIRF_ARC_GHOST)")
        NODES("back-pointers, written at insert:<br/>d_index_node · d_didname_node · qidx_node")
    end

    subgraph INDEXES["🟠 the three structures an entry lives in"]
        DC("dircache hash<br/>key (vid, did) · hkey stored on node")
        DN("index_didname hash<br/>key (vid, pdid, uname)")
        Q("LRU queue / ARC T1·T2·B1·B2<br/>ghosts are full entries")
        DEFER("invalid_dircache_entries<br/>deferred free — pointers handed out<br/>this request stay readable")
    end

    CURDIR("curdir — per-session global<br/>never NULL: replacement entry,<br/>volume root, else rootParent")

    OPS --> ADD & REM & REK & SRCH
    OPS -->|"FPCloseVol / logout"| PUR
    ADD -->|"insert returns the node"| NODES
    REK -->|"insert returns the node"| NODES
    DIR --- NODES
    NODES ==>|"delete by stored node — O(1), identity-exact"| DC
    NODES ==>|"delete by stored node"| DN
    NODES ==>|"unlink by stored node"| Q
    REM -->|"consumes"| NODES
    REM -->|"on success"| DEFER
    PUR --> REM
    SRCH -.->|"read-only keyed lookup"| DC
    SRCH -.->|"read-only keyed lookup"| DN
    ADD & REM -.->|"handoff, never NULL"| CURDIR

    REMOLD("old model: hash_lookup(key) at removal<br/>→ frees whichever node the key resolves to")
    REMOLD -. "⚠ key-equal duplicate ⇒ unlinks the WRONG entry — removed by this PR" .-> DC

    class OPS wire;
    class ADD,REM,REK,PUR,SRCH api;
    class DIR,NODES struct;
    class DC,DN,Q,DEFER index;
    class CURDIR,REMOLD global;

    style WIRE   fill:#f1f5f9,stroke:#cbd5e1,color:#334155;
    style API    fill:#f8fafc,stroke:#cbd5e1,color:#334155;
    style ENTRY  fill:#eef2f6,stroke:#d1d9e0,color:#334155;
    style INDEXES fill:#eef2f6,stroke:#d1d9e0,color:#334155;

    linkStyle 18 stroke:#dc2626,stroke-width:2px;
```

### DIFF
--- a/distrib/docker/netatalk.Dockerfile
+++ b/distrib/docker/netatalk.Dockerfile
@@ -8,6 +8,7 @@ ARG RUN_DEPS="\
     krb5 \
     libevent \
     libgcrypt \
+    libunwind \
     linux-pam \
     mariadb-client \
     mariadb-connector-c \
@@ -27,6 +28,7 @@ ARG BUILD_DEPS="\
     krb5-dev \
     libevent-dev \
     libgcrypt-dev \
+    libunwind-dev \
     linux-pam-dev \
     mariadb-dev \
     meson \
--- a/.github/workflows/containers.yml
+++ b/.github/workflows/containers.yml
@@ -92,6 +92,46 @@ jobs:
           labels: ${{ steps.metadata.outputs.labels }}
           tags: ${{ steps.metadata.outputs.tags }}
 
+  build-container-testsuite-perf:
+    name: Build Alpine testsuite container (release, for perf)
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    # Gated like its only consumers (latencygraph-lantest, perfgraph-speedtest),
+    # so pushes and dependabot PRs do not build an image nothing pulls
+    if: ${{ github.event_name == 'pull_request' && github.actor != 'dependabot[bot]' && !github.event.pull_request.head.repo.fork
+      }}
+    permissions:
+      packages: write
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+      - name: Create image name
+        run: |
+          echo "IMAGE_NAME=${REPO,,}-testsuite-perf" >> ${GITHUB_ENV}
+      - name: Login to container registry
+        uses: docker/login-action@dbcb813823bdd20940b903addbd779551569679f # v4.6.0
+        with:
+          registry: ${{ env.REGISTRY }}
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+      - name: Extract metadata
+        id: metadata
+        uses: docker/metadata-action@dc802804100637a589fabce1cb79ff13a1411302 # v6.2.0
+        with:
+          images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
+          tags: type=raw,value=${{ github.sha }}
+      # As the correctness image but release, so the numbers exclude the asserts
+      - name: Build and push container image
+        uses: docker/build-push-action@53b7df96c91f9c12dcc8a07bcb9ccacbed38856a # v7.3.0
+        with:
+          context: .
+          file: distrib/docker/testsuite_alp.Dockerfile
+          push: true
+          build-args: |
+            BUILDTYPE=release
+          labels: ${{ steps.metadata.outputs.labels }}
+          tags: ${{ steps.metadata.outputs.tags }}
+
   build-container-webmin:
     name: Build Netatalk Webmin Module container
     runs-on: ubuntu-latest
@@ -1917,7 +1957,7 @@ jobs:
 
   latencygraph-lantest:
     name: Lantest (AFP 3.4) - LatencyGraph
-    needs: build-container-testsuite
+    needs: build-container-testsuite-perf
     runs-on: ubuntu-latest
     timeout-minutes: 10
     if: ${{ github.event_name == 'pull_request' && github.actor != 'dependabot[bot]' }}
@@ -1966,7 +2006,7 @@ jobs:
             -e AFP_DIRCACHE_RFORK_MAXSIZE="${{ env.AFP_DIRCACHE_RFORK_MAXSIZE }}" \
             -e AFP_TEST_ITERATIONS="${{ env.AFP_TEST_ITERATIONS }}" \
             -e TEST_FLAGS="-c" \
-            ${{ env.REGISTRY }}/netatalk/netatalk-testsuite:${{ github.sha }} \
+            ${{ env.REGISTRY }}/netatalk/netatalk-testsuite-perf:${{ github.sha }} \
             > lantest.csv 2> lantest.log
       - name: Show lantest log
         if: always()
@@ -2098,7 +2138,7 @@ jobs:
 
   perfgraph-speedtest:
     name: Speedtest (AFP 3.4) - PerfGraph
-    needs: build-container-testsuite
+    needs: build-container-testsuite-perf
     runs-on: ubuntu-latest
     timeout-minutes: 10
     if: ${{ github.event_name == 'pull_request' && github.actor != 'dependabot[bot]' }}
@@ -2148,7 +2188,7 @@ jobs:
             -e AFP_DIRCACHE_RFORK_MAXSIZE="${{ env.AFP_DIRCACHE_RFORK_MAXSIZE }}" \
             -e AFP_TEST_ITERATIONS="${{ env.AFP_TEST_ITERATIONS }}" \
             -e TEST_FLAGS="-c -z 0.00390625,0.0078125,0.015625,0.03125,0.0625,0.125,0.25,0.5,1,2,4,8,16,32,64,128,256" \
-            ${{ env.REGISTRY }}/netatalk/netatalk-testsuite:${{ github.sha }} \
+            ${{ env.REGISTRY }}/netatalk/netatalk-testsuite-perf:${{ github.sha }} \
             > speedtest.csv 2> speedtest.log
       - name: Show speedtest log
         if: always()
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -81,6 +81,7 @@ required at the bare minimum.
 | libldap                    | LDAP support |
 | libpam                     | PAM support |
 | libtirpc **OR** libquota   | Quota support |
+| libunwind                  | symbolized backtraces when a daemon crashes |
 | Perl                       | admin scripts |
 | po4a                       | localization of documentation |
 | stuffit-ffi                | StuffIt support in `nad` |
--- a/bin/nad/ftw.c
+++ b/bin/nad/ftw.c
@@ -49,7 +49,6 @@
 #define mempcpy(D, S, N) ((void *) ((char *) memcpy (D, S, N) + (N)))
 #endif
 
-#define NDEBUG 1
 #include <assert.h>
 
 #ifndef _LIBC
--- a/distrib/docker/debug_alp.Dockerfile
+++ b/distrib/docker/debug_alp.Dockerfile
@@ -1,5 +1,5 @@
 # Debug / profiling Alpine Dockerfile for Netatalk
-# Builds with -Dbuildtype=debugoptimized (-O2 -g) and includes perf + FlameGraph tools
+# Builds with -Dbuildtype=release plus explicit -g, and includes perf + FlameGraph tools
 # See doc/developer/profiling.md for usage guidance
 
 ARG RUN_DEPS="\
@@ -91,11 +91,11 @@ COPY meson_options.txt .
 COPY meson.build .
 RUN rm -rf build
 
-# Build with debugoptimized: -O2 optimization + debug symbols (-g)
-# Also preserve frame pointers for accurate perf stack unwinding
+# Release, so the profile is not dominated by assertions. -g is explicit because
+# release omits it and perf needs symbols; frame pointers for stack unwinding.
 RUN meson setup build \
-    -Dbuildtype=debugoptimized \
-    -Dc_args=-fno-omit-frame-pointer \
+    -Dbuildtype=release \
+    -Dc_args="-fno-omit-frame-pointer -g" \
     -Dwith-appletalk=true \
     -Dwith-dbus-daemon-path=/usr/bin/dbus-daemon \
     -Dwith-docs= \
--- a/distrib/docker/testsuite_alp.Dockerfile
+++ b/distrib/docker/testsuite_alp.Dockerfile
@@ -12,6 +12,7 @@ ARG RUN_DEPS="\
     libedit \
     libevent \
     libgcrypt \
+    libunwind \
     linux-pam \
     localsearch \
     mariadb \
@@ -40,6 +41,7 @@ ARG BUILD_DEPS="\
     libedit-dev \
     libevent-dev \
     libgcrypt-dev \
+    libunwind-dev \
     linux-pam-dev \
     mariadb-dev \
     meson \
@@ -58,6 +60,9 @@ ARG RUN_DEPS
 ARG BUILD_DEPS
 ENV RUN_DEPS=$RUN_DEPS
 ENV BUILD_DEPS=$BUILD_DEPS
+
+# Perf jobs pass release, so their numbers do not include the asserts
+ARG BUILDTYPE=debugoptimized
 
 RUN apk update \
 &&  apk add --no-cache \
@@ -113,7 +118,7 @@ COPY meson.build .
 RUN rm -rf build
 
 RUN meson setup build \
-    -Dbuildtype=debugoptimized \
+    -Dbuildtype=$BUILDTYPE \
     -Dwith-appletalk=true \
     -Dwith-dbus-daemon-path=/usr/bin/dbus-daemon \
     -Dwith-docs= \
--- a/doc/manual/Installation.md
+++ b/doc/manual/Installation.md
@@ -157,6 +157,13 @@ functionality.
     library, netatalk can produce the GSS UAM library for authentication
     with existing Kerberos infrastructure.
 
+- libunwind
+
+    When a daemon hits a fatal fault, netatalk logs a backtrace of the
+    faulting thread, and libunwind names the functions in it. Builds
+    without it fall back to `backtrace_symbols()`, or to bare addresses
+    on C libraries that lack it, such as musl.
+
 - PAM
 
     PAM provides a flexible mechanism for authenticating users. PAM was
--- a/etc/afpd/afp_dsi.c
+++ b/etc/afpd/afp_dsi.c
@@ -504,7 +504,14 @@ static int process_deferred_signals(AFPObj *obj)
         int sig = die_pending;
         die_pending = 0;
         afp_dsi_die(sig);
-        /* afp_dsi_die calls exit() — should not return */
+        /* Reached only with DSI_RECONINPROG set: the handoff failed
+         * (auth.c leaves the flag set) and this child still owns the client
+         * socket, so run the full teardown like the afp_over_dsi() backstop,
+         * keeping afp_dsi_die()'s diagnostic exit mapping. */
+        LOG(log_note, logtype_afpd, "Reconnect-state session terminating");
+        iw_shutdown();
+        afp_dsi_close(obj);
+        exit((sig == SIGTERM || sig == SIGALRM) ? 0 : sig);
     }
 
     /* Primary reconnect (SIGURG) */
--- a/etc/afpd/dircache.c
+++ b/etc/afpd/dircache.c
@@ -20,6 +20,7 @@
 #include <assert.h>
 #include <errno.h>
 #include <signal.h>
+#include <stdint.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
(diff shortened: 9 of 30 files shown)


andylemin:
Hi @rdmark After what started as a quick "let's debug this occasional crash" on #3281 since adding crash backtracing showed the fault was internal and not CI related, it turned into a long hunt... 

It took a lot of CI runs to find the race conditions that triggered it. Once I found the defect class (dircache relying on just one key, which is weaker than the CNID constraints), exploring the fault class exposed a whole bunch of other potential triggers.

This drove an update to the fundamental dircache design to use the same full constraints as CNID, which closed the gap on all the other risk areas, resulting in a much stricter design 🙂

Detected corruption is now fatal as it should be impossible to occur, rather than complex recovery logic. So cleaner overall.

It went unnoticed as the invariant checks were always compiled out - after enabling them the situation was easy to detect as being often present, just rarely reaching a clear exception condition.

Almost ready for review

andylemin:
@rdmark Ready for review

andylemin:
The fixes in here are critical for the upcoming release. #3281 is not critical, but this should merge after that to avoid regressing some things

andylemin:
@rdmark Polished and bottomed out finally on the checks and balances here.. This was quite something but very happy with the outcome. This is true correctness without compromise. I had to really resist compromise until I cracked it 😂

andylemin:
closed and reopened to trigger the CI's one more time for good measure, and still working perfectly. 20 CI runs now without a failure.

distrib/docker/netatalk.Dockerfile
```diff
@@ -8,6 +8,7 @@ ARG RUN_DEPS="\
     krb5 \
     libevent \
     libgcrypt \
+    libunwind \
```
rdmark, line 11: is this library really used in the production container?

rdmark (approved):
one non-blocking request: document the newly introduced libunwind dependency in doc/manual/Installation.md (and maybe INSTALL.md too?)

distrib/docker/netatalk.Dockerfile
andylemin: Yes, but only optional. Without it an crashes will have only basic tracebacks. This provides detailed tracebacks if installed.

andylemin:
> one non-blocking request: document the newly introduced libunwind dependency in doc/manual/Installation.md (and maybe INSTALL.md too?)

Done, thanks for spotting that. Clarified as optional and what it does

PR #3286, opened 2026-09-03
afpd: multi protocol option declares multi-accessor volumes; dircache validation freq defaults to 100

### LINKED ISSUE #3270: RFC: the default settings currently serve neither deployment model — align defaults with single-accessor mode
## The problem

Netatalk now works well for two distinct deployment models:

- **Single-accessor** (Netatalk is the only thing touching the volumes): high-performance, fully cache-backed operation — coherence is memory-bound.
- **Multi-accessor** (Samba, NFS, or local processes also mutate the volumes): strict locking, strict cache validation, interoperable — coherence is storage-bound.

**However, today's defaults are suboptimal for both parties.** For multi-accessor deployments, `strict locking` is off by default — locks are not checked, so Netatalk and Samba can trample each other's files. For single-accessor deployments, the cache is not trusted — every operation still stats the filesystem to revalidate an entry we already know is coherent. We are not serving either group correctly by default, and the two defaults contradict each other: one pays per-access to detect external writers that the other pretends don't exist.

Since most settings already assume the single-accessor model, I propose (alongside #3264–#3267):

1. **Default `dircache validation freq` to 100** (periodic validation for good measure) so that *all* defaults are aligned with single-accessor mode — a Netatalk-only server is then optimal and performant with no tuning at all.
2. **`ea = samba` becomes the single switch for multi-accessor mode.** It already force-defaults everything storage-bound coherence requires: Samba-compatible extended attributes, `strict locking = yes`, `dircache validation freq = 1`, resource-fork cache exclusion, and (Solaris) F_SHARE reservations — with explicit settings still winning, and a logged warning when they weaken coherency (#3223).

We also recommend `ea = samba` even when the other accessor is NFS or a local process other than Samba: it stores metadata in extended attributes instead of AppleDouble sidecar files, which behaves better under NFS and local tooling (fewer files, and a local copy carries the EAs along).

## Why the cache can now be trusted for single-accessor performance

Over the past year the directory cache became internally consistent by construction. It tracks, per entry: filesystem stat metadata (timestamps, size, ownership, permissions), the CNID and parent DID, the name in every needed encoding plus the full path, directory offspring counts, AppleDouble metadata (FinderInfo, dates, AFP attributes), the resource-fork length including a negative "no AppleDouble exists" marker, and resource-fork data itself (Tier 2, #2783).

Every file and directory mutation updates or flushes the affected entries, and mutations are signalled to the other users' afpd processes through a cross-process hint channel (#2733, #2831, #2853), with an idle worker handling maintenance off the hot path (#2808, #3212).

Historically, Netatalk stat()ed the filesystem object on *every single* cache read to compare timestamps against the cache before trusting it — which capped the cache's value at "saves a directory walk". Probabilistic validation (#2447, hardened in #2577) made the paranoia tunable, and the remaining validation cost fell further when it became a relative fstatat against a cached parent directory fd (#3222). Two things stay fully paranoid regardless of the setting: **all critical mutations** (move, delete, rename, copy, exchange) validate against storage every time, and `ea = samba` volumes force validation on every access. The optimistic trust model applies only to non-destructive read-type operations — and it has now had extended production exposure.

## Why strict locking can now be trusted for multi-accessor interoperability

The locking subsystem was overhauled end to end: the release-path, stranding, and share-mode defects that made the old `afp read locks` framework unsafe to enable are fixed (#2645, #3108–#3112, #3145, #3169, #3180), the FPDelete blocking model is normalised to POSIX/macOS/Samba semantics (#3210), and the renamed `strict locking` is exercised by a bidirectional Samba interop test (kernel CIFS mount + AFP client on one shared volume — #3223). This resolved long-standing Samba interoperability issues; the strict path is no longer the buggy one.

### Known follow-up

The `nad` utility does not yet convert metadata between the native Netatalk EA format and the Samba-compatible format; that conversion path is outstanding work for migrating existing volumes into `ea = samba` (or visa versa).

## The request for comments?

- As per above, I propose we change `dircache validation freq = 1` to `dircache validation freq = 100` by default, and declare `ea = samba` the single switch to flip between modes?
 
- "Flipping between modes", implies this is a trivial change which can be done at any time - however it requires an as yet non-existent AD conversion process (or alternatively/additionally, smart logic in Netatalk which dynamically changes ADs to/from Samba format as required when it comes across a format which does not match the afp.conf setting).
   - We could make the Samba format the defacto standard format under `ea = sys`?

- If we proceed: this is a behaviour change for silent configs and needs prominent NEWS/release-note treatment. Eg `Volumes mutated by other processes should set 'ea = samba'` (with a link to a page explaining each of the underlying settings it influences). this is especially pertinent for NAS vendors/users who are the most common multi-accessor mode users.
 
- Are there other deployment shapes that fit neither model — Eg mutation patterns where a 100-access staleness window on read paths is unacceptable but full samba coherence is overkill?

andylemin:
Netatalk's defaults contradicted each other. `strict locking` defaulted to *no*, which assumes nothing else touches the volume. `dircache validation freq` defaulted to *1*, paying a revalidation stat on every cache hit to catch changes by the very processes the locking default assumed away. Neither deployment was served well: an AFP-only server paid for coherence it did not need, and a Samba or NAS server ran with locks its peers could not see.

This PR adds **`multi protocol = yes|no`** as the one place to declare that something else writes the volume, moves every coherency default behind it, returns `ea` to choosing only the on-disk metadata format, and flips `dircache validation freq` to 100 so an untouched config suits a Netatalk-only server.

## Root cause

The coherency defaults were keyed on `ea = samba`. That option answers a storage question — append a NUL to each xattr so Samba's `vfs_streams_xattr` can read it — not a coherence one. Three things followed:

- NFS and local writers (cron, rsync, media post-processing, a NAS web UI) mutate volumes too, but have no reason to want Samba's xattr layout. There was no way to say "something else writes here" without also changing the on-disk format, which cannot be converted in place.
- An operator who wanted the Samba format purely as a storage choice could not decline the locking and revalidation overhead that came with it.
- A format switch silently decided whether AFP reads and writes take POSIX locks, which is the last place anyone looks when debugging a locking problem.

The global `freq = 1` default, meanwhile, only half-protected the undeclared-writer case that `strict locking = no` left open anyway.

## How the defaults got here (4 commits, ~14 years)

Each commit was reasonable on its own; the contradiction is emergent.

- **`8fa8b411`** (Frank Lahm, 2012) adds `afp read locks`, default *no*. Right for a time when the locks cost real throughput and deployments were AFP-only. The defaults start assuming exclusive access here.
- **`2a1f4d31`** (Ralph Boehme, 2015) adds `ea = samba` as a pure storage format for `vfs_streams_xattr` interoperability, with no coherence semantics attached.
- **`bb28fd0e`** (Andy Lemin, 2025) adds `dircache validation freq`, defaulted to *1* to preserve existing behaviour. Conservative for a new tunable, but it fixes the paranoid posture as the global default while the locking default stays optimistic. Both halves of the contradiction now coexist.
- **`ddbf9073`** (Andy Lemin, 2026) sees that safe concurrent access needs four coordinated settings and makes one option default them all, picking `ea = samba` because it was the only existing signal that Samba was involved. Right diagnosis, overloaded switch; this PR keeps the mechanism and re-keys it.

Two commits this cycle raised the stakes: `c455b925` turned the resource-fork cache on by default, and `4ab39fa6` made revalidation a cached-fd `fstatat` rather than a path walk, which is what makes a lazier default worth having.

### Coverage matrix

Which declaration actually governed which setting. Read down the "new" column: every cell is deliberate.

✓ governed correctly · ~ governed only as a side effect · ✗ not governed

| coherency setting | silent config | `ea = samba` | `multi protocol = yes` |
|---|---|---|---|
| `strict locking` | ✗ off | ~ on, via a format switch | ✓ on |
| `dircache validation freq` | ~ 1 for everyone, AFP-only included | ~ forced to 1 | ✓ 1 declared, 100 otherwise |
| resource-fork cache exclusion | ✗ cached | ~ excluded | ✓ excluded, per volume |
| Solaris `F_SHARE` | ✓ on | ✓ on, warns if disabled | ✓ on, warns if disabled |
| EA storage format | ✓ independent | ✗ fused to coherence | ✓ independent |
| NFS / local writers | ✗ undeclarable | ✗ only by adopting Samba's format | ✓ declarable |
| AFP-only performance | ✗ pays for unused coherence | n/a | ✓ cache-backed |

## The switch

`multi protocol = yes|no` (default *no*) resolves volume → preset → `[Global]` through the existing `vdgoption_bool()`, the same precedence `ea` uses. It sets `AFPVOL_MULTIPROTO` and triggers the defaults from `creatvol()`'s commit block, after the volume can no longer fail, so a rejected volume never changes process state. `apply_multiprotocol_defaults()` is the renamed `apply_samba_defaults()`: same mechanism, new trigger, generalised messages. Defaults only, so an explicit setting still wins and logs a warning when it weakens coherency.

The scope is mixed, and now documented as such: `strict locking` and `dircache validation freq` are server-wide, so one declared volume makes every volume stricter. Only the resource-fork cache exclusion is per volume.

`ea` goes back to choosing storage alone. Because the two formats cannot be converted in place, `multi protocol = yes` never changes it; it logs which format the volume uses and that Samba reads only `ea = samba`. That note, and the reverse pointer on an `ea = samba` volume with no declaration, come from the startup load only. `creatvol()` runs again in every session child, so advice about static configuration would otherwise repeat on every login; `pwd` is NULL only on the startup load, which draws the line without new plumbing.

### Defect → fix

Found reviewing the first cut of this branch:

| Defect | Fix |
|---|---|
| The 1-100 range for `dircache validation freq` was enforced three times downstream but never at parse, so an out-of-range value arrived unbounded, silently became validate-always, and suppressed both coherency messages | Pass the real bounds to the existing `safe_atoi()`, which warns and falls back to the default; the range is named once in `globals.h` |
| An unusable value still counted as explicit, so `multi protocol = yes` no longer returned it to 1 and the warning misreported it as "explicitly 100" | `conf_int_key_usable()` beside `conf_key_present()`: explicit now means the operator's value was usable |
| Statistics divided by a copy of the frequency seeded at `dircache_init()`, before volumes load, so the coherency default never reached it — validation behaved correctly while the numbers under-reported it by up to 100x | One `validation_freq()` reader for the decision and the statistics; the shadow copy and `dircache_set_validation_params()` give way to the existing `dircache_reset_validation_counter()` |
| Storage-format advice repeated on every login, and pushed a conversion the man page itself calls potentially data-losing at warning level, including on volumes shared only with NFS | `log_ea_format_advice()`, startup load only, demoted to a note that states the situation instead of recommending a conversion |
| `AFP_MULTI_PROTOCOL` was pasted into afp.conf for the strict boolean parser, so a documented value like `2` was rejected as unset, while `0` would have enabled a server-wide posture change | Non-zero test, matching the documented contract and its siblings `AFP_READONLY` and `DISABLE_TIMEMACHINE` |
| Four `afp.conf(5)` passages and two NEWS entries still credited `ea = samba` with the defaults, and the Webmin module had no control for the new option | Documentation retargeted, with the server-wide scope now stated; Webmin gains select options, a row in both section editors, and labels |

## Behaviour changes

- **`dircache validation freq` defaults to 100.** Volumes other processes modify must declare `multi protocol = yes`, which returns it to 1 with the rest of the set. Move, delete, rename, copy and exchange still validate against storage every time, and internal writes still update or flush the affected entries.
- **`ea = samba` no longer changes any coherency default.** It selects the Samba-compatible xattr format and nothing else. The behaviour it used to imply was added this cycle in `ddbf9073` and never shipped, so no released config depended on it.
- **An unusable `dircache validation freq` is warned about and replaced by the default** instead of silently becoming validate-always, and no longer counts as an explicit setting.
- **NEWS states the migration requirement plainly:** every deployment where anything other than Netatalk writes the storage must set `multi protocol = yes` — Samba or NFS on the same volume, NAS products whose UI or apps write to the share, and hosts where scripts, cron jobs, rsync or shell users touch files.

## Testing

**Spectests.** Both legs pass on a single-container loopback: `ea = sys` with the CI environment, and `ea = ad` with `AFP_DIRCACHE_VALIDATION_FREQ` unset, so that leg runs on the new compiled default of 100 end to end, external-mutation cases included (`FPResolveID` local modify/delete/rename, permission-cache invalidation).

**Unit tests.** `meson test` green. The conf suite covers switch resolution (volume beats preset beats `[Global]`, and a volume-level `no` under a global `yes`), the applied defaults and their notes, explicit settings surviving with warnings, process state left untouched when a volume fails after the option parse, the stock freq-100 posture, parse bounds across unset/valid/zero/negative/over-max/unparsable, an unusable explicit value still being returned to 1, `ea = samba` applying no defaults, and both advice wordings including silence when the format already matches.

**Runtime.** Verified in a container that the advice logs once at startup and stays at one across three logins, while the strict-locking note reaches four (startup plus three sessions) — so the sessions did load volumes, and only the static advice is suppressed.

**Interoperability.** `samba_interop_test.sh` now drives its coherency assertions through `multi protocol = yes` alongside `ea = samba`, and asserts the settings come from the defaults rather than explicit keys. This leg needs the Debian testsuite image with privileged CIFS mounts and has not been run locally.

## Not in this PR

- **`sys` ↔ `samba` EA conversion.** No in-place conversion exists, which is why `multi protocol = yes` only mentions `ea = samba` rather than defaulting it. Transparent on-mismatch conversion plus a `nad` subcommand would make format changes routine, and would let the mention become a default.
- **Per-volume enforcement of the server-wide settings.** `strict locking` and `dircache validation freq` are process-global, so a mixed server applies the strict posture everywhere. Every consuming site already holds the volume pointer, so this is feasible as its own change.
- **Live reconfiguration.** `creatvol()`'s reload early-exit means the option reaches sessions that connect after startup; existing sessions keep the posture they started with.
- **Measurement.** The performance benefit of the lazier default is not quantified here.

```mermaid
flowchart TB
    classDef decl fill:#e0f2fe,stroke:#0284c7,color:#0c4a6e;
    classDef fmt fill:#fef3c7,stroke:#d97706,color:#78350f;
    classDef global fill:#fee2e2,stroke:#dc2626,color:#7f1d1d;
    classDef vol fill:#dcfce7,stroke:#16a34a,color:#14532d;
    classDef advice fill:#f3e8ff,stroke:#9333ea,color:#4c1d95;

    subgraph CONF["afp.conf — two independent questions"]
        direction TB
        MP("<b>multi protocol = yes|no</b> (default no)<br/>Does anything else write this volume?<br/><i>Samba · NFS · cron · rsync · NAS apps</i>")
        EA("<b>ea = sys | samba | ad | none</b><br/>How is metadata stored on disk?")
    end

    subgraph RESOLVE["creatvol() — per volume"]
        direction TB
        VDG("vdgoption_bool()<br/>volume → preset → [Global] → no")
        FLAG("v_flags |= AFPVOL_MULTIPROTO")
        COMMIT("commit block<br/><i>after the volume can no longer fail</i>")
    end

    subgraph DEFAULTS["apply_multiprotocol_defaults() — defaults only; explicit wins and warns"]
        direction TB
        SL("strict locking = yes<br/><i>server-wide</i>")
        FREQ("dircache validation freq = 1<br/><i>server-wide · else 100</i>")
        RFORK("resource-fork cache exclusion<br/><i>per volume · fork.c gate</i>")
        FSHARE("Solaris F_SHARE reservations<br/><i>server-wide</i>")
    end

    subgraph ADVICE["log_ea_format_advice() — startup load only (pwd == NULL)"]
        direction TB
        REC("multi protocol without ea = samba<br/>→ native format, Samba cannot read it")
        PTR("ea = samba without multi protocol<br/>→ declare multi protocol")
    end

    MP --> VDG --> FLAG --> COMMIT
    COMMIT -->|"AFPVOL_MULTIPROTO set"| SL
    COMMIT --> FREQ
    COMMIT --> RFORK
    COMMIT --> FSHARE
    COMMIT --> ADVICE
    EA -->|"AFPVOL_EA_SAMBA · ea_sys.c trailing NUL"| DISK("on-disk metadata<br/><i>native or Samba-compatible xattrs</i>")
    EA -. "never changed implicitly<br/>(no in-place conversion)" .-> REC

    FREQ -->|"live option · single reader"| VF("validation_freq()<br/><i>validation decision + statistics</i>")
    PARSE("afp_config_parse()<br/>safe_atoi() bounds to MIN..MAX") --> VF

    class MP decl;
    class EA,DISK fmt;
    class SL,FREQ,FSHARE,VF,PARSE global;
    class RFORK,FLAG,VDG,COMMIT vol;
    class REC,PTR advice;

    style CONF fill:#f8fafc,stroke:#cbd5e1,color:#334155;
    style RESOLVE fill:#f1f5f9,stroke:#cbd5e1,color:#334155;
    style DEFAULTS fill:#f8fafc,stroke:#cbd5e1,color:#334155;
    style ADVICE fill:#faf5ff,stroke:#d8b4fe,color:#4c1d95;
```

Closes #3270


andylemin:
@rdmark With this one, its a wrap from me until post Golden Gate 🎉
We should now be bug fix and feature milestone complete with 3281, 3285, 3286 🍻
Need to start working on the press release, I have no idea how we can even begin to describe the milestones achieved!

rdmark:
I'm also looking forward to reading your press release draft ;)

contrib/webmin_module/lang/en
```diff
@@ -111,6 +111,7 @@ edit_vol_section_basedir_regex=Base directory
 edit_vol_section_basedir_regex_help=Example: /home
 edit_vol_section_home_path=Relative path
 edit_vol_section_ea=Metadata format
+edit_vol_section_multi_protocol=Multi protocol (modified by other processes)
```
rdmark: this is minor, but when the same label exists for both volume and global level options, I have so far only defined the latter and used it in both places

saves us a duplicated string to maintain!

include/atalk/volume.h
```diff
@@ -148,6 +148,7 @@ typedef enum {
 #define AFPVOL_FORCE_STICKY_XATTR (1 << 11) /*!< write metadata xattr as root on sticky dirs */
 #define AFPVOL_LIMITSIZE (1 << 12)   /*!< limit size for older macs */
 #define AFPVOL_A2VOL     (1 << 13)   /*!< prodos volume */
+#define AFPVOL_MULTIPROTO (1 << 14)  /*!< other processes mutate the volume: coherency defaults apply */
```
rdmark, line 151: out of curiosity I looked up what the 14th bit used to be used for:

```c
#define AFPVOL_DROPBOX   (1 << 14)  /* dropkludge dropbox support */
```

rdmark (approved):

include/atalk/volume.h
andylemin: 🤔

contrib/webmin_module/lang/en
andylemin: Good tip, thanks — fixed. The volume row now uses the existing `edit_global_section_multi_protocol` string, so there is only one label to maintain (matching the 13 other shared labels already used in `edit_vol_section.cgi`).

include/atalk/volume.h
andylemin: Nice archaeology! Safe to reuse: `v_flags` is runtime-only state, rebuilt from afp.conf on every volume load and never written to disk or put on the wire, and `AFPVOL_DROPBOX` has been gone since 2009 (`6dcaba1d`) with no references left in the tree.

There is also no pristine bit to move to — every free slot is a retired one: 15 was `AFPVOL_NOFILEID`, 19 was `AFPVOL_CASEINSEN`, and 21 was `AFPVOL_CACHE`. Happy to shuffle it if you would rather keep distance from the old name.

PR #3290, opened 2026-09-06
ci: hardening of workflow and job level permissions

rdmark:
two changes to address weaknesses in the GitHub workflow permissions configuration: set a workflow level default permission in the VM spectest workflow, and remove checks write permissions from all jobs and workflows

### DIFF
--- a/.github/workflows/containers.yml
+++ b/.github/workflows/containers.yml
@@ -222,7 +222,6 @@ jobs:
     if: ${{ !github.event.pull_request.head.repo.fork && github.actor != 'dependabot[bot]' }}
     permissions:
       contents: read
-      checks: write
     steps:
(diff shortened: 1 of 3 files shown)


rdmark:
@andylemin from what I can tell the "checks: write" permissions aren't needed for the spectest Junit reports to be displayed under Checks -- the Scorecard scanner calls out that they are redundant

andylemin:
@rdmark LGTM. Assume you will squash before merge?

andylemin (approved):

rdmark:
@andylemin I was going to keep the two commits separate because they're doing two distinct things: adding the global permissions to one workflow, while removing some job level permissions in the other

...but happy to squash to reduce commit clutter =)

PR #3291, opened 2026-09-06
distrib: remove default base image from perftest container

rdmark:
having an unpinned default base image may have unexpected side effects, so removing the default and making the BASE_IMAGE env variable mandatory

this expectation is already documented and captured in the shell script harness


rdmark:
@andylemin this is me working through the hardenings flagged by the Scorecard scanner

andylemin (approved):

PR #3292, opened 2026-09-06
spotlight: avoid lossy string length conversion

rdmark:
sl_pack_string() calculated the byte count used in the final octet by subtracting size_t len from an int expression. Although the resulting count is only one to eight, the mixed-width calculation triggers a narrowing-conversion warning and octets * 8 can overflow before the buffer-size check for very large strings.

Derive the value directly from len % 8 and cast only the proven bounded remainder.


rdmark:
@andylemin this is to squash what SonarQube is flagging as "new bugs" to get us back to a zero baseline for the release :)

andylemin (approved):

PR #3293, opened 2026-09-06
testsuite: document FPEnumerate reply buffer and remove stale FIXMEs

rdmark:
Remove stale call-site FIXMEs and replace them with the FPEnumerate reply-buffer contract. Clarify that replies are received in dsi->data, bounded by cmdlen, while datalen remains the request length.

### DIFF
--- a/test/testsuite/Error.c
+++ b/test/testsuite/Error.c
@@ -822,7 +822,6 @@ static void cname_test(char *name)
         test_failed();
     }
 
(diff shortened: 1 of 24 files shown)


rdmark:
@andylemin this should bring us back to a zero bug SonarQube baseline!

andylemin (approved):
lol, after spending time fixing up and refactoring Enumerate I can't believe I forgot all the old FIXME's 😂 thanks!

PR #3298, opened 2026-09-08
cnid: clean up only stale rows the sqlite backend itself wrote

andylemin:
The stale-volume cleanup in `cnid_sqlite_open()` reads `VolUUID` values back out of the `volumes` table and names them in `DROP TABLE IF EXISTS` — a table name cannot be a bound parameter, so those statements are built by string. The backend only ever writes names in one form: `uuid_strip_dashes()` yields exactly 32 hex digits.

This PR makes the cleanup accept only that form. A new `cnid_sqlite_is_table_name()` checks it, and a `volumes` row holding anything else was not written by this backend, so it is left in place and logged rather than acted on. The sibling per-row deletes already use bound parameters and are unchanged.

## Behaviour change

A `volumes` row whose `VolUUID` is not in table-name form is no longer swept up by the stale-volume cleanup; a warning names the volume path. Every name the backend itself writes passes the check, so normal cleanup is unaffected.

## Testing

- Testsuite container build (Alpine, `debugoptimized`, asserts armed): 4 test legs, 0 failures — including the CNID sqlite subtests, which drive `cnid_sqlite_open()` and its stale-volume cleanup with backend-written UUIDs, confirming the check is transparent for every name the backend creates.

### DIFF
--- a/libatalk/cnid/sqlite/cnid_sqlite.c
+++ b/libatalk/cnid/sqlite/cnid_sqlite.c
@@ -74,6 +74,26 @@
 static void cnid_sqlite_set_errno(int sqlite_return);
 
 /*!
+ * @brief Whether a string is one of this backend's per-volume table names
+ *
+ * uuid_strip_dashes() yields exactly 32 hex digits, so a value in any other
+ * form was not written by this backend. A table name cannot be a bound
+ * parameter, so statements that name one accept only this form.
+ */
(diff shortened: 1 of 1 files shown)


andylemin:
@rdmark trivial one for review

rdmark (approved):
nice improvement!

PR #3304, opened 2026-09-12
test: minor cleanup of afpd unit test

rdmark:
use constant declaration in one spot, and remove superfluous newline


rdmark:
@andylemin merging trivial change without review to unblock the pipeline!

PR #3306, opened 2026-09-12
Follow-ups to recent hardening commits: cnid, dbd, adouble, unicode

andylemin:
A review of the commits merged to main between 2026-09-08 and 2026-09-12 found four defects worth fixing before the next release. Each sits at a seam between two commits that were correct on their own; none is an individual mistake.

## Defects and fixes

| # | Module | Defect | Why | Fix |
|---|---|---|---|---|
| 1 | cnid | A volume UUID differing from the stored row only in letter case looked stale, and the cleanup dropped the live CNID table before recreating it empty | SQLite table names are case-insensitive while `volumes.VolUUID` compares bytewise; `afp_voluuid.conf` entries are upper-cased on read, a `volume uuid` in `afp.conf` is stored as typed | A `COLLATE NOCASE` lookup before the stale scan adopts the on-disk spelling for the session. The duplicate table-name predicate and the drop-loop branch it made unreachable are removed |
| 2 | dbd | `dbd -f` aborted the whole rebuild, after wiping the database, when any directory held a regular file or symlink named `.AppleDouble` | 59d1d8a8 opens the directory with `O_NOFOLLOW \| O_DIRECTORY`; the resulting `ENOTDIR` or `O_NOFOLLOW` errno reused the fatal return path | `check_addir()` reports a skip, only a negative result is fatal, ea volumes never stop on it, and the per-file EA check is gated on the same result |
| 3 | adouble | `ad_copy_header()` could fail with earlier entries already copied; `exchangefiles` runs it on live open forks | e788a360 validated each destination inside the copy loop | Validate every entry through the new `ad_entry_fits()`, then copy. `ad_entry()` is built on the same predicate. `copyfile()` maps the failure to `AFPERR_MISC` and keeps an earlier fork-copy errno |
| 4 | unicode | `precompose_w()` left one unwritten unit before its terminator after a composed surrogate pair | The surrogate branch advanced `out` twice; c0a11612 (2011). The NUL always stayed in bounds and callers use the returned length | Extra advance removed. Companion to f43ed6c9 on the `decompose_w()` side |

## Behaviour changes

- A volume whose configured UUID differs only in case from the stored one logs once at info level and keeps its table.
- `dbd -f` logs and skips the AppleDouble checks of a directory whose `.AppleDouble` is not a directory, and completes the volume.
- FPCopyFile of a source whose header entry does not fit the destination returns `AFPERR_MISC` instead of `AFPERR_PARAM`.

## Testing

Alpine container matching CI: zero compiler warnings, full `meson test` green, astyle 3.6.18 compliant. The four SonarCloud findings on the first revision are resolved.

Unit tests, each red with its fix reverted:

- `cnid_open: a case-flipped volume UUID reuses the live table`
- `ad_copy_header leaves the destination untouched when an entry is rejected`

Tool run:

- `dbd -f` on an `adouble = ea` volume with a regular file named `.AppleDouble` and a `v2` volume with a symlink of that name: both exit 0, every later entry is indexed, the stray and its target are untouched.

## Not in this PR

- `write_fork()` computes `offset + reqcount` before its guard, the pattern e16c2495 removed from `read_fork()`.
- A configured `volume uuid` is not checked for hex form before it names a sqlite table.
- The MySQL backend has the same bytewise stale scan. Canonicalising case at the config source would fix both backends but needs a migration path for existing MySQL tables.

### DIFF
--- a/bin/dbd/cmd_dbd_scanvol.c
+++ b/bin/dbd/cmd_dbd_scanvol.c
@@ -21,6 +21,7 @@
 #include <errno.h>
 #include <fcntl.h>
 #include <setjmp.h>
+#include <stdbool.h>
 #include <stdlib.h>
 #include <string.h>
 #include <sys/stat.h>
@@ -231,7 +232,7 @@ static void remove_eafiles(const char *name, struct ea *ea _U_)
     addir_fd = open(ADv2_DIRNAME,
                     O_RDONLY | O_DIRECTORY | O_NOFOLLOW | O_CLOEXEC);
 
-    if (addir_fd == -1) {
+    if (addir_fd < 0) {
         dbd_log(LOGSTD, "Couldn't open '%s/%s': %s",
                 cwdbuf, ADv2_DIRNAME, strerror(errno));
         return;
@@ -322,6 +323,16 @@ static int check_eafiles(const char *fname)
 }
 
 /*!
+ * @brief Whether a non-directory holds the AppleDouble dir name
+ *
+ * A regular file gives ENOTDIR, a symlink the platform's O_NOFOLLOW errno.
+ */
+static bool adouble_dir_unusable(int err)
+{
+    return err == ENOTDIR || err == OPEN_NOFOLLOW_ERRNO;
+}
+
+/*!
  * Check for .AppleDouble folder and .Parent, create if missing
  */
 static int check_addir(int volroot _U_)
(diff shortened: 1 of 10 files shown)


andylemin:
@rdmark FYI

rdmark:
@andylemin a nice lineup of improvements -- let me review later tonight!

moments ago I filed this PR which overhauls afppasswd semantics and does away with `-p` while introducing `-F afp.conf` (root only); I think this makes fix No.5 here obsolete

https://github.com/Netatalk/netatalk/pull/3307

andylemin:
> @andylemin a nice lineup of improvements -- let me review later tonight!
> 
> moments ago I filed this PR which overhauls afppasswd semantics and does away with `-p` while introducing `-F afp.conf` (root only); I think this makes fix No.5 here obsolete
> 
> #3307

Do you want me to drop 5? Happy to do so? I can't do much tomorrow, and its late here now

rdmark:
@andylemin one option is to accept your number 5, and me rebasing / consolidating my PR on top of it after merging

there may be hardening in your version of the code that we should carry over

andylemin:
> @andylemin one option is to accept your number 5, and me rebasing / consolidating my PR on top of it after merging
> 
> there may be hardening in your version of the code that we should carry over

Dropped 5, was easy to remove.. Everything defensive in my fix 5 was present in 3307, in two places 3307 is stricter

andylemin:
@rdmark Ready for review

rdmark (approved):

PR #3307, opened 2026-09-12
afppasswd: harden credential-store management

rdmark:
The goal of this changeset is to allow afppasswd to run unprivileged, that is without the set-uid flag.

Replace SRP's shared flat verifier file with a root-owned directory containing one mode-0600 verifier per numeric uid. The user owns only that user's verifier, allowing local SRP password changes without an elevated afppasswd executable. Update the SRP UAM and configuration plumbing for the new verifier path, retaining 'srp passwd file' as a deprecated alias.

Add a root-only afppasswd -m workflow to migrate a legacy SRP file without resetting passwords. It validates records and local account mappings, stages and synchronizes a complete directory, retains the original under a collision-safe .legacy name, and rolls back publication failures where possible.

To be able to edit the afppasswd password file, we now require root to manage Randnum credentials.

Minor refactoring and hardening of the SRP UAM, moving shared UAM constants into a new global header. Define a minimum libgcrypt library version globally.


rdmark:
@andylemin in the process of tightening and testing this, but I would love to get your early feedback on the high level approach here

rdmark:
@andylemin the migration from flat file to the new per-uid credential store has been tested on a real netatalk installation with accumulated user credentials -- the migration worked as advertised and allowed migration without interruption

do you see any weakness in this new structure where /etc/netatalk/afppasswd.srp is now a world-accessible directory? I have been toying with the idea of having a small auth broker daemon instead (`netatalk-auth`) with a socket interface that handles the writes to afppasswd files, but have been going back and forth about whether it's overkill or not for this usecase.

bin/afppasswd/afppasswd.c
```diff
@@ -965,7 +1109,8 @@ static int update_srp_passwd(const char *path, const char *name, int flags,
         /* Write: username:hex_salt:hex_verifier\n */
         written = fprintf(fp, "%s:%.*s\n", name, (int)sizeof(hex_buf), hex_buf);
 
-        if (written != expected_len || fflush(fp) != 0) {
+        if (written != expected_len || fflush(fp) != 0 ||
+                ftruncate(fd, expected_len) < 0 || fsync(fd) < 0) {
```
andylemin, line 926: I think there's a subtle interaction between the search loop above and this `ftruncate`. The `fgetc(fp) != EOF` check only guarantees nothing *follows* the matching record, so if the user's file happens to hold a stale first line (hand edit, uid-reuse leftover; the user owns the file so this is reachable) the match lands at `pos > 0`. We then write the new record at `pos` but truncate to `expected_len` counted from offset 0, which discards the record we just wrote while still printing "updated SRP verifier".

Since the file is now defined as exactly one record, would it make sense to drop the multi-record search entirely? Something like one `fgets`, require `fgetc(fp) == EOF`, compare the username, then `rewind(fp)` before writing. That removes `pos`, the `found_entry:` label and the goto, and makes the truncate correct by construction. If you'd rather keep the loop, rejecting `pos != 0` on a match would close the gap too.

bin/afppasswd/afppasswd.c
```diff
@@ -1042,17 +1202,27 @@ static int create_srp_file(const char *path, uid_t minuid)
         }
 
         buf[n++] = '\n';
+        fd = open_srp_verifier(dirfd, path, pwd->pw_uid, 1);
```
andylemin, line 1110: A question about the trust model here. Because every placeholder is chowned to the user, the `*` "disabled" state, the old-password proof and the cracklib check are now only enforced inside afppasswd; afpd itself only needs owner == uid, 0600, nlink 1 and a well-formed record. So any local account above `minuid` can write its own `name:salt:verifier` line into its file and log in over SRP, even though afppasswd would tell that user "Your password is disabled".

That may be an acceptable trade for dropping setuid, but I don't think the man page or Upgrading.md currently say the placeholder is advisory. One option that keeps an admin-enforced gate: have `-c`/`-m` create placeholders root-owned, and only `fchown` to the user when root sets a real verifier with `-a`. Ownership then becomes the enrolment bit, and "disable" is `chown root` or deleting the file. If you'd prefer to keep the current behaviour, could we at least document that self-service users can set any password and that removing the file is the only real disable?

bin/afppasswd/afppasswd.c
```diff
@@ -114,174 +118,208 @@ static const unsigned char srp_N_bytes[SRP_NBYTES] = {
...
-                strerror(errno));
-        return -1;
-    }
-
-    if (!S_ISREG(st.st_mode)) {
-        fprintf(stderr, "afppasswd: %s is not a regular file.\n", path);
-        return -1;
-    }
+    if (fd < 0) {
+        if (errno == ENOENT && allow_missing) {
```
andylemin: Before this change a regular user changing their own password never touched afp.conf. Now the open has to succeed and pass the checks below, and only ENOENT falls back to the compiled default. A few fairly common layouts break every non-root `afppasswd` run as a result:

- afp.conf 0600 root:root (reasonable since it may contain `ldap auth pw`) → EACCES
- afp.conf is a symlink (NixOS/Guix, `/usr/local/etc` pointing at a Homebrew prefix) → ELOOP
- 0664 root:admins → the `S_IWGRP` check
- an unprivileged `brew install` leaves afp.conf owned by the brew user → the `st_uid` check (this also stops `sudo afppasswd -c` there)

Non-root can't pass `-F`, so there's no workaround. Would you be open to falling back to the default path on EACCES/ELOOP as well when `allow_missing` is set, or relaxing the metadata check to `S_ISREG && !(S_IWGRP|S_IWOTH)`, since a group/other-unwritable file is what actually protects the value?

bin/afppasswd/afppasswd.c
```diff
@@ -114,174 +118,208 @@ static const unsigned char srp_N_bytes[SRP_NBYTES] = {
...
+            *--end = '\0';
         }
 
-        return -1;
-    }
+        if (*value == '\0') {
+            continue;
+        }
+
+        if (strcasecmp(key, option) == 0) {
```
andylemin: Is there a reason to hand-parse afp.conf here rather than use the same iniparser afpd uses? afppasswd is no longer setuid and already links libatalk; `bin/nad`, `bin/dbd` and `cnid_metad` all do `iniparser_load()` + `INIPARSER_GETSTR(cfg, INISEC_GLOBAL, ...)`. I tried a few inputs the library handles that this loop doesn't, and each makes afpd and afppasswd resolve different directories from the same line:

- `srp verifier path = "/srv/afp/srp"` (iniparser strips the quotes)
- `srp verifier path = /srv/afp/srp ; note` (iniparser truncates at `;`/`#`)
- `[ Global ]`, which afp.conf.5 documents as legal (here `section_len` is 8, so `in_global` never becomes true and we silently fall back to the default)
- backslash line continuation, also documented in the man page
- two `srp passwd file` lines: iniparser is last-wins, here the alias is first-wins

If the concern is keeping the `O_NOFOLLOW` + `fstat` check, iniparser ≥ 4.2 has `iniparser_load_file(FILE *, const char *)`, so we could keep the fd-based open and hand it the stream. It would also let the "new key beats alias" rule live in one place next to the netatalk_conf.c block instead of being duplicated here.

andylemin (commented):
Dropping the setuid bit and moving to per-uid files is a good direction, and the migration path is clearly careful work. I've left inline questions where I think the trust model, the config parsing and a couple of edge cases need a look, plus some smaller simplification asks.

One scope question that doesn't have a natural inline spot: the Randnum change-password removal. The old `randnum_changepw` ran inside afpd as root and didn't depend on afppasswd being setuid, and with `afppasswd -r` now root-only it leaves Randnum users with no self-service path at all. Meanwhile `AFPSRVRINFO_PASSWD` is still advertised whenever `set password = yes`, so clients offer "Change Password" and get `AFPERR_BADUAM`. Is the removal tied to this change, or would it be better split out? If it stays, clearing `PASSWD_SET` in `configinit` when no loaded UAM registered `UAM_SERVER_CHANGEPW` would make the advertisement truthful, and would fix the same pre-existing mismatch for SRP?

bin/afppasswd/afppasswd.c
```diff
@@ -1461,53 +1648,86 @@ int main(int argc, char **argv)
...
+                                   "srp verifier path",
+                                   (flags & OPT_RANDNUM) ? NULL :
+                                   "srp passwd file",
+                                   (flags & OPT_RANDNUM) ? _PATH_AFPDPWFILE :
+                                   _PATH_AFPSRPVERIFIERPATH,
+                                   buf, sizeof(buf), !config_seen) < 0) {
+        return -1;
     }
 
+    path = buf;
```
andylemin: Small one: `buf` is the file-scope scratch buffer that `create_srp_directory`, `update_passwd` and `create_file` later reuse for record data, so `*path` gets clobbered after the opens. No syscall is misrouted today because all the path-taking opens run first, but the later error messages ("problem writing to %s", "can't lock or seek %s", the validator message) end up printing record contents instead of the path. In the RandNum `-a` case that's the `name:<DES ciphertext>` line on stderr. Could the resolved path get its own `char path_buf[MAXPATHLEN + 1]`?

bin/afppasswd/afppasswd.c
```diff
@@ -1000,15 +1144,30 @@ static int update_srp_passwd(const char *path, const char *name, int flags,
...
 
-static int create_srp_file(const char *path, uid_t minuid)
+static int create_srp_directory(const char *path, uid_t minuid)
 {
     struct passwd *pwd;
-    int fd, err = 0;
+    int dirfd, err = 0;
 
-    if ((fd = open_credential_for_replacement(path)) < 0) {
+    if (mkdir(path, 0755) < 0 && errno != EEXIST) {
```
andylemin: Two questions on the `-c -f` path. The man page still says `-f` "allows overwriting an existing password/verifier file", but with a legacy flat file at this path `mkdir` returns EEXIST (ignored), and `open_srp_verifier_directory` then fails with the "run afppasswd -m" hint, so `-f` is effectively inert there. Is that intended, and should the man page text change to match?

Also, when the directory already exists and one uid file fails validation (wrong owner after a `cp -r` restore, say), the loop below `break`s after earlier users have already been reset to placeholders, leaving a half-initialised store with no rollback. Would it be reasonable for `-f` to `fchown`/`fchmod` the existing file back into shape rather than abort?

bin/afppasswd/afppasswd.c
```diff
@@ -734,15 +772,109 @@ static int valid_srp_record(const char *fields)
...
+    return len < 0 || (size_t)len >= size ? -1 : 0;
+}
+
+static int validate_srp_verifier_file(int fd, uid_t uid, const char *path)
 {
-    const char *passwd = NULL;
+    struct stat st;
+
+    if (fstat(fd, &st) < 0 || !S_ISREG(st.st_mode) || st.st_uid != uid ||
+            (st.st_mode & 07777) != 0600 || st.st_nlink != 1) {
```
andylemin: This requires exactly 0600, whereas afpd's `srp_lookup_verifier` and the migrator accept any mode without group/other bits. A 0400 file therefore logs in fine but afppasswd refuses to update it. Could the three sites share one predicate?

Related: root `-a` reaches this via the EEXIST reopen, and nothing ever `fchown`s an existing file, so a mis-owned verifier can't be repaired by root at all. Would it make sense for root (or `-f`) to `fchown` + `fchmod` here instead of failing?

etc/uams/uams_srp.c
```diff
@@ -253,101 +254,158 @@ static uint16_t read_uint16_be(unsigned char **p)
...
-        p++; /* skip colon, now pointing at hex salt */
+    fd = openat(dirfd, uid_name, O_RDONLY | O_CLOEXEC | O_NOFOLLOW);
 
-        /* Check for placeholder (starts with '*') */
-        if (*p == '*') {
-            fclose(fp);
-            return -1;
-        }
+    if (fd < 0 || fstat(fd, &st) < 0 || !S_ISREG(st.st_mode) ||
+            st.st_uid != uid || (st.st_mode & (S_IRWXG | S_IRWXO)) ||
```
andylemin: When the owner/mode/nlink check fails we `goto done` silently, and the only thing the caller logs is "no verifier for user X" at info level. That's indistinguishable from a user who was never enrolled, which will make ownership drift (e.g. `/etc/netatalk` restored with `cp -r` or `scp -r`) quite hard to diagnose. Could we LOG the mismatch at error level, ideally saying which of owner/mode/nlink was off? The docs don't currently mention `chown` as a repair step either.

etc/uams/uams_srp.c
```diff
@@ -408,7 +468,7 @@ static int srp_setup(void *obj, char *ibuf _U_, size_t ibuflen _U_,
     /* Look up verifier */
     session_v = NULL;
 
-    if (srp_lookup_verifier(srpfile, session_username,
+    if (srp_lookup_verifier(verifier_path, session_username, srppwd->pw_uid,
```
andylemin: This is pre-existing structure rather than something this PR introduced, but since `srp_setup` now dereferences `srppwd` I wanted to ask about it. In both `srp_login` and `srp_login_ext`, `srppwd` is assigned from `uam_getname()` before `uam_checkuser()`, and the checkuser failure path returns without calling `srp_session_free()` or touching `session_username`. So after a successful first round for user A, a second FPLogin naming an existing account B that fails checkuser (a service account with a non-login shell) leaves `session_v` as A's verifier but `srppwd` pointing at B. `srp_logincont` only checks `session_v != NULL`, verifies A's M1, and then returns `*uam_pwd = srppwd`, i.e. B.

I may be missing a guard in afpd's login state machine that prevents the second FPLogin, so please correct me if so. If not, resetting session state at the top of both login handlers (or at least `srp_session_free()` on the checkuser failure path, plus clearing `srppwd` inside `srp_session_free`) would close it.

bin/afppasswd/afppasswd_migrate.c
```diff
@@ -0,0 +1,887 @@
...
+           strchr(name, '\r') == NULL;
+}
+
+static int valid_srp_fields(const char *fields)
+{
+    int salt_disabled = 1, verifier_disabled = 1;
+
+    if (strlen(fields) != SRP_FIELDS_LEN ||
+            fields[SRP_HEX_SALT_LEN] != ':' ||
+            fields[SRP_FIELDS_LEN - 1] != '\n') {
```
andylemin: The old afpd parser never looked at the byte after the verifier hex, so a flat file whose last record has no trailing newline (`echo -n`, `vi` with nofixeol, a truncated restore) authenticated fine before. With this check `-m` rejects it as "malformed legacy SRP record N", and since `-m` is the only upgrade path, afpd then refuses all SRP logins until the admin works out that a missing newline is the cause. Could the final `getline()` result have a newline appended before validation (the new per-uid format needs it anyway), or at least could the error say "missing trailing newline"?

NEWS.md
```diff
@@ -128,6 +128,21 @@ great protocol fade away. Get Netatalk 4.6.0 and Netatalk Client 1.0 at
 Changes in 4.6.0
 ----------------
 
+* BREAKING: afppasswd/SRP: replace the shared *afppasswd.srp* file with a
+  root-owned verifier directory at the same path and one mode-0600,
+  user-owned file per numeric uid. This removes the need to install
+  **afppasswd** setuid-root. After stopping **afpd**, migrate an existing flat
+  file without resetting passwords by running `afppasswd -m`; use `-p PATH`
```
andylemin: Tiny doc nit: `-p` is removed in this PR (getopt is now `cfmnra:F:u:w:` and the test suite asserts it's rejected), so this should probably read "use `-F afp.conf` for a non-default location", matching Upgrading.md and the man page.

libatalk/util/netatalk_conf.c
```diff
@@ -3312,9 +3312,25 @@ int afp_config_parse(AFPObj *AFPObj, char *processname)
...
-                                               NULL, _PATH_AFPDSRPPWFILE);
+    {
+        const char *old_srp_option = INIPARSER_GETSTR(config, INISEC_GLOBAL,
+                                                      "srp passwd file", NULL);
+        const char *srp_option = INIPARSER_GETSTR(config, INISEC_GLOBAL,
+                                                  "srp verifier path", NULL);
+
+        if (old_srp_option && *old_srp_option) {
+            LOG(log_warning, logtype_afpd,
+                "Using deprecated 'srp passwd file' option; its value is now a verifier directory. If it names a legacy flat file, stop afpd and run 'afppasswd -m', then use 'srp verifier path'");
```
andylemin: This does the same job as the `strict locking` / `afp read locks` alias block a little further up. Would it be worth pulling that pattern into a small `getoption_strdup_alias()` helper and using it for both, so the project has one alias idiom?

bin/afppasswd/afppasswd_migrate.h
```diff
@@ -0,0 +1,21 @@
+/*
+ * Internal support for migrating the legacy flat SRP verifier file.
+ */
+
+#ifndef NETATALK_AFPPASSWD_MIGRATE_H
+#define NETATALK_AFPPASSWD_MIGRATE_H
+
+#include <sys/types.h>
+
+struct afppasswd_migrate_hooks {
```
andylemin: Question on the hooks struct: the only production caller passes NULL, so this is effectively a test seam in the public interface, and the per-member NULL fill-in in `afppasswd_migrate_srp` isn't exercised by any test (they all pass all three). Since `afppasswd_migrate.c` is compiled straight into the test executable, would `-Wl,--wrap=write,--wrap=fsync,--wrap=getpwnam` on the test target give the same fault injection with no production indirection? If the hooks stay, treating the struct as all-or-nothing would at least drop the untested branches.

bin/afppasswd/afppasswd_migrate.c
```diff
@@ -0,0 +1,887 @@
...
+#include <sys/param.h>
+#include <sys/stat.h>
+#include <sys/types.h>
+#include <unistd.h>
+
+#include <atalk/compat.h>
+
+#include "afppasswd_migrate.h"
+
+#define SRP_HEX_SALT_LEN 32
```
andylemin: These constants, `valid_username` and `valid_srp_fields` are byte-for-byte copies of the definitions in afppasswd.c (which spell them `SRP_SALT_LEN*2` / `SRP_NBYTES*2`), and uams_srp.c has a third set. Both afppasswd files link into one binary, so could they share a header? A small `static inline` header under `include/atalk/` would let afpd use the same definitions too, so a future group-size change can't leave the writer and reader disagreeing about record length.

bin/afppasswd/afppasswd_migrate.c
```diff
@@ -0,0 +1,887 @@
...
+
+    close(directory_fd);
+    unlinkat(parent_fd, temporary_name, AT_REMOVEDIR);
+}
+
+static int create_staging_directory(int parent_fd, const char *basename,
+                                    uid_t administrator_uid,
+                                    char temporary_name[MAXPATHLEN + 1])
+{
+    for (unsigned int attempt = 0; attempt < MIGRATE_ATTEMPTS; attempt++) {
```
andylemin: Am I reading this right that the retry loop can't actually iterate? `reject_stale_staging` runs first and refuses if any `.<name>.migrate.*` sibling exists, and the name embeds `getpid()`, so EEXIST on attempt 0 would need another live process with the same pid. If so, a single `mkdirat` with `.migrate.<pid>` would be simpler, leaving `MIGRATE_ATTEMPTS` just for the `.legacy.N` backup names where it's genuinely needed.

bin/afppasswd/afppasswd_migrate.c
```diff
@@ -0,0 +1,887 @@
...
+        fprintf(stderr,
+                "afppasswd: automatic rollback failed; the original remains at %s.\n",
+                backup_name);
+        return -1;
+    }
+
+    if (unlinkat(parent_fd, backup_name, 0) < 0 ||
+            hooks->sync_fd(parent_fd, hooks->context) < 0) {
+        fprintf(stderr,
+                "afppasswd: rollback restored the source but could not remove %s.\n",
```
andylemin, line 613: When the source has been re-linked successfully but only the backup unlink fails, this returns -1 like the real failure above, so the caller leaves `source_unlinked` set and prints "recovery required; the original remains at ..." even though the original is already back in place. Would a distinct return value (or clearing a flag via an out-parameter) for "restored, backup left behind" let the caller say just that?

bin/afppasswd/afppasswd.c
```diff
@@ -1461,53 +1648,86 @@ int main(int argc, char **argv)
...
+    path = buf;
+
     /* Validate password length for RandNum mode */
     if ((flags & OPT_RANDNUM) && strnlen(pass, PASSWDLEN + 1) > PASSWDLEN) {
         fprintf(stderr, "afppasswd: max RandNum password length is %d.\n", PASSWDLEN);
         return -1;
     }
 
+    if (flags & OPT_MIGRATE) {
+        if (!(flags & OPT_ISROOT)) {
```
andylemin: I think this branch is unreachable: the gate at the top of this block already rejects `OPT_MIGRATE` for non-root. The option-set check just below also mirrors the one above, so a new option would need updating in three places. Could the `-m` exclusivity check be folded into the earlier gate so this reduces to just the `afppasswd_migrate_srp()` call?

bin/afppasswd/afppasswd.c
andylemin: We may need to improve documentation where this is intended

NEWS.md
rdmark: simplified this passage

rdmark:
> One scope question that doesn't have a natural inline spot: the Randnum change-password removal. The old randnum_changepw ran inside afpd as root and didn't depend on afppasswd being setuid, and with afppasswd -r now root-only it leaves Randnum users with no self-service path at all. Meanwhile AFPSRVRINFO_PASSWD is still advertised whenever set password = yes, so clients offer "Change Password" and get AFPERR_BADUAM. Is the removal tied to this change, or would it be better split out? If it stays, clearing PASSWD_SET in configinit when no loaded UAM registered UAM_SERVER_CHANGEPW would make the advertisement truthful, and would fix the same pre-existing mismatch for SRP?

fair point, I got too eager with the hardening here, let me revert that part and we can look closer at remote password changing in a future PR

I am not happy with how remote password changing works right now from a user perspective: whether it is allowed or not depends on the active UAM, and the user has no transparency into why the operation failed. ideally all UAMs should support password changing, or none should, in my opinion.

NEWS.md
rdmark: simplified this passage

bin/afppasswd/afppasswd.c
rdmark: agreed -- the EOF check only rules out content after the match, so a preceding stale line lets us write at a nonzero offset and then truncate the new record while reporting success... refactoring to single-record reading is a good idea, no need to keep the flat file loop around

bin/afppasswd/afppasswd.c
rdmark: Implemented ownership-based enrollment:

- -c creates root-owned placeholders and restores root ownership when reinitializing.
- -m keeps disabled records root-owned while preserving active credentials.
- root’s -a transfers ownership only after writing and synchronizing a real verifier.

bin/afppasswd/afppasswd.c
rdmark: I agree on the Nix and Homebrew scenarios -- not allowing symlinks and user owned afp.conf is too draconian on second thought; how this for a middle road between hardening and portability, in particular the fallback to the compiled default should solve 99% of scenarios I think

- standard config lookup falls back to the compiled-in credential path on ENOENT, EACCES, or ELOOP; the latter two emit a warning.
- symlinks, hard links, user ownership, and group/other write permissions are accepted.
- explicit -F failures remain fatal.
- FIFOs are rejected without hanging.

bin/afppasswd/afppasswd.c
rdmark: indeed, I tried to keep it simple but see now why using iniparser is the right choice for consistency and safety

let me try using the nice iniparser 4.2 API with a fallback for platforms that are stuck on 3.x (NetBSD etc.)

bin/afppasswd/afppasswd.c
rdmark: how about something as simple as this?

```c
--- a/bin/afppasswd/afppasswd.c
+++ b/bin/afppasswd/afppasswd.c
@@ -1604,9 +1604,10 @@ static void print_usage(void)
 int main(int argc, char **argv)
 {
     struct stat st;
+    char path_buf[MAXPATHLEN + 1];
     int flags;
     uid_t uid_min = UID_START, uid;
-    const char *path;
+    const char *path = path_buf;
     const char *config_path = _PATH_CONFDIR "afp.conf";
     int adduser_seen = 0, config_seen = 0, password_seen = 0, uid_seen = 0;
     const char *pass = "";
@@ -1714,12 +1715,10 @@ int main(int argc, char **argv)
     /* Both credential paths are read from the selected afp.conf.
      * Non-root calls can reach only the installed standard file. */
     if (configured_credential_path(config_path, flags & OPT_RANDNUM,
-                                   buf, sizeof(buf), !config_seen) < 0) {
+                                   path_buf, sizeof(path_buf), !config_seen) < 0) {
         return -1;
     }
 
-    path = buf;
-
     /* Validate password length for RandNum mode */
     if ((flags & OPT_RANDNUM) && strnlen(pass, PASSWDLEN + 1) > PASSWDLEN) {
         fprintf(stderr, "afppasswd: max RandNum password length is %d.\n", PASSWDLEN);
```

bin/afppasswd/afppasswd.c
rdmark: this makes me think of the semantics of the _force_ '-f' option and whether it still makes sense for the SRP verifier store? I think it's fine to keep around as a legacy option for Randnum mode, but doing batch operations on the uid files in afppasswd.srp seems unnecessary and fragile -- I cannot think of a legitimate use case.

at the same time, an option for explicitly disabling a single user would be good to add while we are at it, e.g. `-d <user>`

bin/afppasswd/afppasswd.c
rdmark: this turned out to be quite a lot of logic all in all for little benefit, I'm leaning towards dropping this idea of parsing afp.conf in afppasswd and bring back the `-p` option (in a safe manner) for simplicity

can you think of any other afp.conf option we would want afppasswd to parse and make use of?

bin/afppasswd/afppasswd.c
andylemin: Great, thanks. Once it's a single `fgets` + `rewind`, one small follow-on to consider: the write lock is `F_SETLKW`, and the user now owns the file, so a user holding a lock on their own verifier would block root's `afppasswd -a` for them indefinitely. `F_SETLK` with a "busy, try again" message (as the migration path already does) would avoid that.

bin/afppasswd/afppasswd.c
andylemin: That's exactly the gate I was hoping for, nice. Two small things worth keeping an eye on as you wire it up:

- with root-owned placeholders, a not-yet-enrolled user now hits EACCES on the open rather than the `*` check, so it's worth making sure they still get the friendly "see your administrator" message rather than a raw permission error.
- a one-line note in the man page and Upgrading.md that "ownership is enrolment" (root-owned = disabled, user-owned = active) will save admins a lot of guessing, and pairs nicely with the `-d <user>` idea from the other thread.

bin/afppasswd/afppasswd.c
andylemin: That middle road sounds right to me: fall back with a warning for the standard lookup, keep explicit `-F` failures fatal. One thought that might simplify the reasoning: now that the binary isn't setuid, none of these afp.conf checks protect anything for a non-root caller, since the worst they can do is point at a directory whose files they don't own and get EACCES from the kernel. So warn-and-continue is safe as well as friendlier. That said, if you go with dropping afp.conf parsing altogether (other thread), most of this becomes moot anyway.

bin/afppasswd/afppasswd.c
andylemin: I think dropping the parser and bringing `-p` back is the pragmatic call, and you've clearly given the iniparser route a fair go. To your question: I can't think of anything else afppasswd would genuinely want from afp.conf. The only other candidate is `passwd minlen`, and afppasswd already has its own policy via cracklib and `-n`, so it's not worth a parser on its own. `uam list` doesn't help either since the tool can't tell which store the admin is about to enable.

Two suggestions for the "safe manner" part:

- since afppasswd is no longer setuid, there's no privilege for `-p` to abuse, so it could be allowed for non-root too. A user on a non-default install can't self-serve otherwise, and the worst they can do is get EACCES on someone else's file. The kernel enforces the real boundary.
- the compiled-in default already covers packaged installs, so the man page just needs to say clearly that `-p` must match `srp verifier path` (or `passwd file` for `-r`) in afp.conf, and the usage text should say "directory" for the SRP case.

bin/afppasswd/afppasswd.c
andylemin: Yes, that's all it needs. `const char *path = path_buf` works fine since `configured_credential_path` fills it before any use. And if `-p` comes back, `path` just becomes `optarg` or `path_buf` depending on whether it was given.

bin/afppasswd/afppasswd.c
andylemin: Agreed on both counts. Keeping `-f` as a Randnum-only legacy flag and dropping the batch reset of uid files removes the half-initialised-store case entirely, and with ownership-based enrolment `-c` on an existing directory can simply refuse or be a no-op.

`-d <user>` is a good addition and gives admins a documented disable for the first time: chown to root plus a placeholder write, with `-a` as the re-enable. Two small suggestions: make it idempotent (create the root-owned placeholder if the file is missing, so it also works for uid-reuse cleanup), and take the same `F_SETLK` approach so a user can't wedge it.

andylemin:
Thanks for the quick and thoughtful responses across the threads, Daniel. Ownership-based enrolment, the single-record rewrite, the config fallback, and the `-d <user>` idea all move this in a good direction, and I'm glad the migration has already been exercised on a real install with accumulated credentials.

On your earlier question about the world-accessible directory: I don't see a structural weakness. The directory is root-owned and 0755, so users can't create, rename or link entries in it; each verifier is 0600 with `nlink == 1` and opened with `O_NOFOLLOW`, so the classic hard-link and symlink tricks are closed off. What remains is small:

- **enumeration**: anyone can list which uids are enrolled. Only the names leak, never contents, and `/etc/passwd` already exposes the uids. Hiding it would mean 0711, but both afpd and afppasswd open the directory `O_RDONLY | O_DIRECTORY`, which needs read permission, so you'd be into `O_SEARCH`/`O_PATH` portability. Not worth it in my view.
- **self-inflicted lockout**: a user can `chmod` their own file into a state afpd rejects. With the error-level log we discussed on the afpd side, that becomes easy to diagnose.
- **uid reuse**: a departed user's file stays behind for the next holder of that uid. Root `-a` (or `-d`) handles it, and it's worth a sentence in the man page.

On the auth-broker daemon: I'd hold off. It would recreate the trust boundary this PR just removed, in a new root process parsing untrusted socket input, and the per-uid layout already gets the kernel to enforce the only rule that matters. Where a broker would earn its keep is when remote password change is redesigned so that every UAM behaves the same, and at that point afpd is already the root process that can write verifiers, so the need may never materialise.

And thanks for reverting the Randnum change-password removal; agreed that consistent behaviour across UAMs is the right follow-up.

Happy to take another pass once the changes are pushed.

bin/afppasswd/afppasswd.c
rdmark: thanks for supporting my conclusion ;)

the -p option has now made a comeback, eliminating over 500 lines of code

bin/afppasswd/afppasswd.c
rdmark: the changes made!

bin/afppasswd/afppasswd.c
rdmark: implemented a shared permission predicate across afppasswd, afpd, and migration: all reject group/other access and accept 0400 metadata

good call-out about what to do with mis-owned verifiers: I’ll let root’s -a and -d accept any permission mode and normalize it to 0600, while preserving the ownership, file-type, and hard-link checks

etc/uams/uams_srp.c
rdmark: good idea, fixing

etc/uams/uams_srp.c
rdmark: from what I can, tell afpd does block the reported sequence, although at the connection termination after a failed login, rather than rejection of a second FPLogin:

1. B fails uam_checkuser(), so the SRP handler returns AFPERR_NOTAUTH
2. both afp_login() and afp_login_ext() pass that failure to send_reply()
3. send_reply() only preserves the connection for AFP_OK and AFPERR_AUTHCONT. Otherwise, it sends the error and calls obj->exit(0)
4. for DSI, that invokes afp_dsi_die(), which closes the session and exits in this preauthentication flow

but regardless, I agree we should reset state at entry to both SRP login handlers, before parsing or account lookup, and clear srppwd inside srp_session_free()

libatalk/util/netatalk_conf.c
rdmark: having a reusable helper is definitely preferable, borderline out of scope for this PR, but I'll make the change!

bin/afppasswd/afppasswd_migrate.c
rdmark: agreed, let's make this more sturdy

bin/afppasswd/afppasswd_migrate.h
rdmark: how about we use macro substitution in the test code instead? I don't think --wrap is fully portable, couldn't get it to work with the macOS linker at least

bin/afppasswd/afppasswd_migrate.c
rdmark: a shared SRP header has been created

bin/afppasswd/afppasswd_migrate.c
rdmark: refactored away the loop:

- one mkdirat() using .<name>.migrate.<pid>
- MIGRATE_ATTEMPTS now applies only to backup names
- failed staging creation skips caller cleanup, preserving any colliding directory

bin/afppasswd/afppasswd_migrate.c
rdmark: how about we use three named return states:

- restore failed: keep source_unlinked set and retain backup_name
- source restored, cleanup incomplete: clear source_unlinked, retain backup_name, and report the cleanup problem
- restore complete: clear source_unlinked and backup_name

bin/afppasswd/afppasswd.c
rdmark: I see what you're saying, with one distinction:

- only the nested non-root check is unreachable; the earlier gate already returns for every non-root invocation containing OPT_MIGRATE while the outer migration branch remains reachable for root
- the exclusivity check is still necessary: the earlier gate does not constrain root, so combinations such as -m -c would otherwise pass

I’d move migration validation alongside the earlier checks, keeping it separate from the non-root condition so it also applies to root

rdmark:
@andylemin all the code changes are in now, but now experiencing a deterministic error triggered by Valgrind on Alpine only; the first hypothesis was a missing libgcrypt initialization in the unit tests, but that didn't help

it's getting late over here so I'm going to sleep on it and try again tomorrow morning!

andylemin:
Reproduced the Alpine failure locally in an `alpine:3.23.2` container with the CI package set (libgcrypt 1.11.2), and I think I have it.

**What Valgrind is objecting to.** All 69 tests pass; the exit status comes purely from Memcheck. With `--track-origins=yes` every report has the same origin: `Uninitialised value was created by a stack allocation at ... (in /usr/lib/libgcrypt.so.20.6.0)`. So the taint is born inside libgcrypt's own random-pool code, reaches us as the salt from `gcry_randomize()`, and trips the first conditional in `gcry_mpi_scan()`/`gcry_mpi_powm()` inside `srp_compute_verifier`. It's a Memcheck false positive on musl, not an afppasswd bug. You were right that initialization isn't the cause: `initialize_libgcrypt()` is called in both `main` and the test, and it made no difference either way here.

**What I tried:**

| Variant | Memcheck result |
|---|---|
| baseline, CI flags | 186 errors, exit 1 |
| `/etc/gcrypt/random.conf` = `only-urandom` | 202 errors |
| `/etc/gcrypt/random.conf` = `disable-jent` | 186 errors |
| `gcry_control(GCRYCTL_SET_PREFERRED_RNG_TYPE, GCRY_RNG_TYPE_SYSTEM)` before `gcry_check_version` | 0 errors, exit 0, 69/69 ok |

**Suggested fix.** One line at the top of `initialize_libgcrypt()`; I've left it as an inline suggestion on line 93. `GCRY_RNG_TYPE_SYSTEM` makes libgcrypt hand out bytes straight from the kernel CSPRNG (getrandom) instead of routing them through its own pool, which is exactly what we want for a 16-byte salt and the Randnum key file anyway. It only affects afppasswd and its test, never the UAMs inside afpd, and it has to be issued before `gcry_check_version`, which is why it goes first in the helper.

If you'd rather not touch RNG selection at all, the alternative is a Valgrind suppression for `obj:*/libgcrypt.so*` wired into the three `meson test --wrapper` legs, but that's more moving parts for the same result.

On the rest of the update: this is in really good shape. Ownership-based enrolment, `-d`, the shared `srp.h`, the tri-state rollback, the single-record rewrite, newline normalisation in the migrator, `-p` back with 500 fewer lines, and docs that now actually explain the model. I've left three small nits in a second pass and all of them can be follow-ups if you'd prefer to land this as soon as Alpine is green. The two new Sonar flags (the 0755 directory and the test's chmod modes) are both intentional and can be marked accepted.

bin/afppasswd/afppasswd.c
andylemin: 500 lines gone and the man page reads much more simply now. Good call.

bin/afppasswd/afppasswd.c
andylemin: That's the right shape: one predicate everyone agrees on, and root's `-a`/`-d` normalising to 0600 after the ownership and link-count checks. The permission matrix in the new tests (0400 through 07777) gives me confidence it does what it says.

etc/uams/uams_srp.c
andylemin: Thanks for tracing that through `send_reply()`; I'd missed that the connection is torn down on any non-continue result, so you're right that the sequence is blocked today. Resetting at entry and clearing `srppwd` in `srp_session_free()` is the right belt-and-braces regardless, and it's a cheap one.

bin/afppasswd/afppasswd.c
andylemin: Good distinction, you're right that the exclusivity check has to stay for root. The consolidated gate block reads well now.

bin/afppasswd/afppasswd_migrate.h
andylemin: Macro substitution is fine by me, and you're right that `--wrap` is a GNU ld feature, so it was never going to fly on macOS. Nice side effect that the migrator's public interface is now just the one function.

bin/afppasswd/afppasswd_migrate.c
andylemin: The three named states are exactly what I had in mind, and the four rollback-cleanup tests asserting on the diagnostics text nail it down.

bin/afppasswd/afppasswd.c
```diff
@@ -81,38 +80,24 @@ static int unhex(unsigned char x)
...
-    0x83, 0x7C, 0x26, 0x4A, 0xE3, 0xA9, 0xBE, 0xB8,
-    0x7F, 0x8A, 0x2F, 0xE9, 0xB8, 0xB5, 0x29, 0x2E,
-    0x5A, 0x02, 0x1F, 0xFF, 0x5E, 0x91, 0x47, 0x9E,
-    0x8C, 0xE7, 0xA2, 0x8C, 0x24, 0x42, 0xC6, 0xF3,
-    0x15, 0x18, 0x0F, 0x93, 0x49, 0x9A, 0x23, 0x4D,
-    0xCF, 0x76, 0xE3, 0xFE, 0xD1, 0x35, 0xF9, 0xBB,
-};
+static int initialize_libgcrypt(void)
+{
+    if (!gcry_check_version(UAM_NEED_LIBGCRYPT_VERSION)) {
```
andylemin, line 97: Suggested Alpine/Valgrind fix (details in the PR conversation). It has to be issued before `gcry_check_version`, so it goes first:

```suggestion
    gcry_control(GCRYCTL_SET_PREFERRED_RNG_TYPE, GCRY_RNG_TYPE_SYSTEM);

    if (!gcry_check_version(UAM_NEED_LIBGCRYPT_VERSION)) {
```

Verified in an alpine:3.23.2 container: 186 Memcheck errors before, 0 after, all 69 tests still passing.

bin/afppasswd/afppasswd.c
```diff
@@ -955,7 +879,7 @@ static int update_srp_passwd(const char *path, const char *name, int flags,
         lock.l_len = 0;
         lock.l_whence = SEEK_SET;
 
-        if (fcntl(fd, F_SETLKW, &lock) < 0 || fseek(fp, pos, SEEK_SET) < 0) {
+        if (fcntl(fd, F_SETLKW, &lock) < 0 || fseek(fp, 0, SEEK_SET) != 0) {
```
andylemin: Tiny consistency nit, fine as a follow-up: `-d` switched to `F_SETLK` so a user-held lock can't stall root, but root's `-a` still uses `F_SETLKW` here. Same one-word change would give both admin paths the same behaviour.

bin/afppasswd/afppasswd.c
```diff
@@ -688,61 +552,116 @@ static void srp_encode_hex(char *out_hex, const unsigned char *salt,
...
+    if (fd < 0 && create && errno == EEXIST) {
+        fd = openat(dirfd, uid_name, O_RDWR | O_CLOEXEC | O_NOFOLLOW);
+    } else if (fd >= 0) {
+        created = create;
     }
 
-    if (salt_disabled || verifier_disabled) {
-        return salt_disabled && verifier_disabled;
+    if (fd < 0) {
+        fprintf(stderr, "afppasswd: can't open verifier %s/%s: %s\n",
```
andylemin: Follow-up material, not blocking: a not-yet-enrolled user now hits EACCES on this open (root-owned placeholder) and sees the raw `Permission denied` rather than the "see your administrator" message the `*` check used to give. Mapping `EACCES` for non-root callers to that friendlier text would keep the UX the same as before.

test/afppasswd/test_afppasswd.c
```diff
@@ -0,0 +1,947 @@
...
+
+    if (length < 0 || length >= MAXPATHLEN + 1) {
+        return -1;
+    }
+
+    return mkdtemp(directory) == NULL ? -1 : chmod(directory, 0700);
+}
+
+/* Historical afp.conf path-selection tests retained only for reference. */
+#if 0
```
andylemin: Happy either way, but I'd delete this `#if 0` block rather than keep it for reference; git history has it, and it's ~300 lines that will drift from the code. Trivial, can go in a follow-up.

andylemin (commented):
Second pass on 099fdef9. Nearly every earlier thread is resolved and the result is cleaner than what I reviewed the first time. Three small nits inline plus the one-line Valgrind fix; none of the nits need to block landing once Alpine is green.

test/afppasswd/test_afppasswd.c
rdmark: removed!

bin/afppasswd/afppasswd.c
rdmark: fixed!

bin/afppasswd/afppasswd.c
rdmark: I think this should be easy to accomplish, pushing a fix

bin/afppasswd/afppasswd.c
rdmark: thanks for tracking down the false positive on musl; I like this fix over a compiler flag to suppress it

rdmark:
@andylemin fixes committed and pushed! while doing some corner case testing I found an improvement area around how the SRP UAM reports users who are not enrolled, therefore introduced a granular response enum that distinguishes between unenrolled state and other types of bad permissions for the verifier files

andylemin (approved):
Approving. Reviewed 77adb4d5 against the previous head: the Valgrind fix is in and Alpine is green, `-a` and `-d` now share the non-blocking lock, an unenrolled user gets the friendly message again, the retired tests are gone, `-u` is parsed strictly, and the new verifier-status enum in the SRP UAM is a real improvement over the old binary result. I checked that the root-owned-but-active case still ends up rejected as unsafe and that all failure states produce the same wire response, so nothing leaks to a client. I also confirmed the new `-c`/`-u`/`-w` combination rules don't touch any invocation in the docker script, workflows or manual.

One housekeeping item, not a code blocker: the SonarCloud gate is red on a single new alert in `test_srp_lookup.c` (temporary directory under `/tmp`). It's the same S5443 rule already accepted for the other test files, so either accept it in SonarCloud as before or have the test use libatalk's `tmpdir()` like its siblings. Either is fine by me.

Really nice work on this one! It ended up a lot tighter than where it started, and the docs finally explain the model an admin can actually use and reason about. Really good work 🍻

andylemin:
@rdmark Go for merge 🎉

rdmark:
This was a great process, thank you for taking the time to go in depth and work through all the edge cases with me!

I feel much more confident in pushing this into the new release this late into the development cycle. 

The straggler Sonar issues have been resolved now. Ready to ship.

PR #3318, opened 2026-09-16
compat: add header so strlcpy is defined on platforms with older glibc 

nutzman:
Platforms like Debian/bookwork for raspberry pi have an older glibc without strlcpy, strlcat defined.  The presence/absence is detected by mason correctly but some files were missing the include for when it is missing.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved compatibility across supported platforms and build environments.
  - Updated internal components and test utilities to use shared compatibility definitions.

- **Tests**
  - Aligned file-operation, locking, performance, and helper tests with the updated compatibility support.
  - Improved consistency across test environments and supported build configurations.


bin/nad/ftw.c
```diff
@@ -42,6 +42,7 @@
 #include <unistd.h>
 
 #include <atalk/util.h>
+#include <atalk/compat.h>
```
rdmark, line 44: put this above atalk/util.h

etc/atalkd/config.c
```diff
@@ -33,6 +33,7 @@
 #include <atalk/unicode.h>
 #include <atalk/util.h>
 #include <netatalk/at.h>
+#include <atalk/compat.h>
```
rdmark: put this above atalk/logger.h

etc/cnid_dbd/db_param.c
```diff
@@ -19,6 +19,7 @@
 #include <unistd.h>
 
 #include <atalk/logger.h>
+#include <atalk/compat.h>
```
rdmark, line 21: put this above atalk/logger.h

andylemin (dismissed):

rdmark (commented):
thanks for submitting the PR! 

our coding rules require alphabetic order within each cluster of headers; this may seem as a nitpick, but the main reason is to avoid duplicated headers getting added over time (this has been a real, long-running problem in this project)

andylemin:
@rdmark We should consider raspberry pi builds in the CI toolchain?
I run Pi's at home as well 🤓 this should definitely have proper testing

bin/nad/ftw.c
nutzman: moved

rdmark:
@nutzman thanks for the header order adjustments!

can you please squash these into a single commit locally and force push to the PR?

we only do rebase merges in this project :)

rdmark:
@andylemin thats a good idea, and we can do ARM architecture CI jobs as well 

I test a lot on RPis actually, but usually only the latest version of Raspbian

for this particular issue, the trigger is an older toolchain which we can test on regular amd64 containers in CI

andylemin (approved):

andylemin:
@rdmark Approved and merged this even though some of the ordering is still incorrect.
I think we need to run an automated job to sort all the headers in all the files package wide

PR #3322, opened 2026-09-19
doc: programmatically generate TOC index for manpages

rdmark:
rather than hard coding a list of manpages in the manual sidebar index, generate it through the meson logic

also, add the same TOC to the manual index page and rephrase the introduction to read like a guide to the manual rather than a high level introduction to Netatalk

## Summary by CodeRabbit

- **Documentation**
  - Refreshed the manual with a concise overview of installation, configuration, operation, features, and upgrades.
  - Added automatically generated navigation for user commands, configuration files, and administrator commands.
  - Improved manual and website generation for standard and Japanese documentation.
  - Added a tool to generate manpage table-of-contents entries and navigation links.
  - Removed the standalone legacy manual navigation index and its translations.
  - Updated Japanese translations to match the revised structure and generated sections.


rdmark:
@andylemin the intention here is to have a manual index and introduction that fits better in the new context of living side by side with the Netatalk Client manual

besides, explaining what Netatalk is on the landing page of the manual felt redundant since you likely navigate here from another page on the website that also describes what Netatalk is

plus, the programmatic TOC reduces the manual maintenance overhead when adding or removing manpages from the roster

what do you think?

rdmark:
@andylemin there is a side effect worth noting here, that the resulting manpage TOC index is no longer fully alphabetic, but rather grouped by functional area (general->dbd->fce->testsuite->appletalk) which I would argue is actually a minor improvement

andylemin:
Nice! I actually prefer the functional groupings, especially as we've spent some time recently fixing up the groupings 😉

andylemin (approved):

PR #3326, opened 2026-09-20
distrib: fix test suite flag checks in container entrypoint

rdmark:
the container entrypoint script had outdated syntax in a few places:

- the identifier for the spectest is now 'spec'
- afppasswd only takes the -f parameter together with -c -r

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected setup of randomly generated credentials for the second AFP user.
  - Updated test-suite option handling so the `spec` suite correctly applies its share path and strict-locking settings.
  - Improved reliability of environment setup and specification test execution.


rdmark:
@andylemin FYI, merging this trivial PR without review to unblock local testing of other branches

